PR #29, opened 2018-01-01
Fix client consensus to use localhost address for cluster

Eyesonly88:
**What does this PR do?**
It makes the `getmessage` and `putmessage` commands work. The consensus server was hardcoded to "192.168.0.108" and I changed it to the server address. 

Without doing this, the message commands just say `Connecting to "192.168.0.108"`. I tried this by following the guide using IntelliJ Idea.

Tested by manually starting a consensus server then setting a message and getting it as you can see below
image

**Follow up**
I'm a beginner to blockchain development so please let me know if I'm doing something wrong or there is a better way of doing it. 
I'm also not sure this is the best solution either but it seems to work :)


conglinduye:
Thank you very much for your proposal, it is a good way, I will adopt your plan.Hope we can communicate more later.

PR #33, opened 2018-01-03
Fix unit tests by refactoring Blockchain initialization and using Mockito.

seanrobbins:
**What does this PR do?**

Refactor the constructor on Blockchain so there is only a single constructor and the constructor decides whether to initialize a new DB or not. This removes the need for the usage of System.exit(0)

Refactors BlockchainTest so all tests can share the same instance of Blockchain avoiding the issue attempting to initialize Blockchain, when a previous instance had locked access to LevelDB

Uses Mockito to mock the Blockchain in BlockUtilsTest to avoid tests locking access to LevelDB before BlockchainTest runs on a gradle build. Currently only BlockchainTest needs an actual instance of the Blockchain class.

System.exit(0) was causing test processes to terminate which meant the full test suite was not running on the gradle build. Avoiding stopping the entire running process in this way is recommended (as tests were being ignored and so there was a risk of failures not being found when developing).

Previous tests locking access to LevelDB was preventing tests from running together when running a test suite or attempting to run the gradle test task.

- Unit Tests

**Follow up**
This was the fastest way of removing these issues I found. However, it may be desirable to change the constructor on Blockchain from Blockchain(String address) to something like Blockchain(LevelDbDataSourceImpl db)

This would allow passing a mocked LevelDB into Blockchain, which would make DB states much easier to configure for testing the Blockchain functionality.

To allow this initialising the DB could become a responsibility of Peer or maybe Wallet.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ buildscript {
 
 dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    testCompile group: 'org.mockito', name: 'mockito-all', version: '1.9.5'
 
     compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
     compile group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.25'
--- a/src/main/java/org/tron/core/BlockUtils.java
+++ b/src/main/java/org/tron/core/BlockUtils.java
@@ -30,6 +30,7 @@ import java.sql.Timestamp;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.List;
+import java.util.Optional;
 
 import static org.tron.core.Constant.LAST_HASH;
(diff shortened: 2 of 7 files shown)


seanrobbins:
The unit tests failing on travis did not get flagged when I ran a gradle build locally. 

It looks like a similar issue as the one I was attempting to resolve though.

seanrobbins:
Both BlockStore and UTXOStore attempt to separately initialize the connection with LevelDB. On my local the BlockStoreTest and UTXOStoreTest suites run before the BlockchainTests but after when on Travis.

That shouldn't make a difference but it does. Despite my changes, there are still database states between the tests suites which can negatively impact the test run.

seanrobbins:
Closing the DB after BlockchainTest suite has fixed BlockStoresTest but not UTXOStoreTest

https://travis-ci.org/tronprotocol/java-tron/builds/324435064?utm_source=github_status&utm_medium=notification

PR #34, opened 2018-01-03
Use a mock for UTXOSet in Blockchain test

seanrobbins:
**What does this PR do?**

Swaps the UTXOSet in BlockchainTest for a mock

Builds on Travis are failing when attempting to run UTXOStoreTest with a org.fusesource.leveldbjni.internal.NativeDB$DBException

The reason is both UTXOStore and UTXOSet create a lock on the TransactionDB. Using a mock for UTXOSet removes the need for a lock to be set on the TransactionDB to run BlockchainTest suite

- Unit Tests

**Follow up**

UTXOStore and UTXOSet classes both initialize and set a lock on the TransactionDB.

Blockchain and BlockStores classes both initialize and set a lock on the BlockDB.

In all cases these initialization on the DB are done in the constructor of the class.

It seems like these DBs could be initialized outside of the class, perhaps in the Peer class and then the database dependency injected into Blockchain, BlockStores, UTXOStore and UTXOSet in the constructor.

This way, we can ensure the DB connections are created and locked only once during the runtime lifecycle and avoid the NativeDB$DBException.

This would also allow using a mock for TransactionDB and BlockDB, which would allow for easier state management of the data in the unit tests and so better test coverage.

A wrapper class could be created around the initialization of the database so that the type system can make sure the correct DB is injected into correct class. For example, the Blockchain constructor could look like:

`public Blockchain(BlockDB blockDb)`

### DIFF
--- a/src/test/java/org/tron/core/BlockchainTest.java
+++ b/src/test/java/org/tron/core/BlockchainTest.java
@@ -19,6 +19,7 @@ import com.google.protobuf.InvalidProtocolBufferException;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.tron.protos.core.TronBlock.Block;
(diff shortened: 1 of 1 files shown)


seanrobbins:
I created a ticket to collect my thoughts on perhaps injecting the database dependencies: https://github.com/tronprotocol/java-tron/issues/35

PR #36, opened 2018-01-03
refactor string concatentation with StringBuilding in loop

khuezy:
**What does this PR do?**
- Refactors string concatenation with StringBuilder using a for loop.

Performance improvements.
String concatenation is expensive in in-deterministic loops.

### DIFF
--- a/src/main/java/org/tron/command/Cli.java
+++ b/src/main/java/org/tron/command/Cli.java
@@ -27,12 +27,11 @@ public class Cli {
         Scanner in = new Scanner(System.in);
 
         while (true) {
-            String cmd = in.nextLine();
-            cmd = cmd.trim();
+            String cmd = in.nextLine().trim();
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/tron/command/Cli.java
```diff
@@ -27,12 +27,11 @@ public void run(Peer peer) {
...
         while (true) {
-            String cmd = in.nextLine();
-            cmd = cmd.trim();
+            String cmd = in.nextLine().trim();
 
             String[] cmdArray = cmd.split("\\s+");
-
-            if (cmdArray.length == 0 || cmdArray[0].equals("")) {
+            // split on trim() string will always return at the minimum: [""]
+            if ("".equals(cmdArray[0])) {
```
khuezy: Might be clean to handle this in the switch statement:
```
case "":
     break;
```

PR #45, opened 2018-01-09
Add dependency injection, refactor static classes

Rovak:
**What does this PR do?**

* Added dependency injection using Guice
* Changed dependency on commands from peer to application
* Made Client non-static
* Made Server non-static
* Changed some classes to accept dependencies through constructor
* Added shutdown method to application so services can cleanly shutdown instead of a forced terminate using System.exit

* Static classes are harder to test
* Dependency injection makes it easier to wire classes together
* There should be cleanup logic to properly shutdown connections, leave clusters etc

- Unit Tests
- Manual Testing

### DIFF
--- a/src/main/java/org/tron/consensus/server/Server.java
+++ b/src/main/java/org/tron/consensus/server/Server.java
@@ -19,6 +19,7 @@ import io.atomix.catalyst.transport.netty.NettyTransport;
 import io.atomix.copycat.server.CopycatServer;
 import io.atomix.copycat.server.storage.Storage;
 import io.atomix.copycat.server.storage.StorageLevel;
+import org.tron.application.Service;
 import org.tron.consensus.common.GetQuery;
 import org.tron.consensus.common.MapstateMachine;
 import org.tron.consensus.common.PutCommand;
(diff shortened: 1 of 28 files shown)


src/main/java/org/tron/consensus/server/Server.java
```diff
@@ -64,4 +68,9 @@ public static void serverRun() {
             e.printStackTrace();
         }
     }
+
+    @Override
+    public void stop() {
+        server.leave();
```
Rovak: I'm not 100% sure if this is the right way to dispose the server

src/main/java/org/tron/peer/PeerBuilder.java
```diff
@@ -19,22 +23,25 @@
     private Wallet wallet;
     private ECKey key;
     private String type;
+    private Injector injector;
 
-    public PeerBuilder() {
-        Server.serverRun();
```
Rovak: I removed this because starting a server should not be done within a factory class. Starting the server is now being done in `org.tron.example.Tron.run()`

src/main/java/org/tron/core/Blockchain.java
```diff
@@ -57,6 +59,8 @@
     private byte[] lastHash;
     private byte[] currentHash;
 
+    private Client client;
+
     /**
      * create new blockchain
      *
```
seanrobbins: I'm wondering if it is possible to inject the database when initializing the Blockchain for the first time from this constructor:
`public Blockchain(String address, String type) {`

Doing so would avoid making the Blockchain responsible for setting the database connection, which to me isn't an intuitive responsibility of this class.

Doing so would mean the logic for deciding whether to load an existing or create a new blockchain would have to change to some other way.

I wonder if `if(blockDB.getData(LAST_HASH) != null) { ... }` is sufficient for this purpose?

sasaxie: Initialize the database when the program starts. 

Solution A:
```java
// Initialize on the outside
DB blockDB = new DB(); 
// Incoming through formal parameters
Blockchain blockchain = new Blockchain(blockDB); 
```

PR #52, opened 2018-01-10
Don't refer to singleton instance in blockchain

Rovak:
**What does this PR do?**

The code references the singleton instance when it should be refering to his own instance

When there are multiple blockchain instances then it should be refering to his own instance

- Unit Tests
- Manual Testing


sasaxie:
Thanks~

PR #53, opened 2018-01-10
refactor: make wallet less complex to use

JohnnyQQQQ:
**What does this PR do?**

The Wallet class was unnecessary complex as one had to
call init() for no real reason as this could also happen
easily in the constructor.

To make the Wallet class easy understandable 

- Manual Testing

**Follow up**

Do we even need a Wallet class? It only wraps ECKey.

**Extra details**


sasaxie:
We will expand our wallet in the future. Thank you~

PR #54, opened 2018-01-11
Inject block db to blockchain

seanrobbins:
**What does this PR do?**

1) Injects blockDB from Guice injector into the Blockchain constructor, from within PeerBuilder

2) Uses a mock for blockDB in BlockchainTest

3) Removes dbExists feature from Blockchain

1) Should mean the instance of blockDB launched by Guice is the only instance needed. Abstracts Blockchain away from responsibility for maintaining database state.

2) Using a mock in BlockchainTest will make it easier to set up of various test conditions. This will make testing Blockchain class easier.

3) As Guice initializes the blockDB, relying on the database file existing no longer seems to be a reliable way to know if the blockchain has been created.
I encountered NullPointerExceptions where the blockDB existed but the LAST_HASH was null. 
In these changes, if LAST_HASH is null on initialization of Blockchain, the create Blockchain logic will run.

- Unit Tests
- Manual Testing

**Extra details**

Also removes constructor used by Guice to initialize a Blockchain. PeerBuilder has sole responsibility for initializing Blockchain.

### DIFF
--- a/src/main/java/org/tron/core/Blockchain.java
+++ b/src/main/java/org/tron/core/Blockchain.java
@@ -54,7 +54,7 @@ public class Blockchain {
     public static String parentName=Constant.NORMAL;
 
     public static final Logger logger = LoggerFactory.getLogger("BlockChain");
-    private LevelDbDataSourceImpl blockDB = null;
+    private LevelDbDataSourceImpl blockDB;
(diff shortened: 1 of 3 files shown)


sasaxie:
LGTM

PR #55, opened 2018-01-11
Decouple client from blockchain using events

Rovak:
**What does this PR do?**

* Decouples the Client from blockchain by using events
* Removes the Tron.getPeer singleton

* Because the blockchain should not be dependent on the client
* There should not be any singleton reference to the peer, and other classes should not be dependent on the singleton

- [x] Unit Tests
- [x] Manual Testing


Rovak:
Merge #54 before this pull request, there are gonna be some merge conflicts i guess

PR #62, opened 2018-01-15
Fix BlockchainIterator#next and add tests

jromero:
**What does this PR do?**
* Fix the `BlockchainIterator#next()` method to adhere to the defined java contract. next() should return `NoSuchElementException` instead of null when surpassing values.
* Tests added for `BlockchainIterator`
* Define type to generic Iterator to prevent the need for casting

* Currently `BlockchainIterator#next()` returns null instead of throwing an exception as expected based on java docs. 
* Tests are always good 😄 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,8 @@ configurations.getByName('checkstyleConfig') {
 
 dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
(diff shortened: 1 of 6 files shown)


ybhgenius:
Hello, I am a crew member of TRON, and now we are planning to provide an incentive for your contribution. Please leave me your email address. Thanks.

PR #60, opened 2018-01-13
Format code after google's java styleguide, also add codeclimate

JohnnyQQQQ:
**What does this PR do?**

- Major code formatting after [google's java styleguide](https://google.github.io/styleguide/javaguide.html)
- Lint tasks that checks if the code is like defined in the styleguide
- Codeclimate integration for indepth checks of the code quality

**Why are these changes required?**

If we don't use any styleguide for coding, the project might end in a big mess 😄 

**This PR has been tested by:**
- Unit Tests
- Manual Testing

**Follow up**

- A Projectmember has to setup the project on https://www.codeclimate.com
- Discuss in #48 what else we could do to improve the projects quality

**Extra details**

[Stylefile for Intellij](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml)

[Stylefile for Eclipse](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml) Thanks @artinpetrossian

PR #66, opened 2018-01-16
Reformat code to Google java styleguide

jromero:
**What does this PR do?**
* Reformat code to adhere to the pre-approved notion to use the google java styleguide. See https://github.com/tronprotocol/java-tron/pull/60
* Protobuf generated classes are excluded from checkstyle
* Run `lint` as part of the `test` step
* Checks are set to warning except for `Indentation` which have been fixed in this PR

* Code consistency

- Unit Tests
- Manual Testing

**Follow up**
The thought is to follow this PR up with fixes to the current lint violations and changing them from WARNING to ERROR to prevent future deviation.

**Extra details**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,5 +29,7 @@ pernodedata.*
 ringstate.*
 shareddata.*
 
-# svn
-.svn
+-# svn
+-.svn
+
+src/main/gen/
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,4 @@ shareddata.*
 -# svn
 -.svn
 
-src/main/gen/
+generated
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,4 @@ language: java
 jdk: oraclejdk8
 
 script:
-  - ./gradlew lint
   - ./gradlew test
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ apply plugin: 'java'
 apply plugin: 'com.google.protobuf'
 apply plugin: 'application'
 apply plugin: 'checkstyle'
+apply plugin: 'idea'
 
 sourceCompatibility = 1.8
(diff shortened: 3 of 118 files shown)


Rovak:
You should add a travis-ci script that checks if new code complies with the styleguide. I've seen recent commits from other developers that violate the code guidelines.

jromero:
@Rovak Good idea. 

I went ahead and elevated the warnings to errors and made `test` depend on `lint`. 

One thing I must note is that there are A LOT of violations. I suppressed some locally via annotations and others globally by commenting out the rule in the `google_checks.xml`. Without doing that I would think this PR would be extremely large. This allows those disabled rules to be tackled as independent PRs.

JohnnyQQQQ:
I don't think that this is an approach that scales very well. I would opt-in to go with a checkstyle on the changeset of every pull-request.

Ebay did a neat article on how this can be archieved https://ebaytech.berlin/checkstyle-on-changed-files-with-gradle-1619e49dbe4e

jromero:
I think the scenarios are very different. They had a large code base to start off. This repo is still relatively small that all issues can be fixed fairly easily. The only reason why I didn't was to limit the changeset.

After all issues are fixed both solutions work about the same.

AlmightyKerk:
Update  consistancy to consistency

jromero:
> Update consistancy to consistency

Everyone's a critic... 😛

sasaxie:
Dear developers:                                                                                                                                
Thank you for your enthusiasm and support to TRON. Recently we will make a standard form of submitting codes, so the new pull requests will not handled temporarily.

jromero:
I've updated the PR an minimized the scope. I've also update the PR description to give a better understanding on what the changes are and what is still to come.

I've seen a lot of chatter in the gitter room about people wanting to help. Addressing lint warnings and setting the rule to error out in the future might be a good low hanging fruit for many.

PR #74, opened 2018-01-20
Fix test bytes to long

seanrobbins:
**What does this PR do?**

1) Calls buffer.clear() before writing to the ByteBuffer in TypeConversion 
2) Adds assertions into the TypeConversionTest

This test was failing in build runs from Gradle: 

org.tron.common.utils.TypeConversionTest > testBytesToLong FAILED
    java.nio.BufferOverflowException at TypeConversionTest.java:39

Was not failing when the same test is run in IntelliJ, but as it was being caused by the buffer containing bytes from a previous use of the TypeConversion class, may have also become an issue for the running application in the future.

- Unit Tests

### DIFF
--- a/src/main/java/org/tron/common/utils/TypeConversion.java
+++ b/src/main/java/org/tron/common/utils/TypeConversion.java
@@ -25,11 +25,13 @@ public class TypeConversion {
   private static ByteBuffer buffer = ByteBuffer.allocate(8);
 
   public static byte[] longToBytes(long x) {
+    buffer.clear();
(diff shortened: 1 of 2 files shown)


zhaohong:
thanks!

ybhgenius:
Hello, I am a crew member of TRON, and now we are planning to provide an incentive for your contribution. Please leave me your email address. Thanks.

seanrobbins:
Hello, that sounds very interesting. You can contact me using sean.m.robbins at gmail.com

PR #77, opened 2018-01-21
Add protobuf generated classes to .gitignore file

seanrobbins:
**What does this PR do?**

Adds protobuf generated classes to .gitignore file

If someone working in the code base does a 'git add' after the Protobuf code has been generated, the generated code will get added and then possibly committed. 
I expect the repository only needs to keep track of the .proto files and not the generated code.

- Manual Testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ shareddata.*
 
 # svn
 .svn
+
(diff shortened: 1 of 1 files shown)


zhaohong:
Thanks!

PR #83, opened 2018-01-24
Add services init to constructor when running with injector

svenanders:
**What does this PR do?**
When running the demo app as described in the README, I get a new Nullpointer when trying to execute the app (it fails at `public void addService(Service service)` in `Application.java`.

This is because the constructor calls to initialize `this.services` and `p2pnode` is not called when running the Application with `new Application(buildGuice())`.

In this PR I added the init calls to the constructor so that the project can be started as described in the README.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**
I changed the readme to point to `org.tron.program.example.Tron` for the Main Class. The gif needs to be updated as well (though I haven't done so).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ TRON is a product of Web 4.0 and the decentralized internet of next generation.
 
 IDEA: 
 - [Edit Configurations...] -> [Add New Configuration] -> [Application]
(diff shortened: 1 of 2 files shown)


drevie:
thanks!

fayimora:
Would be nice to get this in @sasaxie

PR #123, opened 2018-02-12
Potential Deadlock fixed 

sdargutev:
**What does this PR do?**
Fixes a bug in LevelDbDataSourceImpl.closeDB, that can cause deadlock and memory leak when called multiple times.

To avoid blocking threads and leaking memory. 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


Rovak:
I've  fixed this one before in #71, weird that the previous fix got reverted :/

The unittests you've added should ensure this bug won't fly under the radar again

zhaohong:
@Rovak  i'll check why got reverted

@sdargutev  thanks!

PR #131, opened 2018-02-14
NodeImpl improvements

Rovak:
* Remove duplicate `loopFetchBlocks.join()`
* Await advertise thread when closing node
* Mark `isAdvertiseActive` as volatile to prevent potential variable caching problems
* Replace `Function` in `ExecutorLoop` with `Consumer` because it was never returning a result
* Marked `trxToAdvertise` and `blockToAdvertise` `final` because they are being `synchronized`


zhaohong:
thanks

PR #132, opened 2018-02-14
Remove Apache Gossip because it was replaced with Scalecube

Rovak:
Remove Apache Gossip from libs because it was replaced with Scalecube


zhaohong:
thanks

PR #133, opened 2018-02-14
Remove deprecated Kafka and Raft (Phase 1 and 2) Consensus

Rovak:
## Cleanup old consensus phases

* Remove Kafka consensus
* Remove CopyCat (Raft) Consensus
* Remove Kafka docker containers
* Remove kafka and copycat unittests
* Remove `PendingStateImpl` because it wasn't being used and most likely will be replaced in the futur


zhaohong:
thanks

PR #134, opened 2018-02-15
use class.getName() instead of string

Rovak:
Minor improvement, use the `class.getName()` to retrieve the classname instead of a string


zhaohong:
thanks!

PR #135, opened 2018-02-15
Move NodeDelegate logic to seperate class

Rovak:
The `NodeDelegate` logic should not be in the application class and has been moved to a `NodeDelegateImpl` class


Rovak:
fixed travis

olenheim:
@Rovak Thanks for your pull. The NodeDelegate here is in charge of some logic event about blockchain such as sync, verify trx, handle block etc. Actually in our design Application should do this thing. Node is in charge of more basal logic event of  communication with peers.

Rovak:
@olenheim thanks for the clarification, but i still believe that moving the `NodeDelegate` implementation out of the `ApplicationImpl` allows for better separation of concerns. My `NodeDelegate` also has fewer dependencies which makes it easier to unittest in the future.

zhaohong:
@olenheim  @Rovak    this pr looks good to me .

olenheim:
@Rovak  I think maybe you are right, we will accept your PR，please fix the bug and conflict because of  new commit , then I will merge it, Thanks

Rovak:
I'll fix this later this week

PR #149, opened 2018-02-27
Feature/handle vote transaction

zhangheng1536:
**What does this PR do?**
Deal with voting transactions.
Make statistics on the votes.
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


src/main/java/org/tron/core/db/Manager.java
```diff
@@ -145,6 +150,42 @@ public boolean processTrx(TransactionCapsule trxCap) {
     return true;
   }
 
+  private void voteWitessCount(Transaction trx) {
+    try {
+      int parameterSize = trx.getParameterList().size();
+      if (parameterSize <= 0) {
```
KamilZielinski: You can use `isEmpty()` method to check if list has nothing inside

src/main/java/org/tron/core/db/Manager.java
```diff
@@ -145,6 +150,42 @@ public boolean processTrx(TransactionCapsule trxCap) {
     return true;
   }
 
+  private void voteWitessCount(Transaction trx) {
```
KamilZielinski: `voteWitessCount` -> voteWitnessCount

src/main/java/org/tron/core/db/Manager.java
```diff
@@ -145,6 +150,42 @@ public boolean processTrx(TransactionCapsule trxCap) {
...
+            countVoteWitess(voteAddress, voteAdd);
+          });
+        }
+      }
+    } catch (InvalidProtocolBufferException e) {
+      e.printStackTrace();
+    }
+  }
+
+  private void countVoteWitess(ByteString voteAddress, int countAdd) {
```
KamilZielinski: `countVoteWitess` -> `countVoteWitness`

zhangheng1536: All right, this is really my mistake, thank you for your correction, and I will fix it immediately.
@KamilZielinski

src/main/java/org/tron/core/db/Manager.java
```diff
@@ -145,6 +150,40 @@ public boolean processTrx(TransactionCapsule trxCap) {
     return true;
   }
 
+  private void voteWitnessCount(Transaction trx) {
+    try {
+      if (trx.getParameterList() == null || trx.getParameterList().isEmpty()) {
```
KamilZielinski: Actually, my bad here. I didn't notice that Apache Commons Collections are included into the project, so you can use `CollectionUtils.isEmpty()` which already has null check

zhangheng1536: @KamilZielinski 
We will be happy to accept your PR which  have not  bugs if you wish.
Thank you for your attention.

PR #172, opened 2018-03-05
Reduce the amount of code a little

AhnSinYong:
**What does this PR do?**
**Args.java**
- new String() create a new distinct instance of String object but literal reuse an instance from the string pool if one is available.

**BlockCapsule.java**
- Replace list null check with CollectionUtils.isEmpty().

**UtxoStore.java**
- Replace if condition code reformat for readable.
- Replace the way add a value to long array.

- Reduce the amount of code a little


zhaohong:
thanks!

PR #173, opened 2018-03-05
Reduce the amount of code a little #2

AhnSinYong:
**What does this PR do?**
**ByteArrayTest.java**
- Reformat some code
- Add 2 test methods about int

**ByteArray.java**
- Reformat some code
- Replace Try-Catch-Finally with Try-with-resources
- Replace if statement with conditional ternary operator
- Replace if condition with third party libraries utilities

**ActuatorFactory.java**
- Remove duplicate calling same method
- Replace return null of createActuator method with return actuatorList
  Because return null does not seem appropriate
  
**Manager.java**
- Reformat some code

Reduce the amount of code a little #2

- Unit Tests (partly)

### DIFF
--- a/src/main/java/org/tron/common/utils/ByteArray.java
+++ b/src/main/java/org/tron/common/utils/ByteArray.java
@@ -18,14 +18,18 @@
 
 package org.tron.common.utils;
 
+import com.google.common.primitives.Ints;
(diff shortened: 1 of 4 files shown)


zhaohong:
thanks!

trastvolt:
gh pr checkout 173

PR #176, opened 2018-03-06
Modify the code slightly

AhnSinYong:
**What does this PR do?**
**RpcApiService.java**
- Replace Anonymous inner class by lambda

**BlockStore.java**
- Replace if condition with third-party utilities
- Replace if statement with ternary operator
- Simplify logical operation (containBlock method)
- Replace creation empty arraylist and add elemnt with stream collect (getBlockChainHashesOnFork method)

Modify the code slightly


zhaohong:
thanks!

PR #258, opened 2018-03-18
[DO NOT MERGE] Untangling static dependencies

Rovak:
This pull request is created to allow for review, this is a ongoing refactor

* Inject dependencies rather than referencing static instances
* Move `WalletApi` to `api` namespace
* Made grpc port configurable from config


src/main/java/org/tron/common/application/Module.java
```diff
@@ -15,21 +15,53 @@
...
 
 public class Module extends AbstractModule {
 
+  private static final Logger logger = LoggerFactory.getLogger("Module");
+
+
+  private final Config config;
+  private Args args;
+
+  public Module(Config config, Args args) {
```
Rovak: `Args` dependency should be removed

PR #264, opened 2018-03-19
BlockCapsule refactoring

KamilZielinski:
**What does this PR do?**
Refactoring BlockCapsule class and its tests (important question down below)

Code was messy.

- Unit Tests

**Extra details**
- Lombok used
- I decided to remove unpacking since it's always unpacked when you create it with any contructor. The one which takes `data` as an argument is also parsing `byte[]` to `block` so it's never null
- Added finals

**Important question!** 
Not sure but if there would be a case that someone could call `getData()`, right after adding first transaction, then if second one was added the `data` won't update anymore. 
Is this a real usecase? Should `data` be always up-to-date? Was it done like that previously due to some performance issues?


zhaohong:
Dividing this pr into different pr is more better, one is about  typos correcting and  other is about the logic.

Issue #337, opened 2018-03-31
Build Failed in Ubuntu 16.04.4 LTS


WendySanarwanto:
Please answer these questions before submitting your issue. Thanks!

1. What did you do?
When I followed instruction on `README.md` file for building the source by running `./gradlew build`, the build process was ended as __BUILD FAILED__

2. What did you expect to see?
The build process should be ended as successful.

3. What did you see instead?
![image](https://user-images.githubusercontent.com/18456618/38158972-b98d7d46-34d1-11e8-8143-0e9fbee8a939.png)

My environment settings:
![image](https://user-images.githubusercontent.com/18456618/38158974-d84f2b62-34d1-11e8-9d1b-45d5c0c88c7d.png)

AydenRennaker:
https://www.digitalocean.com/community/tutorials/how-to-install-java-with-apt-get-on-ubuntu-16-04

Install the 1.8 jdk its a dependency issue the error comes from a function that has been deprecated in the latest java version.

WendySanarwanto:
Hi @AydenRennaker,

I've followed your suggestion by installing `Oracle JDK 8`, replacing my `OpenJDK 8`, and then re-ran the build. The build's outcome is `SUCCESS`. I will edit the `README.md` file to tell the reader that they should use `Oracle JDK 8` over `OpenJDK 8` so that they won't get failing build.

Cheers.

lacasseio:
I suggest installing `openjfx` using `sudo apt-get install openjfx` instead of forcing users to install Oracle JDK 8 as the issue is related to missing JavaFX package. See https://stackoverflow.com/a/19529820.

WendySanarwanto:
Hi @lacasseio ,

Thanks for your feedback. I will retest your suggestion and will get back with the result.

vniche:
Real pain to work with Oracle JDK now as it always been, any new words on the dependency on it? I would really like to go for at least OpenJDK being officially supported...

Nguyentiensy99:
Cảm ơn các bạn làm ơn thêm tên tôi OK nguyentiensy. Tôi không hiểu về công nghệ nên các bạn giúp đỡ cảm ơn nhiều nhiều

gen2brain:
I managed to compile on Fedora with OpenJDK and OpenJFX, I just installed distro packages and build with the gradle. Just wanted to let you know Oracle JDK is not necessary.
Not related to Ubuntu though, sorry, just think this info can be useful to someone else.

[...]

PR #339, opened 2018-03-31
Added info under `Prepared dependencies` section, to suggest Linux Ubuntu's users, using Oracle JDK 8, instead of OpenJDK 8, to avoid error build

WendySanarwanto:
**What does this PR do?**
Changed `README.md` file by adding info under `Prepared dependencies` section, to suggest Linux Ubuntu's users, using Oracle JDK 8, instead of OpenJDK 8, to avoid error build.

Some Linux Ubuntu users are using `OpenJDK 8` instead of using `Oracle JDK 8`. Before this PR, the `Prepared dependencies` section only mention `JDK` , not mentioning specific `JDK` flavor. Also, building the source code using `OpenJDK 8` will result in `BUILD FAILED` outcome (see: https://github.com/tronprotocol/java-tron/issues/337). By adding more info in `Prepared dependencies` section, we could help Linux Ubuntu user (whom use OpenJDK 8), avoiding this build error.

- Manual Testing

**Follow up**

**Extra details**
Machine environment:
- Ubuntu OS version 16.04.4 LTS.
- Oracle JDK 8

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ git clone https://github.com/tronprotocol/java-tron.git
 ## Prepare dependencies
 
 * JDK 1.8 (JDK 1.9+ are not supported yet)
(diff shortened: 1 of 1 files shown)


Phum7895 (changes requested):
@neweromask

PR #478, opened 2018-04-19
Disconnect peer liang

lvs007:
**What does this PR do?**
If the main chain of the current node is higher than the synchronous node and the synchronization block is not found, then the peer link is disconnected.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/src/main/java/org/tron/core/net/node/NodeImpl.java
+++ b/src/main/java/org/tron/core/net/node/NodeImpl.java
@@ -332,7 +332,6 @@ public class NodeImpl extends PeerConnectionDelegate implements Node {
                       peer.getAdvObjWeSpread().put(idToSpread.getKey(), Time.getCurrentMillis());
                       sendPackage.add(idToSpread, peer);
                     });
-//                peer.cleanInvGarbage();
               });
 
           sendPackage.sendInv();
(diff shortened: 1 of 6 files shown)


src/main/java/org/tron/common/overlay/message/ReasonCode.java
```diff
@@ -82,6 +82,8 @@
 
     RESET(0x11),
 
+    SYNC_FAIL(0x12),
```
olenheim: this reason need modify proto files

src/main/java/org/tron/core/net/node/NodeImpl.java
```diff
@@ -748,7 +749,14 @@ private void onHandleSyncBlockChainMessage(PeerConnection peer, SyncBlockChainMe
     }
 
     if (blockIds.isEmpty()) {
-      peer.setNeedSyncFromUs(false);
+      if (CollectionUtils.isNotEmpty(summaryChainIds) && summaryChainIds.get(0).getNum() < del
+          .getHeadBlockId().getNum()) {
```
olenheim: when peer's number id lower than our solid block number , It's really sync fail, in this case , it maybe can sync again

src/main/java/org/tron/common/overlay/message/ReasonCode.java
lvs007: add to proto files

src/main/java/org/tron/core/net/node/NodeImpl.java
lvs007: modify use LatestBlockHeaderNumber compare

PR #588, opened 2018-05-01
fix: change invalid if condition

AhnSinYong:
**What does this PR do?**
- change invalid if condition

- "if condition" and "else if condition" are the same.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


AhnSinYong:
I do not know what the intention of the person who wrote this code is, so I can not give a concrete basis.
However, I wrote this PR because I think it will be possible to judge whether there is a problem when the code writer confirms it.

AhnSinYong:
If it is an error, it will be a merge, otherwise the PR will just be closed.

paulvi:
OK, so it is like questioning authors "hey, what is this for, is there error here?"

Yrp:
@AhnSinYong @paulvi thank you for the help. We have fixed the issue. https://github.com/tronprotocol/java-tron/pull/645

PR #600, opened 2018-05-03
remove gradle use of jars in libs folder

paulvi:
remove gradle use of jars in libs folder
(is not present and is not recommended)

**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


paulvi:
Excuse if it was not clear, but this was over line

`compile fileTree(dir: 'libs', include: '*.jar')` inf gradle

PR #731, opened 2018-05-15
when connect the node add to active peer after finish handshake finish

lvs007:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


src/main/java/org/tron/common/overlay/server/ChannelManager.java
```diff
@@ -170,32 +116,33 @@ public boolean isRecentlyDisconnected(InetAddress peerAddr) {
   }
 
   public void add(Channel peer) {
-    newPeers.add(peer);
+    if (!peer.isProtocolsInitialized()) {
```
wubin12: 这个逻辑不需要判断了

PR #894, opened 2018-05-30
rm logback to STDOUT

ithinker1991:
**What does this PR do?**
avoid deadlock because of logback 

- Manual Testing

**Follow up**

**Extra details**


jotto:
Is there any additional context around this deadlock? Even a stacktrace would help.

Log output to STDOUT is a very useful feature.

PR #954, opened 2018-06-06
Make the stest can collect the coverage,and add case account007(creat…

wangzichichi:
…eaccount)

**What does this PR do?**
Add case about account007, create account function.
Modify the file build.gradle, let the task "stest" can collect jacoco coverage data.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


wangzichichi:
1.Modify build.gradle, to make the Task "stest" can collect coverage.
2.Set the create witness case to false.

PR #962, opened 2018-06-07
Fix bad transaction

Yrp:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


pangpangfeng:
background:
1. Deploy the beta environment 5witnesses + 1 soliditynode.
2. After block height 6000 blocks.
3. There is a transaction background.
step:
1. Account A initiates a bad transaction and broadcasts, checking all the nodes that received the broadcast. There is no abnormality in the node.
2. Modify the signature of the block After making a bad block, broadcast to check all the nodes that received the broadcast. There is no abnormality in the node.

PR #1063, opened 2018-06-21
feature: add GR withdraw limit

nanfengpo:
**What does this PR do?**
add GR withdraw limit in  withdrawBalance validation
a guard representative is not allowed to withdraw Balance
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


kookiekrak:
@nanfengpo do GRs need to make a new address to come back as SRs?

PR #1274, opened 2018-08-09
[cpu usage] Incorporate CPU Usage code

cryptoannavan:
**What does this PR do?**
incorporeate CPU usage

speed improvements

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


tjchern:
There are two logic: 
1. check the cpu max that vm can use before vm startup, and 
2. generate bill that the account actually should pay after vm end.
the code you modify is the step 1.
Thank you very much

PR #1282, opened 2018-08-09
Revert "Fix validation on AssetIssue abbr"

zhaohong:
Reverts tronprotocol/java-tron#1273

some transaction has already  break the rules.


nfleury:
@zhaohong just to understand, why does the fact that some tx already broke the rule make that we can't fix it now ? What is restricting it from being fixed now exactly ?

PR #1352, opened 2018-08-23
Include AccountCreateContract fees in the black hole account

Rovak:
**What does this PR do?**

Add the fee for the `AccountCreateContract` to the blackhole account

The `AccountCreateContract` fees aren't taken into account in the blackhole account so the total burned TRX isn't accurate


huzhenyuan:
It will cause data not sync for all node.
New node calc the fee with all  AccountCreateContract.
But old node is not but just newerly AccountCreateContract

Rovak:
I don't fully understand how adjusting the balance of the blackhole account prevents the node from syncing. Where is this checking being done?

zhaohong:
reopen.

PR #1477, opened 2018-09-08
Timeout not on chain

tjchern:
**What does this PR do?**

1. delete tolerant
2. timeout throw out, not push to chain

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


src/main/java/org/tron/common/runtime/Runtime.java
```diff
@@ -277,30 +276,18 @@ private long getEnergyLimit(AccountCapsule creator, AccountCapsule caller,
...
-      {
-        if (trx.getRet(0).getContractRet() == contractResult.OUT_OF_TIME) {
-          thisTxCPULimitInUsRatio = Args.getInstance().getMinTimeRatio();
-        } else {
-          thisTxCPULimitInUsRatio = Args.getInstance().getMaxTimeRatio();
-        }
-      }
+    if (ET_NORMAL_TYPE == executorType && null != this.blockCap &&
+        !this.blockCap.getInstance().getBlockHeader().getWitnessSignature().isEmpty()) {
+      thisTxCPULimitInUs = 1000_000L;
```
huzhenyuan: use constant config

src/main/java/org/tron/common/runtime/Runtime.java
```diff
@@ -562,6 +523,7 @@ public void go() {
       result.setException(e);
       runtimeError = result.getException().getMessage();
       logger.error("runtime error is :{}", result.getException().getMessage());
+      throw new VMTimeOutException(e.getMessage());
     } catch (Throwable e) {
```
Yrp: catch (Throwable e) will catch the VMTimeOutException instance

PR #1800, opened 2018-12-01
Add Raw Data Hex

Rovak:
For hardware wallet integration the `raw_data_hex` is required, this can be send to the hardware wallet where it can be decoded and the transaction information can be read which then will be presented to the user


zhaohong:
looks good,.

PR #1838, opened 2018-12-11
hard fork is effective when maintenance

Yrp:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


src/main/java/org/tron/core/config/Parameter.java
```diff
@@ -107,6 +107,7 @@
 
     int START_NEW_TRANSACTION = 4;
     int ENERGY_LIMIT = 5;
+    int version_3_2_2 = 6;
```
KamilZielinski: `final` should be capitalized

PR #1934, opened 2019-01-12
Use original gnu license document

lazaridiscom:
The copyright affects the license-text itself, so it must be

**Copyright (C) 2007 Free Software Foundation, Inc. https://fsf.org/**

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
                    GNU LESSER GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
- Copyright (C) [2007] [TRON Foundation], Inc. <https://fsf.org/>
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 


tzdybal:
Few comments regarding licensing:
* GNU LGPL-3.0 is "canonically" placed in `COPYING.LESSER` file, not `LICENSE`.
* Copyright and license statements/notices should be included in all files (see: https://www.gnu.org/licenses/gpl-howto.en.html).
* Currently LGPL-3.0 is applied in MIT license style (Copyright statement and license in `LICENCE` file).

lazaridiscom:
> GNU LGPL-3.0 is "canonically" placed in COPYING.LESSER file, not LICENSE.

Personally, I find this "COPYING.LESSER" thingy highly annoying. I prefer LICENSE or better LICENSE.txt.

> Copyright and license statements/notices should be included in all files (see: https://www.gnu.org/licenses/gpl-howto.en.html).

Again, very annoying, especially if its in the header. The first thing i want to see is "what does this file contain / do", not some repeated babbling re license.

> Currently LGPL-3.0 is applied in MIT license style (Copyright statement and license in LICENCE file).

This is actually the smallest problem with the license.

The biggest is that it becomes not immediately clear that the code is based on Ethereum. I would expect at minimum a "CREDITS.md" file in root.

-

But having said all that: the Tron repos on github are dead since 14 days. It is unbelievable how sloppy the project and its docs are.

lazaridiscom:
One can discuss oneself to death about "code organization".

But within this project, this detail-level does not matter anyways.

(and this PR is about just this frickin line "Copyright (C) 2007 Free Software Foundation, Inc. https://fsf.org/", which should be fixed. Just file another issue about the other issues you raise, and see how long a response from the team will take...)

PR #1948, opened 2019-01-18
Enable shadowJar for jitpack dependency

nfleury:
**What does this PR do?**
Restore shadowJar build

So that java-tron can be used as dependency with jitpack

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,8 @@ apply plugin: "jacoco"
 apply plugin: 'maven-publish'
 
 jar.enabled = false
-shadowJar.enabled = false
+//allow shadowJar to enable jitpack dependency
+shadowJar.enabled = true
 
 sourceCompatibility = 1.8
 [compileJava, compileTestJava]*.options*.encoding = 'UTF-8'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -264,7 +265,7 @@ run {
 shadowJar {
     baseName = 'java-tron'
     classifier = null
-    version = null
+    version = '1.0.0'
```
wubinTron: Not need to set a special version number. removing shadowJar.enabled could also generate java-tron.jar lib.

nfleury:
> it's not necessary to build a "java-tron" jar.
> What u need is FullNode.jar or SolidityNode.jar

What I needed was to include java-tron as a lib, I don't care about the the distrib jar (=same lib, different bundle)

nfleury:
> setting shadowJar.enabled = true can generate java-tron.jar. Not need a special version number.

Ok thanks, you can close this request anyway.

wubinTron:
We already adopted your suggestion.

PR #2153, opened 2019-04-22
Eliminate dependency on JavaFX

mbashov:
**What does this PR do?**

The class `javafx.util.Pair`, which is the only reason for the JavaFX dependency, is replaced with an internal `Pair` implementation.

JavaFX is a huge dependency that is quite nontrivial to install and configure on common GNU/Linux distributions (for example, Ubuntu 18.04 provides only Java 11 binaries in its openjfx package). None of JavaFX features are actually used in the code; as far as I can see, the only methods from `javafx.util.Pair` that are used are the constructor, `getKey()` and `getValue()`, and they are trivial.

- Unit Tests
- Manual Testing (I haven't done much testing. At least block creation on a private network, broadcasting transactions and querying transactions worked as I expected).


jiangyy0824:
Thanks for your suggestion. This has been fixed in PR#2510. We will test the PR and approve it.

PR #2323, opened 2019-06-21
address add isContract member

llwslc:
**What does this PR do?**
add new opcode isContract

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

[TIP44](https://github.com/tronprotocol/TIPs/issues/44)


sullof:
@Sh11thead 
To improve TronWeb and stop users from trying to transfer TRX and assets to a smart contract account we need an API that can be called to check if that address is actually a smart contract or not. Do you plan to add a new HTTP API for that?

Sh11thead:
@sullof  
This API may be a solution .
`/wallet/getcontract?value=`
If the non-smart contract address, this interface will report an error.

We are envisioning a new API or a unified transfer method that can transfer both the normal address and the smart contract address.

PR #2359, opened 2019-07-17
Hide creating witness apis for Shasta

shenyongri110:
**What does this PR do?**
Use isOneWitness in the config.conf to control if user can create witness for shasta network.

Since shasta won't allow user to create SR, we should hide those create witness api.

- Unit Tests
No
- Manual Testing
Test Env: Private Net

Http API (Postman)
Scenario1: When isOneWitness = true
Expected Result: Should return 405 http code and write log info.
Actual Result: Return 405 (Method is not supported by this URL) and write log info.

Scenario2: When isOneWitness = false || isOneWitness is empty
Expected Result: Should execute original Java-Tron create witness logic.
Actual Result: Can execute witness original logic.

Grpc API (Grpcc client tool)
Scenario1: When isOneWitness = true
Expected Result: Should return null transaction.
Actual Result: Return null transaction.

Scenario2: When isOneWitness = false || isOneWitness is empty
Expected Result: Should execute original Java-Tron create witness logic.
Actual Result: Can execute witness original logic.

**Follow up**

**Extra details**

### DIFF
--- a/src/main/java/org/tron/core/config/args/Args.java
+++ b/src/main/java/org/tron/core/config/args/Args.java
@@ -956,9 +956,8 @@ public class Args {
             : Runtime.getRuntime().availableProcessors();
 
     INSTANCE.isOneWitness =
-            config.hasPath("node.isOneWitness") ? config
-                    .getBoolean("node.isOneWitness")
-                    : false;
+            config.hasPath("node.isOneWitness") && config
+                    .getBoolean("node.isOneWitness");
 
     initBackupProperty(config);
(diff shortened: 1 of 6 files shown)


Yrp:
Thank you for the report.
Regular updates to shasta will not be merged into java-tron unless there are major bugs.

sullof:
@Yrp Someone was able to become an SR today. Shasta is not working anymore and we have to reset it. This will cause serious problems to all the developers and all the explorer, included TronScan. You don't like our PR, fine. But you guys MUST find a solution for private networks.

PR #2374, opened 2019-07-19
Feature/multi validate signature

ithinker1991:
**What does this PR do?**

fix length errror

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


sercangenc:
The problem is that someone is scamming people using multi-signature, giving out their private key and people are investing to withdraw the balance, but they can't withdraw, how to withdraw the balance here

sercangenc:
And how to return these victim addresses

PR #2396, opened 2019-08-01
Change method name 'build' to 'createApplication', 'size' to 'getStreamCount'

pdhung3012:
The class ApplicationFactory is used to represent ApplicationFactory.  This method named 'build' is to create an instance of the application. Thus, the method name 'createApplication' is more intuitive than 'build'.

The class TronStoreWithRevoking is used to represent TronStoreWithRevoking.  This method named 'size' is to count number of streams. Thus, the method name 'getStreamCount' is more intuitive than 'size'.

### DIFF
--- a/src/main/java/org/tron/common/application/ApplicationFactory.java
+++ b/src/main/java/org/tron/common/application/ApplicationFactory.java
@@ -22,7 +22,7 @@ public class ApplicationFactory {
   /**
    * Build a new application.
    */
-  public Application build() {
+  public Application createApplication() {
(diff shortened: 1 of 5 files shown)


Yrp:
Thanks for the report!
createApplication is a bit better, we will consider modifying it.
Stream.count is just an implementation of size calculation, size is more suitable.

The pull request will be closed, open it if you have some updates.

pdhung3012:
Thank you for your help. I have reverted the getStreamCount method name to size

PR #2542, opened 2019-10-14
Odyssey v3.6.5 kafka plugin

heipacker:
**What does this PR do?**
add more info into transactionLogTrigger struct, because i want to parse internal transaction info from transactionLog when i consume message from kafka plugin

we need these info

- Unit Tests
- Manual Testing
yes, it is running on production
**Follow up**

**Extra details**


heipacker:
> 1. Please resolve the code conflicts
ok, i will
> 2. Have you added the changes of Kafka plugin?
kafka plugin no need change code

jiangyy0824:
@heipacker The address in internal transaction is in HEX string, you can convert the address format to base58  with WalletUtil.encode58Check. So this PR is not necessary.
We will support the convert in the next release of java-tron. Thanks!

PR #2583, opened 2019-10-18
Http/add data field

fbsobreira:
**What does this PR do?**
Add data (optional) field on http api transactions (transfer/transfer assets/triggersmartcontract)

We are using tronweb with http API and we want to add extra data to transactions.
Transaction DATA field is not available on HTTP API transaction builder.

- Unit Tests (Yes)
- Manual Testing (Yes)

**Follow up**
telegran: @fbsobreia

**Extra details**
We have a new feature to be release in our product that makes use of the data field.
Our MVP works already works with local build transactions but the use of tronweb/http api transaction builder will speed up release

Several other developer have been asking for this feature in discord channel.


zk19862018:
Thanks for your contribution, we will review it.v

PR #2639, opened 2019-11-11
UDP packet check to avoid udp packe attack

tomos01:
**What does this PR do?**
UDP packet check to avoid udp packe attack

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/common/net/udp/message/discover/DiscoverMessageInspector.java
+++ b/framework/src/main/java/org/tron/common/net/udp/message/discover/DiscoverMessageInspector.java
@@ -1,0 +1,82 @@
+package org.tron.common.net.udp.message.discover;
+
+import java.util.regex.Pattern;
+import org.springframework.util.StringUtils;
+import org.tron.common.net.udp.message.Message;
+import org.tron.common.overlay.discover.node.Node;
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/tron/common/net/udp/message/discover/DiscoverMessageInspector.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.regex.Pattern;
+import org.springframework.util.StringUtils;
+import org.tron.common.net.udp.message.Message;
+import org.tron.common.overlay.discover.node.Node;
+import org.tron.common.overlay.discover.table.KademliaOptions;
+
+public class DiscoverMessageInspector {
+
+  public final static Pattern PATTERN_IP = Pattern.compile("\\d+\\.\\d+\\.\\d+\\.\\d+");
```
shydesky: this is not a perfect regular match pattern

framework/src/main/java/org/tron/common/overlay/discover/table/KademliaOptions.java
```diff
@@ -1,34 +1,13 @@
-/*
```
shydesky: these GNU comments can not be removed

PR #2646, opened 2019-11-14
remove unused variables and improve style

Parachuteuk:
**What does this PR do?**
remove unused variables and improve style

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/common/overlay/discover/node/Node.java
```diff
@@ -33,19 +33,19 @@
 
   private boolean isFakeNodeId = false;
 
-  public Node(String enodeURL) {
+  public Node(String encodeURL) {
```
shydesky: should be encodeURI

PR #2683, opened 2019-11-28
remove unusefull code and optimized code

M-blockcoder:
**What does this PR do?**
Remove unuse file
Remove unuse file
local


common/src/main/java/org/tron/common/utils/CollectionUtils.java
```diff
@@ -11,22 +11,6 @@
 
 public class CollectionUtils {
 
-  public static <K, V> List<V> collectList(Collection<K> items, Function<K, V> collector) {
```
shydesky: these general methods should be reserved

PR #2707, opened 2019-12-05
move all args to common

geb789:
**What does this PR do?**
move all args  to common
move all args  to common


framework/src/main/java/org/tron/common/application/ApplicationImpl.java
```diff
@@ -35,7 +36,7 @@ public void setOptions(Args args) {
 
   @Override
   @Autowired
-  public void init(Args args) {
+  public void init(CommonParameter args) {
```
jiangyy0824: rename "args"

framework/src/main/java/org/tron/common/application/ApplicationImpl.java
```diff
@@ -46,7 +47,7 @@ public void addService(Service service) {
   }
 
   @Override
-  public void initServices(Args args) {
+  public void initServices(CommonParameter args) {
```
jiangyy0824: rename "args"

framework/src/main/java/org/tron/common/overlay/discover/DiscoverServer.java
```diff
@@ -45,7 +46,7 @@
   @Autowired
   private WireTrafficStats stats;
 
-  private Args args = Args.getInstance();
+  private CommonParameter args = Args.getInstance();
```
jiangyy0824: rename "args"

PR #2767, opened 2019-12-19
remove wallet reference from servlets. Move transaction related function from wallet to TransactionUtil.

tronalex:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


tronalex:
> pls supplement more unit test, thank you

Hey guys, there is no new function imported. And the logic is not affected also. The PR is used to move transaction related functions to TransactionUtil. Then servlets could be de-coupled from wallet. Thanks.

> pls supplement more unit test, thank you

PR #2786, opened 2019-12-20
use ChainbaseManager instead of Manager in commonly used place. Reduce the dependency of framework module. 

tronalex:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


tronalex:
> This PR can not pass all the checks. Please modify it.

The PR is ready. Please review. Thanks

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -281,16 +285,17 @@ public static String encode58Check(byte[] input) {
...
 
   public Account getAccount(Account account) {
-    AccountStore accountStore = dbManager.getAccountStore();
+    AccountStore accountStore = chainBaseManager.getAccountStore();
     AccountCapsule accountCapsule = accountStore.get(account.getAddress().toByteArray());
     if (accountCapsule == null) {
       return null;
     }
-    BandwidthProcessor processor = new BandwidthProcessor(dbManager);
+    BandwidthProcessor processor = new BandwidthProcessor(dbManager, chainBaseManager);
```
wubinTron: Why pass chain base manger to BandwidthProcessor.  You can get chain base manage from dbmanage.

framework/src/main/java/org/tron/core/db/BandwidthProcessor.java
```diff
@@ -24,10 +25,12 @@
 public class BandwidthProcessor extends ResourceProcessor {
 
   private Manager dbManager;
+  private ChainBaseManager chainBaseManager;
 
-  public BandwidthProcessor(Manager manager) {
+  public BandwidthProcessor(Manager manager, ChainBaseManager chainBaseManager) {
```
shydesky: the variable dbManager of BandwidthProcessor should be removed completely. The implementation that both Manager and ChainBaseManager reserved used as the combination inheritance is redundancy. So I advise that implementing the function getHeadSlot() in the ChainbaseManager and then removing the Manager from the  BandwidthProcessor construction function.

tronalex: > getHeadSlot

Your suggestion is right. But the BandwidthProcessor still has some dependency with Manager. Please refer to genesisBlock. Another PR is needed.

shydesky: yes, I know. So, I think you should determine the role of the Chainbase at first. If it is the container of the XXXStore, I think there can be a delegate function using the genesisBlock as an input parameter to implement the getHeadSlot().

tronalex: Currently, you can't implement getHeadSlot in ChainBaseManager, it has dependency on args. The genesisBlock should also be moved to ChainBaseManager. And the logic is quite complexed. So I plan to implement it in the next PR.

shydesky: Ok, let's make a review deeply and determine whether there can be another reasonable way.

tronalex: > Ok, let's make a review deeply and determine whether there can be another reasonable way.

Waiting for your quick reply. Thanks.

tronalex: > the variable dbManager of BandwidthProcessor should be removed completely. The implementation that both Manager and ChainBaseManager reserved used as the combination inheritance is redundancy. So I advise that implementing the function getHeadSlot() in the ChainbaseManager and then removing the Manager from the BandwidthProcessor construction function.

I have removed dbManager from BandwidthProcessor completely. Please check.

chainbase/src/main/java/org/tron/common/utils/DBConfig.java
```diff
@@ -124,4 +124,8 @@
   @Getter
   @Setter
   private static String transactionHistoreSwitch;
+
+  @Getter
+  @Setter
+  private static boolean saveInternalTx;
```
wubinTron: don't need to pass the value to dbconfig. You could directly use config from parameter

tronalex: @wubinTron Thanks for your suggestions. I have update the PR. Please check again.

PR #2796, opened 2019-12-24
reformat fork controller and move store to chainbase

helloboy12345:
What does this PR do?
move all stores to chain base and delete duplicate file
Why are these changes required?
move all stores to chain base and delete duplicate file

This PR has been tested by:

Unit Tests
Manual Testing


helloboy12345:
this diff only reformats fork controller and the main logic is the same. In addition,  it is hard to add unit test for reformating code

Issue #2848, opened 2020-01-09
Contract can be created with null bytecode


tlatkdgus1:
### 1. What did you do?
When I deployed test contract, fill empty bytecode in payload, I can expected success response

and can i attach in tronide
![image](https://user-images.githubusercontent.com/13446909/72047344-ac79ce80-32fd-11ea-9c18-e178162bf3f2.png)

here is my test contract 
[https://shasta.tronscan.org/#/contract/TCmeKAnkHjm6y9xDxLmE1u3YLXbLfo5R1E/code](url)

### 2. What did you expect to see?
meaning error

### 3. What did you see instead?
response from **/wallet/deploycontract**

`{ 
"visible":false,
"txID":"968be1b28778253a82198e658663e49c872564e4e14b79cadac8eb1997a99f28",
"contract_address":"416e9efeffc3aa63c36ca9b70a8144e8be8477afda",
"raw_data":{ 
"contract":[ 
{ 
"parameter":{ 
"value":{ 
"owner_address":"41babf033a8794cb3e46127d1ca76e669fa2d28887",
"new_contract":{ 
"origin_address":"41babf033a8794cb3e46127d1ca76e669fa2d28887",
"abi":{ 
"entrys":[ 
{ 
"outputs":[ 
{ 
"type":"uint256"
}
],
"constant":true,
"name":"blogMain_getValue",
"stateMutability":"View",
"type":"Function"
},
{ 
"inputs":[ 
{ 
"name":"_value",
"type":"uint256"
}
],
"name":"blogMain_setValue",
"stateMutability":"Nonpayable",
"type":"Function"
}
]
},
"origin_energy_limit":10000000
}
},
"type_url":"type.googleapis.com/protocol.CreateSmartContract"
},
"type":"CreateSmartContract"
}
],
"ref_block_bytes":"0cd1",
"ref_block_hash":"5d7062e8d88de899",
"expiration":1578553863000,
"fee_limit":300000000,
"timestamp":1578553805674
},
[...]

PR #2872, opened 2020-01-14
Add validate when deploy contract

tlatkdgus1:
**What does this PR do?**
Check bytecode when deploy contract

can not validate bytecode when deploy contract, so Add Exception about empty bytecode case

**Issue**
#2848

### DIFF
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -350,8 +350,13 @@ public class Wallet {
       CreateSmartContract contract = ContractCapsule
           .getSmartContractFromTransaction(trx.getInstance());
       long percent = contract.getNewContract().getConsumeUserResourcePercent();
+      byte[] data = contract.getNewContract().getBytecode().toByteArray();
+
       if (percent < 0 || percent > 100) {
         throw new ContractValidateException("percent must be >= 0 and <= 100");
+      }
+      if(ArrayUtils.isEmpty(data)){
+        throw new ContractValidateException("bytecode must have");
       }
     }
     setTransaction(trx);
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -350,9 +350,14 @@ public TransactionCapsule createTransactionCapsule(com.google.protobuf.Message m
       CreateSmartContract contract = ContractCapsule
           .getSmartContractFromTransaction(trx.getInstance());
       long percent = contract.getNewContract().getConsumeUserResourcePercent();
+      byte[] data = contract.getNewContract().getBytecode().toByteArray();
+
       if (percent < 0 || percent > 100) {
         throw new ContractValidateException("percent must be >= 0 and <= 100");
       }
+      if (ArrayUtils.isEmpty(data)) {
+        throw new ContractValidateException("bytecode must have");
```
andelf: `s/bytecode must have/bytecode must not be empty/`

tlatkdgus1: change this one

tlatkdgus1:
> Hello,The root cause of the problem is that TVM itself can accept and execute empty bytecodes and return a normal execution result. Whether this feature continues to be retained requires further evaluation. If the creation of empty bytecode contracts is prohibited, changes need to be made in TVM.
> Simple verification in Wallet is not enough,this need Hard fork in a future release.

Yes I think so too, it is nothing but a temporary measure

shydesky:
Thanks for your PR! We will fix this problem with a hard-fork in the future. We do not only add the restrictions on the client-side. So we close this PR temporarily.

PR #2877, opened 2020-01-15
move getAddress method from two servlets to util file

Parachuteuk:
**What does this PR do?**
move getAddress method from two servlets to util file

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -439,4 +441,27 @@ public static void printAccount(Account reply, HttpServletResponse response, Boo
...
+    String addressParam = "address";
+    String addressStr = request.getParameter(addressParam);
+    if (org.apache.commons.lang3.StringUtils.isBlank(addressStr)) {
+      String input = request.getReader().lines()
+              .collect(Collectors.joining(System.lineSeparator()));
+      Util.checkBodySize(input);
+      JSONObject jsonObject = JSONObject.parseObject(input);
+      addressStr = jsonObject.getString(addressParam);
+    }
+    if (org.apache.commons.lang3.StringUtils.isNotBlank(addressStr)) {
```
jiangyy0824: Why use "org.apache.commons.lang3.StringUtils" instead of "StringUtils"?

Parachuteuk: it conflicts with StringUtils already used in Util.java

jiangyy0824: you should not use org.apache.commons.lang3.StringUtils directly. Import the org.apache.commons.lang3.StringUtils. And remove the useless import files in "GetBrokerageServlet.java" and "GetRewardServlet.java"

Parachuteuk: Thanks for your advice, will do.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -372,7 +375,7 @@ public static String parseMethod(String methodSign, String input) {
     byte[] selector = new byte[4];
     System.arraycopy(Hash.sha3(methodSign.getBytes()), 0, selector, 0, 4);
     //System.out.println(methodSign + ":" + Hex.toHexString(selector));
-    if (StringUtils.isNullOrEmpty(input)) {
+    if (StringUtils.isEmpty(input)) {
```
jiangyy0824: why you use isEmpty instead of isNullOrEmpty

Parachuteuk: they provide the same implementation, isNullOrEmpty  can be replaced with isEmpty, also, no need to import org.pf4j.util.StringUtils

PR #2889, opened 2020-01-16
remove duplicate code

TracyHesiCecil:
**What does this PR do?**
remove duplicate code for dbconfig
remove duplicate code for dbconfig

- Unit Tests
- Manual Testing


chainbase/src/main/java/org/tron/common/utils/DBConfig.java
```diff
@@ -72,48 +72,4 @@
   @Getter
   @Setter
   private static long blockNumForEneryLimit;
```
jiangyy0824: Why not delete the DBConfig.java?

PR #2900, opened 2020-01-21
remove dulplicate code

geb789:
**What does this PR do?**
remove duplicate code for vm config
remove duplicate code for vm config

- Unit Tests
- Manual Testing

### DIFF
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -985,16 +985,8 @@ public class Args extends CommonParameter {
   }
 
   public static void initConfig(CommonParameter parameter) {
-    initVMConfig(parameter);
     initDBConfig(parameter);
   }
-
-  public static void initVMConfig(CommonParameter parameter) {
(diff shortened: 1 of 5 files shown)


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -985,17 +985,9 @@ public static void logConfig() {
   }
 
   public static void initConfig(CommonParameter parameter) {
-    initVMConfig(parameter);
     initDBConfig(parameter);
   }
 
-  public static void initVMConfig(CommonParameter parameter) {
-    VMConfig.setMaxTimeRatio(parameter.getMaxTimeRatio());
```
shydesky: where is this logic in the new implementation?

geb789: maxTimeRatio in vm config already delate, directly using the variable in common parameter.

shydesky: ok, I see

PR #2904, opened 2020-01-23
add TRX_SYMBOL_BYTES

linktechchi:
**What does this PR do?**
This PR tries to replace the "_".getBytes() with TRX_SYMBOL_BYTES
Using a constant to replace a function is better.
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


andelf:
You should fill at least the first two fields:

- What does this PR do?
- Why are these changes required?

BTW, it seems that you forget to change test cases. There are many `*Test.java` files that test against underscore "_".

PR #2948, opened 2020-02-19
add metrics into monitor

chengtx01:
**What does this PR do?**
add metrics into monitor

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/TronNetDelegate.java
+++ b/framework/src/main/java/org/tron/core/net/TronNetDelegate.java
@@ -48,6 +48,7 @@ import org.tron.core.net.message.BlockMessage;
 import org.tron.core.net.message.MessageTypes;
 import org.tron.core.net.message.TransactionMessage;
 import org.tron.core.net.peer.PeerConnection;
+import org.tron.core.services.monitor.MonitorMetric;
 import org.tron.core.store.WitnessScheduleStore;
 import org.tron.protos.Protocol.Inventory.InventoryType;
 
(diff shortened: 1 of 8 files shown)


framework/src/main/java/org/tron/core/net/TronNetDelegate.java
```diff
@@ -191,6 +195,7 @@ public void processBlock(BlockCapsule block) throws P2pException {
     BlockId blockId = block.getBlockId();
     synchronized (blockLock) {
       try {
```
spidemen: what if there is a exception happen?? In this case, you code will collect this process time, which mean would miss some data????

framework/src/main/java/org/tron/core/net/TronNetDelegate.java
```diff
@@ -202,10 +207,12 @@ public void processBlock(BlockCapsule block) throws P2pException {
...
           if (!backupServerStartFlag
-              && System.currentTimeMillis() - block.getTimeStamp() < BLOCK_PRODUCED_INTERVAL) {
+              && now - block.getTimeStamp() < BLOCK_PRODUCED_INTERVAL) {
             backupServerStartFlag = true;
             backupServer.initServer();
           }
+          monitorMetric.getHistogram(MonitorMetric.NET_BLOCK_LATENCY)
+                  .update(now - block.getTimeStamp());
         }
       } catch (ValidateSignatureException
```
spidemen: try add some thing inside catch since we also need the fail block process time.

PR #2960, opened 2020-02-21
Monitor

chengtx01:
**What does this PR do?**
1. set block latency info
2. set error protocol count
3. fix TPS

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/net/TronNetService.java
```diff
@@ -107,6 +111,7 @@ protected void onMessage(PeerConnection peer, TronMessage msg) {
           throw new P2pException(TypeEnum.NO_SUCH_MESSAGE, msg.getType().toString());
       }
     } catch (Exception e) {
+      monitorMetric.getCounter(MonitorMetric.NET_ERROR_PROTO_COUNT).inc();
```
spidemen: This will collect all message exceptions, like TCP, UDP.   However,  some http request would give wrong format of contract or transactionID and so on, this will not be covered  by this piece of code.

PR #2974, opened 2020-02-25
optimize metrics

chengtx01:
**What does this PR do?**
optimize metrics

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/services/filter/HttpInterceptor.java
```diff
@@ -66,28 +66,28 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
       if (EndpointCount.containsKey(endpoint)) {
         obj = EndpointCount.get(endpoint);
       } else {
-        obj.put(monitorMetric.TOTAL_REQUST, 0);
-        obj.put(monitorMetric.FAIL_REQUST, 0);
+        obj.put(metricsService.TOTAL_REQUST, 0);
```
spidemen: this is specific endpoint statics data, cannot apply metricsSever.TOTL_REQUEST since this is for total number of request, not a specific endpoint.

PR #3005, opened 2020-03-03
single core adjustment

MikumoGuynemer:
**What does this PR do?**
single core adjustment


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -543,7 +543,7 @@ public static void setParam(final String[] args, final String confFileName) {
     PARAMETER.validateSignThreadNum =
         config.hasPath(Constant.NODE_VALIDATE_SIGN_THREAD_NUM) ? config
             .getInt(Constant.NODE_VALIDATE_SIGN_THREAD_NUM)
-            : Runtime.getRuntime().availableProcessors() / 2;
+            : (Runtime.getRuntime().availableProcessors() / 2 == 0 ? 1 : Runtime.getRuntime().availableProcessors() / 2);
```
andelf: use `(Runtime.getRuntime().availableProcessors() + 1) / 2`

PR #3017, opened 2020-03-05
Metric API

spidemen:
**What does this PR do?**
1. fix naming on blockchaininfo , nodeinfo
2. fix blockprocess time
3. add sort map for meter
4. fix missTx calculation 
5. fix naming on Metric Key
6. fix interval 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/metrics/MetricsUtil.java
```diff
@@ -14,6 +14,16 @@
 
   private static MetricRegistry metricRegistry = new MetricRegistry();
 
+  private static long startTime;
```
chengtx01: A default value is recommended

PR #3021, opened 2020-03-06
move startTime to MetricsApiService

Starsakary:
**What does this PR do?**

move startTime to MetricsApiService


framework/src/main/java/org/tron/core/metrics/MetricsUtil.java
```diff
@@ -14,14 +14,8 @@
 
   private static MetricRegistry metricRegistry = new MetricRegistry();
 
-  private static long startTime;
-
   public static long getInterval() {
```
spidemen: remove this function

PR #3023, opened 2020-03-06
modify blockNum type

Starsakary:
**What does this PR do?**
modify blockNum type


framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainInfo.java
```diff
@@ -22,11 +22,11 @@
   private String failProcessBlockReason;
   private List<DupWitnessInfo> dupWitness;
 
-  public int getHeadBlockNum() {
+  public long getHeadBlockNum() {
     return headBlockNum;
   }
 
-  public void setHeadBlockNum(int headBlockNum) {
+  public void setHeadBlockNum(long headBlockNum) {
```
chengtx01: Where the function is called also needs to be modified synchronously

Starsakary: done

PR #3027, opened 2020-03-08
Feature/refine dataword

dpneko:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


actuator/src/main/java/org/tron/core/vm/VM.java
```diff
@@ -547,10 +547,11 @@ public void step(Program program) {
           }
 
           if (word1.xor(word2).isZero()) {
-            word1.and(DataWord.ZERO);
+
```
dpneko: need review

PR #3029, opened 2020-03-10
remove duplicate code

Parachuteuk:
**What does this PR do?**
remove duplicate code
to improve the code 
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -1612,7 +1612,7 @@ public class Wallet {
     }
 
     // output
-    ShieldedTransactionOutput(shieldedReceives, builder, ovk);
+    shieldedOutput(shieldedReceives, builder, ovk);
 
     TransactionCapsule transactionCapsule = null;
     try {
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -1740,6 +1722,21 @@ public TransactionCapsule createShieldedTransactionWithoutSpendAuthSig(
     return transactionCapsule;
 
   }
+  
+  private void ShieldedTransactionOutput (List<ReceiveNote> shieldedReceives,
```
andelf: Class method naming style matters.
should be a verbal phrase. and should be in `lowerCamelCase`

framework/src/main/java/org/tron/core/db/PendingManager.java
```diff
@@ -31,29 +31,25 @@ public PendingManager(Manager db) {
...
-      } catch (InterruptedException e) {
-        logger.error(e.getMessage());
-        Thread.currentThread().interrupt();
-      }
+      TxIteration(tx);
     }
     dbManager.getPoppedTransactions().clear();
   }
+  
+  private void TxIteration(TransactionCapsule tx) {
```
andelf: ditto

framework/src/main/java/org/tron/core/Wallet.java
Parachuteuk: done, thanks

framework/src/main/java/org/tron/core/db/PendingManager.java
Parachuteuk: done

PR #3034, opened 2020-03-11
change api "listnode" "nodeinfo" "metrics" path from "wallet" to "monitor"

Starsakary:
**What does this PR do?**

change api "listnode" "nodeinfo" "metrics" path from "wallet" to "monitor"

### DIFF
--- a/framework/src/main/java/org/tron/core/services/http/solidity/SolidityNodeHttpApiService.java
+++ b/framework/src/main/java/org/tron/core/services/http/solidity/SolidityNodeHttpApiService.java
@@ -184,7 +184,7 @@ public class SolidityNodeHttpApiService implements Service {
       context.addServlet(new ServletHolder(triggerConstantContractServlet),
           "/walletsolidity/triggerconstantcontract");
 
-      context.addServlet(new ServletHolder(getNodeInfoServlet), "/wallet/getnodeinfo");
+      context.addServlet(new ServletHolder(getNodeInfoServlet), "/monitor/getnodeinfo");
       context.addServlet(new ServletHolder(getBrokerageServlet), "/walletsolidity/getBrokerage");
       context.addServlet(new ServletHolder(getRewardServlet), "/walletsolidity/getReward");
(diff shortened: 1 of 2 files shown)


chengtx01:
Old interfaces need to be compatible

framework/src/main/java/org/tron/core/services/http/solidity/SolidityNodeHttpApiService.java
```diff
@@ -184,7 +184,7 @@ public void start() {
       context.addServlet(new ServletHolder(triggerConstantContractServlet),
           "/walletsolidity/triggerconstantcontract");
 
-      context.addServlet(new ServletHolder(getNodeInfoServlet), "/wallet/getnodeinfo");
+      context.addServlet(new ServletHolder(getNodeInfoServlet), "/monitor/getnodeinfo");
```
andelf: To be compared with the scope name `wallet`.
`getnodeinfo` may be under the scope of `node`, `system`, `network` or `status`.

> monitor: *n.* An instrument or device used for observing, checking, or keeping a continuous record of a process or quantity.

Starsakary: We are changing some apis to a new namespace, this is only a temporary name

framework/src/main/java/org/tron/core/services/http/solidity/SolidityNodeHttpApiService.java
```diff
@@ -184,9 +184,11 @@ public void start() {
       context.addServlet(new ServletHolder(triggerConstantContractServlet),
           "/walletsolidity/triggerconstantcontract");
 
-      context.addServlet(new ServletHolder(getNodeInfoServlet), "/wallet/getnodeinfo");
+      context.addServlet(new ServletHolder(getNodeInfoServlet), "/walletsolidity/getnodeinfo");
```
chengtx01: Whether it is necessary to change ‘wallet’ to ‘walletsolidity’ needs to be considered again

PR #3107, opened 2020-04-22
Optimize solid http interface 

kaiserlinda:
**What does this PR do?**
optimize solid http interface 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/services/http/solidity/GetTransactionByIdSolidityServlet.java
```diff
@@ -69,4 +58,14 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
     }
   }
 
+  private void fillResponse(ByteString txId, boolean visible, HttpServletResponse response)
```
andelf, line 61: `fill` means that information is at hand, and one should transcribe the detail into another (empty/uninitialized) object.

So, if you use the name `fill`, the function's signature would better be

```java
private void fillResponse(Transaction txn, boolean visible, HttpServletResponse response)
```

BTW, the core logic is `getTransactionById`, which shouldn't hide behind a `fillResponse`.

PR #3114, opened 2020-04-26
remove KademliaOptions from common module

linktechchi:
**What does this PR do?**
remove KademliaOptions from common module
too many useless variables in KademliaOptions
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


common/src/main/java/org/tron/common/overlay/discover/node/Node.java
```diff
@@ -57,7 +56,7 @@ public static Node instanceOf(String hostPort) {
 
   public static byte[] getNodeId() {
     Random gen = new Random();
-    byte[] id = new byte[KademliaOptions.NODE_ID_LEN];
+    byte[] id = new byte[64];
```
andelf, line 59: make it a local const with meaningful name.

PR #3144, opened 2020-05-11
Optimization: remove sortWitness() method

guoquanwu:
**What does this PR do?**

Optimization: remove sortWitness() method

sortWitness() takes about 7-8% percent CPU time of the whole thread, it is invoked by payReward() when processing the block, but the result of sortWitness() only changed when the maintenance is executed which interval is 6 hour, so store the result of sortWitness() in a list and update the list when maintenance happened.

- Unit Tests
- Manual Testing


guoquanwu:
to be merged into branch ->release_tronz_4.0, close

PR #3292, opened 2020-07-17
protocol: Fix spelling

nguyenhuedang:
**What does this PR do?**
- Fix spelling in protos/api

- To correct meaning of return field

- Not yet


mattopolitan:
I'm afraid Tron will not accept this pull request, aside from it fails all the tests:
1. constant_result is used in TriggerConstantContract. The result is said to be constant as any actions made by triggerconstantcontract() won't edit any the on-chain data. 
2. This fix requires a lot of changes on other parties' applications as well, e.g. tronscan, tronweb, trongrid, etc. It may crash a lot of potential services. The cost is way too high for a single fix spelling.

nguyenhuedang:
Got it, thank you @mattopolitan. This is a misunderstanding from my side.

PR #3294, opened 2020-07-20
[for test]Cross chain dev slot

guoquanwu:

### DIFF
--- a/protocol/src/main/protos/core/contract/balance_contract.proto
+++ b/protocol/src/main/protos/core/contract/balance_contract.proto
@@ -55,10 +55,19 @@ message CrossToken {
   bytes chain_id = 5;
 }
 
+message CrossChainInfo {
+  bytes owner_address = 1;
+  bytes chain_id = 2;
+  bytes sr_list = 3;
+  int64 begin_sync_height = 4;
+  int64 maintenance_time_interval = 5;
+  bytes parent_block_hash = 6;
+}
+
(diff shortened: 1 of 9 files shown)


protocol/src/main/protos/core/contract/balance_contract.proto
```diff
@@ -55,10 +55,19 @@ message CrossToken {
...
+  bytes parent_block_hash = 6;
+}
+
 message RegisterCrossContract {
-  bytes chain_id = 1;
-  bytes sr_list = 2;
-  int64 begin_sync_height = 3;
-  int64 maintenance_time_interval = 4;
-  bytes parent_block_hash = 5;
+  CrossChainInfo cross_chain_info = 1;
```
andelf, line 68: Currently, all `**Contract` has almost the same layout.

e.g. The first field is always `owner_address`. 

Embedding a `CrossChainInfo` breaks this rule.

protocol/src/main/protos/core/contract/balance_contract.proto
```diff
@@ -55,10 +55,19 @@ message CrossToken {
...
+}
+
+message UpdateCrossContract {
+  CrossChainInfo cross_chain_info = 1;
```
andelf, line 72: Is every field of `CrossChainInfo` changable?
If not, only list ones that are editable.
And do not use a `CrossChainInfo`

PR #3303, opened 2020-07-24
Pbft 4.1 dev

lvs007:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


chainbase/src/main/java/org/tron/core/net/message/MessageTypes.java
```diff
@@ -33,6 +33,10 @@
 
   TRX_INVENTORY(0x13),
 
+  PBFT_BLOCK_MSG(0x14),
```
KI5FPL, line 36: You'd better start a new message type scope. Start your `PBFT_` message type from 0x40.

PR #3384, opened 2020-09-03
Release v4.1

Yrp:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


actuator/src/main/java/org/tron/core/utils/ProposalUtil.java
```diff
@@ -317,8 +317,19 @@ public static void validator(DynamicPropertiesStore dynamicPropertiesStore,
         }
         break;
       }
+      case ALLOW_TVM_ISTANBUL: {
```
andelf, line 320: wrong order

PR #3424, opened 2020-09-26
Add /wallet/getblockinfobylimitnext API

kevholder:
**What does this PR do?**

Add a `/wallet/getblockinfobylimitnext` API which is like
`/wallet/getblockbylimitnext` but returns transaction info list instead
of transaction list.

Minimizes the number of API requests required when scanning transaction logs using the HTTP API.

### DIFF
--- a/protocol/src/main/protos/api/api.proto
+++ b/protocol/src/main/protos/api/api.proto
@@ -404,6 +404,15 @@ service Wallet {
   //Use this function instead of GetBlockByLimitNext.
   rpc GetBlockByLimitNext2 (BlockLimit) returns (BlockListExtention) {
   }
+  rpc GetBlockInfoByLimitNext (BlockLimit) returns (BlockInfoList) {
+    option (google.api.http) = {
+      post: "/wallet/getblockinfobylimitnext"
+      body: "*"
+      additional_bindings {
+        get: "/wallet/getblockinfobylimitnext"
+      }
+    };
+  }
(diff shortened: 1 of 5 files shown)


protocol/src/main/protos/api/api.proto
```diff
@@ -1189,10 +1198,20 @@ message BlockExtention {
   bytes blockid = 3;
 }
 
+message BlockInfo {
```
shydesky, line 1201: These changes of the protocol need to be synchronized here：  https://github.com/tronprotocol/protocol

kevholder: I'm not sure how to do it :/

shydesky: > I'm not sure how to do it :/
Modify the proto file in the corresponding position of https://github.com/tronprotocol/protocol and submit a PR in the protocol repo.

Tizzit-max: I got this my bad

Tizzit-max (approved):
I got it I'm so forgetful

Tizzit-max:
Pull_Request_Target #3424

Benson0224:
Thanks for your contribution to java-tron, we will close this PR Since no update for a long time.

Please re-open this PR after the corresponding change was done in https://github.com/tronprotocol/protocol, and please do link the related change in this PR. thanks.

PR #3472, opened 2020-10-27
fix dataword

taihaofu:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/program/Program.java
+++ b/actuator/src/main/java/org/tron/core/vm/program/Program.java
@@ -1178,7 +1178,7 @@ public class Program {
       BlockCapsule blockCapsule = contractState.getBlockByNum(index);
 
       if (Objects.nonNull(blockCapsule)) {
-        return new DataWord(blockCapsule.getBlockId().getBytes());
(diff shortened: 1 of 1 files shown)


actuator/src/main/java/org/tron/core/vm/program/Program.java
```diff
@@ -1178,7 +1178,7 @@ public DataWord getBlockHash(int index) {
       BlockCapsule blockCapsule = contractState.getBlockByNum(index);
 
       if (Objects.nonNull(blockCapsule)) {
-        return new DataWord(blockCapsule.getBlockId().getBytes());
+        return new DataWord(blockCapsule.getBlockId().getBytes()).clone();
```
KI5FPL, line 1181: The constructor of `DataWord` should be changed, instead of any caller position.

A `DataWord` must not point to any non-owned memory.

zhangheng1536 (approved):
LGTM

actuator/src/main/java/org/tron/core/vm/program/Program.java
inmortalbobz: Annyminous hackers love it

PR #3523, opened 2020-11-26
V4.1.2

Lredhdx:
support solidity update to 0.6.0
add 47th proposal to change MAX_FEE_LIMIT

### DIFF
--- a/actuator/src/main/java/org/tron/core/utils/ProposalUtil.java
+++ b/actuator/src/main/java/org/tron/core/utils/ProposalUtil.java
@@ -401,6 +401,16 @@ public class ProposalUtil {
         }
         break;
       }
+      case MAX_FEE_LIMIT: {
+        if (!forkController.pass(ForkBlockVersionEnum.VERSION_4_1_2)) {
+          throw new ContractValidateException("Bad chain parameter id [MAX_FEE_LIMIT]");
+        }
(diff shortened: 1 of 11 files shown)


actuator/src/main/java/org/tron/core/utils/ProposalUtil.java
```diff
@@ -401,8 +401,18 @@ public static void validator(DynamicPropertiesStore dynamicPropertiesStore,
         }
         break;
       }
+      case MAX_FEE_LIMIT: {
+        if (!forkController.pass(ForkBlockVersionEnum.VERSION_4_1_2)) {
+          throw new ContractValidateException("Bad chain parameter id [MAX_FEE_LIMIT]");
+        }
+        if (value < 0 || value > 10_000_000_000L) {
+          throw new ContractValidateException(
+              "Bad MAX_FEE_LIMIT parameter value, valid range is [0,10_000_000_000L]");
```
andelf, line 410: The maximum value is equal to the current setting,
which is meaningless.

andelf (dismissed):

andelf:
BTW, the timeout mechanism should be changed too.
The more fee_limit(energy_limit), the longer the time limit.
Or else this change is meaningless. 
The TVM is slow while doing internal transactions. A larger amount of internal calls will timeout the transaction.

taihaofu (approved):

PR #3538, opened 2020-12-03
Support account balance

Yrp:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


guoquanwu (approved):

protocol/src/main/protos/core/contract/balance_contract.proto
```diff
@@ -33,4 +33,50 @@ message TransferContract {
...
+  bytes transaction_identifier = 1;
+  repeated Operation operation = 2;
+  string type = 3;
+  string status = 4;
+}
+
+
+
+message BlockBalanceTrace {
+  message BlockIdentifier {
```
andelf, line 54: should reuse BlockId

tronfamily:
To the point:
wallet/getaccountbalance
Description： Get the account balance in a specific block. 

**Why is it necessary to query the block hash as a 2nd argument?**

In practice, this means that this block hash must first be determined by wallet/getblockbynum .  This would be an unnecessary additional query.

PR #3545, opened 2020-12-04
add create3 to tvm

dpneko:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/OpCode.java
+++ b/actuator/src/main/java/org/tron/core/vm/OpCode.java
@@ -621,6 +621,15 @@ public enum OpCode {
   CREATE2(0xf5, 4, 1, OpCode.Tier.SpecialTier),
 
   /**
+   * (0xf6) CREATE3, same as CREATE2. It will compute the address from the caller address of the
+   * creating contract instead of the address of the creating contract. The reason of creating this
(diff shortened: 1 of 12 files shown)


actuator/src/main/java/org/tron/core/vm/OpCode.java
```diff
@@ -620,6 +620,15 @@
...
   CREATE2(0xf5, 4, 1, OpCode.Tier.SpecialTier),
 
+  /**
+   * (0xf6) CREATE3, same as CREATE2. It will compute the address from the caller address of the
+   * creating contract instead of the address of the creating contract. The reason of creating this
+   * op is that the change of the address computation method of CREATE2 in version 4.1.0 may cause
+   * loss of assets. In order to avoid this situation, CREATE3 uses the old way to compute the
+   * address. So that users can deploy the contract to the right address which seems not right now.
+   */
+  CREATE3(0xf6, 4, 1, Tier.SpecialTier),
```
andelf, line 630: Since this is a TRON only extension, the opcode should not be 0xf6. Which is very likely to be occupied by a future version of EVM.
Use the slot after the 0xd0-0xd4 range. TRON's "inventions" are already there.

andelf (changes requested):

PR #3658, opened 2021-03-01
account asset optimization

forfreeday:
### 1. What did you do?
Optimize your Account account. Create a new store of account assets.

Compare with [release_4.1.2](https://github.com/tronprotocol/java-tron/releases/tag/GreatVoyage-v4.1.2) and  [optimized version](https://github.com/tronprotocol/java-tron/tree/feature/account_asset_optimization) 

The serialization speed is optimized on the original side according to the Account account volume, and as much as possible, other than the necessary fields of the account, such as address, balance, accountName and other related fields are kept, other business fields that can be taken out separately are extracted.

This time, the Asset asset-related storage is optimized, stripped from the Account and moved to the new Asset-related storage.

### 2. What did you expect to see?
Account transfer speed further optimized
**1. Serialization speed**
Account reduces Asset-related fields during serialization, and after serialization fields are performed, a performance comparison of 1 million serializations is performed:

**Results of comparing the results of 4 tests**
|    version   |  improvements  |
| ---- | ---- |
|   1      |  14.77%  |
|    2    |  37.31%  |
|    3    |  28.69%  |
|    4     |  23.15%  |

**2.Transaction processing speed**
Account deserialization 1 million times compared, the optimized version will be slightly faster in deserialization about 16%

**Results of comparing the results of 4 tests**
|    version |  improvements  |
| ---- | ---- |
|    1    |  28.75% |
|    2    |  39.35% |
|    3   |  37.92%  |
|    4   |  30.04%  |

**3.Transaction processing speed**
There is a significant increase in average speed when comparing both transaction and block processing speeds

|    version  |  txs/ms   |  block/ms    |  
| ---- | ---- | ---- |
|    optimized   |  23.26%| 33.91% |

### 3. What did you see instead?
Account transfer speeds have improved, as have transaction and block packing speeds
Based on the above comparison, it can be determined that the version that extracts the fields will be faster in the serialization operation.


andelf:
Is this "optimization" records all TRC10 token balances under a single K-V indexed by token_name?

You might read and write large key-value pairs over and over again.

Have you ever tested the "actual performance gain" while some large TRC10 token doing airdrop, like BTT?

RonSherfey (approved):
Approved changes

PR #3761, opened 2021-04-23
style: remove unused field and method

yanghang8612:


Lredhdx (approved):
remove unused field and method

PR #3827, opened 2021-05-31
optimization processBlock tps

forfreeday:
**What does this PR do?**
Optimize processBlock processing speed
Optimize payReward reward processing speed

By locating the block processing time, we found that the payReward() method takes more time to process the block, so we can speed up the block by reducing unnecessary operations


xq-lu (approved):
LGTM

PR #3927, opened 2021-07-16
Test case adapt http  default return

wangzichichi:
**What does this PR do?**
Test case adapt http  default return
Daily build task should be passed 100%
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


wangqqqqq (approved):
LGTM

PR #4047, opened 2021-09-09
fix case to adapt java-tron v4.4.0(call can use 63/64 left energy)

wangzichichi:
**What does this PR do?**
adapt case to java-tron v4.4.0

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


wangqqqqq (approved):
LGTM

PR #4149, opened 2021-11-20
Optimize the heap usage by making the Jackson ObjectMapper a static m…

WoonchanCho:
…ember

**What does this PR do?**
This makes the instance of ObjectMapper static member of org.tron.common.utils.JsonUtil and reuse the object rather than instantiating one whenever the class's static method is called.

org.tron.common.utils.JsonUtil has two static methods that transform the string JSON representation to a Java object vice versa.
The two methods instantiate the ObjectMapper object whenever they are called.
The problem is that this instantiation operation is expensive and the best practice is to reuse the ObjectMapper object in the multiple operations.

As the doc explains, this object is thread-safe as long as the configuration is not changed.

Please refer to the javadoc for the detailed explanation: https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/ObjectMapper.html

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/common/src/main/java/org/tron/common/utils/JsonUtil.java
+++ b/common/src/main/java/org/tron/common/utils/JsonUtil.java
@@ -4,12 +4,12 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.util.StringUtils;
 
 public class JsonUtil {
+  private static final ObjectMapper objectMapper = new ObjectMapper();
 
   public static final <T> T json2Obj(String jsonString, Class<T> clazz) {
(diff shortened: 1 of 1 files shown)


WoonchanCho:
Please see the sample performance comparison between the existing strategy and the new strategy: https://github.com/WoonchanCho/ObjectMapperPerfTest.
In my local test (Mac) the new strategy was much performant (28 times) than the existing.

halibobo1205:
Hi, @WoonchanCho , `ObjectReader` and `ObjectWriter` may be a better choice,  it is recommended to try. This PR closed for now

PR #4168, opened 2021-12-03
fix default branch when build with docker

guoquanwu:
**What does this PR do?**

fix default branch when build with docker

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN set -o errexit -o nounset \
     && cd $TMP_DIR \
     && git clone https://github.com/tronprotocol/java-tron.git \
     && cd java-tron \
-    && git checkout docker \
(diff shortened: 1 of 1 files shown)


yagop:
Why not using the local code? That allows to easily build any version with Docker, not only the master.

```bash
git clone https://github.com/tronprotocol/java-tron.git
cd java-tron
docker build .
```

guoquanwu:
> Why not using the local code? That allows to easily build any version with Docker, not only the master.
> 
> ```shell
> git clone https://github.com/tronprotocol/java-tron.git
> cd java-tron
> docker build .
> ```

It's just a template, people can change it when building. The building is automatic with only one command and the master branch is also the stable&latest branch.

Issue #4371, opened 2022-05-13
Git ssh in start.sh


aghamir:
### 1. What did you do?
Run `./start.sh --release --run`

### 2. What did you expect to see?
Run the fullnode with built jar file

### 3. What did you see instead?
Error because git@github.com: Permission denied (publickey).

It seems the script write with git ssh instead of https. Therefore, other people who doesn't add the ssh key to this repo cannot use this script

forfreeday:
HTTPS support can be added

ethan1844:
Thanks for your contribution to java-tron, this issue will be closed as no update for a long time.

Please feel free to re-open it if you still see the issue, thanks.

PR #4372, opened 2022-05-13
Fix git url to be access by all user

aghamir:
**What does this PR do?**
Fix git url use https one instead of ssh one
All user can run start.sh
- Manual Testing

**Follow up**
#4371


aghamir:
This fixed in ba1734bc2.

PR #4415, opened 2022-05-30
Fix: change tobin fee type from int64 to string

guoquanwu:
**What does this PR do?**
change tobin fee type from int64 to string

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


guoquanwu:
The `System.out` will be deleted in the future, just for the inner test

PR #4602, opened 2022-08-22
feat(db): remove method getValuesPrev

halibobo1205:
**What does this PR do?**
    1. this method is not used, and is wrong in some case
    2. re-add correct method if it may be used in the future

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


guoquanwu (approved):
LGTM

PR #4733, opened 2022-10-26
refactor(db): remove useless revokingDB

lurais:
**What does this PR do?**

To remove useless revokingDB

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


chainbase/src/main/java/org/tron/core/db/TronStoreWithRevoking.java
```diff
@@ -56,10 +54,7 @@
   protected TronStoreWithRevoking(String dbName) {
     int dbVersion = CommonParameter.getInstance().getStorage().getDbVersion();
     String dbEngine = CommonParameter.getInstance().getStorage().getDbEngine();
-    if (dbVersion == 1) {
-      this.revokingDB = new RevokingDBWithCachingOldValue(dbName,
-          getOptionsByDbNameForLevelDB(dbName));
-    } else if (dbVersion == 2) {
+    if (dbVersion == 2) {
```
halibobo1205, line 57: If set dbVersion = 1 ? Is it still necessary to check the version?

guoquanwu, line 57: should reserve the check for v1, and throw an exception when people try to use v1 to start the node?

framework/src/main/java/org/tron/core/config/DefaultConfig.java
```diff
@@ -48,9 +47,7 @@ public RevokingDatabase revokingDatabase() {
     int dbVersion = Args.getInstance().getStorage().getDbVersion();
     RevokingDatabase revokingDatabase;
     try {
-      if (dbVersion == 1) {
-        revokingDatabase = RevokingStore.getInstance();
-      } else if (dbVersion == 2) {
+      if (dbVersion == 2) {
```
guoquanwu, line 50: same with above

chainbase/src/main/java/org/tron/core/db/TronStoreWithRevoking.java
lurais: It will throw a runtime exception when dbVersion set to 1 as the else logic will take effect.

lurais: It will throw a runtime exception when dbVersion set to 1 as the else logic will take effect.

framework/src/main/java/org/tron/core/config/DefaultConfig.java
lurais: It will throw a runtime exception when dbVersion set to 1 as the else logic will take effect.

PR #4734, opened 2022-10-26
refactor(db): optimize db lock

lurais:
**What does this PR do?**
To optimize the way of db lock

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


chainbase/src/main/java/org/tron/common/storage/leveldb/LevelDbDataSourceImpl.java
```diff
@@ -257,14 +279,17 @@ public Set<byte[]> allKeys() {
   @Override
   public Set<byte[]> allValues() {
     resetDbLock.readLock().lock();
-    try (DBIterator iterator = getDBIterator()) {
-      Set<byte[]> result = Sets.newHashSet();
-      for (iterator.seekToFirst(); iterator.hasNext(); iterator.next()) {
-        result.add(iterator.peekNext().getValue());
+    try {
+      checkDbAlive();
+      try (DBIterator iterator = getDBIterator()) {
```
halibobo1205, line 284: move checkDbAlive into getDBIterator()?

chainbase/src/main/java/org/tron/common/storage/rocksdb/RocksDbDataSourceImpl.java
```diff
@@ -102,37 +106,36 @@ public void closeDB() {
...
     resetDbLock.readLock().lock();
-    Set<byte[]> result = Sets.newHashSet();
-    try (final RocksIterator iter = getRocksIterator()) {
-      for (iter.seekToFirst(); iter.isValid(); iter.next()) {
-        result.add(iter.key());
+    try {
+      checkDbAlive();
+      Set<byte[]> result = Sets.newHashSet();
+      try (final RocksIterator iter = getRocksIterator()) {
+        for (iter.seekToFirst(); iter.isValid(); iter.next()) {
```
halibobo1205, line 134: same to leveldb

halibobo1205 (changes requested):

chainbase/src/main/java/org/tron/common/storage/leveldb/LevelDbDataSourceImpl.java
lurais: check db shall be before any db operations and just need check once actually for one public method ,if put checkdb operation in a private method , the times of checkDb is not sure

halibobo1205: getDBIterator() is not a db operation?

lurais: getDBIterator is a db operation and before this method should checkdb but not in this method as we will not use this private method directly

PR #4779, opened 2022-11-10
feat(freezeV2): fix bug, add query api to get delegate max resource size

chaosuper12321:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


actuator/src/main/java/org/tron/core/actuator/DelegateResourceActuator.java
```diff
@@ -138,12 +139,18 @@ public boolean validate() throws ContractValidateException {
       throw new ContractValidateException("delegateBalance must be more than 1TRX");
     }
 
+    boolean isTransactionCreate = this.getTx().isTransactionCreate();
```
chaosuper12321, line 142: 没必要创建

PR #4811, opened 2022-11-23
fix(freezeV2): fix some problems

lxcmyf:
**What does this PR do?**
1. fix deflation or expansion problems caused by precision
2. deduct resource of old freezing first when calculate usage

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


chainbase/src/main/java/org/tron/core/capsule/AccountCapsule.java
```diff
@@ -1255,9 +1299,12 @@ public void updateAccountType(AccountType accountType) {
   public void clearDelegatedResource() {
     Builder builder = account.toBuilder();
     AccountResource newAccountResource = getAccountResource().toBuilder()
-        .setAcquiredDelegatedFrozenBalanceForEnergy(0L).build();
+        .setAcquiredDelegatedFrozenBalanceForEnergy(0L)
```
lxcmyf, line 1302: TODO

PR #4884, opened 2022-12-29
feat(DynamicEnergy): optimize the return result of estimate energy

daniel-cao-byte:
Optimize the return result of estimate energy APIs


zhangheng1536 (approved):
LGTM

PR #5096, opened 2023-03-28
feat(api): remove insecure APIs

halibobo1205:
**What does this PR do?**
     remove some APIs for security concerns
    possible exposure of sensitive information
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


forfreeday (approved):

forfreeday (approved):

guoquanwu (approved):

Crypto2:
Why have these been removed? They are widely used and going to break a ton of integrations.

StevenProhashing:
> Why have these been removed? They are widely used and going to break a ton of integrations. It's the job of the network/firewall configuration to prevent any leaking not the coin node.

Agreed. We upgraded to the latest Tron version until we discovered these changes break our integrations👎. We use the API on our local network.

Crypto2:
Using one of the SDKs or rolling your own, we used them because back in the day there was no SDK out. To get the key from your password it is just a sha256 hash of your password I believe.

FrankWooster:
Are you familiar with the concept of backward compatibility? Apparently not. 
Great approach. 
👎

ExOCeo:
@Crypto2, it seems all the removed are about address and PK. There used to be some warning in TRON docs for the APIs, it's not safe to get such info transmitted through the network indeed. So now, we have to use the SDK. It is impractical for normal developers to build a node, configure a firewall and DMZ to make a safe box.

ExOCeo:
@Crypto2, it seems all the removed are about address and PK. There used to be some warnings in TRON docs for the APIs, it's not safe to get such transmitted through the network indeed. It is impractical for normal developers to build a node, configure a firewall and DMZ to make a safe box. So now, we have to use the SDK.

Crypto2:
That takes all of 15 minutes and a DMZ wouldn't be involved in the first place.

ExOCeo:
@Crypto2, it is not just a matter of time, check how many users are rolling with third-party providers, there are many reasons to not maintain a node while acquiring the service to access TRON. The integrations are impacted, and the notice to adapt to it is rather short. TRON should make a much earlier advance notice before eliminating those APIs.

FrankWooster:
> That takes all of 15 minutes and a DMZ wouldn't be involved in the first place and they would have to install and configure the node either way. Also the integrations most people are talking about probably were built years ago when Tron came out and that's all there was to use unless you were using one of the like 1-2 languages with an SDK.

The TRON team needs to work seriously on the key storage and security architecture. And not delete methods. DMZ closes these problems with the proper alignment of the DMZ.

PR #5117, opened 2023-04-07
feature(test): use random port in unit test

guoquanwu:
**What does this PR do?**

Using random port in unit test.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/test/java/org/tron/common/utils/PublicMethod.java
+++ b/framework/src/test/java/org/tron/common/utils/PublicMethod.java
@@ -1,7 +1,11 @@
 package org.tron.common.utils;
 
 import com.google.protobuf.ByteString;
+
+import java.io.IOException;
 import java.math.BigInteger;
+import java.net.InetAddress;
+import java.net.Socket;
 import java.util.Random;
 
 import org.tron.api.GrpcAPI;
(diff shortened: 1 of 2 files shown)


framework/src/test/java/org/tron/common/utils/PublicMethod.java
```diff
@@ -105,11 +109,30 @@ public static GrpcAPI.Return broadcastTransaction(
...
+
   public static int chooseRandomPort(int min, int max) {
-    return new Random().nextInt(max - min + 1) + min;
+    int port = new Random().nextInt(max - min + 1) + min;
+    try {
+      while (!checkPortAvailable(port)) {
+        port = new Random().nextInt(max - min + 1) + min;
+      }
+    } catch (IOException e) {
+      return new Random().nextInt(max - min + 1) + min;
```
lurais, line 123: Why should get a new random number when an IOException catched? Does it means the port is available if an IOException catched?

317787106 (approved):

framework/src/test/java/org/tron/common/utils/PublicMethod.java
guoquanwu: This method just picks a random port, if the check logic can not return a value, we should return an alternative port.

lurais: So shall we check the alternative port again?

guoquanwu: I think it will be too complicated, if the probability of picking a useless port is 1%, double check will decrease the probability to 0.01%, most of time it will work well

lurais (commented):
ok

PR #5193, opened 2023-05-04
made nix shell for old java/gradle

dzmitry-lahoda:
**What does this PR do?**

allows anybody build tron

it is hard to get old gradle/java other way

- Unit Tests
- Manual Testing +

**Follow up**

more nix for running tron

**Extra details**

nix is standard on reproducible deterministic software

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,15 @@
+{}:
+let
+  pkgs = import (fetchTarball https://github.com/NixOS/nixpkgs/archive/619d52a37ef190017a71b26469ec434f1861c20e.tar.gz)
+    { };
+in
+pkgs.mkShell {
+  buildInputs = [
+    pkgs.jdk8
+    pkgs.gradle_4
+    pkgs.grpcurl
+  ];
+  shellHook = ''
+    export NIX_ENV=dev
+  '';
+}


halibobo1205:
Hi, @dzmitry-lahoda , can you provide more details on when the old Java or Gradle is needed?

dzmitry-lahoda:
do you one other one liner to setup tron build env? or even how to download gradle which is not released anymore(not even downloadable btw).  

nix-shell does that.

halibobo1205:
> latest gradle and java fail to compile tron. both should be of some earlier versions.

Do you mean that [gradle 5.6.4](https://github.com/tronprotocol/java-tron/blob/develop/gradle/wrapper/gradle-wrapper.properties#L6) is no longer available for download? By the way, java-tron only supports jdk8 for now.

dzmitry-lahoda:
to enter this shell use `nix-shell`.  to be honest have no idea why i was not able to setup build for tron using official guide or latest nixes.

forfreeday:
@dzmitry-lahoda 
Have you used the `. /gradlew` to compile?
Use the `./gradlew clean build` command, you can build java-tron.
It is not necessary to bring in the `nix-shell`, and you can show the problems that arise from using the gradlew build

PR #5282, opened 2023-06-19
fix: add retry for batch validate units

yanghang8612:


yanghang8612:
Since the unit tests will retry up to 3 times during execution, this fix is no longer needed.

PR #5366, opened 2023-07-20
feat(log): optimize log print

lurais:
**What does this PR do?**
Optimize log print in Wallet.java, print the stack trace when an unknown exception be catched

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -2058,7 +2058,7 @@ public class Wallet {
             .parseFrom(chainBaseManager.getMerkleTreeIndexStore().get(blockNum));
       }
     } catch (Exception ex) {
-      logger.error(ex.getMessage());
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -4304,7 +4304,7 @@ public long getEnergyFee(long timestamp) {
 
       return energyFee;
     } catch (Exception e) {
-      logger.error("getEnergyFee timestamp={} failed, error is {}", timestamp, e.getMessage());
+      logger.error("getEnergyFee timestamp={} failed, error is {}", timestamp, e.getMessage(), e);
```
halibobo1205, line 4307: duplicate?

lurais: Add the e will print out the stack trace, and the e.getMessage will not.Actually, the e.getMessage is for log analyze and stack trace will print in the next line

halibobo1205 (approved):
LGTM

PR #5367, opened 2023-07-20
feat(Servlet): optimize the http request interface

forfreeday:
**What does this PR do?**
1. Optimized Util getAddress method
2. add `GetBrokerageServletTest` test case
3. add `GetRewardServletTest` test case

Fix incorrect interface returns, add test cases

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -499,6 +500,12 @@ public static byte[] getAddress(HttpServletRequest request) throws Exception {
     byte[] address = null;
     String addressParam = "address";
     String addressStr = request.getParameter(addressParam);
+
+    if (APPLICATION_FORM_URLENCODED.equals(request.getContentType())
```
halibobo1205, line 504: It's not clear. Is this the only scene?

forfreeday:
@halibobo1205 

With the addition of parameter checks, if the uploaded parameter is an illegal parameter, an error will be returned.

request:
```
curl --location 'http://127.0.0.1:8090/wallet/getReward' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'address='
```

expected Result:
```
{
    "Error": "INVALID address, invalid request parameter"
}
```

Otherwise, normal parameter requests for `JSON`, `GET`, and `POST` are not affected.

framework/src/main/java/org/tron/core/services/http/Util.java
halibobo1205: It is recommended to make a clear distinction between GET, POST (application/json, application/x-www-form-urlencoded)

forfreeday: There is no need to add other types of content types, and these types are supported even if other types of judgments are not explicitly added.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -518,6 +518,57 @@ public static byte[] getAddress(HttpServletRequest request) throws Exception {
...
   }
 
+  private static String checkGetParam(HttpServletRequest request, String key) throws Exception {
+    String method = request.getMethod();
+    String value;
+
+    if (HttpMethod.GET.toString().toUpperCase() .equalsIgnoreCase(method)) {
+      value = request.getParameter(key);
+      if (StringUtils.isBlank(value)) {
+        throw new IllegalArgumentException("Invalid request parameter");
```
halibobo1205, line 528: Need to keep compatibility

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -518,6 +518,57 @@ public static byte[] getAddress(HttpServletRequest request) throws Exception {
...
+        }
+      } else if (APPLICATION_FORM_URLENCODED.toLowerCase().contains(contentType)) {
+        value = getValue(request);
+        String[] keyValue = value.split("=");
+        if (keyValue.length <= 1) {
+          throw new IllegalArgumentException("Invalid request parameter");
+        }
+        if (keyValue[0].equals("address")) {
+          return keyValue[1];
+        }
```
halibobo1205, line 554: Need to keep compatibility

forfreeday:
@halibobo1205 
Adjusted the level of interface validation, keeping the original checksums

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -518,6 +514,67 @@ public static byte[] getAddress(HttpServletRequest request) throws Exception {
...
+        if (StringUtils.isBlank(value)) {
+          return null;
+        }
+
+        JSONObject jsonObject = JSON.parseObject(value);
+        if (jsonObject != null) {
+          return jsonObject.getString(key);
+        }
+      } else if (APPLICATION_FORM_URLENCODED.toLowerCase().contains(contentType)) {
+        return getParam(getRequestValue(request));
```
halibobo1205, line 539: Why not just use `request.getParameter(addressParam)`?

forfreeday:
`getParam` doesn't need to be a generic method, it's just that the name of the method looks generic, and the only correct entry for this method is address, while all the others are incorrect.

halibobo1205:
> @halibobo1205 `getParam` doesn't need to be a generic method, it's just that the name of the method looks generic, and the only correct entry for this method is address, while all the others are incorrect.

@forfreeday Why does `getParam` need an input?

forfreeday:
@halibobo1205 
Do you understand what this method does?
getParam is the input parameter of the POST method.

halibobo1205:
> @halibobo1205 Do you understand what this method does? getParam is the input parameter of the POST method. What would happen if you didn't use getParam?

@forfreeday I really don't get it. @tomatoishealthy Please take a look.

framework/src/test/java/org/tron/core/services/http/GetBrokerageServletTest.java
tomatoishealthy, line 1: For historical reasons, this logic does not have a unit test.

Better add comprehensive unit tests for all situations after changing the whole logic of the param extraction to make sure all the new logic works well, such as: good requests for both form&json

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -518,6 +514,47 @@ public static byte[] getAddress(HttpServletRequest request) throws Exception {
...
+  private static String checkGetParam(HttpServletRequest request, String key) throws Exception {
+    String method = request.getMethod();
+    String value = null;
+
+    if (HttpMethod.GET.toString().toUpperCase() .equalsIgnoreCase(method)) {
+      return request.getParameter(key);
+    }
+    if (HttpMethod.POST.toString().toUpperCase().equals(method)) {
+      String contentType = request.getContentType();
+      if (StringUtils.isBlank(contentType)) {
```
halibobo1205, line 526: Need to keep compatibility

framework/src/test/java/org/tron/core/services/http/GetBrokerageServletTest.java
```diff
@@ -0,0 +1,46 @@
...
+    getRewardServlet.doPost(request, response);
+    try {
+      String contentAsString = response.getContentAsString();
+      JSONObject result = JSONObject.parseObject(contentAsString);
+      int reward = (int)result.get("reward");
+      Assert.assertEquals(0, reward);
+      String content = (String) result.get("Error");
+      Assert.assertNull(content);
+    } catch (UnsupportedEncodingException e) {
+      e.printStackTrace();
```
halibobo1205, line 43: Assert.fail(); is a good choice

framework/src/test/java/org/tron/core/services/http/GetRewardServletTest.java
```diff
@@ -0,0 +1,44 @@
...
+    getRewardServlet.doPost(request, response);
+    try {
+      String contentAsString = response.getContentAsString();
+      JSONObject result = JSONObject.parseObject(contentAsString);
+      int reward = (int)result.get("reward");
+      Assert.assertEquals(0, reward);
+      String content = (String) result.get("Error");
+      Assert.assertNull(content);
+    } catch (UnsupportedEncodingException e) {
+      e.printStackTrace();
```
halibobo1205, line 41: Please test all cases of checkGetParam.

framework/src/test/java/org/tron/core/services/http/GetBrokerageServletTest.java
forfreeday: I'm not expecting an exception to be thrown; the end result of this use case is a normal return of the

framework/src/test/java/org/tron/core/services/http/GetRewardServletTest.java
forfreeday: That's good advice.

framework/src/test/java/org/tron/core/services/http/GetBrokerageServletTest.java
halibobo1205: Does it mean that even if an UnsupportedEncodingException is thrown, it's still normal?

forfreeday: Using a statement when pointing out a problem is a good way to respond; if you use a rhetorical question, I can't be sure of your intentions and it doesn't help to accurately describe the problem.

forfreeday: The point of this use case is not to test for the UnsupportedEncodingException, but rather to determine the value of null.

halibobo1205: `Assert.fail()` is added where it is not possible to reach in the expectation, once it is reached, indicates that the test has failed, the result is not as expected.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -498,16 +503,7 @@ public static void printAccount(Account reply, HttpServletResponse response, Boo
...
-    if (StringUtils.isBlank(addressStr)) {
-      String input = request.getReader().lines()
-          .collect(Collectors.joining(System.lineSeparator()));
-      Util.checkBodySize(input);
-      JSONObject jsonObject = JSON.parseObject(input);
-      if (jsonObject != null) {
-        addressStr = jsonObject.getString(addressParam);
-      }
-    }
+    String addressStr = checkGetParam(request, addressParam);
```
halibobo1205, line 506: `getParam` or `checkAndGetParam` ?

framework/src/test/java/org/tron/core/services/http/GetRewardServletTest.java
```diff
@@ -0,0 +1,127 @@
...
+      Assert.assertEquals(0, reward);
+      String content = (String) result.get("Error");
+      Assert.assertNull(content);
+    } catch (UnsupportedEncodingException e) {
+      Assert.fail(e.getMessage());
+    }
+  }
+
+  @After
+  public void deleteDatabase() {
```
halibobo1205, line 121: `BaseTest` has already done the release.

framework/src/main/java/org/tron/core/services/http/Util.java
forfreeday: Use `checkGetParam` as the method name, because there are some checking judgments

framework/src/test/java/org/tron/core/services/http/GetRewardServletTest.java
forfreeday: Yes, it can be dispensed with here

halibobo1205 (approved):
LGTM

PR #5378, opened 2023-07-28
feature(db): optimize properties query frequency

### LINKED ISSUE #5375: Improve Query Performance of Property Store
# Rationale

Currently, when the get operation is performed on the snapshot of the property store, the maximum number of layers traversed by the snapshot impl is 20 times. The idea of optimization is that when a snapshot commits, it integrates the full data of the previous snapshot with the incremental data of the current snapshot so that the latest snapshot at the time of commit has the full data. In this way, during the get operation of the property store, the number of layers traversed by the snapshot will be less than or equal to 2 times.

Improve property store query performance and increase TPS.

# Implementation

During initialization, the full data of the snapshotroot is merged into the next snapshot. When the snapshot commits, the incremental data of the current snapshot is merged with the full data in the previous snapshot.

After optimization, when the property store performs a get operation, the value can be found from the incremental data of the snapshot and the full data in the previous snapshot, so that the value can be obtained by traversing at most 2 times. Before the optimization, it needs to traverse up to 20 times to get the value. After optimization, the property store query performance is improved, thereby improving TPS.

CarlChaoCarl:
close #5375


chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
```diff
@@ -83,6 +99,20 @@ public void merge(Snapshot from) {
     Streams.stream(fromImpl.db).forEach(e -> db.put(e.getKey(), e.getValue()));
   }
 
+  public void mergeFullData(Snapshot from) {
```
guoquanwu, line 102: As I see, the `merge` method is used for merging the `snapshotImpl` which is behind itself,  this method `mergeFullData` is used for merging the `snapshotImpl` which is ahead, and only one `snapshot`. it only can merge all data by another outside iteration.

So renaming the method name to `mergeAhead` or `mergeBefore` or `mergeFront` may be better?

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
```diff
@@ -233,6 +233,12 @@ public synchronized void commit() {
     }
 
     --activeSession;
+
+    dbs.forEach(db -> {
```
guoquanwu, line 237: Why set the reload logic in `commit()`, `commit()` only invoked by the end of pushing a block, not a transaction.

Are there other reasons for doing this, can you explain?

guoquanwu:
Any issue associated with this PR? So we can have a comprehensive background

CarlChaoCarl:
> Any issue associated with this PR? So we can have a comprehensive background

This Issue is associated with the current pr
https://github.com/tronprotocol/java-tron/issues/5375

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
halibobo1205: +1

CarlChaoCarl: This is because if the transaction ends without a commit,
Revoke is needed at this time,
session.close is called,
Then call snapshotManager.revoke,
Then there is no need to perform reload logic.
So the reload logic is only executed when the transaction commits. 

@tomatoishealthy @halibobo1205

halibobo1205:
@CarlChaoCarl Please fix CI.

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
```diff
@@ -30,6 +30,12 @@ public class SnapshotImpl extends AbstractSnapshot<Key, Value> {
     }
     previous = snapshot;
     snapshot.setNext(this);
+    isOptimized = snapshot.isOptimized();
+    if (isOptimized) {
+      if (root == previous) {
+        Streams.stream(root.iterator()).forEach( e -> put(e.getKey(),e.getValue()));
```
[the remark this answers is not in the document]
CarlChaoCarl: 👌🏻

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
```diff
@@ -40,6 +46,16 @@ public byte[] get(byte[] key) {
...
+    // for properties optimized
+    /*
+    if (isOptimized) {
+      value = db.get(Key.of(key));
+      if (value !=null) {
+        return value.getBytes();
+      }
+    }
+     */
+
```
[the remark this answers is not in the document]
CarlChaoCarl: I removed this part of the commented code

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
```diff
@@ -83,6 +99,20 @@ public void merge(Snapshot from) {
...
   }
 
+  public void mergeFullData(Snapshot from) {
+    if (from instanceof SnapshotRoot) {
+      return ;
+    }
+    SnapshotImpl fromImpl = (SnapshotImpl) from;
+    Streams.stream(fromImpl.db).forEach(e ->
+        {
+          if (db.get(e.getKey()) == null && e.getValue() != null ) {
```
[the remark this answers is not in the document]
CarlChaoCarl: 👌🏻

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
CarlChaoCarl: OK, I renamed the mergeFullData function to mergeAhead

CarlChaoCarl:
>

CarlChaoCarl:
> @CarlChaoCarl Please fix CI.

OK, I have solved the CI problem

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
```diff
@@ -83,6 +88,19 @@ public void merge(Snapshot from) {
     Streams.stream(fromImpl.db).forEach(e -> db.put(e.getKey(), e.getValue()));
   }
 
+  public void mergeAhead(Snapshot from) {
+    if (from instanceof SnapshotRoot) {
+      return ;
+    }
+    SnapshotImpl fromImpl = (SnapshotImpl) from;
+    Streams.stream(fromImpl.db).forEach(e -> {
+      if (db.get(e.getKey()) == null && e.getValue() != null) {
```
halibobo1205, line 97: How would happen for `e.getValue() != null`?

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
```diff
@@ -233,6 +233,12 @@ public synchronized void commit() {
...
+      if (db.getHead().isOptimized()) {
+        db.getHead().reloadToMem();
```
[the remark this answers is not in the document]
CarlChaoCarl: OK,I get it

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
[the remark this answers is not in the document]
CarlChaoCarl: db.get(e.getKey()) == null && e.getValue() != null
Passing these two judgments means:
1. The current db does not have this key, that is, db.get(e.getKey()) is null,
2. At the same time, the snapshotimpl from has this key and the value is not empty, that is, e.getValue() != null.

@wubin01

chainbase/src/main/java/org/tron/core/db2/core/SnapshotImpl.java
CarlChaoCarl: db.get(e.getKey()) == null && e.getValue() != null
Passing these two judgments means:
1. The current db does not have this key, that is, db.get(e.getKey()) is null,
2. At the same time, the snapshotimpl from has this key and the value is not empty, that is, e.getValue() != null.

@wubin01  @halibobo1205

halibobo1205: @yuekun0707 , It shouldn't be null.
image

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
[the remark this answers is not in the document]
CarlChaoCarl: I have fixed this

wubin01 (approved):

halibobo1205 (approved):

CarlChaoCarl:
> 



> Better to add the unit test for these changes

@tomatoishealthy  Yeah, I have added unit tests.

PR #5392, opened 2023-08-02
feat(net):supplement disconnect reasons

317787106:
**What does this PR do?**

- supplement two disconnect reasons, including RelayService and shutdown node

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


protocol/src/main/protos/core/Tron.proto
```diff
@@ -601,6 +601,7 @@ enum ReasonCode {
   TOO_MANY_PEERS_WITH_SAME_IP = 0x22;
   LIGHT_NODE_SYNC_FAIL = 0x23;
   BELOW_THAN_ME = 0X24;
+  NOT_WITNESS = 0x25;
```
halibobo1205, line 604: Is it compatible with older versions?

317787106: All of the newly added unrecognized ReasonCode will be converted to UNKNOWN, and no exception will be thrown.

chengtx01 (approved):

protocol/src/main/protos/core/Tron.proto
317787106: @wubin01 there is no MESSAGE_WITH_WRONG_LENGTH really. But there is NO_SUCH_MESSAGE. Do you means add NO_SUCH_MESSAGE in Protocol.ReasonCode?

317787106: I have added new ReasonCode NO_SUCH_MESSAGE in proto.

ss3344520 (approved):

Nidacoke (approved):
Everyone have any idea for more updates.

PR #5393, opened 2023-08-02
feat(net): improve chain inventory generating logic

chengtx01:
**What does this PR do?**
Retrying inventory generating if first blockId does not match the chain summary.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandler.java
@@ -87,6 +87,18 @@ public class SyncBlockChainMsgHandler implements TronMsgHandler {
 
   private LinkedList<BlockId> getLostBlockIds(List<BlockId> blockIds) throws P2pException {
 
+    BlockId unForkId = getUnForkId(blockIds);
+    LinkedList<BlockId> ids = getBlockIds(unForkId.getNum());
+
+    if (ids.isEmpty() || !unForkId.equals(ids.peekFirst())) {
+      unForkId = getUnForkId(blockIds);
+      ids = getBlockIds(unForkId.getNum());
+    }
+
+    return ids;
+  }
+
+  private BlockId getUnForkId(List<BlockId> blockIds) throws P2pException {
     BlockId unForkId = null;
     for (int i = blockIds.size() - 1; i >= 0; i--) {
       if (tronNetDelegate.containBlockInMainChain(blockIds.get(i))) {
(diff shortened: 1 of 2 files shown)


guoquanwu:
Is there any issue associated with this PR?

framework/src/main/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandler.java
```diff
@@ -87,6 +87,18 @@ private boolean check(PeerConnection peer, SyncBlockChainMessage msg) throws P2p
 
   private LinkedList<BlockId> getLostBlockIds(List<BlockId> blockIds) throws P2pException {
 
+    BlockId unForkId = getUnForkId(blockIds);
+    LinkedList<BlockId> ids = getBlockIds(unForkId.getNum());
+
+    if (ids.isEmpty() || !unForkId.equals(ids.peekFirst())) {
+      unForkId = getUnForkId(blockIds);
```
halibobo1205, line 94: If I'm not mistaken, is it just a retry?

chengtx01: Yes, it can be solved by retrying.

framework/src/test/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandlerTest.java
```diff
@@ -64,6 +65,21 @@ public void testProcessMessage() throws Exception {
     method.setAccessible(true);
     boolean f = (boolean)method.invoke(handler, peer, message);
     Assert.assertTrue(!f);
+
+    Method method1 = handler.getClass().getDeclaredMethod(
```
halibobo1205, line 69: Is it possible to reproduce this scenario without using reflection?

tomatoishealthy:
It only happened with the forks which are not finalized?  

If so, all the blocks on both(or more) forks are stored in memory?

If all the above is right, just sending the corresponding blocks can solve this thoroughly?

chengtx01:
> It only happened with the forks which are not finalized?
> 
 yes

> If so, all the blocks on both(or more) forks are stored in memory?
> 
This problem will only involve non-solidified blocks.
> If all the above is right, just sending the corresponding blocks can solve this thoroughly?
>
After the chain switch is completed, the generated inventory is without any problems.

tomatoishealthy:
> After the chain switch is completed, the generated inventory is without any problems.

I mean generating the inventory by iterating the specified fork instead of using `tronNetDelegate.getBlockIdByNum(i)` may solve this thoroughly, but seems an interface for querying the fork chain doesn't exist.

tomatoishealthy (approved):
Approved, @wubin01 should do a double check

framework/src/test/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandlerTest.java
chengtx01: Okay, further optimization.

wubin01 (approved):

framework/src/main/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandler.java
```diff
@@ -87,6 +87,18 @@ private boolean check(PeerConnection peer, SyncBlockChainMessage msg) throws P2p
 
   private LinkedList<BlockId> getLostBlockIds(List<BlockId> blockIds) throws P2pException {
```
srbaeza-ai, line 88: https://github.com/tronprotocol/java-tron/pull/5393

srbaeza-ai (approved):

PR #5403, opened 2023-08-08
fix(db): fix getHeadBlockId abnormal return problem

### LINKED ISSUE #5389: getHeadBlockId abnormal return
# System information

Java-tron version: v4.7.2
OS & Version: Linux & macOS

# Expected behavior

The `getHeadBlockId` function should return the correct block header information.

# Actual behavior

When synchronizing blocks, an abnormal block ID appears, as shown in the following log:
image

This ID: `00000000032b9cc50deb2777f7e751ade0245ace23c2cc8579d733af34401832` is composed of two parts, the number of block 53189829 and the hash of block 53189830. The reason for this phenomenon is that there is a concurrency problem in this function. The function definition is as follows:
```
public synchronized BlockId getHeadBlockId() {
    return new BlockId(
        dynamicPropertiesStore.getLatestBlockHeaderHash(),
        dynamicPropertiesStore.getLatestBlockHeaderNumber());
  }
```
When calling this function, if only the block hash is written at this time, and the block number has not been written in time, the above-mentioned situation will occur.

wubin01:
**What does this PR do?**
Fix getHeadBlockId abnormal return problem, fix #5389.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/chainbase/src/main/java/org/tron/core/ChainBaseManager.java
+++ b/chainbase/src/main/java/org/tron/core/ChainBaseManager.java
@@ -316,9 +316,7 @@ public class ChainBaseManager {
   }
 
   public synchronized BlockId getHeadBlockId() {
-    return new BlockId(
-        dynamicPropertiesStore.getLatestBlockHeaderHash(),
-        dynamicPropertiesStore.getLatestBlockHeaderNumber());
+    return new BlockId(dynamicPropertiesStore.getLatestBlockHeaderHash());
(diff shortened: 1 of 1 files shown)


halibobo1205 (approved):
LGTM

halibobo1205:
close #5389.

chengtx01 (approved):

PR #5421, opened 2023-08-14
feat(all):make @PreDestroy work

### LINKED ISSUE #5420: @PreDestroy does not work
#### System information

OS : Mac OS X
JVM : Oracle Corporation 1.8.0_321 x86_64
Git : 9e95599de2a9d60f4e706d18d3b71ac8850a0c01
Version : 4.7.2
Code : 18031

OS : Linux
JVM : Red Hat, Inc. 1.8.0_302 amd64
Git : 9e95599de2a9d60f4e706d18d3b71ac8850a0c01
Version : 4.7.2
Code : 18031

#### Expected behaviour

The PreDestroy annotation is used on methods as a callback notification to signal that the instance is in the process of being removed by the container. The method annotated with PreDestroy is typically used to release resources that it has been holding. This annotation MUST be supported by all container managed objects that support PostConstruct except the application client container in Java EE 5.

As documented in [Bean.destroyMethod](https://docs.spring.io/spring-framework/docs/5.3.18/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod--):

  > public abstract [String](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html?is-external=true) destroyMethod
The optional name of a method to call on the bean instance upon closing the application context, for example a close() method on a JDBC DataSource implementation, or a Hibernate SessionFactory object. The method must have no arguments but may throw any exception.
As a convenience to the user, the container will attempt to infer a destroy method against an object returned from the @Bean method. For example, given an @Bean method returning an Apache Commons DBCP BasicDataSource, the container will notice the close() method available on that object and automatically register it as the destroyMethod. This 'destroy method inference' is currently limited to detecting only public, no-arg methods named 'close' or 'shutdown'. The method may be declared at any level of the inheritance hierarchy and will be detected regardless of the return type of the @Bean method (i.e., detection occurs reflectively against the bean instance itself at creation time).
To disable destroy method inference for a particular @Bean, specify an empty string as the value, e.g. @Bean(destroyMethod=""). Note that the [DisposableBean](https://docs.spring.io/spring-framework/docs/5.3.18/javadoc-api/org/springframework/beans/factory/DisposableBean.html) callback interface will nevertheless get detected and the corresponding destroy method invoked: In other words, destroyMethod="" only affects custom close/shutdown methods and [Closeable](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html?is-external=true)/[AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html?is-external=true) declared close methods.
Note: Only invoked on beans whose lifecycle is under the full control of the factory, which is always the case for singletons but not guaranteed for any other scope.
See Also:
[DisposableBean](https://docs.spring.io/spring-framework/docs/5.3.18/javadoc-api/org/springframework/beans/factory/DisposableBean.html), [ConfigurableApplicationContext.close()](https://docs.spring.io/spring-framework/docs/5.3.18/javadoc-api/org/springframework/context/ConfigurableApplicationContext.html#close--)
Default:
"(inferred)"

According to [DisposableBeanAdapter.inferDestroyMethodIfNecessary](https://github.com/spring-projects/spring-framework/blob/5.3.x/spring-beans/src/main/java/org/springframework/beans/factory/support/DisposableBeanAdapter.java#L394):
> If the current value of the given beanDefinition's "destroyMethodName" property is AbstractBeanDefinition.INFER_METHOD, then attempt to infer a destroy method. Candidate methods are currently limited to public, no-arg methods named "close" or "shutdown" (whether declared locally or inherited). The given BeanDefinition's "destroyMethodName" is updated to be null if no such method is found, otherwise set to the name of the inferred method. This constant serves as the default for the @Bean#destroyMethod attribute and the value of the constant may also be used in XML within the  or  attributes.
Also processes the java.io.Closeable and AutoCloseable interfaces, reflectively calling the "close" method on implementing beans as well.

When the process receives the `kill -15` signal, the [@PreDestroy](https://docs.oracle.com/javase/8/docs/api/javax/annotation/PreDestroy.html) qualified method should be executed.

#### Actual behaviour
 
`@PreDestroy` does not work

Terminate the process with `kill -15`.

halibobo1205:
close #5420

### DIFF
--- a/framework/src/main/java/org/tron/common/application/TronApplicationContext.java
+++ b/framework/src/main/java/org/tron/common/application/TronApplicationContext.java
@@ -2,8 +2,7 @@ package org.tron.common.application;
 
 import org.springframework.beans.factory.support.DefaultListableBeanFactory;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
-import org.tron.core.db.Manager;
-import org.tron.core.net.TronNetService;
+import org.tron.program.FullNode;
 
 public class TronApplicationContext extends AnnotationConfigApplicationContext {
 
@@ -23,19 +22,13 @@ public class TronApplicationContext extends AnnotationConfigApplicationContext {
   }
 
   @Override
-  public void destroy() {
-
+  public void doClose() {
+    logger.info("******** start to close ********");
     Application appT = ApplicationFactory.create(this);
     appT.shutdownServices();
     appT.shutdown();
-
-    TronNetService tronNetService = getBean(TronNetService.class);
-    tronNetService.close();
-
-    Manager dbManager = getBean(Manager.class);
-    dbManager.stopRePushThread();
-    dbManager.stopRePushTriggerThread();
-    dbManager.stopFilterProcessThread();
-    super.destroy();
+    super.doClose();
+    logger.info("******** close end ********");
+    FullNode.shutDownSign = true;
   }
 }
--- a/framework/src/test/java/org/tron/core/services/filter/LiteFnQueryGrpcInterceptorTest.java
+++ b/framework/src/test/java/org/tron/core/services/filter/LiteFnQueryGrpcInterceptorTest.java
@@ -3,23 +3,23 @@ package org.tron.core.services.filter;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.StatusRuntimeException;
-import java.io.File;
+import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
-
-import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Assert;
-import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import org.junit.rules.TemporaryFolder;
 import org.tron.api.GrpcAPI;
 import org.tron.api.WalletGrpc;
 import org.tron.api.WalletSolidityGrpc;
 import org.tron.common.application.Application;
 import org.tron.common.application.ApplicationFactory;
 import org.tron.common.application.TronApplicationContext;
-import org.tron.common.utils.FileUtil;
 import org.tron.common.utils.PublicMethod;
 import org.tron.core.ChainBaseManager;
 import org.tron.core.Constant;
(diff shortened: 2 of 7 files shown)


framework/src/main/java/org/tron/common/application/TronApplicationContext.java
```diff
@@ -23,19 +22,13 @@ public TronApplicationContext(String... basePackages) {
...
-    tronNetService.close();
-
-    Manager dbManager = getBean(Manager.class);
-    dbManager.stopRePushThread();
-    dbManager.stopRePushTriggerThread();
-    dbManager.stopFilterProcessThread();
-    super.destroy();
+    super.doClose();
+    logger.info("******** close end ********");
+    FullNode.shutDownSign = true;
```
lurais, line 32: Should the shutDownSign be assigned once the validate executor be closed? Is this reasonable to be here?

framework/src/test/java/org/tron/core/services/filter/LiteFnQueryGrpcInterceptorTest.java
```diff
@@ -89,40 +93,45 @@ public void init() {
...
    */
-  @After
-  public void destroy() throws InterruptedException {
+  @AfterClass
+  public static void destroy() throws InterruptedException {
     if (channelFull != null) {
       channelFull.shutdown().awaitTermination(5, TimeUnit.SECONDS);
     }
+    if (channelSolidity != null) {
+      channelSolidity.shutdown().awaitTermination(5, TimeUnit.SECONDS);
```
lurais, line 102: What if the channelSolidity have not been shut down after 5 seconds?

framework/src/main/java/org/tron/common/application/TronApplicationContext.java
halibobo1205: You can find something in here, https://github.com/tronprotocol/java-tron/pull/4020.

framework/src/test/java/org/tron/core/services/filter/LiteFnQueryGrpcInterceptorTest.java
halibobo1205: Waits for the channel to become terminated, giving up if the timeout is reached.

framework/src/main/java/org/tron/common/application/TronApplicationContext.java
```diff
@@ -23,19 +22,13 @@ public TronApplicationContext(String... basePackages) {
   }
 
   @Override
-  public void destroy() {
```
chengtx01, line 26: Shouldn't all the places where the destroy method is called be replaced, including all test code?

halibobo1205: destroy ->close -> doClose
image

PR #5425, opened 2023-08-15
feat(all):name thread pools

### LINKED ISSUE #5422: Organize all ThreadExecutors with explicit name and time-wait shutdown
# Rationale

1. At present, the java-tron `thread pool` name is not clearly defined, a thread is created with the following (default) name: Thread[pool-1-thread-1], which isn’t helpful and doesn’t let you differentiate threads by name. As you can see, this isn't useful.
2.  Some thread pools don't have a graceful shutdown.

What are the use-cases?

In case of performance issues: thread dumps and using the top command. In both cases, it’s useful to know what function does a thread service, as the stacktrace in the dump might not always be revealing.

A tool for quick investigation is top. More precisely, top -H -p <pid>. This shows the usual top table, but the -H flag means that threads for the chosen process should be printed. You basically get the most CPU-heavy and currently active threads, by name. In those cases, it’s extremely useful to have custom names.

1. [DynamicArgs.java#L29](https://github.com/tronprotocol/java-tron/blob/develop/framework/src/main/java/org/tron/core/config/args/DynamicArgs.java#L29)
2. [Manager.java#L536](https://github.com/tronprotocol/java-tron/blob/develop/framework/src/main/java/org/tron/core/db/Manager.java#L536)
3.  [TransactionsMsgHandler.java#L55](https://github.com/tronprotocol/java-tron/blob/develop/framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java#L55)

# Implementation
All of these executors have a thread factory. It’s hidden in the default factory method, but you can supply a thread factory. If not supplied, a default thread factory is used whenever a thread is needed.

1. Name threads and thread-pools of ExecutorService.
```java
ThreadFactory factory = new ThreadFactoryBuilder().setNameFormat("my-sad-thread-%d").build();
 ExecutorService es = Executors.newFixedThreadPool(4, factory);
```
2. Make time-wait shutdown(awaitTermination) for thread-pools of ExecutorService.
```java
 es.shutdown();
 es.awaitTermination(60, java.util.concurrent.TimeUnit.SECONDS);
```

Yes.

halibobo1205:
close #5422 .

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java
+++ b/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java
@@ -26,7 +26,6 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import lombok.AllArgsConstructor;
@@ -46,6 +45,7 @@ import org.tron.common.crypto.zksnark.BN128G1;
 import org.tron.common.crypto.zksnark.BN128G2;
 import org.tron.common.crypto.zksnark.Fp;
 import org.tron.common.crypto.zksnark.PairingCheck;
+import org.tron.common.es.ExecutorServiceManager;
 import org.tron.common.parameter.CommonParameter;
 import org.tron.common.runtime.ProgramResult;
 import org.tron.common.runtime.vm.DataWord;
@@ -983,11 +983,13 @@ public class PrecompiledContracts {
   public static class BatchValidateSign extends PrecompiledContract {
 
     private static final ExecutorService workers;
+    private static final String workersName = "validate-sign-contract";
     private static final int ENGERYPERSIGN = 1500;
     private static final int MAX_SIZE = 16;
 
     static {
-      workers = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() / 2 + 1);
+      workers = ExecutorServiceManager.newFixedThreadPool(workersName,
+          Runtime.getRuntime().availableProcessors() / 2 + 1);
     }
 
     @Override
@@ -1290,10 +1292,12 @@ public class PrecompiledContracts {
     private static final Integer[] SIZE = {2080, 2368, 2464, 2752};
     private static final ExecutorService workersInConstantCall;
     private static final ExecutorService workersInNonConstantCall;
+    private static final String constantCallName = "verify-transfer-constant-call";
+    private static final String nonConstantCallName = "verify-transfer-non-constant-call";
 
     static {
-      workersInConstantCall = Executors.newFixedThreadPool(5);
-      workersInNonConstantCall = Executors.newFixedThreadPool(5);
+      workersInConstantCall = ExecutorServiceManager.newFixedThreadPool(constantCallName, 5);
+      workersInNonConstantCall = ExecutorServiceManager.newFixedThreadPool(nonConstantCallName, 5);
     }
 
     @Override
--- a/chainbase/src/main/java/org/tron/common/storage/metric/DbStatService.java
+++ b/chainbase/src/main/java/org/tron/common/storage/metric/DbStatService.java
@@ -1,11 +1,10 @@
 package org.tron.common.storage.metric;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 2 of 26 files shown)


chainbase/src/main/java/org/tron/common/storage/metric/DbStatService.java
```diff
@@ -1,22 +1,20 @@
...
 import org.tron.core.db2.common.DB;
 
 @Slf4j(topic = "metrics")
 @Component
 public class DbStatService {
-  private static final ScheduledExecutorService statExecutor  =
-      Executors.newSingleThreadScheduledExecutor(
-          new ThreadFactoryBuilder().setNameFormat("db-stats-thread-%d").build());
-
+  private final String esName = "db-stats";
```
lurais, line 15: Should this be static?

halibobo1205: For singleton, there should be no need.

actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java
```diff
@@ -983,11 +983,13 @@ public Pair<Boolean, byte[]> execute(byte[] rawData) {
   public static class BatchValidateSign extends PrecompiledContract {
 
     private static final ExecutorService workers;
+    private static final String workersName = "validate-sign-contract";
```
lurais, line 986: Should this managed by ExecutorServiceManager?

halibobo1205: Yes, do you have any good ideas?

lurais: So we can just move the excutors name and the construction into ExecutorServiceManager, then the lifecycle and all the executors is controlled in ExecutorServiceManager , a certain executor shall be obtained by using a name(may be obtained by spring injection).

halibobo1205: Thread pool shutdown is better controlled by the service, not the manager but this one is a special case, so maybe it can be done this way.

framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java
```diff
@@ -102,7 +103,7 @@ private void check(PeerConnection peer, TransactionsMessage msg) throws P2pExcep
   private void handleSmartContract() {
     smartContractExecutor.scheduleWithFixedDelay(() -> {
       try {
-        while (queue.size() < MAX_SMART_CONTRACT_SUBMIT_SIZE) {
+        while (queue.size() < MAX_SMART_CONTRACT_SUBMIT_SIZE && smartContractQueue.size() > 0) {
```
wubin01, line 106: Why add this logic, if the queue has no elements, it will enter an infinite loop.

wubin01, line 106: Try not to change the original logic, if it is a bug, it is recommended to submit a pr separately.

halibobo1205: `TrxEvent event = smartContractQueue.take();`
Retrieves and removes the head of this queue, waiting if necessary until an element becomes available.
if queue is empty, will block `kill -15`.

halibobo1205: @wubin01 , As explained above,  `smartContractExecutor.shutdown()` -> `ExecutorServiceManager.shutdownAndAwaitTermination(smartContractExecutor, smartEsName)`
if queue is empty, will block 60s

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
```diff
@@ -117,18 +116,6 @@ public void init() {
     exitThread.start();
   }
```
chengtx01, line 117: Why not adjust exitThread to SingleThreadExecutor?

halibobo1205: `exitThread` will only run `once`, when it receives the exit signal ` LockSupport.unpark(exitThread) ` then `System.exit(1)`, otherwise it blocks in here `  LockSupport.park() `, unlike other single-thread that has while(true) logic.

chengtx01 (approved):

wubin01 (approved):

317787106 (approved):

PR #5442, opened 2023-08-22
test(api): add test cases of Wallet

317787106:
**What does this PR do?**

- add shield transaction's test cases of Wallet

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -2054,8 +2054,16 @@ public IncrementalMerkleTree getMerkleTreeOfBlock(long blockNum) throws ZksnarkE
...
+        byte[] treeRoot = chainBaseManager.getMerkleTreeIndexStore().get(blockNum);
+        if (treeRoot == null) {
+          return null;
+        }
+        IncrementalMerkleTreeCapsule treeCapsule = chainBaseManager.getMerkleTreeStore()
+            .get(treeRoot);
+        if (treeCapsule == null) {
+          return null;
+        }
+        return treeCapsule.getInstance();
```
chengtx01, line 2066: Why do you need to modify the logic code in the PR for adding test cases?

317787106: There is a bug in method getMerkleTreeOfBlock,  but this mehod is not invoked by any other class.

jwrct (approved):

ss3344520 (approved):

PR #5446, opened 2023-08-23
feat(sign): optimize signing exception log

### LINKED ISSUE #5423: Optimize error logs in signature verification
# Rationale

At present, in the process of verifying the signature of a transaction, if the signature format is wrong, an error log will also be printed out. But the current form of this error log makes it not clear for developers. And it should be captured and converted to a more understandable exception.
The current error message format is as follows:

> 17:16:23.017 WARN [qtp250219336-22365] o.t.c.Wallet Broadcast transaction fc6a47670a46bf01ad006320b90bb9bd572f45e8d0f199c4f3b222dd737cbff8 failed
java.lang.IllegalArgumentException: Invalid point compression
at [org.bouncycastle.math.ec](http://org.bouncycastle.math.ec/).ECCurve$AbstractFp.decompressPoint(Unknown Source)
at [org.bouncycastle.math.ec](http://org.bouncycastle.math.ec/).ECCurve.decodePoint(Unknown Source)
at org.tron.common.crypto.ECKey.decompressKey(ECKey.java:636)
at org.tron.common.crypto.ECKey.recoverPubBytesFromSignature(ECKey.java:552)
at org.tron.common.crypto.ECKey.signatureToKeyBytes(ECKey.java:426)

To make the error logs more standardized.
# Implementation
Just catch the exception which is only argument exception type and convert to another more standardized one.

Yes.

lurais:
**What does this PR do?**
Optimize signing exception log， close #5423

Standardize the signing related exception 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/crypto/src/main/java/org/tron/common/crypto/SignUtils.java
+++ b/crypto/src/main/java/org/tron/common/crypto/SignUtils.java
@@ -26,10 +26,14 @@ public class SignUtils {
   public static byte[] signatureToAddress(
       byte[] messageHash, String signatureBase64, boolean isECKeyCryptoEngine)
       throws SignatureException {
-    if (isECKeyCryptoEngine) {
-      return ECKey.signatureToAddress(messageHash, signatureBase64);
+    try {
+      if (isECKeyCryptoEngine) {
+        return ECKey.signatureToAddress(messageHash, signatureBase64);
+      }
+      return SM2.signatureToAddress(messageHash, signatureBase64);
+    } catch (Exception e) {
+      throw new SignatureException(e);
     }
-    return SM2.signatureToAddress(messageHash, signatureBase64);
   }
 
   public static SignatureInterface fromComponents(
--- a/framework/src/test/java/org/tron/common/crypto/BouncyCastleTest.java
+++ b/framework/src/test/java/org/tron/common/crypto/BouncyCastleTest.java
@@ -8,6 +8,7 @@ import java.security.SignatureException;
 import java.util.Arrays;
 import org.bouncycastle.crypto.digests.SM3Digest;
 import org.bouncycastle.util.encoders.Hex;
+import org.junit.Assert;
 import org.junit.Test;
 import org.tron.common.crypto.sm2.SM2;
 import org.tron.common.utils.Sha256Hash;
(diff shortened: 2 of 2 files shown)


crypto/src/main/java/org/tron/common/crypto/SignUtils.java
```diff
@@ -26,10 +26,17 @@ public static SignInterface fromPrivate(byte[] privKeyBytes, boolean isECKeyCryp
...
       throws SignatureException {
-    if (isECKeyCryptoEngine) {
-      return ECKey.signatureToAddress(messageHash, signatureBase64);
+    try {
+      if (isECKeyCryptoEngine) {
+        return ECKey.signatureToAddress(messageHash, signatureBase64);
+      }
+      return SM2.signatureToAddress(messageHash, signatureBase64);
+    } catch(IllegalArgumentException e) {
+      if (e.getMessage().contains("Invalid point compression")) {
```
halibobo1205, line 35: Why special treatment, I understand that you can throw a SignatureException whenever `signatureToAddress` is incorrect.

lurais: Currently only the exception with message like "Invalid point compression" can be sure to deal like this way, and other kind of messages can not be found now, that means just treat it like this may be not correct and that will not be noticed in this way.

halibobo1205: I'm not sure I agree with this.

halibobo1205:
@lurais Please re-check your PR and don't bring any irrelevant information.

crypto/src/main/java/org/tron/common/crypto/SignUtils.java
lurais: From the method signatureToKeyBytes in class ECKey, other kinds of error messages has been judged, that means not treat them the same, and it is core method ,any error messages deserved to be noticed, if we are not sure how to deal, we should throw it.

lurais:
> 

updated.

crypto/src/main/java/org/tron/common/crypto/SignUtils.java
lurais: > IllegalArgumentException

When the signature is not correct but the length is right, for example, if someone send a transaction , and the correct signature is "abc", but it was replaced to "001", that will bring it a IllegalArgumentException.

lurais: > Actually, I think all exceptions in this scene are SignatureException even `001`

yes , it is clear to throw an SignatureException in this scene, while the point is that  shall we just throw a SignatureException in any unknown scene. @tomatoishealthy

lurais: updated.

halibobo1205 (approved):
LGTM

chengtx01 (approved):

PR #5469, opened 2023-09-04
restore jsonrpc api consistency to display only TRX as value

alexqrid:
**What does this PR do?**
This PR restores consistency in JSON-RPC API. Previously `value` field in `JSON-RPC` response displayed various currencies(TRC-10 or TRX) and this is incorrect.  

As we can't distinguish the transaction type in `JSON-RPC` response it's weird to return various currencies amount in `value` field or vote amount.
For instance, for transaction types `UnfreezeAssetContract`,`TransferAssetContract` `value` field indicates token amount, for `VoteWitnessContract` it indicates vote count, for `Exchange*Contract` it indicates unknown currency at all as the currency depends on exchange.

- Manual Testing

**Follow up**
#### System information

java-tron version: `GreatVoyage-v4.7.2`
OS & Version: Linux

#### Expected behaviour
In order to keep consistent API, `value` field in JSON-RPC response should indicate TRX(TRON) amount, neither asset amount nor vote count.

#### Actual behaviour
`value` field indicates not only TRX(TRON) value but TRC-10 token value as well.

Check for example transaction `64e594ccb2cb47b23ade85adbcb3404a501d2627291b09a0ed49364049c5abfa`:

<details><summary>eth_getTransactionByHash response</summary>

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "blockHash": "0x0000000003197e6a40297fe291280f9c7ad3cc970564b93f96ced133d305a55e",
        "blockNumber": "0x3197e6a",
        "from": "0x4da5ddaaea7f1138c030ac8d97017d5020dd6536",
        "gas": "0x0",
        "gasPrice": "0x1a4",
        "hash": "0x64e594ccb2cb47b23ade85adbcb3404a501d2627291b09a0ed49364049c5abfa",
        "input": "0x",
        "nonce": "0x0000000000000000",
        "r": "0x49a16c9c202a9ab7b2b71e172e440d475733169a90514dc9fb6c89be576a57ad",
        "s": "0x150276b20b7aa21f96afbde95d4f31b30a10b0839bb8f7af85c99e274de248d3",
        "to": "0x49d216cb7b1cc41f166be76241301781f87e039a",
        "transactionIndex": "0x5",
        "type": "0x0",
        "v": "0x1c",
        "value": "0x87a238"
    }
}
```
</details>

```bash
curl --location 'http://node/jsonrpc' \
--header 'Content-Type: application/json' \
--data '{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "eth_getTransactionByHash",
    "params": [
        "0x64e594ccb2cb47b23ade85adbcb3404a501d2627291b09a0ed49364049c5abfa"
    ]
}
```

**Extra details**
With this PR jsonrpc response is consistent and returns correct value of tron for the transacion above (`0x64e594ccb2cb47b23ade85adbcb3404a501d2627291b09a0ed49364049c5abfa`):

```json
{
    "jsonrpc": "2.0",
    "id": 7704110,
    "result": {
        "blockHash": "0x0000000003197e6a40297fe291280f9c7ad3cc970564b93f96ced133d305a55e",
        "blockNumber": "0x3197e6a",
        "from": "0x4da5ddaaea7f1138c030ac8d97017d5020dd6536",
        "gas": "0x0",
        "gasPrice": "0x1a4",
        "hash": "0x64e594ccb2cb47b23ade85adbcb3404a501d2627291b09a0ed49364049c5abfa",
        "input": "0x",
        "nonce": "0x0000000000000000",
        "r": "0x49a16c9c202a9ab7b2b71e172e440d475733169a90514dc9fb6c89be576a57ad",
        "s": "0x150276b20b7aa21f96afbde95d4f31b30a10b0839bb8f7af85c99e274de248d3",
        "to": "0x49d216cb7b1cc41f166be76241301781f87e039a",
        "transactionIndex": "0x5",
        "type": "0x0",
        "v": "0x1c",
        "value": "0x0"
    }
}
```

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
@@ -242,17 +242,6 @@ public class JsonRpcApiUtil {
         case TransferContract:
           amount = contractParameter.unpack(TransferContract.class).getAmount();
           break;
-        case TransferAssetContract:
-          amount = contractParameter.unpack(TransferAssetContract.class).getAmount();
-          break;
-        case VoteWitnessContract:
-          List<Vote> votesList = contractParameter.unpack(VoteWitnessContract.class).getVotesList();
-          long voteNumber = 0L;
-          for (Vote vote : votesList) {
-            voteNumber += vote.getVoteCount();
-          }
-          amount = voteNumber;
-          break;
         case WitnessCreateContract:
           amount = 9999_000_000L;
           break;
@@ -261,21 +250,18 @@ public class JsonRpcApiUtil {
           amount = 1024_000_000L;
           break;
         case ParticipateAssetIssueContract:
-          break;
+        case UnfreezeAssetContract:
+        case VoteWitnessContract:
+        case TransferAssetContract:
+        case ExchangeWithdrawContract:
+        case ExchangeInjectContract:
+        case ExchangeTransactionContract:
+              break;
         case FreezeBalanceContract:
           amount = contractParameter.unpack(FreezeBalanceContract.class).getFrozenBalance();
           break;
         case TriggerSmartContract:
           amount = contractParameter.unpack(TriggerSmartContract.class).getCallValue();
-          break;
-        case ExchangeInjectContract:
-          amount = contractParameter.unpack(ExchangeInjectContract.class).getQuant();
-          break;
-        case ExchangeWithdrawContract:
-          amount = contractParameter.unpack(ExchangeWithdrawContract.class).getQuant();
-          break;
-        case ExchangeTransactionContract:
-          amount = contractParameter.unpack(ExchangeTransactionContract.class).getQuant();
           break;
         case AccountPermissionUpdateContract:
           amount = 100_000_000L;
@@ -295,10 +281,6 @@ public class JsonRpcApiUtil {
         case UnfreezeBalanceV2Contract:
         case CancelAllUnfreezeV2Contract:
           amount = getAmountFromTransactionInfo(hash, contract.getType(), transactionInfo);
-          break;
-        case UnfreezeAssetContract:
-          amount = getUnfreezeAssetAmount(contractParameter.unpack(UnfreezeAssetContract.class)
-              .getOwnerAddress().toByteArray(), wallet);
           break;
         default:
       }
@@ -354,33 +336,6 @@ public class JsonRpcApiUtil {
     }
     return amount;
   }
-
-  public static long getUnfreezeAssetAmount(byte[] addressBytes, Wallet wallet) {
-    long amount = 0L;
-    try {
-      if (addressBytes == null) {
-        return amount;
-      }
-
-      AssetIssueList assetIssueList = wallet
-          .getAssetIssueByAccount(ByteString.copyFrom(addressBytes));
-      if (assetIssueList != null) {
-        if (assetIssueList.getAssetIssueCount() != 1) {
-          return amount;
-        } else {
-          AssetIssueContract assetIssue = assetIssueList.getAssetIssue(0);
-          for (FrozenSupply frozenSupply : assetIssue.getFrozenSupplyList()) {
-            amount += frozenSupply.getFrozenAmount();
-          }
-        }
-      }
-    } catch (Exception e) {
-      logger.warn("Exception happens when get token10 frozenAmount. Exception = [{}]",
-          Throwables.getStackTraceAsString(e));
-    }
-    return amount;
-  }
-
   /**
    * convert 40 or 42 hex string of address to byte array, compatible with "41"(T) ahead,
    * padding 0 ahead if length is odd.


317787106:
I don't agree with this modification. It's not compatible with the data ago. And tronscan uses this value though it's not TRX. I think it's not necessary to do it.

alexqrid:
> And tronscan uses this value though it's not TRX. I think it's not necessary to do it.

Thank you for the answer. Let me clarify some points:
Actually nobody can use it without querying `/wallet/gettransaction*`/`/wallet/gettransactioninfo*`, as they need to identify the type of transaction(interacted system contract in the transaction), but all needed values with more identifiable currency information are located in the response of the mentioned endpoints, but not in jsornpc. 
If tronscan uses it, it just make an extra api call to the node to retrieve data that can be parsed from previous responses.
If we make the transaction type identifiable in `JSON-RPC` response, then one can use it as you mentioned, but at the moment it's just an extra API call.

halibobo1205:
@317787106 Use `type` to make the difference? ,see https://github.com/tronprotocol/java-tron/pull/4232/commits/83a6d5847b78fcae971557943b4385ba47b98f34.  What is `type` used for, I don't see it being used at the moment.

317787106:
> @317787106 Use `type` to make the difference? ,see [83a6d58](https://github.com/tronprotocol/java-tron/commit/83a6d5847b78fcae971557943b4385ba47b98f34). What is `type` used for, I don't see it being used at the moment.

@halibobo1205  `type` is not used. May be it was used ago but not deleted later.

halibobo1205:
@317787106 Help figure out what this field was used for before, or add a new field?

alexqrid:
Seems that `type` was kept to make API Ethereum `JSON-RPC` compatible.
And I agree with @halibobo1205 if we want to make use of various `value` types, the `type` should be defined and documented.
But what will be the number of type for each transaction, the same as [protobuf number of `enum ContractType`](https://github.com/tronprotocol/java-tron/blob/437c12e1fc6f12593bd6fdf7d159eb6461d8a3e8/protocol/src/main/protos/core/Tron.proto#L339)?

alexqrid:
Also I'm not quite sure with correctness of [this part](https://github.com/tronprotocol/java-tron/blob/2eee9cc1475a22a8f1d4cd82ff6d8c1020818dfa/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java#L270).
Imagine if one uses json rpc to parse transacted value and construct the following event:
```
{
"sender" : response.result.from,
"recipient": response.result.to,
"value": response.result.value
}
```
And this would be wrong for ShieldedTransaction, as the value can be either from or to, currently actual information for ShieldedTransaction types isn't displayed correctly for such a case.

317787106:
> Seems that `type` was kept to make API Ethereum `JSON-RPC` compatible. And I agree with @halibobo1205 if we want to make use of various `value` types, the `type` should be defined and documented. But what will be the number of type for each transaction, the same as [protobuf number of `enum ContractType`](https://github.com/tronprotocol/java-tron/blob/437c12e1fc6f12593bd6fdf7d159eb6461d8a3e8/protocol/src/main/protos/core/Tron.proto#L339)?

@alexqrid You are right, `type` is the transaction type. There are 3 types of transaction in ethereum, includes Legacy transactions、Access list transactions、EIP-1559 transactions. More details on https://docs.infura.io/networks/ethereum/concepts/transaction-types. 0x0 refers to Legacy transactions, and tron only has Legacy transactions. column `type` can get from TransactionReceipt and it cannot be deleted.

317787106:
@alexqrid Any transaction's type in java-tron is 0x0.

317787106:
@alexqrid The reason that we use amount in TransferAssetContract: and other contract is to be compatible with tronscan and trongrid api. You can look at this https://apilist.tronscanapi.com/api/transaction-info?hash=af1094b946e947795656a9df3da7fef311b790c6ee067b29251b13fd5463c510. There is column `amount`. I am not sure if there is any problem if not used.

halibobo1205:
@alexqrid For compatibility reasons, no changes will be considered for now, cc @317787106  @jwrct .

halibobo1205:
> Also I'm not quite sure with correctness of [this part](https://github.com/tronprotocol/java-tron/blob/2eee9cc1475a22a8f1d4cd82ff6d8c1020818dfa/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java#L270). Imagine if one uses json rpc to parse transacted value and construct the following event:
> 
> ```
> {
> "sender" : response.result.from,
> "recipient": response.result.to,
> "value": response.result.value
> }
> ```
> 
> And this would be wrong for ShieldedTransaction, as the value can be either from or to, currently actual information for ShieldedTransaction types isn't displayed correctly for such a case.

@317787106  @jwrct  Please Take A Look

halibobo1205:
> Seems that `type` was kept to make API Ethereum `JSON-RPC` compatible. And I agree with @halibobo1205 if we want to make use of various `value` types, the `type` should be defined and documented. But what will be the number of type for each transaction, the same as [protobuf number of `enum ContractType`](https://github.com/tronprotocol/java-tron/blob/437c12e1fc6f12593bd6fdf7d159eb6461d8a3e8/protocol/src/main/protos/core/Tron.proto#L339)?

I think this is a good idea, let's discuss it,  maybe we could add a new field. @317787106 @jwrct

317787106:
@alexqrid There may be compatible problem with previous version. Let's suppose that one user has used this value (same as amount) of the api eth_getTransactionByHash, if your changes take effect, vaule will be zero except TRX related. If this user don't acquire this changes, the data will be wrong.

If you only want value only related to TRX, you can use other api to get the ContractType. This is not time consuming.

PR #5483, opened 2023-09-12
feat(http): optimize parameter parsing

forfreeday:
**What does this PR do?**
optimize incorrect content-type comparisons.

When a client requests an interface, if it is an application/json request, it may be written in a variety of ways, including:
1. `application/json`
2. `application/json;charset=iso-8859-1`
3. `application/json;charset=utf-8`

When determining the content-type of a request,  the interface needs to be compatible with the variable writing ways of the requests.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/http/Util.java
+++ b/framework/src/main/java/org/tron/core/services/http/Util.java
@@ -524,7 +524,6 @@ public class Util {
 
   private static String checkGetParam(HttpServletRequest request, String key) throws Exception {
     String method = request.getMethod();
-    String value = null;
 
     if (HttpMethod.GET.toString().toUpperCase().equalsIgnoreCase(method)) {
       return request.getParameter(key);
@@ -534,8 +533,10 @@ public class Util {
       if (StringUtils.isBlank(contentType)) {
         return null;
       }
-      if (contentType.contains(MimeTypes.Type.APPLICATION_JSON.asString())) {
-        value = getRequestValue(request);
+      if (contentType.contains(MimeTypes.Type.FORM_ENCODED.asString())) {
+        return request.getParameter(key);
+      } else {
+        String value = getRequestValue(request);
         if (StringUtils.isBlank(value)) {
           return null;
         }
@@ -544,13 +545,10 @@ public class Util {
         if (jsonObject != null) {
           return jsonObject.getString(key);
         }
-      } else if (contentType.contains(MimeTypes.Type.FORM_ENCODED.asString())) {
-        return request.getParameter(key);
-      } else {
         return null;
       }
     }
-    return value;
+    return null;
   }
 
   public static String getRequestValue(HttpServletRequest request) throws IOException {
--- a/framework/src/main/java/org/tron/core/services/http/Util.java
+++ b/framework/src/main/java/org/tron/core/services/http/Util.java
@@ -534,9 +534,7 @@ public class Util {
       if (StringUtils.isBlank(contentType)) {
         return null;
       }
-      if (contentType.contains(MimeTypes.Type.APPLICATION_JSON.asString())
-              || contentType.contains(MimeTypes.Type.APPLICATION_JSON_UTF_8.asString())
-              || contentType.contains(MimeTypes.Type.APPLICATION_JSON_8859_1.asString())) {
+      if (contentType.contains(MimeTypes.Type.APPLICATION_JSON.asString())) {
         value = getRequestValue(request);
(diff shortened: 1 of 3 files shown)


lurais (approved):

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -535,7 +534,9 @@ private static String checkGetParam(HttpServletRequest request, String key) thro
       if (StringUtils.isBlank(contentType)) {
         return null;
       }
-      if (APPLICATION_JSON.toLowerCase().contains(contentType)) {
+      if (contentType.contains(MimeTypes.Type.APPLICATION_JSON.asString())
```
halibobo1205, line 537: if `contentType.contains(MimeTypes.Type.APPLICATION_JSON_UTF_8.asString()) == ture`
then  `contentType.contains(MimeTypes.Type.APPLICATION_JSON.asString() must be true.

halibobo1205:
@forfreeday Please add test for validation.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -535,7 +534,9 @@ private static String checkGetParam(HttpServletRequest request, String key) thro
       if (StringUtils.isBlank(contentType)) {
         return null;
       }
-      if (APPLICATION_JSON.toLowerCase().contains(contentType)) {
```
lurais, line 538: Should the contentType to lower case too ?

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -545,7 +546,7 @@ private static String checkGetParam(HttpServletRequest request, String key) thro
         if (jsonObject != null) {
           return jsonObject.getString(key);
         }
-      } else if (APPLICATION_FORM_URLENCODED.toLowerCase().contains(contentType)) {
+      } else if (contentType.contains(MimeTypes.Type.FORM_ENCODED.asString())) {
```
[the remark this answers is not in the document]
forfreeday: The characteristics of `String.contains()` dictate that when using `contains()`, you need to pay attention to its their containment relationship; longer characters can contain shorter characters, and conversely, shorter characters cannot contain longer characters.
```java
String str1 = "abc";
String str2 = "ab";
str1.contains(str2);
```
is valid, the reverse is not.
The `content-type` of the request could be any of the following:
`application/json;`
`application/json;charset=utf-8;`
`application/json; charset=utf-8;`
So you need to compare the constants with the content-type of the request.

framework/src/main/java/org/tron/core/services/http/Util.java
forfreeday: The `content-type` will be true if the condition is met.

framework/src/main/java/org/tron/core/services/http/Util.java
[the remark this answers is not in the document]
halibobo1205: `charset=utf-8;application/json; ` is also legal.

framework/src/main/java/org/tron/core/services/http/Util.java
forfreeday: defaults to lowercase.

framework/src/main/java/org/tron/core/services/http/Util.java
halibobo1205: I mean the last two rulings are not necessary,  if you insist on doing this, all the scenes need to be added

framework/src/main/java/org/tron/core/services/http/Util.java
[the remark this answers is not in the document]
forfreeday: Actually `contentType.contains(MimeTypes.Type.APPLICATION_JSON.asString()` is compatible with all scenarios, more judgment seems redundant

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -535,8 +533,10 @@ private static String checkGetParam(HttpServletRequest request, String key) thro
       if (StringUtils.isBlank(contentType)) {
         return null;
       }
-      if (APPLICATION_JSON.toLowerCase().contains(contentType)) {
-        value = getRequestValue(request);
+      if (contentType.contains(MimeTypes.Type.FORM_ENCODED.asString())) {
```
lurais, line 536: Shall we add some explaination about this change?

PR #5523, opened 2023-10-07
feat(db/trans-cache): verify cached files with crc32c

halibobo1205:
Use **CRC32**  to catch accidents like  disk write errors, etc.


jwrct (approved):

chainbase/src/main/java/org/tron/core/db2/common/TxCacheDB.java
```diff
@@ -288,6 +295,7 @@ private boolean recovery() {
   }
 
   private boolean recovery(int index, Path file) {
+    checkCrc32c(index, file);
```
lurais, line 298: Has this error happened in tests? Why will it happen? Will some error info printed when this kind of errors be found?

halibobo1205: To catch accidents like disk write errors, service crash etc.

> Will some error info printed when this kind of errors be found?

Yes,  there will be an exception.

jwrct (approved):

PR #5571, opened 2023-11-03
dependency(jetty): upgrade for security

halibobo1205:
CVE-2023-40167

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -70,8 +70,8 @@ dependencies {
     compile group: 'com.google.api.grpc', name: 'proto-google-common-protos', version: '2.15.0'
 
     // http
-    compile 'org.eclipse.jetty:jetty-server:9.4.49.v20220914'
-    compile 'org.eclipse.jetty:jetty-servlet:9.4.49.v20220914'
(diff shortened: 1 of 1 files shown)


framework/build.gradle
```diff
@@ -70,8 +70,8 @@ dependencies {
     compile group: 'com.google.api.grpc', name: 'proto-google-common-protos', version: '2.15.0'
 
     // http
-    compile 'org.eclipse.jetty:jetty-server:9.4.49.v20220914'
-    compile 'org.eclipse.jetty:jetty-servlet:9.4.49.v20220914'
+    compile 'org.eclipse.jetty:jetty-server:9.4.53.v20231009'
+    compile 'org.eclipse.jetty:jetty-servlet:9.4.53.v20231009'
     compile 'com.alibaba:fastjson:1.2.83'
```
317787106, line 75: Why not solve other dependency security also but only jetty? such as bcprov-jdk15on in https://github.com/tronprotocol/libp2p/pull/74/files.

halibobo1205: if  `bcprov-jdk15on` is required, it will be a stand-alone PR.

PR #5575, opened 2023-11-08
feat(transaction/validate): advance transaction validation when producing blocks

### LINKED ISSUE #5564: Advance transaction validation when producing blocks
# Rationale

Transaction execution requires some resources to be requested, and this process can be time-consuming, so transaction validation should be done in advance to avoid unnecessary waste of resources.

## Current Transaction Validation
Tapos
TooBig
Expiration
ContractSize
Signature
Duplication

## Current Block Production Flow
image

# Implementation

 Transactions are validated first and executed after.
## New Block Production Flow
image

Yes

halibobo1205:
close #5564 .

### DIFF
--- a/framework/src/main/java/org/tron/common/validator/TransactionValidator.java
+++ b/framework/src/main/java/org/tron/common/validator/TransactionValidator.java
@@ -2,26 +2,34 @@ package org.tron.common.validator;
 
 import javax.annotation.PostConstruct;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.Pair;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import org.tron.api.GrpcAPI.Return.response_code;
 import org.tron.core.capsule.TransactionCapsule;
+import org.tron.core.exception.ContractSizeNotEqualToOneException;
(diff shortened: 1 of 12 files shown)


framework/src/main/java/org/tron/common/validator/TransactionValidator.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @PostConstruct
+  private void prepare() {
+    dupTransactionValidator
+        .nextValidator(bigTransactionValidator)
+        .nextValidator(expiredTransactionValidator)
+        .nextValidator(contractSizeValidator)
+        .nextValidator(taposTransactionValidator)
+        .nextValidator(signatureValidator);
+  }
```
forfreeday, line 34: The way you assembled the validator, the coupling is relatively high, can it be changed to a dynamic loading form to reduce the code coupling

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -260,6 +261,9 @@ public class Manager {
   private ExecutorService filterEs;
   private static final String filterEsName = "filter";
 
+  @Autowired
+  TransactionValidator transactionValidator;
```
forfreeday, line 265: Add a private modifier for control

PR #5583, opened 2023-11-14
feat(rpc): gRPC reflection service support

### LINKED ISSUE #5391: Expected to support gRPCurl tool for grpc interface invocation easily
# Rationale
Expected Java-tron to support gRPCurl tool for grpc interface invocation.

Java-tron supports HTTP、json-rpc and grpc APIs,  some users may usually test them through command line tool, For example, curl and grpcurl and so on.  For http and json-rpc API, they may use curl command, but for grpc call, gRPCurl is not supported by Java-tron. Although other tools can be used: postman or bloomRPC, but they may generally need to import proto files, which is not convenient enough for users to use.

tomatoishealthy:
close #5391

### DIFF
--- a/framework/src/main/java/org/tron/core/services/interfaceOnSolidity/RpcApiServiceOnSolidity.java
+++ b/framework/src/main/java/org/tron/core/services/interfaceOnSolidity/RpcApiServiceOnSolidity.java
@@ -2,6 +2,7 @@ package org.tron.core.services.interfaceOnSolidity;
 
 import com.google.protobuf.ByteString;
 import io.grpc.netty.NettyServerBuilder;
+import io.grpc.protobuf.services.ProtoReflectionService;
 import io.grpc.stub.StreamObserver;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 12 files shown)


halibobo1205 (approved):
LGTM

317787106:
@tomatoishealthy What does this service do?

framework/src/main/java/org/tron/core/services/interfaceOnSolidity/RpcApiServiceOnSolidity.java
```diff
@@ -124,6 +125,10 @@ public void start() {
       // add lite fullnode query interceptor
       serverBuilder.intercept(liteFnQueryGrpcInterceptor);
 
+      if (parameter.isRpcReflectionServiceEnable()) {
```
lurais, line 128: Should the parameter rpcReflectionServiceEnable control all the RpcApiServices? It seems to be not flexible to use only one parameter to control all .

jwrct (approved):

PR #5594, opened 2023-11-21
feat(net): fix bug of method toString in BlockCapsule

317787106:
**What does this PR do?**

- fix bug of method toString in BlockCapsule

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
@@ -56,7 +56,6 @@ public class BlockCapsule implements ProtoCapsule<Block> {
 
   private Block block;
   private List<TransactionCapsule> transactions = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


halibobo1205 (approved):

chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
```diff
@@ -314,7 +313,7 @@ public boolean hasWitnessSignature() {
 
   @Override
   public String toString() {
-    toStringBuff.setLength(0);
+    StringBuilder toStringBuff = new StringBuilder();
```
lurais, line 316: why not replace the origin field with StringBuffer?

317787106: @lurais toStringBuff is only used in toString() method, using local variable makes code easier to understand.

jwrct (approved):

PR #5597, opened 2023-11-23
feat(net): delete config item node.discovery.bind.ip

### LINKED ISSUE #5579: Delete network config item - node.discovery.bind.ip
# Rationale

In the configuration file ```config.conf```, you can configure the local IP address through configuration items, including LAN IP and WAN IP. The main role of IP is to be used in the node discovery phase to construct UDP messages that are sent to peers. Generally, the LAN address is configured through node.discovery.bind.ip, and the WAN address is configured through node.discovery.external.ip.

Setting the IP of P2pConfig in java-tron is divided into as following priorities:

1.  The external network IP / WAN IP obtained from the query URL through the libp2p module;
2.  The value obtained by reading the configuration item node.discovery.external.ip;
3.  The value obtained by reading the configuration item node.bind.bind.ip;

This IP is used to fill in the address or addressIpv6 field of the Endpoint from in the following 5 messages:

1.  HelloMessage
2.  PingMessage
3.  PongMessage
4.  FindNeighbors
5.  Neighbors

For Ethernet networking (main chain), an external network connection must be ensured, so that priority 1 is met. For local area networking (private chain), assume that it cannot obtain the external network IP. During the actual connection, the node obtains peer's IP through the data packet socket , and then encapsulates the IP into a ```Node``` and adds it to the peer's k bucket. 

See the method org.tron.p2p.discover.socket.P2pPacketDecoder#decode:
```
@Override
public void decode(ChannelHandlerContext ctx, DatagramPacket packet, List<Object> out)
    throws Exception {
  ByteBuf buf = packet.content();
  int length = buf.readableBytes();
  ...
  byte[] encoded = new byte[length];
  buf.readBytes(encoded);
  try {
    UdpEvent event = new UdpEvent(Message.parse(encoded), packet.sender());
    out.add(event);
  } catch (P2pException pe) {
    ...
  }
}
```
Local node doesn’t get peer’s ip from above messages. Therefore, it is actually not necessary to configure the LAN IP.

# Implementation

Delete the configuration item node.discovery.bind.ip in the configuration file to reduce redundant items. For nodes without ethernet access, the local ip is 127.0.0.1.

317787106:
**What does this PR do?**

- delete config item node.discovery.bind.ip
- update libp2p version, include snappy-java version

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -635,7 +635,7 @@ public class Args extends CommonParameter {
         config.hasPath(Constant.NODE_LISTEN_PORT)
             ? config.getInt(Constant.NODE_LISTEN_PORT) : 0;
 
-    bindIp(config);
+    bindIp();
     externalIp(config);
 
     PARAMETER.nodeP2pVersion =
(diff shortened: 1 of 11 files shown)


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1521,22 +1521,12 @@ private static FilterQuery getEventFilter(final com.typesafe.config.Config confi
...
-        }
-      }
-    } else {
-      PARAMETER.nodeDiscoveryBindIp = config.getString(Constant.NODE_DISCOVERY_BIND_IP).trim();
+  private static void bindIp() {
+    try (Socket s = new Socket("www.baidu.com", 80)) {
+      PARAMETER.nodeDiscoveryBindIp = s.getLocalAddress().getHostAddress();
+    } catch (IOException e) {
+      logger.warn("Can't get bind IP. Fall back to 127.0.0.1: " + e);
+      PARAMETER.nodeDiscoveryBindIp = "127.0.0.1";
```
jwrct, line 1529: Why should this logic be retained?

317787106: if the external ip/W is not exist, use LAN IP.

xxo1shine (approved):

framework/src/main/java/org/tron/core/config/args/Args.java
halibobo1205: Why hardcoded into `www.baidu.com`?

317787106: @halibobo1205 It has been moved to libp2p module. This url is not necessary and we can use alternative way.

jwrct (approved):

PR #5603, opened 2023-11-27
test(plugins/dbLite): add test for merge function

halibobo1205:


plugins/src/test/java/org/tron/plugins/DbLiteTest.java
```diff
@@ -139,7 +139,7 @@ void testTools(String dbType, int checkpointVersion)
     }
     // start and validate the snapshot
     startApp();
-    generateSomeTransactions(6);
+    generateSomeTransactions(checkpointVersion == 1 ? 18 : 6);
```
lurais, line 142: Is it reasonable for the same function to produce the same input but have inconsistent actual execution logic?

halibobo1205: This PR is designed to test different logic.

Issue #5492, opened 2023-09-13
Slow precompiles for alt_bn128 curve


r0wdy1:
# Timeout issue

### Intro

Timeout restrictions on Tron have previously prevented the ability to deploy certain zkSNARK-based applications. This issue has been identified in the past and the root problem has not yet been addressed ( [https://github.com/tronprotocol/java-tron/issues/4311](https://github.com/tronprotocol/java-tron/issues/4311)). 

In this prs, we directly address this issue and propose an implementation that solves it ( [https://github.com/zkBob/java-tron/pull/1](https://github.com/zkBob/java-tron/pull/1),[https://github.com/zkBob/zksnark-java-sdk/pull/1](https://github.com/zkBob/zksnark-java-sdk/pull/1) ). By optimizing the bn128 precompiles, zk-based apps and zkSNARKs verification processes become usable on Tron. 

It’s important to note that these changes will not only enable the zkBob application on Tron, but also a whole class of applications and protocols based on elliptic curve cryptography and pairings in particular. The possibilities brought by ZKP are not limited to privacy focused protocols, some other examples include zk based auth for AA wallets, light clients, zk based bridges, computational integrity enforcing contracts, and many others.

### System information

java-tron version: `java -jar FullNode.jar -v`
OS & Version: Linux
Branch: master

### Expected behaviour

The `verifyProof`  method of a [contract](https://gist.github.com/r0wdy1/4609aaac16f9643ef5e97d70782d8992) for checking pairing over alt_bn128 curve as a part of Groth16 proof system which includes 6 point addition, 5 point multiplication and pairing check itself gives an appropriate result given a valid input. 

### Actual behaviour

A contract fails to return the result due to timeout error because of an unoptimized implementation of the bn128 precompiles ([Multiplication ,](https://github.com/tronprotocol/java-tron/blob/9e95599de2a9d60f4e706d18d3b71ac8850a0c01/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java#L772) [Addition](https://github.com/tronprotocol/java-tron/blob/9e95599de2a9d60f4e706d18d3b71ac8850a0c01/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java#L718) , [Pairing](https://www.notion.so/Timeout-issue-a20fc82aebd74f81afcb1f6d7888c8a0?pvs=21))

### Steps to reproduce the behaviour

********Reproducing on private network********

[...]

Issue #4311, opened 2022-02-28
Timeout in calls to comparatively low energy consuming contracts


raidshift:
### 1. What did you do?
* I would like to create a clone of Tornado Cash (https://tornado.cash/) on TRON  
* Therefore I needed to deploy a ZK SNARK Verifier Contract (which performs some calls to precompiled ECDSA contracts) on TRON Blockchain:
https://tronscan.io/#/contract/TAvhRwrobjBsNLpKpTTsNfSVMAkELzuufB

### 2. What did you expect to see?
* _function verifyProof(bytes memory proof, uint256[6] memory input) public view returns (bool)_
to be executed without errors like in any other EVM compatible Network

* Test Parameters (for a simple test which should return the value _false_):
**proof** = 0x0c3f55fb97b02d2c507f8c5b6b87b8f8872939f07959f0f671bd83c767d42ca52e718ff1621ebe51cc7c831d39f5a3da25d1cd7db0b4d9e49d216762a3385f14273b6f3daf5b6f3e3ed77c0508117f3d651ddd6e8ac4f2b350dfca929d005504132acfbef922dff2af5a98e4f0d2115fbd8af99a30ebe18420b85ee2961b6dc905a114775273fcccfea094fa06ca9e0325b71bae23c6830f0a2c10737a5543f52072054a099fe527a0941005a44e8d0b055aa4065df5d90bd3966f242e5ad2ec0a9227d4ecdf1fbbfde91b96110e2259fba35fda009e3eaf2126af0d4d242c912e70a7083394fee6a6bc05ebd0d78f45621caa1053bbd2c1f0efde16e90cc59f
**input** = ["0","0","0","0","0","0"]

### 3. What did you see instead?

* When executing via [TronScan](https://tronscan.io/#/contract/TAvhRwrobjBsNLpKpTTsNfSVMAkELzuufB):
class org.tron.core.vm.program.Program$OutOfTimeException : CPU timeout for 'SWAP1' operation executing

* When executing via [EtherScan](https://etherscan.io/address/0xce172ce1f20ec0b3728c9965470eaf994a03557a#readContract):
a valid response: _false_

To sum it up it seems that it is not possible to call this kind of contracts (even if there are calls to low energy cosuming precompiled ECDSA contracts within the contract) due to harsh timeout restrictions on TRON

vivian-kang:
[...]

PR #5611, opened 2023-12-11
feat(precompiles): optimize native implementation of the bn128 precompiles

AllFi:
**What does this PR do?**
This PR optimizes `BN128Multiplication`, and `BN128Pairing` precompiled contracts by implementing Montgomery form arithmetic. It is worth mentioning that this PR makes `BN128Addition` a little bit slower because of the overhead of the conversion to and from the Montgomery form.

The current implementation of these precompiles is relatively slow. In general, this hinders on-chain zkSNARKs verification and, consequently, makes zk-based apps almost unfeasible (https://github.com/tronprotocol/java-tron/issues/4311). This PR is aiming to mitigate this [issue](https://github.com/tronprotocol/java-tron/issues/5492). 

This PR is far worse than https://github.com/tronprotocol/java-tron/pull/5507 in terms of optimization but it doesn't require JNI and doesn't introduce a lot of changes in the implementation.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**
I've performed some benchmarks before and after these modifications. The results are below:

The average time of operations before (Intel(R) Core(TM) i7-10750H CPU, 32 GB RAM):
```
BN128Addition cost: 41165 ns
BN128Multiplication cost: 3490156 ns
BN128Pairing cost (2 pairs): 87665281 ns
```
The average time of operations after:
```
BN128Addition cost: 48147 ns (+17%)
BN128Multiplication cost: 2393792 ns (-31%)
BN128Pairing cost (2 pairs): 57961905 ns (-33%)
```


Hsenaai (approved):

Federico2014 (dismissed):

3for:
Hi, @AllFi , I add a test case `FpMulMontTest` in [https://github.com/3for/java-tron/tree/feature/bn128_montgomery_multiplication](https://github.com/3for/java-tron/tree/feature/bn128_montgomery_multiplication), and there's an uncatched exception "java.lang.NullPointerException". Maybe it would be better to add some code to handle the 'null Fp value' case.

AllFi:
Hi, @3for, @Federico2014 ! Thank you for the review!

There's indeed an uncaught exception, but it is only possible because access modifiers in the `Fp` were modified.

The current implementation doesn't expose any constructors/methods to instantiate `Fp` (`Fp2`, etc.) from outside the package. The only way to instantiate `Fp` objects from outside the package is by creating G1/G2 points. The code in that section checks that the created `Fp` values are not equal to null. 
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/crypto/src/main/java/org/tron/common/crypto/zksnark/BN128Fp.java#L50-L53
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/crypto/src/main/java/org/tron/common/crypto/zksnark/BN128Fp2.java#L55-L58
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/crypto/src/main/java/org/tron/common/crypto/zksnark/BN128G1.java#L44-L46
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/crypto/src/main/java/org/tron/common/crypto/zksnark/BN128G2.java#L61-L63

Precompiled contracts also check that the created points are not equal to null.
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java#L745-L753
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java#L799-L801
https://github.com/zkBob/java-tron/blob/410cce47e9375b89811bdde427d0126b301bc4b9/actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java#L865-L867

In summary, the current implementation assumes that it is the responsibility of the caller to check that the created `Fp` values are not equal to null. And the code in the `BN128Fp.create`, `BN128Fp2.create`, `BN128G1.create`, and `BN128G2.create` handles this case. If I'm not missing something, it seems impossible to cause a `NullPointerException` by calling precompiled contracts using any possible data.

When we perform operations within the package, we use constructors directly, so it also seems impossible to encounter a `NullPointerException`. We still use static methods for constants, but in this case, we can be certain that the values are correct.

Am I missing something?

3for:
I think it's okay for the caller to handle the 'null Fp value' case. There're another two issues:
1) It's better to check that `v` is non-negative: `0<=v<P`.
```
static Fp create(BigInteger v) {
    if (v.compareTo(P) >= 0 || v.signum() == -1) {
      // Only the values less than P, and not be negative are valid
      return null;
    }
    return new Fp(toMontgomery(v));
  }
```
2) It's better to check that `0<=s<order`, for `sP == (s+order)P`, and there exists such case in 32 bytes.
```
@Override
    public Pair<Boolean, byte[]> execute(byte[] data) {

      if (data == null) {
        data = EMPTY_BYTE_ARRAY;
      }

      byte[] x = parseWord(data, 0);
      byte[] y = parseWord(data, 1);

      byte[] s = parseWord(data, 2);

      BN128<Fp> p = BN128Fp.create(x, y);
      if (p == null) {
        return Pair.of(false, EMPTY_BYTE_ARRAY);
      }

      BN128<Fp> res = p.mul(BIUtil.toBI(s)).toEthNotation();

      BigInteger value = new BigInteger(1, s);
      BigInteger order = new BigInteger(
              "21888242871839275222246405745257275088548364400416034343698204186575808495617");
      byte[] s_a = value.add(order).toByteArray();
      BN128<Fp> res_a = p.mul(BIUtil.toBI(s_a)).toEthNotation();
      boolean flag_x= Arrays.equals(res.x().bytes(), res_a.x().bytes()); // is true
      boolean flag_y= Arrays.equals(res.y().bytes(), res_a.y().bytes()); // is true

      return Pair.of(true, encodeRes(res.x().bytes(), res.y().bytes()));
    }
  }
```

AllFi:
> 1. It's better to check that v is non-negative: 0<=v<P.

Yes, that does make sense. Thank you for pointing it out. Fixed in https://github.com/tronprotocol/java-tron/pull/5611/commits/04a00c2d2e6abf1fc75908994c87df6e6bf1ccb1.

> 2. It's better to check that 0<=s<order, for sP == (s+order)P, and there exists such case in 32 bytes.

I am not sure about it. I have the following concerns:
1. This check compromises backward compatibility. All the modifications made in this PR thus far maintain backward compatibility.
2. EIP196 specifies that multiplication with a scalar greater than the order should be successful.
   > Multiply point with scalar that is larger than the field order (should succeed).
3. There are some tests in go-ethereum where a scalar greater than the order is treated as valid input. https://github.com/ethereum/go-ethereum/blob/7124057bad16694d2b1f15dfe68a6109961b34ab/core/vm/testdata/precompiles/bn256ScalarMul.json#L59

That's why I believe it's better to keep it as is, but the decision is up to you.

3for:
It looks good to me now. Keeping `s` unchecked is okay.

Federico2014 (approved):

r0wdy1:
@3for @Federico2014 @lxcmyf @tomatoishealthy  is there any additional help or information that we can provide in regards of this PR? 
Is there any plan on going forwards with it?

lxcmyf:
> @3for @Federico2014 @lxcmyf @tomatoishealthy is there any additional help or information that we can provide in regards of this PR? Is there any plan on going forwards with it?

@r0wdy1 
Can this optimization meet your needs if the current configuration is set to a timeout of 80ms? Or can you optimize to complete the transaction within 80ms as much as possible?

lxcmyf:
> @lxcmyf , happy to hear from you again I wouldn't give any exact estimates until we do a thorough testing on Nile. As is the case with all of the optimizations the final results also depend on node configuration Despite that our first predictions were around 120 ms we have achieved a significant progress on our end and we've seen the pairing work for about 80 ms, but that should be proven on the testnet As you can see all of the github actions are stuck in pending state so we're patiently waiting for this PR to drop on Nile for further testing and optimizations.

@r0wdy1 
Okay, it's great to see your further breakthrough. We will confirm the specific merging branch and time, as well as some preparation work. If there is any progress, we will synchronize it as soon as possible.

Murphytron:
> @lxcmyf can or should we participate in the call too?

Hi @r0wdy1, if you feel you would contribute to the meeting by your attendance, please leave a comment here about the topics to be discussed: https://github.com/tronprotocol/pm/issues/75, and meanwhile email to [Murphy](mailto:murphy.zhang@tron.network), so we could invite you to the call, thank you.

xiangjie256329:
预计什么时候合并呢,我们想再复测一下之前跑不了的合约,合并到nile后会把超时时间改成80ms吗,不然120ms不会报错,之前的问题不会复现,或者把这个pr先更新到shasta上也行,总得有个和主网完全相同的环境吧

lxcmyf:
> > 预计什么时候合并呢,我们想再复测一下之前跑不了的合约,合并到nile后会把超时时间改成80ms吗,不然120ms不会报错,之前的问题不会复现,或者把这个pr先更新到shasta上也行,总得有个和主网完全相同的环境吧
> 
> Good advice, also concerned about future planning @lxcmyf

@tomatoishealthy 
At present, the PR is still undergoing further review and testing. If there are any latest developments, we will synchronize them here as soon as possible.

barbatos2011:
Hi @r0wdy1, there are two things we need to do before going to nile testnet.
1. code review [doing]
2. assess the impact of BN128Addition's duration increase [discussing]
Once these are finished, we can deploy the code to nile testnet and confirm the effect of this optimization.
Then one more thing before releasing the code, we need to revaluate the fee of these op codes to avoid some DOS attacks.
We will update the progress here later.

AllFi:
Hello! 

I would like to provide a bit more context about benchmarks after reading the [discussion](https://github.com/tronprotocol/pm/blob/master/TRON%20Core%20Devs%20Meetings/Meeting%2012.md). I've used "average" test cases (not involving infinity and low scalars) from go-ethereum set to benchmark the code. You can find the exact test that I've used to perform benchmark here: https://github.com/zkBob/java-tron/commit/1555a372547b8fc5b5d218d0b5d56d8e2fded0b4. The only difference is that I've performed more iterations than specified in this commit.

barbatos2011:
> Hello!
> 
> I would like to provide a bit more context about benchmarks after reading the [discussion](https://github.com/tronprotocol/pm/blob/master/TRON%20Core%20Devs%20Meetings/Meeting%2012.md). I've used "average" test cases (not involving infinity and low scalars) from go-ethereum set to benchmark the code. You can find the exact test that I've used to perform benchmark here: [zkBob@1555a37](https://github.com/zkBob/java-tron/commit/1555a372547b8fc5b5d218d0b5d56d8e2fded0b4). The only difference is that I've performed more iterations than specified in this commit.

Hi @AllFi , thanks for your information! Are the benchmarks' results mentioned above from this commit's test code, right?

AllFi:
Hi @barbatos2011! Yes, they are. To be more precise, I used the `bn128Bench` test from the branch `zkBob:feature/bn128_montgomery_multiplication` from this PR to measure the durations of precompiles after the optimization and I used the same test from the [branch](https://github.com/zkBob/java-tron/commits/bn128-native-bench/) mentioned earlier to measure durations of existing implementations of precompiles.

barbatos2011:
> Hi @barbatos2011! Yes, they are. To be more precise, I used the `bn128Bench` test from the branch `zkBob:feature/bn128_montgomery_multiplication` from this PR to measure the durations of precompiles after the optimization and I used the same test from the [branch](https://github.com/zkBob/java-tron/commits/bn128-native-bench/) mentioned earlier to measure durations of existing implementations of precompiles.

@AllFi OK, thanks again for this, I will share this information to other developers and will try it if necessary.

GuipaiQigong111:
Has the code within this pull request been implemented on other blockchains and undergone live deployment previously? Additionally, is there available documentation that provides an overview of the optimized computational logic?

AllFi:
Hi, @GuipaiQigong111! 

I am not sure what do you mean exactly. If you are asking if other blockchains use Montgomery form, then the answer is yes. All efficient implementations of these precompiles use Montgomery form as far as I know. Though they implement more complicated multiprecision integer arithmetics. If you are asking about the exact code from this PR, then the answer is no as far as I know. The original implementation of these precompiles was taken from https://github.com/ethereum/ethereumj and I am not aware of any project that uses it now apart from the Tron. I am also not aware of any more efficient implementations on Java. I tried implementing these precompiles using well-known arkworks library written on Rust but that [PR ](https://github.com/tronprotocol/java-tron/pull/5507)was postponed. That's why the goal of this PR was to optimize existing implementation by introducing as few modifications as possible.

The technique used in this PR is called Montgomery modular multiplication and there are a lot of resources that you can check. The implementation in this PR is probably the simplest one (since it doesn't involve multiprecision integers arithmetics) so this [Wikipedia page](https://en.wikipedia.org/wiki/Montgomery_modular_multiplication) might be enough to understand the code.

GuipaiQigong111:
@AllFi 
Okay, thank you for helping me clarify.

lxcmyf:
@AllFi 
Firstly, considering the readability and maintainability of the code, we would like to obtain a code level documentation. If other chains (such as Ethereum clients) have the same implementation, it can be noted in the documentation and compared if necessary. This kind of document not only helps us to have a deeper understanding of the principles and logic behind optimization, but also provides great convenience for future maintenance and expansion work. We would greatly appreciate it if you could provide some assistance in this regard.

Secondly, in order to ensure that the optimized code remains functionally consistent with the original code, we would like to know how you evaluate data consistency, as well as the evaluation methods and data you use. This helps us to use optimized code with more confidence and reduce potential risks.

Finally, we are also very concerned about the performance comparison data between the optimized code and the Ethereum Java client under the same opcode. Such data can intuitively reflect the effectiveness of your optimization and help us better evaluate the value of this improvement in practical applications.

We understand that these requests may involve some additional work for you, but please believe that these tasks are crucial for our project. We would greatly appreciate it if you could provide this information at your convenience.

AllFi:
Hello, @lxcmyf @yanghang8612! Sorry for the late reply.

### Documentation
As far as I know, most clients use the Montgomery form, but they implement multi-precision arithmetics, so the code in these implementations differs significantly from the current implementation. Even after optimization, the current implementation remains quite naive and unique in this regard. Therefore, I am not sure if comparing it with other implementations would be useful. Regarding the code-level documentation, could you please provide some examples for reference? It would also be helpful if you could point out which parts are unclear and require a better explanation.

### Tests
I've added test cases from [`go-ethereum`](https://github.com/ethereum/go-ethereum/tree/master/core/vm/testdata/precompiles) to evaluate the correctness of the implementation. Previously, there were no tests for these precompiles, or I might have missed them. You can find these test cases in `bn256Add.json`, `bn256Pairing.json`, and `bn256ScalarMul.json`. These test cases are executed in `PrecompiledContractsTest.bn128AdditionTest`, `PrecompiledContractsTest.bn128MultiplicationTest`, and `PrecompiledContractsTest.bn128PairingTest`, respectively.

### Benchmarks
I've conducted a simple comparative benchmark between `java-tron` and Hyperledger Besu. You can reassess Besu's performance by running these commands on your local machine.
```
git clone --recursive https://github.com/hyperledger/besu
cd ./besu
./gradlew installDist
cd ./build/install/besu/bin
./evmtool benchmark
./evmtool benchmark --nonative
```
It's worth noting that Hyperledger Besu has two implementations of these precompiles: the Java version and the native version. I've used the input data from Besu benchmarks in the `java-tron` benchmarks.

Here are the results:
* java-tron without any optimizations
```
BN128Addition cost 6.2 µs
BN128Multiplication cost 2,069.6 µs
BN128Pairing cost 29,369.5 µs
```
* java-tron with optimizations from this PR
```
BN128Addition cost 10.2 µs
BN128Multiplication cost 1,478.7 µs
BN128Pairing cost 20,105.1 µs
```
* `besu` with Java implementation:
```
AltBN128 Add    500 gas @    6.4 µs /    78.1 MGps
AltBN128 Mul 40,000 gas @7,273.9 µs /     5.5 MGps
AltBN128 2 pairing 180,000 gas @478,523.7 µs /     0.4 MGps
```
* `besu` with native implementation:
```
AltBN128 Add    500 gas @    2.8 µs /   179.5 MGps
AltBN128 Mul 40,000 gas @  113.5 µs /   352.5 MGps
AltBN128 2 pairing 180,000 gas @1,717.5 µs /   104.8 MGps
```
It seems that the Java implementation of `java-tron` is generally better than the Java implementation of Besu, but it significantly lags behind Besu's native implementation. I believe Besu uses the native implementation by default, so in practice, the `java-tron` implementation (even optimized with this PR) will be significantly slower than Besu's implementation. I've also conducted a benchmark with the [proposed](https://github.com/tronprotocol/java-tron/pull/5507) native implementation:
* `java-tron` with native implementation:
```
BN128Addition cost 1.9 µs
BN128Multiplication cost 109.3 µs
BN128Pairing cost 1,187.1 µs
```
As you can see, this implementation is comparable and even slightly faster than Besu's native implementation. I believe the only way to achieve comparable speed with other clients for these precompiles is by using a native implementation.

### BN128Addition precompile
I've tried to figure out something that might help but didn't manage to mitigate the constant overhead related to conversion between forms. There's always the option to keep the old implementation and use it only for this precompile. It will be a bit cumbersome, but it might be done.

yanghang8612:
> Hello, @lxcmyf @yanghang8612! Sorry for the late reply.
> 
> ### Documentation
> As far as I know, most clients use the Montgomery form, but they implement multi-precision arithmetics, so the code in these implementations differs significantly from the current implementation. Even after optimization, the current implementation remains quite naive and unique in this regard. Therefore, I am not sure if comparing it with other implementations would be useful. Regarding the code-level documentation, could you please provide some examples for reference? It would also be helpful if you could point out which parts are unclear and require a better explanation.
> 
> ### Tests
> I've added test cases from [`go-ethereum`](https://github.com/ethereum/go-ethereum/tree/master/core/vm/testdata/precompiles) to evaluate the correctness of the implementation. Previously, there were no tests for these precompiles, or I might have missed them. You can find these test cases in `bn256Add.json`, `bn256Pairing.json`, and `bn256ScalarMul.json`. These test cases are executed in `PrecompiledContractsTest.bn128AdditionTest`, `PrecompiledContractsTest.bn128MultiplicationTest`, and `PrecompiledContractsTest.bn128PairingTest`, respectively.
> 
> ### Benchmarks
> I've conducted a simple comparative benchmark between `java-tron` and Hyperledger Besu. You can reassess Besu's performance by running these commands on your local machine.
> 
> ```
> git clone --recursive https://github.com/hyperledger/besu
> cd ./besu
> ./gradlew installDist
> cd ./build/install/besu/bin
> ./evmtool benchmark
> ./evmtool benchmark --nonative
> ```
> 
> It's worth noting that Hyperledger Besu has two implementations of these precompiles: the Java version and the native version. I've used the input data from Besu benchmarks in the `java-tron` benchmarks.
> 
> Here are the results:
> 
> * java-tron without any optimizations
> 
> ```
> BN128Addition cost 6.2 µs
> BN128Multiplication cost 2,069.6 µs
> BN128Pairing cost 29,369.5 µs
> ```
> 
> * java-tron with optimizations from this PR
> 
> ```
> BN128Addition cost 10.2 µs
> BN128Multiplication cost 1,478.7 µs
> BN128Pairing cost 20,105.1 µs
> ```
> 
> * `besu` with Java implementation:
> 
> ```
> AltBN128 Add    500 gas @    6.4 µs /    78.1 MGps
> AltBN128 Mul 40,000 gas @7,273.9 µs /     5.5 MGps
> AltBN128 2 pairing 180,000 gas @478,523.7 µs /     0.4 MGps
> ```
> 
> * `besu` with native implementation:
> 
> ```
> AltBN128 Add    500 gas @    2.8 µs /   179.5 MGps
> AltBN128 Mul 40,000 gas @  113.5 µs /   352.5 MGps
> AltBN128 2 pairing 180,000 gas @1,717.5 µs /   104.8 MGps
> ```
> 
> It seems that the Java implementation of `java-tron` is generally better than the Java implementation of Besu, but it significantly lags behind Besu's native implementation. I believe Besu uses the native implementation by default, so in practice, the `java-tron` implementation (even optimized with this PR) will be significantly slower than Besu's implementation. I've also conducted a benchmark with the [proposed](https://github.com/tronprotocol/java-tron/pull/5507) native implementation:
> 
> * `java-tron` with native implementation:
> 
> ```
> BN128Addition cost 1.9 µs
> BN128Multiplication cost 109.3 µs
> BN128Pairing cost 1,187.1 µs
> ```
> 
> As you can see, this implementation is comparable and even slightly faster than Besu's native implementation. I believe the only way to achieve comparable speed with other clients for these precompiles is by using a native implementation.
> 
> ### BN128Addition precompile
> I've tried to figure out something that might help but didn't manage to mitigate the constant overhead related to conversion between forms. There's always the option to keep the old implementation and use it only for this precompile. It will be a bit cumbersome, but it might be done.

Great job. We'll look into it carefully.

lxcmyf:
> @lxcmyf
> 
> > Firstly, considering the readability and maintainability of the code, we would like to obtain a code level documentation. If other chains (such as Ethereum clients) have the same implementation
> 
> Actually the only [Ethereum Java client ](https://ethereum.org/en/developers/docs/nodes-and-clients/#execution-clients) [Besu](https://github.com/hyperledger/besu/) [uses](https://github.com/hyperledger/besu-native/tree/main/altbn128) Rust implementation via JNI for that
> 
> As you can see there's a whole besu-native project dedicated to using optimized math implementations in Besu Java code
> 
> Getting back to Montgomery multiplication implementation, do you need a math breakdown or a separate implementation write up? The core of the implementation is located in a single file (https://github.com/tronprotocol/java-tron/pull/5611/files#diff-99ee7c44ca170212da69fa123475bf66c7768696b855620b3e75ed73ba8ad695) and almost every line of code has a corresponding comment.
> 
> @yanghang8612 Basically, the idea boils down to transforming all the integers into the Montgomery form before doing any modulo additions and multiplications and transforming them back when everything is done. In Montgomery form the standard `mod` replaced by a more effective right shift operation. You can think of it as a kind of "batch" processing for modular operations: when multiple consecutive operations are executed ( such as in mutiplying an EC point by a scalar ) the speed gain is greater than the overhead of creating a batch ( i.e. transforming back and forth ) itself. With addition on the other hand the situation is the opposite - transformation overhead outweighs the gains.
@AllFi @r0wdy1 
Yes, I see that the relevant annotations are already complete enough. Thank you for your efforts in this regard. Can the comparison of data consistency before and after this optimization be reflected in unit testing? Finally, please forgive my intrusion. Is it possible for this optimization implementation to be submitted on Ethereum Besu?

yanghang8612:
@r0wdy1 Understood. Thanks for the explanation.

AllFi:
Hello, @lxcmyf!

I tested the old implementation (in a separate branch) with test cases mentioned earlier, and there weren't any differences compared to the current implementation. It is possible, but I suppose it would require keeping the old implementation intact. So, there would be two almost identical `org.tron.common.crypto.zksnark` packages and also two implementations of precompiles. If you think that it is necessary, it might be done, but as I mentioned earlier, the code will probably be a bit cumbersome.

I don't think this implementation makes a lot of sense for Besu. They already have a native implementation that works orders of magnitude faster than the Java implementation. I don't see any reason for them to be interested in optimizing the Java implementation by 30% or so.

lxcmyf:
> Hello, @lxcmyf!
> 
> I tested the old implementation (in a separate branch) with test cases mentioned earlier, and there weren't any differences compared to the current implementation. It is possible, but I suppose it would require keeping the old implementation intact. So, there would be two almost identical `org.tron.common.crypto.zksnark` packages and also two implementations of precompiles. If you think that it is necessary, it might be done, but as I mentioned earlier, the code will probably be a bit cumbersome.
> 
> I don't think this implementation makes a lot of sense for Besu. They already have a native implementation that works orders of magnitude faster than the Java implementation. I don't see any reason for them to be interested in optimizing the Java implementation by 30% or so.

So that's how it is. Thank you for informing me.

ghost (approved):

1charm1 (approved):

1charm1 (approved):

1charm1 (approved):

SeriousCoding789:
Thanks @AllFi for this work. We're closing this PR in favor of #6374.

As noted in the PR description itself, this approach is dominated by the arkworks PR in #5507, and the BN128 benchmarks collected in #6374 show that the chosen `besu-gnark` implementation outperforms both alternatives by a wide margin (BN128Pairing ~656μs vs ~35,772μs here). The non-JNI advantage of this PR may no longer load-bearing because #6374 addresses the JNI safety concern at the architectural leve.

Please follow #6374 for ongoing progress and update.

PR #5654, opened 2024-01-08
feat(db): optimize calculation services for reward when the node starts

halibobo1205:
We can start working on computational tasks immediately without waiting for a minimum of one block to synchronize if the new reward algorithm is already in effect when the node starts.

### DIFF
--- a/chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
+++ b/chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
@@ -176,7 +176,7 @@ public class RewardViCalService {
 
     Sha256Hash rewardViRootLocal = MerkleTree.getInstance().createTree(ids).getRoot().getHash();
     if (!Objects.equals(rewardViRoot, rewardViRootLocal)) {
-      logger.error("merkle root mismatch, expect: {}, actual: {}",
+      logger.warn("merkle root mismatch, expect: {}, actual: {}",
           rewardViRoot, rewardViRootLocal);
     }
     logger.info("calcMerkleRoot: {}", rewardViRootLocal);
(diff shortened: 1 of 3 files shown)


chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
```diff
@@ -75,8 +74,14 @@ public RewardViCalService(@Autowired  DynamicPropertiesStore propertiesStore,
...
-  @PostConstruct
-  private void init() {
+  public void init() {
+    // after init, we can get the latest block header number from db
+    this.newRewardCalStartCycle = this.getNewRewardAlgorithmEffectiveCycle();
+    boolean ret = this.newRewardCalStartCycle != Long.MAX_VALUE;
+    if (ret) {
+      // checkpoint is flushed to db, we can start rewardViCalService immediately
+      lastBlockNumber = Long.MAX_VALUE;
+    }
```
ss3344520, line 84: There is no need to define this variable `ret`.
The style needs to be consistent,  `lastBlockNumber` -> `this.lastBlockNumber`

chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
```diff
@@ -75,8 +74,14 @@ public RewardViCalService(@Autowired  DynamicPropertiesStore propertiesStore,
...
     es.scheduleWithFixedDelay(this::maybeRun, 0, 3, TimeUnit.SECONDS);
```
ss3344520, line 85: What is the significance of this code in the enableNewRewardAlgorithm function? 

if (ret && lastBlockNumber == -1) {
      lastBlockNumber = this.getLatestBlockHeaderNumber();
    }

halibobo1205: For the block-catching case, record the block for which NewRewardAlgorithm is in effect, and wait for the block to be flushed before data computation to ensure data consistency.

jwrct (approved):

chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
ss3344520: This function getLatestBlockHeaderNumber obtains the data of the underlying database, right?

halibobo1205: Yes.

ss3344520 (approved):

PR #5667, opened 2024-01-12
feat(log): verify columns' length of HelloMessage

317787106:
**What does this PR do?**

-  verify columns of hellomessage, including address, sig & codeVersio, length should not be over 200

- avoid too long log for HelloMessage

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/message/handshake/HelloMessage.java
+++ b/framework/src/main/java/org/tron/core/net/message/handshake/HelloMessage.java
@@ -4,6 +4,7 @@ import com.google.protobuf.ByteString;
 import lombok.Getter;
 import org.apache.commons.lang3.StringUtils;
 import org.tron.common.utils.ByteArray;
+import org.tron.common.utils.DecodeUtil;
 import org.tron.common.utils.StringUtil;
 import org.tron.core.ChainBaseManager;
 import org.tron.core.capsule.BlockCapsule;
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/tron/core/net/message/handshake/HelloMessage.java
```diff
@@ -169,6 +170,22 @@ public boolean valid() {
       return false;
     }
 
+    int maxByteSize = 200;
```
jwrct, line 173: Is a maximum byte size of 200 too large? Can it be set smaller?

317787106: The target of method `valid` is to make sure that log's content is not too large, so the value of  maxByteSize  is not very important. The length of address may be 42 bytes or more (not certain), sig is 65, codeVersion may not stable(20 bytes?). Can you give a suggestion?

jwrct: Just take the maximum of the three values, for example 65.

framework/src/main/java/org/tron/core/net/message/handshake/HelloMessage.java
```diff
@@ -169,6 +170,22 @@ public boolean valid() {
...
+    ByteString sig = this.helloMessage.getSignature();
+    if (!sig.isEmpty() && sig.toByteArray().length > maxByteSize) {
+      return false;
+    }
+
+    ByteString codeVersion = this.helloMessage.getCodeVersion();
+    if (!codeVersion.isEmpty() && codeVersion.toByteArray().length > maxByteSize) {
+      return false;
+    }
+
```
ss3344520, line 188: The length judgment of address and version can be more specific.
If there are 3 multi-signatures, how long can the length be?

317787106: this signature only come from SR's sig of timestamp in RelayService. it's 65 stablely, not related to multi-signature. You can read the code:

```
      String sig =
          TransactionCapsule.getBase64FromByteString(msg.getSignature());
      byte[] sigAddress = SignUtils.signatureToAddress(hash.getBytes(), sig,
          Args.getInstance().isECKeyCryptoEngine());
```

framework/src/main/java/org/tron/core/net/message/handshake/HelloMessage.java
317787106: In fact, I don't test the true length of address ang sig, because i have no fastforword node to use. Only give a theoretical value. Give an upper value can be ok, but give an exact value is not necessary, because method checkHelloMessage will verify the address and sig again. The main purpose is to not record too many message in log file.

jwrct (approved):

PR #5406, opened 2023-08-09
feat(db): optimize old rewards withdrawal

halibobo1205:
close #5354

PR #5742, opened 2024-03-04
feat(*): optimize old rewards withdrawal

halibobo1205:
**What does this PR do?**
  1. remove useNewRewardAlgorithmFromStart
  2. code clear
  3. exit when the calculation task gets an exception
  4. tools: add `db root` to calculate the data root
  5. make reward-vi root configurable

  see  https://github.com/tronprotocol/java-tron/pull/5406#issuecomment-1968923924.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/chainbase/src/main/java/org/tron/core/db/common/iterator/StoreIterator.java
+++ b/chainbase/src/main/java/org/tron/core/db/common/iterator/StoreIterator.java
@@ -93,14 +93,12 @@ public final class StoreIterator implements org.tron.core.db.common.iterator.DBI
 
   @Override
   public byte[] getKey() {
-    checkState();
     checkValid();
     return dbIterator.peekNext().getKey();
   }
 
   @Override
   public byte[] getValue() {
-    checkState();
     checkValid();
     return dbIterator.peekNext().getValue();
   }
--- a/chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
+++ b/chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
@@ -25,9 +25,9 @@ import org.tron.common.error.TronDBException;
 import org.tron.common.es.ExecutorServiceManager;
 import org.tron.common.parameter.CommonParameter;
 import org.tron.common.utils.ByteArray;
+import org.tron.common.utils.MerkleRoot;
 import org.tron.common.utils.Pair;
 import org.tron.common.utils.Sha256Hash;
-import org.tron.core.capsule.utils.MerkleTree;
 import org.tron.core.db.common.iterator.DBIterator;
 import org.tron.core.db2.common.DB;
 import org.tron.core.store.DelegationStore;
@@ -178,7 +178,7 @@ public class RewardViCalService {
         .map(this::getHash)
         .collect(Collectors.toCollection(ArrayList::new));
 
-    Sha256Hash rewardViRootLocal = MerkleTree.getInstance().createTree(ids).getRoot().getHash();
+    Sha256Hash rewardViRootLocal = MerkleRoot.root(ids);
     if (!Objects.equals(rewardViRoot, rewardViRootLocal)) {
       logger.warn("Merkle root mismatch, expect: {}, actual: {}."
               + " If you are quite sure that there is no miscalculation (not on the main network)"
--- a/chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
+++ b/chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
@@ -3,7 +3,6 @@ package org.tron.core.service;
 import static org.tron.core.store.DelegationStore.DECIMAL_OF_VI_REWARD;
 import static org.tron.core.store.DelegationStore.REMARK;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Streams;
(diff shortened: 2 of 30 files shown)


chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
```diff
@@ -176,8 +180,11 @@ private void calcMerkleRoot() {
 
     Sha256Hash rewardViRootLocal = MerkleTree.getInstance().createTree(ids).getRoot().getHash();
     if (!Objects.equals(rewardViRoot, rewardViRootLocal)) {
-      logger.warn("merkle root mismatch, expect: {}, actual: {}",
-          rewardViRoot, rewardViRootLocal);
+      logger.warn("Merkle root mismatch, expect: {}, actual: {}."
+              + " If you are quite sure that there is no miscalculation (not on the main network)"
+              + ", please configure 'storage.merkleRoot.reward-vi = {}'"
+              + "(for a specific network such as Nile, etc.) to config.conf to fix the hints",
```
eodiandie, line 186: a small spelling error.'to config.conf' should be 'in config.conf'.

chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
```diff
@@ -55,10 +53,11 @@ public class RewardViCalService {
 
   private volatile long lastBlockNumber = -1;
 
-  @VisibleForTesting
-  @Setter
-  private Sha256Hash rewardViRoot = Sha256Hash.wrap(
-      ByteString.fromHex("9debcb9924055500aaae98cdee10501c5c39d4daa75800a996f4bdda73dbccd8"));
+  private static final String MAIN_NET_ROOT_HEX =
+      "9debcb9924055500aaae98cdee10501c5c39d4daa75800a996f4bdda73dbccd8";
```
eodiandie, line 57: this root hex is already configured in conf. here it is declared again. Should it be possible for this value declared in one place instead of two different files？

halibobo1205: This is to ensure that the main-net node does not need to change the configuration file.

chainbase/src/main/java/org/tron/core/db/common/iterator/StoreIterator.java
```diff
@@ -93,14 +93,12 @@ public boolean valid() {
...
   @Override
   public byte[] getKey() {
-    checkState();
     checkValid();
     return dbIterator.peekNext().getKey();
   }
 
   @Override
   public byte[] getValue() {
-    checkState();
```
317787106, line 103: There is a varible method `close` and a method with the same name, maybe you should use different names to distinguish them.

chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
```diff
@@ -99,25 +98,30 @@ private boolean isDone() {
...
           }
+        } else {
+          clearUp(false);
+          logger.info("rewardViCalService is no need to run");
         }
-      } else {
-        clearUp(false);
-        logger.info("rewardViCalService is no need to run");
       }
+    } catch (Exception e) {
```
lurais, line 122: In which case shall it throw an Exception?

plugins/src/main/java/org/tron/plugins/utils/MerkleTree.java
```diff
@@ -0,0 +1,94 @@
...
+
+@Getter
+public class MerkleTree {
+
+  private static volatile MerkleTree instance;
+  private List<Sha256Hash> hashList;
+  private List<Leaf> leaves;
+  private Leaf root;
+
+  public static MerkleTree getInstance() {
```
eodiandie, line 17: here it uses Double-Checked Locking to ensure that only one instance is created in a multi-threaded environment. However, this code suffers from concurrency issues. Double-Checked Locking is not always reliable in Java because it depends on the memory model and compiler optimizations.
A safer implementation of the singleton pattern uses static inner classes in a way that is thread-safe and efficient。
`   private static class MerkleTreeHolder {
        private static final MerkleTree instance = new MerkleTree();
    }

    public static MerkleTree getInstance() {
        return MerkleTreeHolder.instance;
    }`

chainbase/src/main/java/org/tron/core/service/RewardViCalService.java
halibobo1205: Such as `too many open files`, `not enough disk space`, etc.

plugins/src/main/java/org/tron/plugins/utils/MerkleTree.java
halibobo1205: This class has member variables and can not be used for concurrency,  need to modify the usage scenarios while labeling them as not being able to be concurrent.

chainbase/src/main/java/org/tron/core/db/common/iterator/StoreIterator.java
halibobo1205: No changes for now.

jwrct (approved):

OnlineshopDimond46 (approved):

OnlineshopDimond46 (approved):

PR #5688, opened 2024-01-22
fix(api): fix the bug of determine if triggerConstantContract succeeds

317787106:
**What does this PR do?**

- fix the bug of determine if triggerConstantContract succeeds

**Why are these changes required?**

- If contractResult from triggerConstantContract is revert , the method `getShieldedContractScalingFactor` mistakenly believes it to be correct. If we give a random parameter contractAddress to `getShieldedContractScalingFactor`, transaction's result of triggerConstantContract will be revert. 
-  This bug also occurs in method `isShieldedTRC20NoteSpent`.

**This PR has been tested by:**
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**
 There are logs as following in version 4.7.3:
```
08:35:47.595 ERROR [rpc-full-executor-0] [API](RpcApiService.java:2485) createShieldedContractParameters: 
java.lang.ArithmeticException: BigInteger: modulus not positive
        at java.math.BigInteger.mod(BigInteger.java:2458)
        at org.tron.core.Wallet.checkPublicAmount(Wallet.java:4090)
        at org.tron.core.Wallet.createShieldedContractParameters(Wallet.java:3504)
        at org.tron.core.Wallet$$FastClassBySpringCGLIB$$3f6b5781.invoke(<generated>)
        at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
        at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:783)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
        at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:753)
        at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed(MethodInvocationProceedingJoinPoint.java:89)
        at org.tron.common.prometheus.MetricAspect.walletAroundAdvice(MetricAspect.java:43)
        at sun.reflect.GeneratedMethodAccessor27.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs(AbstractAspectJAdvice.java:634)
[...]

PR #5746, opened 2024-03-05
feat(api): check if scalingFactor is bigger than 0

317787106:
**What does this PR do?**

- rollback the changes in https://github.com/tronprotocol/java-tron/pull/5688
- throw exception if scalingFactor <= 0 in checkPublicAmount.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -3862,12 +3862,11 @@ public class Wallet {
     TransactionExtention.Builder trxExtBuilder = TransactionExtention.newBuilder();
     Return.Builder retBuilder = Return.newBuilder();
     TransactionExtention trxExt;
-    Transaction trx;
 
     try {
       TransactionCapsule trxCap = createTransactionCapsule(trigger,
           ContractType.TriggerSmartContract);
-      trx = triggerConstantContract(trigger, trxCap, trxExtBuilder, retBuilder);
+      Transaction trx = triggerConstantContract(trigger, trxCap, trxExtBuilder, retBuilder);
 
       retBuilder.setResult(true).setCode(response_code.SUCCESS);
       trxExtBuilder.setTransaction(trx);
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -4127,12 +4129,11 @@ public byte[] getShieldedContractScalingFactor(byte[] contractAddress)
     TransactionExtention.Builder trxExtBuilder = TransactionExtention.newBuilder();
     Return.Builder retBuilder = Return.newBuilder();
     TransactionExtention trxExt;
-    Transaction trx;
 
     try {
       TransactionCapsule trxCap = createTransactionCapsule(trigger,
           ContractType.TriggerSmartContract);
-      trx = triggerConstantContract(trigger, trxCap, trxExtBuilder, retBuilder);
+      Transaction trx = triggerConstantContract(trigger, trxCap, trxExtBuilder, retBuilder);
```
halibobo1205, line 4136: Doesn't this need to be handled?

317787106: This part is rollback to 4.7.3.

halibobo1205: What is the reason for this part of the direct rollback ? I see the previous analysis which is also wrong

317787106: Only this part is added: 
```
if (scalingFactor.compareTo(BigInteger.ZERO) <= 0) {
      throw new ContractValidateException("scalingFactor must be positive");
}
```

Other part is still same as before. Check the condition outside this method is not only easy to understand and but also include all conditions.

317787106: @halibobo1205 what's wrong with previous analysis? This optimization is to coverage this additional case: contract and method `scalingFactor` are both exist, but it really return 0.

jwrct (approved):

halibobo1205 (approved):

Issue #4692, opened 2022-09-29
Java-tron - Release linux/arm64 builds


andreclaro:
## Description
Release `linux/arm64` builds for java-tron.

## Related Issues
- https://github.com/tronprotocol/java-tron/issues/4248
- https://github.com/tronprotocol/java-tron/issues/4525

## Why this feature is important
- Take advantages of running the nodes on top Arm CPU architecture machines, such as, AWS Graviton2/3. The main benefits are costs reduction and performance increase. For further details consult: https://aws.amazon.com/ec2/graviton and https://about.gitlab.com/blog/2021/08/05/achieving-23-cost-savings-and-36-performance-gain-using-gitlab-and-gitlab-runner-on-arm-neoverse-based-aws-graviton2-processor
" Our latest cost analysis and performance benchmarks show that customers can realize cost savings of up to 23% and performance gains of up to 36% by deploying the GitLab application and GitLab Runner on the Arm-based Graviton2 when compared to the x86 based M5/C5 EC2 instances." 

- Enable developers to run `java-tron` on their MacBook Pro M1/M2 laptops.

halibobo1205:
Due [ExchangeProcessor](https://github.com/tronprotocol/java-tron/blob/master/chainbase/src/main/java/org/tron/core/capsule/ExchangeProcessor.java#L19) use `Math.pow`  to calculate instead of `StrictMath.pow`,  `Math.pow` for jdk1.8_x86 being specially implemented.

In java8, the default behavior of Math is using any available FPU like 8087. The results of float operations might be platform-dependent.
In later versions, the default behavior changes and will use IEEE-float(SSE2 instructions). So you can't update JDK/JRE without a proposal-fork.

In detail, java-tron's double arithmetic will use x87 instructions. And is difficult to port to other languages or  platform, since all modern languages use ieee-compatible and stable arithmetic for float numbers.

ethan1844:
Thanks for your contribution to java-tron, this issue will be closed as no update for a long time.

Please feel free to re-open it if you still see the issue, thanks.

microcai:
so, tron use java, but can only run on x86_64 ?

PR #5845, opened 2024-05-31
RocksDB Linux ARM support.

donbeave:
**What does this PR do?**

Upgraded version to the latest minor version of rocksdbjni.

This upgrade added support for Linux ARM64. Related to https://github.com/tronprotocol/java-tron/issues/4692.

- Unit Tests
- Manual Testing

Can try this docker image, it built with these changes, it works on amd64/arm64 linux: https://hub.docker.com/r/donbeave/docker-tron-full-node

Here are Dockerfiles for this docker build: 
1) ARM64/AArch64: https://github.com/ChainArgos/docker-tron-full-node/blob/main/Dockerfile.arm64
2) x86-64: https://github.com/ChainArgos/docker-tron-full-node/blob/main/Dockerfile.amd64

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -37,7 +37,7 @@ dependencies {
     compile group: 'com.beust', name: 'jcommander', version: '1.72'
     compile group: 'com.typesafe', name: 'config', version: '1.3.2'
     compile group: leveldbGroup, name: leveldbName, version: leveldbVersion
-    compile group: 'org.rocksdb', name: 'rocksdbjni', version: '5.15.10'
+    compile group: 'org.rocksdb', name: 'rocksdbjni', version: '5.18.4'
     // https://mvnrepository.com/artifact/org.quartz-scheduler/quartz
     compile group: 'org.quartz-scheduler', name: 'quartz', version: '2.3.2'
     compile group: 'io.prometheus', name: 'simpleclient', version: '0.15.0'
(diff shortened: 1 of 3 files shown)


Joey139 (approved):

halibobo1205:
As far as I know, the following work is needed to support the ARM architecture:
`Unlike some of the numeric methods of class StrictMath, all implementations of the equivalent functions of class Math are not defined to return the bit-for-bit same results.`
 Because TRON uses `math.pow` instead of `strictMath.pow` in jdk8_x86_64, some numerical calculations are inconsistent across architectures or higher jdk. The code is shown below:
```java
    logger.debug("balance: " + balance);
    long newBalance = balance + quant;
    logger.debug("balance + quant: " + newBalance);

    double issuedSupply = -supply * (1.0 - Math.pow(1.0 + (double) quant / newBalance, 0.0005));
    logger.debug("issuedSupply: " + issuedSupply);
    long out = (long) issuedSupply;
    supply += out;

    return out;
  }
```
The test cases are as follows:
```java
 @Test
  public void testPow() {
    double x = 29218;
    double q = 4761432;
    double ret = Math.pow(1.0 + x / q, 0.0005);
    double ret2 = StrictMath.pow(1.0 + x / q, 0.0005);

    System.out.printf("%s%n", doubleToHex(ret)); //  3ff000033518c576
    System.out.printf("%s%n", doubleToHex(ret2)); // 3ff000033518c575
    Assert.assertEquals(0, Double.compare(ret, ret2)); // fail in jdk8_X86.
  }

  public static String doubleToHex(double input) {
    // Convert the starting value to the equivalent value in a long
    long doubleAsLong = Double.doubleToRawLongBits(input);
    // and then convert the long to a hex string
    return Long.toHexString(doubleAsLong);
  }
```

We are trying to solve this problem but don't have good ideas.

halibobo1205:
Trace on :
 - #5954

PR #5879, opened 2024-06-24
feat(net) : optimize the isIdle method

zeusoo001:
**What does this PR do?**
Optimize the isIdle method. refer to issue https://github.com/tronprotocol/java-tron/issues/5913

During the block synchronization process, if the broadcast list has not been received, the synchronization may fail. The detailed process is as follows:
1. After processing the `chain inventory` message, set `fetchFlag` to true.
2. The scheduler will execute the `startFetchSyncBlock` method to fetch the block, and at this time, `fetchFlag` will be set to false.
```
etchExecutor.scheduleWithFixedDelay(() -> {
 try {
   if (fetchFlag) {
     fetchFlag = false;
     startFetchSyncBlock();
   }
 } catch (Exception e) {
   logger.error("Fetch sync block error", e);
 }
}, 10, 1, TimeUnit.SECONDS);
```
3. Since `advInvRequest` is not empty at this time, `peer.isIdle()` returns false, so after this scheduling, the block is not obtained, but `fetchFlag` is set to false and cannot be set back, so the block cannot be obtained later.
```
private void startFetchSyncBlock() {
 HashMap<PeerConnection, List<BlockId>> send = new HashMap<>();
 tronNetDelegate.getActivePeer().stream()
     .filter(peer -> peer.isNeedSyncFromPeer() && peer.isIdle())
     .filter(peer -> peer.isFetchAble())
```
4. Since the block is not obtained, the peer status cannot be updated, the status check will fail, and the connection will be disconnected.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
+++ b/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
@@ -167,6 +167,10 @@ public class PeerConnection {
   }
 
   public boolean isIdle() {
+    return advInvRequest.isEmpty() && isSyncIdle();
+  }
+
+  public boolean isSyncIdle() {
     return syncBlockRequested.isEmpty() && syncChainRequested == null;
   }
 
--- a/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
+++ b/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
@@ -167,7 +167,7 @@ public class PeerConnection {
   }
 
   public boolean isIdle() {
-    return advInvRequest.isEmpty() && syncBlockRequested.isEmpty() && syncChainRequested == null;
+    return syncBlockRequested.isEmpty() && syncChainRequested == null;
   }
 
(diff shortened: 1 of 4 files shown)


jwrct (approved):

framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
```diff
@@ -167,7 +167,7 @@ public void setBlockBothHave(BlockId blockId) {
   }
 
   public boolean isIdle() {
-    return advInvRequest.isEmpty() && syncBlockRequested.isEmpty() && syncChainRequested == null;
```
halibobo1205, line 170: What were the considerations of this previous judgment?

zeusoo001: This logic has been around for a long time, and this special scenario may not have been considered before.

lxcmyf (approved):

PR #5884, opened 2024-06-25
feat(metric): add interceptors to monitor gRPC using Prometheus

halibobo1205:
**What does this PR do?**
  Add interceptors to monitor gRPC services using Prometheus, inspired by https://github.com/grpc-ecosystem/java-grpc-prometheus
    1. Decoupled from the ApiAccessInterceptor.
    2. Fix the response latency monitor of RPCs handled by the server, the `startCall` is asynchronous, so the latency is not recorded correctly.
  
  ```java
      Histogram.Timer requestTimer = Metrics.histogramStartTimer(
            MetricKeys.Histogram.GRPC_SERVICE_LATENCY, endpoint);
        Listener<ReqT> res = next.startCall(call, headers); //  Starts asynchronous processing of an incoming call.
        Metrics.histogramObserve(requestTimer);
        return res;
   ```
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/ratelimiter/PrometheusInterceptor.java
+++ b/framework/src/main/java/org/tron/core/services/ratelimiter/PrometheusInterceptor.java
@@ -1,0 +1,46 @@
+
+package org.tron.core.services.ratelimiter;
+
+import io.grpc.ForwardingServerCall;
+import io.grpc.Metadata;
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
+import io.grpc.Status;
+import io.prometheus.client.Histogram;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import org.tron.common.prometheus.MetricKeys;
+import org.tron.common.prometheus.Metrics;
+
+/**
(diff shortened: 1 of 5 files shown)


framework/src/main/java/org/tron/core/services/ratelimiter/PrometheusInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+    MonitoringServerCall(ServerCall<R, S> delegate) {
+      super(delegate);
+      this.requestTimer = Metrics.histogramStartTimer(
+          MetricKeys.Histogram.GRPC_SERVICE_LATENCY, getMethodDescriptor().getFullMethodName());
+    }
+
+    @Override
+    public void close(Status status, Metadata responseHeaders) {
+      Metrics.histogramObserve(requestTimer);
+      super.close(status, responseHeaders);
```
ss334452, line 43: After the call is completed, will the close method be executed? Where is it triggered?

ss334452 (approved):

CodeNinjaEvan (approved):

Joj501 (approved):

halibobo1205:
close with #6097.

PR #5895, opened 2024-07-01
feat(net): optimize fetch inventory message broadcast processing logic

zeusoo001:
**What does this PR do?**
Optimize fetch inventory message broadcast processing logic.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
@@ -55,7 +55,9 @@ public class FetchInvDataMsgHandler implements TronMsgHandler {
 
     FetchInvDataMessage fetchInvDataMsg = (FetchInvDataMessage) msg;
 
-    check(peer, fetchInvDataMsg);
+    boolean isAdv = isAdvInv(peer, fetchInvDataMsg);
+
+    check(peer, fetchInvDataMsg, isAdv);
 
     InventoryType type = fetchInvDataMsg.getInventoryType();
     List<Transaction> transactions = Lists.newArrayList();
@@ -64,6 +66,10 @@ public class FetchInvDataMsgHandler implements TronMsgHandler {
 
     for (Sha256Hash hash : fetchInvDataMsg.getHashList()) {
       Item item = new Item(hash, type);
+      if (isAdv) {
+        peer.getAdvInvSpread().invalidate(item);
+      }
+
       Message message = advService.getMessage(item);
       if (message == null) {
         try {
@@ -127,7 +133,21 @@ public class FetchInvDataMsgHandler implements TronMsgHandler {
     }
   }
 
-  private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg) throws P2pException {
+  public boolean isAdvInv(PeerConnection peer, FetchInvDataMessage msg) {
+    MessageTypes type = msg.getInvMessageType();
+    if (type == MessageTypes.TRX) {
+      return true;
+    }
+    for (Sha256Hash hash : msg.getHashList()) {
+      if (peer.getAdvInvSpread().getIfPresent(new Item(hash, InventoryType.BLOCK)) == null) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg,
+                     boolean isAdv) throws P2pException {
     MessageTypes type = fetchInvDataMsg.getInvMessageType();
 
     if (type == MessageTypes.TRX) {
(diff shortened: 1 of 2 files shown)


halibobo1205 (approved):

jwrct (approved):

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -127,7 +133,21 @@ private void sendPbftCommitMessage(PeerConnection peer, BlockCapsule blockCapsul
...
   }
 
-  private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg) throws P2pException {
+  public boolean isAdvInv(PeerConnection peer, FetchInvDataMessage msg) {
+    MessageTypes type = msg.getInvMessageType();
+    if (type == MessageTypes.TRX) {
+      return true;
+    }
+    for (Sha256Hash hash : msg.getHashList()) {
+      if (peer.getAdvInvSpread().getIfPresent(new Item(hash, InventoryType.BLOCK)) == null) {
```
daniel-cao-byte, line 142: 多次请求 BLOCK 仅是 `isAdvInv` 方法返回 `false`，并不会导致 `check` 方法失败吧？

ss334452: The second request will be treated as synchronous and the check will fail.

abc-x-t:
Hi! Is there a related issue that explains the background of this PR? Or is this PR still a work in progress?

troncommons:
The `release_v4.8.1` branch has been created. Please make sure to update the target branch of your PRs if they are intended for the next release.

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -64,6 +66,10 @@ public void processMessage(PeerConnection peer, TronMessage msg) throws P2pExcep
 
     for (Sha256Hash hash : fetchInvDataMsg.getHashList()) {
       Item item = new Item(hash, type);
+      if (isAdv) {
+        peer.getAdvInvSpread().invalidate(item);
```
barbatos2011, line 75: Could you explain the reason for adding this operation here?

xxo1shine: A broadcast list can only be requested once. After a peer has requested a list, it can be removed from the cache. If it is requested again the next time, the check will fail, thus preventing the same list from being requested multiple times.

barbatos2011: If requests are allowed multiple times, is it possible that it could lead to a potential attack?

xxo1shine: If multiple requests are allowed, there is a DDoS attack.

lvs0075 (approved):

lvs0075 (approved):

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -127,7 +133,21 @@ private void sendPbftCommitMessage(PeerConnection peer, BlockCapsule blockCapsul
     }
   }
 
-  private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg) throws P2pException {
+  public boolean isAdvInv(PeerConnection peer, FetchInvDataMessage msg) {
```
King31T, line 141: I have a question, `isAdvInv` function takes into account the scenario of block broadcasting. Why doesn't the `check` function consider the scenario of block broadcasting?

xxo1shine: The broadcast scenario is also considered. For example, transactions belong to broadcast. The block will determine whether it is broadcast or synchronized. However, the broadcast block does not require other checking logic.

King31T: Actually, the real question for me lies in your last sentence., why **the broadcast block does not require other checking logic** ? Is there a risk of duplicate requests of broadcast block as broadcast transaction?

King31T: No need to reply, I got it, Thx.

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
317787106: The logic is too jumpy, and there are no comments here, which makes it confusing. You could add comment like this: cache the Inventory sent to the peer; once a FetchInvData message is received from the peer, remove this Inventory from the cache. If the same FetchInvData request is received from the peer again and it is no longer in the cache, then reject the request.

xxo1shine: OK, annotation added.

PR #5944, opened 2024-08-07
feat(net): add some log and adjustment of disconnection strategy

317787106:
**What does this PR do?**

- add some log for isolated2 disconnection
- only disconnect from broadcasting peers when node is a lan node or isOpenFullTcpDisconnect is true.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
+++ b/framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
@@ -80,7 +80,7 @@ public class ResilienceService {
           .collect(Collectors.toList());
       if (!peers.isEmpty()) {
         int index = new Random().nextInt(peers.size());
-        disconnectFromPeer(peers.get(index), ReasonCode.RANDOM_ELIMINATION);
+        disconnectFromPeer(peers.get(index), ReasonCode.RANDOM_ELIMINATION, "random");
       }
     }
   }
@@ -98,14 +98,14 @@ public class ResilienceService {
             .filter(peer -> !peer.getChannel().isTrustPeer())
             .collect(Collectors.toList());
         Optional<PeerConnection> one = getEarliestPeer(peers);
-        one.ifPresent(peer -> disconnectFromPeer(peer, ReasonCode.BAD_PROTOCOL));
+        one.ifPresent(peer -> disconnectFromPeer(peer, ReasonCode.BAD_PROTOCOL, "lan node"));
       }
     }
   }
 
   private void disconnectIsolated2() {
     if (isIsolateLand2()) {
-      logger.info("Node is isolated, try to disconnect from peers");
+      logger.warn("Node is isolated, try to disconnect from peers");
       int peerSize = tronNetDelegate.getActivePeer().size();
 
       //disconnect from the node whose lastActiveTime is smallest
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
```diff
@@ -137,11 +138,14 @@ private void disconnectIsolated2() {
           logger.warn("Sort disconnectIsolated2 peers failed: {}", e.getMessage());
           return;
         }
-
+        int candidateSize = peers.size();
         if (peers.size() > disconnectSize) {
           peers = peers.subList(0, disconnectSize);
         }
-        peers.forEach(peer -> disconnectFromPeer(peer, ReasonCode.BAD_PROTOCOL));
+        logger.info("All peer Size:{}, plan size:{}, candidate size:{}, real size:{}", peerSize,
```
xxo1shine, line 145: This log will refresh the screen.

317787106: No, only isolated node will print this log but this condition is rare. Even in this case, its period is 30 seconds.

xxo1shine (approved):

framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
```diff
@@ -74,75 +74,80 @@ private void disconnectRandom() {
...
-        Optional<PeerConnection> one = getEarliestPeer(peers);
-        one.ifPresent(peer -> disconnectFromPeer(peer, ReasonCode.BAD_PROTOCOL));
-      }
+    if (!isLanNode()) {
+      return;
+    }
+    // disconnect from the node that has keep inactive for more than inactiveThreshold
+    // and its lastActiveTime is smallest
+    int peerSize = tronNetDelegate.getActivePeer().size();
+    if (peerSize >= CommonParameter.getInstance().getMinConnections()) {
```
fyyhtx, line 96: Is this conditional statement redundant?

317787106: It's not redundant, condition of disconnection is different:  peerSize > CommonParameter.getInstance().getMinActiveConnections().

fyyhtx: Why is it '>', shouldn't lan nodes be '=' in most scenarios?

317787106: It's ">=", not ">".

fyyhtx: I'm talking about why the comparison in the isLanNode method is '>'

317787106: You are right, I will revise it.

fyyhtx: So let's go back to the initial question: how do you think?

317787106: Even it's a lan node, it will disconnect from some nodes only if peerSize >= CommonParameter.getInstance().getMinConnections(). Later condition is more strict. It's not redundant.

CodeNinjaEvan (approved):

lvs007 (approved):

PR #6064, opened 2024-11-01
doc(README): update free space required

halibobo1205:
A 2TB drive is no longer sufficient to sync mainnet with default settings. A freshly synced node will consume about 2.2T without pruning. It's not reasonable to expect users to turn off their node every few weeks to prune, especially for SR where an offline node means no generating block. A freshly synced node will consume about 2.2 T today. It's currently increasing by about 1.67 GiB a day.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,13 +91,13 @@ Minimum:
 
 - CPU with 8 cores
 - 16GB RAM
-- 2TB free storage space to sync the Mainnet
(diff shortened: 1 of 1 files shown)


lvs007 (approved):

Jake151987 (approved):

CodeNinjaEvan (approved):

Joj501 (commented):
README update Free space

PR #6157, opened 2025-01-27
chore: changed shield link to chat gitter

bravesasha:
the old link was broken


bravesasha:
I understand, that's your right.

PR #6360, opened 2025-06-25
feat(net): add volatile modifier to peer attribute variables

### LINKED ISSUE #6310: gt lastNum error reported when synchronizing blocks
#### Software Versions
OS : Linux
Version : 4.8.0

#### Expected behavior
The node is running normally and no exception should be thrown.

#### Actual behavior
I deployed a node online and found an exception while checking the logs. I want to know what the specific error is, whether it has any impact on the system, and whether it will affect block synchronization or data consistency?

Image

#### Frequency
Only appeared once.

zeusoo001:
**What does this PR do?**
Add volatile modifier to peer attribute variables, to avoid concurrent access problems, Fixes #6310

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
+++ b/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
@@ -85,7 +85,7 @@ public class PeerConnection {
 
   @Getter
   @Setter
-  private TronState tronState = TronState.INIT;
+  private volatile TronState tronState = TronState.INIT;
 
   @Autowired
   private TronNetDelegate tronNetDelegate;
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
```diff
@@ -146,7 +146,7 @@ public class PeerConnection {
   private Map<BlockId, Long> syncBlockRequested = new ConcurrentHashMap<>();
   @Setter
   @Getter
```
SeriousCoding789, line 148: why `syncBlockToFetch` here not add this `volatile`, as you change the logic order at `SyncService.java`

xxo1shine: This field does not have a second assignment, so it does not need to be modified by volatile.

lvs0075 (approved):

PR #6368, opened 2025-06-30
fix(crypto): optimize the null witness

### LINKED ISSUE #6281: Invalid  witness address when the localwitness is null
### Software Versions

OS : Mac OS
Version : 4.7.7

### Expected behavior

When launching the full-node with the null localwitness config,  the node should run normally without an witness address .
```
localwitness = [
]
```

### Actual behavior

It generates an invalid witness address: `41dcc703c0e500b653ca82273b7bfad8045d85a470` even when the private key is null.

Image

After analysis, I find that even if the localwitness is null,  it will still execute `initWitnessAccountAddress` function.
```
 else if (config.hasPath(Constant.LOCAL_WITNESS)) {
      localWitnesses = new LocalWitnesses();
      List<String> localwitness = config.getStringList(Constant.LOCAL_WITNESS);
      localWitnesses.setPrivateKeys(localwitness);
      witnessAddressCheck(config);
      localWitnesses.initWitnessAccountAddress(PARAMETER.isECKeyCryptoEngine());
      logger.debug("Got privateKey from config.conf");
    }
```
From the `initWitnessAccountAddress` function, it will invoke the `fromPrivate` function.
```
  public void initWitnessAccountAddress(boolean isECKeyCryptoEngine) {
    if (witnessAccountAddress == null) {
      byte[] privateKey = ByteArray.fromHexString(getPrivateKey());
      final SignInterface ecKey = SignUtils.fromPrivate(privateKey,
          isECKeyCryptoEngine);
      this.witnessAccountAddress = ecKey.getAddress();
    }
  }
```
In this case, the `privKeyBytes` is [] instead of null, so it continues to execute `fromPrivate(new BigInteger(1, privKeyBytes))`. So, the empty private key is finally passed to the Bouncycastle library, which leads to an implicit error and outputs an invalid address.
```
  public static ECKey fromPrivate(byte[] privKeyBytes) {
    if (Objects.isNull(privKeyBytes)) {
      return null;
    }
    return fromPrivate(new BigInteger(1, privKeyBytes));
  }
```

Federico2014:
**What does this PR do?**
Closes #6281


bladehan1 (approved):

3for (approved):

framework/src/main/java/org/tron/core/metrics/node/NodeMetricManager.java
```diff
@@ -36,8 +37,10 @@ private void setNodeInfo(NodeInfo nodeInfo) {
 
     nodeInfo.setIp(Args.getInstance().getNodeExternalIp());
 
-    ByteString witnessAddress = ByteString.copyFrom(Args.getLocalWitnesses()
-        .getWitnessAccountAddress(CommonParameter.getInstance().isECKeyCryptoEngine()));
+    byte[] witnessAccountAddress = Args.getLocalWitnesses()
+        .getWitnessAccountAddress(CommonParameter.getInstance().isECKeyCryptoEngine());
+    ByteString witnessAddress = !ByteUtil.isNullOrZeroArray(witnessAccountAddress) ? ByteString
```
317787106: witnessAddress may be null, so it also should be considered in method `isActiveWitness`

Federico2014: If `witnessAddress` is null,   `isActiveWitness` will always return false and won't throw the exception. If necessary, we can add the test:
```
  @Test
  public void testContains() {
    List<ByteString> list = new ArrayList<>();
    Assert.assertFalse(list.contains(null));
    list.add(null);
    Assert.assertTrue(list.contains(null));
  }
````

Comment (Federico2014) [reply framework/src/main/java/org/tron/core/metrics/node/NodeMetricManager.java]: test added.

**Review by barbatos2011 — APPROVED** [review]

File: framework/src/main/java/org/tron/core/consensus/ConsensusService.java [hunk 2]
```diff
@@ -72,6 +72,8 @@ public void start() {
       if (null == witnessCapsule) {
         logger.warn("Witness {} is not in witnessStore.", Hex.toHexString(witnessAddress));
       }
+      // In multi-signature mode, the address derived from the private key may differ from
+      // witnessAddress.
       Miner miner = param.new Miner(privateKey, ByteString.copyFrom(privateKeyAddress),
           ByteString.copyFrom(witnessAddress));
       miners.add(miner);
```
lxcmyf, line 78: How to handle it when privateKeys.size() == 0 ?

Federico2014: If the `privateKeys.size() == 0`,  `miners` will be empty, `DposTask` will not execute. So we need not to handle it.
```
    if (!dposService.isEnable() || ObjectUtils.isEmpty(dposService.getMiners())) {
      return;
    }
```

chainbase/src/main/java/org/tron/common/utils/LocalWitnesses.java
```diff
@@ -85,8 +85,8 @@ private void validate(String privateKey) {
       privateKey = privateKey.substring(2);
     }
 
-    if (StringUtils.isNotBlank(privateKey)
-        && privateKey.length() != ChainConstant.PRIVATE_KEY_LENGTH) {
+    if (StringUtils.isBlank(privateKey) || (StringUtils.isNotBlank(privateKey)
+        && privateKey.length() != ChainConstant.PRIVATE_KEY_LENGTH)) {
```
raymondliu0711: I think this condition can be simplified

Federico2014: Done

halibobo1205 (approved):

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -405,31 +405,33 @@ public static void setParam(final Config config) {
...
+      byte[] witnessAddress = null;
       if (StringUtils.isNoneBlank(PARAMETER.witnessAddress)) {
-        byte[] bytes = Commons.decodeFromBase58Check(PARAMETER.witnessAddress);
-        if (bytes != null) {
-          localWitnesses.setWitnessAccountAddress(bytes);
+        witnessAddress = Commons.decodeFromBase58Check(PARAMETER.witnessAddress);
+        if (witnessAddress != null) {
           logger.debug("Got localWitnessAccountAddress from cmd");
         } else {
           PARAMETER.witnessAddress = "";
```
xxo1shine: If `privateKey` is configured incorrectly, an exception will be thrown and the system will exit. If `witnessAddress` is configured incorrectly, is it necessary to consider system exit?

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -405,31 +405,33 @@ public static void setParam(final Config config) {
...
           logger.warn(IGNORE_WRONG_WITNESS_ADDRESS_FORMAT);
         }
       }
-      localWitnesses.initWitnessAccountAddress(PARAMETER.isECKeyCryptoEngine());
+      localWitnesses.initWitnessAccountAddress(witnessAddress, PARAMETER.isECKeyCryptoEngine());
       logger.debug("Got privateKey from cmd");
     } else if (config.hasPath(Constant.LOCAL_WITNESS)) {
```
xxo1shine: In the following two scenarios, why not consider the `witnessAddress` configuration of the startup command, which should have the highest priority?
1. else if (config.hasPath(Constant.LOCAL_WITNESS))
2. (config.hasPath(Constant.LOCAL_WITNESS_KEYSTORE))

framework/src/main/java/org/tron/core/config/args/Args.java
Federico2014: You make a good point. I’ve already made the changes. Please check again.

framework/src/main/java/org/tron/core/config/args/Args.java
Federico2014: According to previous logic, the private key and `witnessAddress` should either both be read from the command or both from the configuration file, without considering any other cases.

PR #6369, opened 2025-06-30
feat(jsonrpc): jsonrpc set error resolver

### LINKED ISSUE #6330: `eth_call` does not return data like the `triggerconstantcontract` interface does
I'm testing a contract with a custom Error.

```
// SPDX-License-Identifier: GPL-3.0
pragma solidity ^0.8.4;

/// Insufficient balance for transfer. Needed `required` but only
/// `available` available.
/// @param available balance available.
/// @param required requested amount to transfer.
error InsufficientBalance(uint256 available, uint256 required);

contract TestToken {
    mapping(address => uint) balance;
    function transfer(address to, uint256 amount) public {
        if (amount > balance[msg.sender])
            revert InsufficientBalance({
                available: balance[msg.sender],
                required: amount
            });
        balance[msg.sender] -= amount;
        balance[to] += amount;
    }
    // ...
}
```

When I test the `transfer(address,uint256)` method of a contract deployed on a private chain through the `eth_call` interface, if a revert exception occurs, I can't get the returned data data and thus can't resolve the custom error. The interface returns the following:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "REVERT opcode executed",
        "data": "{}"
    }
}
```

But when testing through the `triggerconstantcontract` interface, there is a return result, as follows

```
{
    "result": {
        "result": true,
        "message": "REVERT opcode executed"
    },
    "energy_used": 1267,
    "constant_result": [
        "cf47918100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001"
    ],
    "transaction": {
        "ret": [
            {
                "ret": "FAILED"
            }
        ],
        "visible": true,
...
```

0xbigapple:
**What does this PR do?**
Support setting of data field when jsonrpc return error， from #6330 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


common/src/main/java/org/tron/core/exception/JsonRpcInternalException.java
waynercheung, line 1: Just modify JsonRpcInternalException itself? 
Do not modify TronException.

0xbigapple: > Just modify JsonRpcInternalException itself? Do not modify TronException.

Thank you for your suggestion. I have added a JsonRpcException as the parent class for JSON-RPC exceptions, which makes it easier to extend the data field for all JSON-RPC exceptions.   https://github.com/tronprotocol/java-tron/pull/6369/commits/d5752417d2772b75f8d723ea07c8bbfdea6800d9

common/src/main/java/org/tron/core/exception/jsonrpc/JsonRpcException.java
```diff
@@ -0,0 +1,32 @@
+package org.tron.core.exception.jsonrpc;
+
+import lombok.Getter;
+import org.tron.core.exception.TronException;
+
+@Getter
+public class JsonRpcException extends TronException {
+  private Object data = null;
```
317787106, line 8: why not define data's type as byte[] ?

SeriousCoding789: Object is more general than byte[], I think

0xbigapple: Yes, use the Object type because it is more flexible and can support various kinds of data.

317787106: get it.

317787106 (approved):

bladehan1 (approved):

barbatos2011 (approved):

kuny0707 (approved):

lvs0075 (approved):

PR #6371, opened 2025-06-30
feat(crypto):  optimize zksnark config

### LINKED ISSUE #6334: Optimize the configuration switch used for zkSNARK and shielded transaction
A switch option exists in the configuration used to determine whether to allow shielded transactions API.
```
node.fullNodeAllowShieldedTransaction
```
### zkSnark parameters initialization

The default value is true.  However, if we set `fullNodeAllowShieldedTransaction = false`, the FullNode will skip the zkSnark parameters initialization process, as follows:
```
public static void librustzcashInitZksnarkParams() {
  logger.info("init zk param begin");
  if (!JLibrustzcash.isOpenZen()) {
    logger.info("zen switch is off, zen will not start.");
    return;
  }
  String spendPath = getParamsFile("sapling-spend.params");
  String spendHash = "25fd9a0d1c1be0526c14662947ae95b758fe9f3d7fb7f55e9b4437830dcc6215a7ce3ea465"
      + "914b157715b7a4d681389ea4aa84438190e185d5e4c93574d3a19a";
  String outputPath = getParamsFile("sapling-output.params");
  String outputHash = "a1cb23b93256adce5bce2cb09cefbc96a1d16572675ceb691e9a3626ec15b5b546926ff1c"
      + "536cfe3a9df07d796b32fdfc3e5d99d65567257bf286cd2858d71a6";
  try {
    JLibrustzcash.librustzcashInitZksnarkParams(
        new LibrustzcashParam.InitZksnarkParams(spendPath, spendHash, outputPath, outputHash));
  } catch (ZksnarkException e) {
    logger.error("librustzcashInitZksnarkParams fail!", e);
  }
  logger.info("init zk param done");
}
```
In this case, the FullNode cannot process shielded transactions. An error will occur if it attempts to synchronize blocks that contain shielded transactions. Therefore, the zkSnark parameters must be initialized regardless of whether `fullNodeAllowShieldedTransaction` is set to true or false.

`fullNodeAllowShieldedTransaction` should be kept to control the shielded transaction API. So it is necessary to remove the following code snippet from the zkSnark parameters initialization function:
```
if (!JLibrustzcash.isOpenZen()) {
  logger.info("zen switch is off, zen will not start.");
  return;
}
```
### pushTransaction

Furthermore, `fullNodeAllowShieldedTransaction` also affects the behavior of `pushTransaction`. However, relying on it is not recommended, as the execution of shielded transactions is ultimately governed by the `ALLOW_SHIELDED_TRANSACTION` proposal.
```
if (isShieldedTransaction(trx.getInstance()) && !Args.getInstance()
    .isFullNodeAllowShieldedTransactionArgs()) {
  return true;
}
```
So the above code snippet should be replaced by:

```
if (isShieldedTransaction(trx.getInstance()) && !chainBaseManager.getDynamicPropertiesStore()
    .supportShieldedTransaction()) {
  return true;
}
```

Federico2014:
**What does this PR do?**
Close #6334

### DIFF
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -991,9 +991,18 @@ public class Args extends CommonParameter {
     PARAMETER.eventFilter =
         config.hasPath(Constant.EVENT_SUBSCRIBE_FILTER) ? getEventFilter(config) : null;
 
-    PARAMETER.allowShieldedTransactionApi =
-        !config.hasPath(Constant.ALLOW_SHIELDED_TRANSACTION_API)
-            || config.getBoolean(Constant.ALLOW_SHIELDED_TRANSACTION_API);
+    if (config.hasPath(Constant.ALLOW_SHIELDED_TRANSACTION_API)) {
+      PARAMETER.allowShieldedTransactionApi =
+          config.getBoolean(Constant.ALLOW_SHIELDED_TRANSACTION_API);
+    } else if (config.hasPath(Constant.NODE_FULLNODE_ALLOW_SHIELDED_TRANSACTION)) {
+      // for compatibility with previous configuration
+      PARAMETER.allowShieldedTransactionApi =
+          config.getBoolean(Constant.NODE_FULLNODE_ALLOW_SHIELDED_TRANSACTION);
+      logger.warn("Configuring [node.fullNodeAllowShieldedTransaction] will be deprecated. "
+          + "Please use [node.allowShieldedTransactionApi] instead.");
+    } else {
+      PARAMETER.allowShieldedTransactionApi = true;
+    }
 
     PARAMETER.zenTokenId = config.hasPath(Constant.NODE_ZEN_TOKENID)
         ? config.getString(Constant.NODE_ZEN_TOKENID) : "000000";
--- a/common/src/main/java/org/tron/core/Constant.java
+++ b/common/src/main/java/org/tron/core/Constant.java
@@ -248,6 +248,9 @@ public class Constant {
 
   public static final String EVENT_SUBSCRIBE_FILTER = "event.subscribe.filter";
 
+  public static final String NODE_FULLNODE_ALLOW_SHIELDED_TRANSACTION = "node"
+      + ".fullNodeAllowShieldedTransaction";
+
   public static final String ALLOW_SHIELDED_TRANSACTION_API = "node"
       + ".allowShieldedTransactionApi";
 
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -134,6 +134,7 @@ test {
     }
     maxHeapSize = "1024m"
     doFirst {
+        // Restart the JVM after every 100 tests to avoid memory leaks and ensure test isolation
         forkEvery = 100
         jvmArgs "-XX:MetaspaceSize=128m","-XX:MaxMetaspaceSize=256m", "-XX:+UseG1GC"
     }


bladehan1 (approved):

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -868,9 +868,9 @@ public boolean pushTransaction(final TransactionCapsule trx)
       TooBigTransactionException, TransactionExpirationException,
       ReceiptCheckErrException, VMIllegalException, TooBigTransactionResultException {
 
-    if (isShieldedTransaction(trx.getInstance()) && !Args.getInstance()
-        .isFullNodeAllowShieldedTransactionArgs()) {
-      return true;
+    if (isShieldedTransaction(trx.getInstance()) && !chainBaseManager.getDynamicPropertiesStore()
```
317787106, line 871: So how can we support shielded transaction through config item `committee.xxx` ?

Federico2014: To support shielded transactions, we need to turn on ` ALLOW_SHIELDED_TRANSACTION` proposal.  However, the corresponding code is commented out.

framework/src/main/java/org/tron/core/zen/ZksnarkInitService.java
```diff
@@ -24,11 +24,6 @@ private void init() {
   public static void librustzcashInitZksnarkParams() {
```
halibobo1205, line 27: Can this be optimized so that it is only initialized once in the JVM? Currently, a large number of unit tests repeatedly generate related files, causing insufficient space in the tmp folder.

Image

Image

Federico2014: ok, I will fix it.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -1831,7 +1831,7 @@ public BytesMessage getNullifier(ByteString id) {
 
   private long getBlockNumber(OutputPoint outPoint)
       throws BadItemException, ZksnarkException {
-    if (!getFullNodeAllowShieldedTransaction()) {
+    if (!getAllowShieldedTransactionApi()) {
       throw new ZksnarkException(SHIELDED_ID_NOT_ALLOWED);
     }
     ByteString txId = outPoint.getHash();
```
3for, line 1830: Replace the code with `checkAllowShieldedTransactionApi` would be cleaner.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1030,10 +1030,6 @@ public static void setParam(final Config config) {
         config.hasPath(Constant.NODE_SHIELDED_TRANS_IN_PENDING_MAX_COUNTS) ? config
             .getInt(Constant.NODE_SHIELDED_TRANS_IN_PENDING_MAX_COUNTS) : 10;
 
-    if (PARAMETER.isWitness()) {
-      PARAMETER.fullNodeAllowShieldedTransactionArgs = true;
```
3for, line 1034: This logic should not be removed?

framework/src/main/java/org/tron/core/zen/ZksnarkInitService.java
halibobo1205:
```java
private static final AtomicBoolean initialized = new AtomicBoolean(false);
If (initialized.compareAndSet(false, true)){
  do InitZksnark
}

```

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: the redundant `getAllowShieldedTransactionApi` is removed.

framework/src/main/java/org/tron/core/config/args/Args.java
Federico2014: I think it should be removed.  The witness can allow the `shieldedTransactionApi` freely.

framework/src/main/java/org/tron/core/zen/ZksnarkInitService.java
Federico2014: After optimization, the `initZksnarkParams` method is invoked only 6 times during tests, significantly fewer than before.

barbatos2011 (approved):

Issue #6339, opened 2025-06-04
Add a new API to query the real-time vote count of witness


317787106:
# Background
Currently, HTTP/gRPC has an interface `/wallet/listwitnesses` for querying the number of witness votes in the previous epoch. During the maintenance period, the node calculates the votes of all witnesses and selects the first 27 witnesses as SRs for the next epoch block production. There are the following problems:

1. This interface returns the JSON-format voting results of the previous epoch, which is unchanged in the current epoch, and lacks an interface for real-time query of the number of votes in the current epoch;

2. This interface sorts by the address of the witness instead of the number of votes. Users cannot view the overall ranking and need to sort manually; witnesses cannot easily view their ranking in the previous epoch;

We plan to modify the results of `/wallet/listwitnesses`, namely return results sorted in descending vote order of the witness; and paln to add a new interface to query all real-time witnesses, also sort them in descending order by the number of votes.

# Rationale

Why should this feature exist?

- For ordinary users, they can easily view the historical ranking and real-time ranking of each witness through the API to vote for themselves;

- For witnesses who want to become Super Represent (SR), they urgently need to know their real-time votes and rankings in the current epoch so as to attract votes and enter the top 27; for SRs who have entered the top 27 in the previous epoch, they urgently need to understand their leading advantage over the 28th witness and provide an early warning mechanism to remind them to increase votes and avoid losing.

What are the use-cases?

- Developers, SRs & SR candidates need this feature urgently.

# Specification

1. For the existing interface `/wallet/listwitnesses` for querying the voting results of the previous epoch, the current proto is:

```
rpc ListWitnesses (EmptyMessage) returns (WitnessList) {...}
```

Since no parameters can be taken in EmptyMessage, the list of witnesses returned by `getWitnessList` function is sorted in reverse order by vote count by default; if the vote counts are the same, they are sorted in lexicographical order by witness address.

2. A new interface `/wallet/listrealtimewitness` is added, which supports HTTP/gRPC and is used to query the real-time vote count of all witnesses in the current epoch. The results are sorted in reverse order by the vote count. If the vote count is the same, they are sorted in lexicographical order of the witness address. In order to avoid too large query results, pagination query is introduced:
```
rpc ListRealtimeWitness (PaginatedMessage) returns (WitnessList) {
};
```
[...]

PR #6373, opened 2025-07-01
feat(API): return witness list with latest vote count in desc order

Sunny6889:
**What does this PR do?**
Close https://github.com/tronprotocol/java-tron/issues/6339

Added 4 API endpoints:
- HttpFullNode, `/wallet/getpaginatednowwitnesslist`
- HttpFullNodeSolidity,` /walletsolidity/getpaginatednowwitnesslist`
- RpcFullNode, `protocol.Wallet/GetPaginatedNowWitnessList `
- RpcFullNodeSolidity, `protocol.WalletSolidity/GetPaginatedNowWitnessList `

- Unit Tests
  -  Logic
  -  Servlet
  -  API Layers for both Rpc and Http requests.
- Manual Testing

**Follow up**

**Extra details**
Performance testing result https://github.com/tronprotocol/java-tron/issues/6339#issuecomment-3018489707


common/src/main/java/org/tron/core/config/Parameter.java
```diff
@@ -107,6 +107,7 @@ public class DatabaseConstants {
     public static final int PROPOSAL_COUNT_LIMIT_MAX = 1000;
     public static final int EXCHANGE_COUNT_LIMIT_MAX = 1000;
     public static final int MARKET_COUNT_LIMIT_MAX = 1000;
+    public static final int WITNESS_LIST_LIMIT_MAX = 1000;
```
317787106: use WITNESS_COUNT_LIMIT_MAX instead ?

SeriousCoding789: How about WITNESS_LIST_COUNT_LIMIT_MAX ? cause witness_count may misleading others like the max number of witness on BlockChain

317787106: General not misleading. PROPOSAL_COUNT_LIMIT_MAX also refers to proposal list.

barbatos2011 (approved):

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -764,6 +770,96 @@ public WitnessList getWitnessList() {
...
+      return null;
+    }
+    if (limit > WITNESS_COUNT_LIMIT_MAX) {
+      limit = WITNESS_COUNT_LIMIT_MAX;
+    }
+
+    // It contains the final vote count at the end of the last epoch.
+    List<WitnessCapsule> witnessCapsuleList = chainBaseManager.getWitnessStore().getAllWitnesses();
+    if (offset >= witnessCapsuleList.size()) {
+      return null;
```
lxcmyf, line 796: Return null or empty list?

SeriousCoding789: @lxcmyf I checked, return both is fine.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -770,14 +771,23 @@ public WitnessList getWitnessList() {
...
+    
+    /*
+      In the maintenance period, the VoteStores will be cleared.
+      To avoid the race condition of VoteStores deleted but Witness vote counts not updated, return retry error.
+    */
+    if (chainBaseManager.getDynamicPropertiesStore().getStateFlag() == 1) {
+      long maintenanceLogicTime = chainBaseManager.getDynamicPropertiesStore().getMaintenanceSkipSlots() * BLOCK_PRODUCED_INTERVAL / 1000;
+      String message = "Maintenance clearing, please try again later after " + maintenanceLogicTime + " seconds.";
+      throw new MaintenanceClearingException(message);
+    }
```
SeriousCoding789, line 792: @lxcmyf @317787106 @barbatos2011 @kuny0707 please review again for this new corner case.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -764,6 +771,105 @@ public WitnessList getWitnessList() {
...
+   * In the current epoch, X changes votes to:
+   * - W2: 60 votes
+   * - W3: 80 votes
+   *
+   * Resulting vote changes:
+   * - W1: -100 (votes removed)
+   * - W2: +60 (new votes)
+   * - W3: +80 (new votes)
+   */
+  private Map<ByteString, Long> countVote(VotesStore votesStore) {
```
xxo1shine, line 843: Why not use the countVote function defined in MaintenanceManager? It feels a bit repetitive to implement it again.

SeriousCoding789: there is a delete VoteStore operation in MaintenanceManager logic

framework/src/main/java/org/tron/core/services/http/GetPaginatedNowWitnessListServlet.java
```diff
@@ -0,0 +1,52 @@
...
+      JsonFormat.merge(params.getParams(), build, params.isVisible());
+      fillResponse(build.getOffset(), build.getLimit(), params.isVisible(), response);
+    } catch (Exception e) {
+      Util.processError(e, response);
+    }
+  }
+
+  private void fillResponse(long offset, long limit, boolean visible, HttpServletResponse response)
+      throws IOException, MaintenanceClearingException {
+    GrpcAPI.WitnessList reply = wallet.getPaginatedNowWitnessList(offset, limit);
```
halibobo1205, line 45: Due to the real-time nature of witness vote count updates during query execution, pagination-based data retrieval may result in the following data integrity issues:

* **Data Duplication**: When new witness nodes are inserted into previously queried pages due to vote count changes, subsequent page results may shift, causing the same records to appear across multiple pages.

* **Missing Records**: When witness nodes are removed from earlier pages due to ranking changes, the forward shift of subsequent page data may cause certain records to be skipped entirely.

* **Temporal Inconsistency**: Each paginated request may capture a different temporal snapshot of the dataset, resulting in an incoherent view of the witness list across the entire pagination sequence.

framework/src/main/java/org/tron/core/services/interfaceOnSolidity/http/GetPaginatedNowWitnessListOnSolidityServlet.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.servlet.http.HttpServletResponse;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.tron.core.services.http.GetPaginatedNowWitnessListServlet;
+import org.tron.core.services.interfaceOnSolidity.WalletOnSolidity;
+
+@Component
+@Slf4j(topic = "API")
+public class GetPaginatedNowWitnessListOnSolidityServlet extends GetPaginatedNowWitnessListServlet {
```
halibobo1205, line 13: Miss GetPaginatedNowWitnessListOnPBFTServlet

framework/src/main/java/org/tron/core/services/RpcApiService.java
```diff
@@ -1872,6 +1885,18 @@ public void listWitnesses(EmptyMessage request,
       responseObserver.onCompleted();
     }
 
+    @Override
+    public void getPaginatedNowWitnessList(PaginatedMessage request,
```
halibobo1205, line 1862: L401 and L1889 use the same `getPaginatedNowWitnessListCommon `?

protocol/src/main/protos/api/api.proto
```diff
@@ -499,6 +499,8 @@ service Wallet {
     };
   };
 
+  rpc GetPaginatedNowWitnessList (PaginatedMessage) returns (WitnessList) {
```
halibobo1205, line 502: **Subsequent Requirements:**

1. **Upstream Protocol Documentation**: These changes require synchronization with the official protocol specifications maintained at https://github.com/tronprotocol/protocol

2. **Wallet-CLI Implementation Support**: Implementation updates should be integrated into the wallet command-line interface at https://github.com/tronprotocol/wallet-cli to support the query.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -764,6 +771,105 @@ public WitnessList getWitnessList() {
...
+      In the maintenance period, the VoteStores will be cleared.
+      To avoid the race condition of VoteStores deleted but Witness vote counts not updated, return retry error.
+    */
+    if (chainBaseManager.getDynamicPropertiesStore().getStateFlag() == 1) {
+      long maintenanceLogicTime = chainBaseManager.getDynamicPropertiesStore().getMaintenanceSkipSlots() * BLOCK_PRODUCED_INTERVAL / 1000;
+      String message = "Maintenance clearing, please try again later after " + maintenanceLogicTime + " seconds.";
+      throw new MaintenanceClearingException(message);
+    }
+    // It contains the final vote count at the end of the last epoch.
+    List<WitnessCapsule> witnessCapsuleList = chainBaseManager.getWitnessStore().getAllWitnesses();
```
halibobo1205, line 794: Given the current pseudo-pagination implementation on the server side, adopting a session-level caching mechanism could effectively address the identified issues without performance degradation.

framework/src/main/java/org/tron/core/services/interfaceOnSolidity/http/GetPaginatedNowWitnessListOnSolidityServlet.java
SeriousCoding789: PBFT I think will not be supported

framework/src/main/java/org/tron/core/services/http/GetPaginatedNowWitnessListServlet.java
SeriousCoding789: For **Data Inconsistency**, I think that is expected, actually here one page(1000 witness) will return all witness in real cases, as performance testing also mentioned the witness accounts increase really slowly.  For reliable data,  the client should use `/WalletSolidity/getpaginatednowwitnesslist`

framework/src/main/java/org/tron/core/services/RpcApiService.java
SeriousCoding789: One is WalletApi, the other is WalletSolidityApi

protocol/src/main/protos/api/api.proto
SeriousCoding789: Agree, will do that after this finalised.

framework/src/main/java/org/tron/core/Wallet.java
SeriousCoding789: I think the third-parties like TronScan can do this in their part. As for java-tron, this API is controlled by FullNode to open or not, setting the rate-limiter. For reliable data, I would suggest the client just use `/WalletSolidity/getpaginatednowwitnesslist`. Considering the WitnessVote transaction volume and increasing speed, I would prefer a simple solution.

common/src/main/java/org/tron/core/exception/MaintenanceClearingException.java
```diff
@@ -0,0 +1,20 @@
+package org.tron.core.exception;
+
+/**
+ * Maintenance clearing exception - thrown when system is in maintenance clearing state
+ * Please try again later
+ */
+public class MaintenanceClearingException extends TronException {
```
317787106: maybe you can rename the exception to MaintenanceUnavailableException.

framework/src/main/java/org/tron/core/Wallet.java
317787106: Set message as "It's unavailable in maintenance, please try again later". Don't need to specify the time, because it's not accurate.

common/src/main/java/org/tron/core/exception/MaintenanceClearingException.java
SeriousCoding789: Sounds good.

framework/src/main/java/org/tron/core/Wallet.java
SeriousCoding789: ok

Issue #5910, opened 2024-07-10
Implement eth_getBlockReceipts method


Isarafanikov:
# Background
I noticed that method `eth_getBlockReceipts` is not implemented in the java-tron node, making it incompatible with tooling that expects proper specification implementation (for example, TheGraph)

# Rationale
`eth_getBlockReceipts` is a part of the Ethereum [JSON-RPC Specification ](https://ethereum.github.io/execution-apis/api-documentation/) since [August 2023](https://github.com/ethereum/execution-apis/pull/438), and is based on the method with the same name present in various node implementations (geth, erigon) as well as in QuickNode and Alchemy.

# Specification
See `eth_getBlockReceipts` in https://ethereum.github.io/execution-apis/api-documentation/

# Implementation
Not yet implemented

317787106:
@Isarafanikov You can now use `/wallet/gettransactioninfobyblocknum` temporarily. Further progress may be discussed here.

Isarafanikov:
Thanks for the info. Sadly, using the custom method would require too much code modifications (and, in some cases, impossible due to closed source tooling), so I will have to wait until the endpoint is properly implemented

317787106:
@Isarafanikov We will read Ethereum's implementation of `eth_getBlockReceipts ` carefully to see if there are any difficulties and then incorporate it into the plan.

abn2357:
What is ‘the Graph’  of your tooling example in Background

DongDongSunny:
there is a jsonRpc method here in tron https://developers.tron.network/reference/eth_gettransactionreceipt, will it help?

317787106:
The only problem is that eth_getBlockReceipts may cost a lot of time, as it needs to convert the transactioninfo in one block to transactionReceipt, specially if there is too many events in this block.

simbadMarino:
Hello everyone!, I want to follow up on development plans to implement the eth_getBlockReceipts, is there a plan already to implement it?  Implementing this will significantly help to reduce processing costs for Data platforms like Token Terminal, Alchemy, QuickNode, The Graph, etc, thus closing gaps in new platform integrations.

Thanks!

simbadMarino:
> there is a jsonRpc method here in tron https://developers.tron.network/reference/eth_gettransactionreceipt, will it help?

[...]

PR #6379, opened 2025-07-02
feat(jsonrpc): implement eth_getBlockReceipts

0xbigapple:
**What does this PR do?**
Implement eth_getBlockReceipts method, from #5910 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
```diff
@@ -9,179 +9,155 @@
...
+      BlockCapsule blockCapsule,
+      TransactionInfo txInfo,
+      TransactionContext context,
+      long energyFee) {
+    // Set basic fields
+    this.blockHash = ByteArray.toJsonHex(blockCapsule.getBlockId().getBytes());
+    this.blockNumber = ByteArray.toJsonHex(blockCapsule.getNum());
+    this.transactionHash = ByteArray.toJsonHex(txInfo.getId().toByteArray());
+    this.transactionIndex = ByteArray.toJsonHex(context.index);
+    // Compute cumulativeGasTillTxn
```
SeriousCoding789: modify to "// Compute cumulative gas until this transaction"

0xbigapple: Thank you, this comment is much clearer. I have adopted it and made modifications. https://github.com/tronprotocol/java-tron/pull/6379/commits/e64fc1fe8fc78871d09fdc30a37568e992d964c6

bladehan1 (approved):

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -777,7 +789,116 @@ public TransactionReceipt getTransactionReceipt(String txId)
...
+   * @param txId the transaction ID
+   * @return TransactionContext containing index and cumulative values, or null if not found
+   */
+  private TransactionContext findTransactionContext(TransactionInfoList infoList,
+      ByteString txId) {
+
+    long cumulativeGas = 0;
+    long cumulativeLogCount = 0;
+
+    for (int index = 0; index < infoList.getTransactionInfoCount(); index++) {
```
lxcmyf, line 826: Suggest judging the size of the list as empty and then recycling it.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -777,7 +789,116 @@ public TransactionReceipt getTransactionReceipt(String txId)
...
+      TransactionInfo info = infoList.getTransactionInfo(index);
+      ResourceReceipt resourceReceipt = info.getReceipt();
+
+      if (info.getId().equals(txId)) {
```
lxcmyf, line 830: Need to check if `info.getId()` is null.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
lxcmyf, line 822: Use an enhanced for loop for cleaner iteration.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -777,7 +789,116 @@ public TransactionReceipt getTransactionReceipt(String txId)
...
+        return new TransactionContext(index, cumulativeGas, cumulativeLogCount);
+      } else {
+        cumulativeGas += resourceReceipt.getEnergyUsageTotal();
+        cumulativeLogCount += info.getLogCount();
```
lxcmyf, line 830: cumulativeGas and cumulativeLogCount only update but no use.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -777,7 +789,116 @@ public TransactionReceipt getTransactionReceipt(String txId)
...
+   * null if block not found
+   * @throws JsonRpcInvalidParamsException if the parameter format is invalid
+   * @throws JsonRpcInternalException if there's an internal error
+   */
+  @Override
+  public List<TransactionReceipt> getBlockReceipts(String blockNumOrTag)
+      throws JsonRpcInvalidParamsException, JsonRpcInternalException {
+    Block block = wallet.getByJsonBlockId(blockNumOrTag);
+    if (block == null) {
+      return null;
```
lxcmyf, line 853: Return null or empty list？

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -777,7 +789,116 @@ public TransactionReceipt getTransactionReceipt(String txId)
...
+   * @return List of TransactionReceipt objects for all transactions in the block
+   */
+  private List<TransactionReceipt> getTransactionReceiptsFromBlock(BlockCapsule blockCapsule,
+      TransactionInfoList transactionInfoList, long energyFee) {
+
+    List<TransactionReceipt> receipts = new ArrayList<>();
+    long cumulativeGas = 0;
+    long cumulativeLogCount = 0;
+
+    for (int index = 0; index < transactionInfoList.getTransactionInfoCount(); index++) {
```
lxcmyf, line 891: Need to check if the `transactionInfoList` is empty. If it is empty, it can be returned directly without going down.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: To be consistent with ethereum eth_getBlockReceipts logic, return null for blocks that are not found.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: > Suggest judging the size of the list as empty and then recycling it.

Hi, If the list is empty, the loop condition `index < 0` will skip the loop entirely and return null

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: > Need to check if `info.getId()` is null.

Thank you for your suggestion. Adding null checks can improve code robustness, but I believe TransactionInfo objects should contain valid IDs. A null ID typically indicates data corruption or a system error, which points to serious issues in java-tron. In most cases, this is an unnecessary check, and adding it in every loop will increase overhead.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: Hi, If the `transactionInfoList` is empty, the loop condition index < 0 will skip the loop entirely and return receipts which is empty ArrayList<>()

PR #6383, opened 2025-07-03
feat(tvm): implement TIP-6780

### LINKED ISSUE #765: TIP-6780: SELFDESTRUCT only in same transaction
```
tip: 6780
title: SELFDESTRUCT only in same transaction
author: lei19942016@hotmail.com
status: Draft
type: Standards Track
category: VM
created: 2025-05-29
```

## Summary

Ethereum has incorporated [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) in the Cancun upgrade, modifying the behavior of the `SELFDESTRUCT` opcode. To maintain compatibility with Ethereum's implementation of this opcode modification, TRON needs to introduce this EIP and correspondingly update the `SELFDESTRUCT` opcode behavior in the TVM.

## Abstract

This TIP changes the functionality of the `SELFDESTRUCT` opcode. The new functionality will be only to send all TRX in the account to the target, except that the current behaviour is preserved when `SELFDESTRUCT` is called in the same transaction a contract was created.

## Motivation

To maintain compatibility with Ethereum EVM instruction modifications, we are implementing corresponding changes to the `SELFDESTRUCT` opcode behavior in TRON.

**Original motivation from EIP-6780:**

The `SELFDESTRUCT` opcode requires large changes to the state of an account, in particular removing all code and storage. This will not be possible in the future with Verkle trees: Each account will be stored in many different account keys, which will not be obviously connected to the root account.

This EIP implements this change. Applications that only use `SELFDESTRUCT` to retrieve funds will still work. Applications that only use `SELFDESTRUCT` in the same transaction as they created a contract will also continue to work without any changes.

## Specification

The behaviour of `SELFDESTRUCT` is changed in the following way:

1. When `SELFDESTRUCT` is executed in a transaction that is not the same as the contract calling `SELFDESTRUCT` was created:

    - The current execution frame halts.
    - `SELFDESTRUCT` does not delete any data (including storage keys, code, or the account itself).
    - `SELFDESTRUCT` transfers the entire account balance to the target.
    - Note that if the target is the same as the contract calling `SELFDESTRUCT` there is no net change in balances.  Unlike the prior specification, TRX will not be burnt in this case.

2. When `SELFDESTRUCT` is executed in the same transaction as the contract was created:

    - `SELFDESTRUCT` continues to behave as it did prior to this TIP, this includes the following actions
        - The current execution frame halts.
        - `SELFDESTRUCT` deletes data as previously specified.
        - `SELFDESTRUCT` transfers the entire account balance to the target
        - The account balance of the contract calling `SELFDESTRUCT` is set to `0`.
    - Note that if the target is the same as the contract calling `SELFDESTRUCT` that TRX will be burnt.

A contract is considered created at the beginning of a create transaction or when a CREATE series operation begins execution (CREATE, CREATE2, and other operations that deploy contracts in the future).  If a balance exists at the contract's new address it is still considered to be a contract creation.

The `SELFDESTRUCT` opcode remains deprecated as specified in [TIP-652](https://github.com/tronprotocol/tips/blob/master/tip-652.md). Any use in newly deployed contracts is strongly discouraged even if this new behaviour is taken into account, and future changes to the TVM might further reduce the functionality of the opcode.

## Rationale

Getting rid of the `SELFDESTRUCT` opcode has been considered in the past, and there are currently no strong reasons to use it. This TIP implements a behavior that will attempt to leave some common uses of `SELFDESTRUCT` working, while reducing the complexity of the change on TVM implementations that would come from contract versioning.

Handling the account creation and contract creation as two distinct and possibly separate events is needed for use cases such as counterfactual accounts.  By allowing the `SELFDESTRUCT` to delete the account at contract creation time it will not result in stubs of counterfactually instantiated contracts that never had any on-chain state other than a balance prior to the contract creation. These accounts would never have any storage and thus the trie updates to delete the account would be limited to the account node, which is the same impact a regular transfer of TRX would have.

## Backwards Compatibility

This TIP requires a hard fork, since it modifies consensus rules.

Contracts that depended on re-deploying contracts at the same address using `CREATE2` (after a `SELFDESTRUCT`) will no longer function properly if the created contract does not call `SELFDESTRUCT` within the same transaction.

Previously it was possible to burn TRX by calling `SELFDESTRUCT` targeting the executing contract as the beneficiary.  If the contract existed prior to the transaction the TRX will not be burned.  If the contract was newly created in the transaction the TRX will be burned, as before.

## Test Cases

Test cases for this TIP can reference EIP-6780 test cases [`eip6780_selfdestruct`](https://github.com/ethereum/execution-spec-tests/tree/1983444bbe1a471886ef7c0e82253ffe2a4053e1/tests/cancun/eip6780_selfdestruct).

## Security Considerations

The following applications of `SELFDESTRUCT` will be broken and applications that use it in this way are not safe anymore:

1. Where `CREATE2` is used to redeploy a contract in the same place in order to make a contract upgradable. This is not supported anymore and other types of proxy contracts should be used instead.

2. Where a contract depended on burning TRX via a `SELFDESTRUCT` with the contract as beneficiary, in a contract not created within the same transaction.

## Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

raymondliu0711:
**What does this PR do?**
The implementation of TIP-6780 (tronprotocol/tips#765).

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


CodeNinjaEvan (approved):

lvs0075 (approved):

kuny0707 (approved):

paxospace:
Implement Tip-6780 merged 6 commit

paxospace:
raymondliu07111:feature/tip_6780_selfdestruct

PR #6393, opened 2025-07-09
feat(net): add rate limiting logic for P2P messages

### LINKED ISSUE #6297: P2P message rate limit
# Background
Java-tron currently does not limit the rate of processing P2P messages, but the processing capacity of network nodes is limited by physical resources such as bandwidth, CPU, and memory. Processing a large number of P2P messages will cause excessive consumption of resources. Competition for resources among different applications may cause the following problems:
- Risk of resource overload: Burst traffic or high concurrent requests may cause the node processing capacity to be saturated, affecting normal service response.
- Degradation of service quality (QoS): Critical traffic (such as transaction/block broadcast) may be delayed or lost because non-critical traffic occupies resources.

For example, SyncBlockChainMessage is sent to request synchronization of block lists, and FetchInvDataMessage is sent to request blocks or transactions. Java-tron currently does not limit the frequency of processing these request messages. If a large number of these request messages are received in a short period of time, when processing these messages, since the size of the reply message is much larger than the size of the request message, it is very likely that the node's bandwidth, CPU, memory and other system resources will be over-consumed, thereby affecting the stability of the node service.

By comparing with the code implementation of Ethereum, it is found that Ethereum has set corresponding qps limits for related messages. Before processing a message, it will first check whether the rate exceeds the limit. If it exceeds the limit, the message will no longer be processed and the corresponding error code will be responded.
- Status request frequency cannot exceed 5 qps
- Goodbye message frequency cannot exceed 1 qps
- Ping message frequency cannot exceed 5 qps
- GetMetaData request frequency cannot exceed 5 qps
- Block synchronization request no more than 64 blocks within 30s

# Rationale
You can use the frequency of normal message sending as a benchmark to set appropriate frequency limits for different messages without affecting normal functions. The current P2P messages of java-tron are sorted out as follows:
- SYNC_BLOCK_CHAIN: When remainNum = 0 (only a few blocks to be synchronized, less than 4000), it has been controlled by code logic and no rate limit is required. When remainNum > 0, the SYNC_BLOCK_CHAIN ​​message will be sent twice in succession under normal logic. At least 2000 blocks need to be processed before the next SYNC_BLOCK_CHAIN ​​message can be sent. Each block takes at least 2ms to process, and it takes at least 4s to process 2000 blocks. The rate of SYNC_BLOCK_CHAIN ​​messages can be limited, and the specific threshold will be determined after verification.
- FETCH_INV_DATA (synchronization phase): The block acquisition task is scheduled once per second, so under normal logic, it will only be acquired once within 1 second. The rate of FETCH_INV_DATA messages can be limited, and the specific threshold will be determined after verification.
- Under normal logic, the number of blocks at a time will not exceed 100 blocks. The number of blocks at a time can be limited to no more than 100 blocks.
- P2P_HELLO: Handshake message, which is sent only once in the entire life cycle of a peer, has been controlled by code logic.
- P2P_DISCONNECT: Disconnect message. After receiving the message, the connection will be disconnected and no response will be given. You don't need to consider the frequency limit for now.
- BLOCK_CHAIN_INVENTORY: SYNC_BLOCK_CHAIN ​​response message. If no SYNC_BLOCK_CHAIN ​​message is sent, an error will be reported and the connection will be disconnected when BLOCK_CHAIN_INVENTORY is received. This has been controlled by code logic.
- INVENTORY: In theory, if the other node sends an invalid list (such as a list of historical transaction blocks), the bandwidth cost is higher and it is difficult to limit the frequency. It can be controlled by limiting the transaction tps speed (but the current threshold is relatively large).
- FETCH_INV_DATA(Broadcast phase): You can only request lists that have been sent to the other party, and you cannot request the same list repeatedly. This has been controlled by code logic.
- BLOCK: The response message of the FETCH_INV_DATA message has been controlled by code logic.
- TRXS: The response message of the FETCH_INV_DATA message has been controlled by code logic.
- PBFT_COMMIT_MSG: PBFT messages. The PBFT function will not be enabled for the time being. After receiving the message, it will be discarded directly without any restrictions for the time being.

When the limit is exceeded, the message is discarded directly and the number of times the peer exceeds the limit is recorded for use by the peer scoring logic.

# Implementation

yes
yes

fyyhtx:
**What does this PR do?**
Add rate limiting logic for P2P messages. Closes #6297 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
@@ -161,7 +161,7 @@ public class FetchInvDataMsgHandler implements TronMsgHandler {
               + " message exceeds the rate limit");
         }
         if (fetchInvDataMsg.getHashList().size() > NetConstants.MAX_BLOCK_FETCH_PER_PEER) {
-          throw new P2pException(TypeEnum.BAD_MESSAGE, "fetch too more blocks, size:"
+          throw new P2pException(TypeEnum.BAD_MESSAGE, "fetch too many blocks, size:"
               + fetchInvDataMsg.getHashList().size());
         }
         for (Sha256Hash hash : fetchInvDataMsg.getHashList()) {
--- a/framework/src/test/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandlerTest.java
+++ b/framework/src/test/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandlerTest.java
@@ -126,7 +126,7 @@ public class FetchInvDataMsgHandlerTest {
     try {
       fetchInvDataMsgHandler.processMessage(peer, msg);
     } catch (Exception e) {
-      Assert.assertEquals("fetch too more blocks, size:101", e.getMessage());
+      Assert.assertEquals("fetch too many blocks, size:101", e.getMessage());
     }
     try {
       fetchInvDataMsgHandler.processMessage(peer, msg);


framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
```diff
@@ -178,9 +178,11 @@ private void processMessage(PeerConnection peer, byte[] data) {
           handshakeService.processHelloMessage(peer, (HelloMessage) msg);
           break;
         case P2P_DISCONNECT:
-          peer.getChannel().close();
-          peer.getNodeStatistics()
-                  .nodeDisconnectedRemote(((DisconnectMessage)msg).getReason());
+          if (peer.getP2pRateLimiter().tryAcquire(type.asByte())) {
```
SeriousCoding789, line 181: why disconnect need a rate limiter? As in your issue mentioned "After receiving the message, the connection will be disconnected and no response will be given."

framework/src/main/java/org/tron/core/net/P2pRateLimiter.java
```diff
@@ -0,0 +1,32 @@
+package org.tron.core.net;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.util.concurrent.RateLimiter;
+
+public class P2pRateLimiter {
+  private final Cache<Byte, RateLimiter> rateLimiters = CacheBuilder.newBuilder()
+      .maximumSize(32).build();
```
SeriousCoding789, line 9: here is a hidden size 32 limit, can make it a large one so it normally won't never hit the limit?

framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
```diff
@@ -164,6 +171,9 @@ public void setChannel(Channel channel) {
     }
     this.nodeStatistics = TronStatsManager.getNodeStatistics(channel.getInetAddress());
     lastInteractiveTime = System.currentTimeMillis();
+    p2pRateLimiter.register(SYNC_BLOCK_CHAIN.asByte(), 2);
+    p2pRateLimiter.register(FETCH_INV_DATA.asByte(), 2);
+    p2pRateLimiter.register(P2P_DISCONNECT.asByte(), 1);
```
SeriousCoding789: make this hardcode 2,2,1 number to use a static variable

framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
SeriousCoding789: Another question: is this rate limit set per Peer or shared by all Peers? Will this rate limiter effect the normal disconnect logic?

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -156,6 +164,10 @@ private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg) thr
         if (!peer.isNeedSyncFromUs()) {
           throw new P2pException(TypeEnum.BAD_MESSAGE, "no need sync");
         }
+        if (fetchInvDataMsg.getHashList().size() > MAX_FETCH_SIZE) {
+          throw new P2pException(TypeEnum.BAD_MESSAGE, "fetch too more blocks, size:"
```
Sunny6889, line 168: How about add a more customised message such as `TypeEnum.EXCEED_SIZE_LIMIT`

framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
bladehan1: Will relay nodes trigger speed limits? What about busy sync nodes on the network?

framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
fyyhtx: > Another question: is this rate limit set per Peer or shared by all Peers? Will this rate limiter effect the normal disconnect logic?

This rate limit is set per peer. This rate limiter does not affect the normal disconnection logic.

fyyhtx: > why disconnect need a rate limiter? As in your issue mentioned "After receiving the message, the connection will be disconnected and no response will be given."

To prevent the peer from sending a large number of repeated disconnect messages at once, it will not respond to the message, but will execute the disconnection logic.

framework/src/main/java/org/tron/core/net/P2pRateLimiter.java
fyyhtx: The message types defined by the current protocol are very limited, and the limit of 32 is sufficient for the foreseeable future.

framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
fyyhtx: > Will relay nodes trigger speed limits? What about busy sync nodes on the network?

For synchronization logic, relay nodes are no different from ordinary full nodes.

xxo1shine (approved):

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -156,6 +156,14 @@ private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg) thr
         if (!peer.isNeedSyncFromUs()) {
           throw new P2pException(TypeEnum.BAD_MESSAGE, "no need sync");
         }
+        if (!peer.getP2pRateLimiter().tryAcquire(fetchInvDataMsg.getType().asByte())) {
```
317787106, line 159: `consumerInvToFetch` task is scheduled every 30ms, but the limit  FETCH_INV_DATA_RATE is 3 times/s, there will be conflict.

fyyhtx: This rate limit is based on the premise of block synchronization.

common/src/main/java/org/tron/core/Constant.java
```diff
@@ -318,6 +318,9 @@ public class Constant {
 
   public static final String RATE_LIMITER_HTTP = "rate.limiter.http";
   public static final String RATE_LIMITER_RPC = "rate.limiter.rpc";
+  public static final String RATE_LIMITER_P2P_SYNC_BLOCK_CHAIN = "rate.limiter.p2p.syncBlockChain";
```
SeriousCoding789, line 321: I wonder should config file to add these new configurations with default values?

jwrct: I don't think it's necessary, they are not critical configuration items. The rate limit parameter configuration is a precautionary solution, these configuration items may never be used in practice.

fyyhtx: I don't think it's necessary, they are not critical configuration items. The rate limit parameter configuration is a precautionary solution, these configuration items may never be used in practice.

SeriousCoding789: ok

framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
```diff
@@ -259,6 +261,7 @@ private void processException(PeerConnection peer, TronMessage msg, Exception ex
           code = Protocol.ReasonCode.NO_SUCH_MESSAGE;
           break;
         case BAD_MESSAGE:
+        case RATE_LIMIT_EXCEEDED:
```
halibobo1205, line 264: This mapping doesn't seem quite right. `RATE_LIMIT_EXCEEDED` is a server capacity issue rather than a protocol violation. Should we use a more specific error type or create a dedicated rate limit handler?

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1041,6 +1044,18 @@ public static void setParam(final Config config) {
 
     PARAMETER.rateLimiterInitialization = getRateLimiterFromConfig(config);
 
+    PARAMETER.rateLimiterSyncBlockChain =
+        config.hasPath(Constant.RATE_LIMITER_P2P_SYNC_BLOCK_CHAIN) ? config
+            .getDouble(Constant.RATE_LIMITER_P2P_SYNC_BLOCK_CHAIN) : 3.0;
```
halibobo1205, line 1049: Please add a comment specifying the rate limit unit: is this per second, per minute, per hour, or something?

framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
fyyhtx: According to normal protocol logic, RATE_LIMIT_EXCEEDED will not appear. Once it appears, it is considered a protocol violation.

framework/src/main/java/org/tron/core/config/args/Args.java
fyyhtx: If no special instructions are given, the default is qps.

common/src/main/java/org/tron/core/Constant.java
```diff
@@ -318,6 +318,9 @@ public class Constant {
...
+  public static final String RATE_LIMITER_P2P_FETCH_INV_DATA = "rate.limiter.p2p.fetchInvData";
+  public static final String RATE_LIMITER_P2P_DISCONNECT = "rate.limiter.p2p.disconnect";
```
waynercheung, line 323: As user can set the value, so add this configuration to config file? such as config.conf.

framework/src/main/java/org/tron/core/net/P2pRateLimiter.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public void register(Byte type, double rate) {
+    RateLimiter rateLimiter = RateLimiter.create(Double.POSITIVE_INFINITY);
```
waynercheung, line 12: what about create the RateLimter with rate, such as:
`RateLimiter rateLimiter = RateLimiter.create(rate);`

framework/src/main/java/org/tron/core/net/P2pRateLimiter.java
```diff
@@ -0,0 +1,32 @@
...
+    if (rateLimiter == null) {
+      return;
+    }
+    rateLimiter.acquire();
+  }
+
+  public boolean tryAcquire(Byte type) {
+    RateLimiter rateLimiter = rateLimiters.getIfPresent(type);
+    if (rateLimiter == null) {
+      return true;
```
waynercheung, line 28: Is there any risk if return true for the unregistered type(rateLimter == null)?

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -156,6 +156,14 @@ private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg) thr
...
+          throw new P2pException(TypeEnum.RATE_LIMIT_EXCEEDED, fetchInvDataMsg.getType()
+              + " message exceeds the rate limit");
+        }
+        if (fetchInvDataMsg.getHashList().size() > NetConstants.MAX_BLOCK_FETCH_PER_PEER) {
+          throw new P2pException(TypeEnum.BAD_MESSAGE, "fetch too more blocks, size:"
```
waynercheung: use "fetch too many blocks" instead of "fetch too more blocks".

framework/src/test/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandlerTest.java
```diff
@@ -93,4 +102,36 @@ public void testSyncFetchCheck() {
...
+    P2pRateLimiter p2pRateLimiter = new P2pRateLimiter();
+    p2pRateLimiter.register(FETCH_INV_DATA.asByte(), 1);
+    p2pRateLimiter.acquire(FETCH_INV_DATA.asByte());
+    Mockito.when(peer.getP2pRateLimiter()).thenReturn(p2pRateLimiter);
+    FetchInvDataMsgHandler fetchInvDataMsgHandler = new FetchInvDataMsgHandler();
+
+    try {
+      fetchInvDataMsgHandler.processMessage(peer, msg);
+    } catch (Exception e) {
+      Assert.assertEquals("fetch too more blocks, size:101", e.getMessage());
```
waynercheung: use "fetch too many blocks" instead of "fetch too more blocks".

common/src/main/java/org/tron/core/Constant.java
fyyhtx: This is a repeated question; please refer to the previous response for details.

framework/src/main/java/org/tron/core/net/P2pRateLimiter.java
fyyhtx: The approach you mentioned has a problem: the initial capacity of the token bucket is only 1. Once the peer completes the handshake, it immediately enters the synchronization state. If the initial token count is 1, it may lead to insufficient tokens.

framework/src/main/java/org/tron/core/net/P2pRateLimiter.java
fyyhtx: Unregistered types return true, meaning no speed limit.

PR #6394, opened 2025-07-09
fix(net): fix for ambiguous reason code in disconnect messages

### LINKED ISSUE #5989: Ambiguous Reason Code in Disconnection Message
#### Software Versions

```
OS : Linux
Git : a8ad2a169e58946b5b8de6ecf7b7ef5b8db05aff
Version : 4.7.5
Code : 18306
```

#### Expected behaviour
When disconnected, the exact reason for the disconnection can be obtained through the disconnection message.

#### Actual behaviour
When running Java-tron, I found that if I wanted to disconnect a peer in certain scenarios, the disconnect message sent carried an ambiguous reason code. You can take a look at the example below.

Example:
The node receives the block sent by the peer and then disconnects the peer due to block signature verification failure, but the reason for the disconnection message being sent is UNKNOWN.
```
10:17:39.014 WARN  [peerWorker-1] [net](P2pEventHandlerImpl.java:258) Message from /172.31.0.220:10850 process failed, type: BLOCK Num:58698013,ID:00000000037fa91d2bbac58ad77bd462a2878178e730eb93a4d3e00e21cbe52e, trx size: 4 type: (15, block sign error), detail: valid signature failed.

10:17:39.014 INFO  [peerWorker-1] [net](PeerConnection.java:175) Send peer /172.31.0.220:10850 message type: P2P_DISCONNECT reason: UNKNOWN
```
Therefore, after receiving the disconnection message, the peer cannot accurately know the specific reason for being disconnected. I think the reason code carried in the disconnect message in this example can be clearer. In addition, do you think it is necessary to check whether there are other disconnection scenarios with similar problems?

#### Frequency
It occurs every time in the above scenarios.

Send a block with an invalid signature to the node.

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

### PR DESCRIPTION
**Pull request by jwrct**
**What does this PR do?**
Closes #5989 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/test/java/org/tron/core/net/services/HandShakeServiceTest.java
+++ b/framework/src/test/java/org/tron/core/net/services/HandShakeServiceTest.java
@@ -21,7 +21,6 @@ import org.junit.rules.TemporaryFolder;
 import org.mockito.Mockito;
 import org.springframework.context.ApplicationContext;
 import org.tron.common.application.TronApplicationContext;
-import org.tron.common.utils.ByteArray;
 import org.tron.common.utils.ReflectUtils;
 import org.tron.common.utils.Sha256Hash;
 import org.tron.core.ChainBaseManager;
@@ -34,6 +33,7 @@ import org.tron.core.net.TronNetService;
 import org.tron.core.net.message.handshake.HelloMessage;
 import org.tron.core.net.peer.PeerConnection;
 import org.tron.core.net.peer.PeerManager;
+import org.tron.core.net.service.handshake.HandshakeService;
 import org.tron.p2p.P2pConfig;
 import org.tron.p2p.base.Parameter;
 import org.tron.p2p.connection.Channel;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by Sunny6889 — APPROVED** [review]

**Review by bladehan1 — APPROVED** [review]

File: framework/src/test/java/org/tron/core/net/services/HandShakeServiceTest.java [hunk 1]
```diff
@@ -126,13 +126,27 @@ public void testInvalidHelloMessage() {
     //block hash is empty
     try {
       BlockCapsule.BlockId hid = ChainBaseManager.getChainBaseManager().getHeadBlockId();
-      Protocol.HelloMessage.BlockId hBlockId = Protocol.HelloMessage.BlockId.newBuilder()
-          .setHash(ByteString.copyFrom(new byte[0]))
+      Protocol.HelloMessage.BlockId okBlockId = Protocol.HelloMessage.BlockId.newBuilder()
```
waynercheung, line 129: what about using "validBLockId" instead of "okBlockId"?

fyyhtx: Test case adjustment only.

jwrct: Test case adjustment only.

PR #6397, opened 2025-07-10
feat(*): Enrich FullNode command-line-options

### LINKED ISSUE #6218: Enrich  FullNode command-line-options
# Background
Currently, the main service JARs in Java-Tron primarily include FullNode.jar and SolidityNode.jar. FullNode.jar supports three types of nodes: SR Node, FullNode, and LightFullNode. SolidityNode is not supported, but it is recommended that it be supported.

| JAR | Description |
|-----|-------------|
| FullNode.jar | full node, which is capable of creating and broadcasting transactions, synchronizing and broadcasting blocks |
| SolidityNode.jar | solidity node, is capable of synchronizing solidified blocks from trusted nodes, can neither create nor broadcast transactions nor can it broadcast blocks. |

# Rationale

To enrich FullNode command-line options, support SolidityNode, merge FullNode.jar and SolidityNode.jar.

# Specification
Merge SolidityNode.jar into FullNode.jar, and start it via `java -jar FullNode.jar --solidity`

# Scope Of Impact
1. Build process: SolidityNode.jar will no longer be built anymore.
2. Tool usage: FullNode.jar will support the `--solidity` command.

halibobo1205:
close #6218.


halibobo1205:
TODO:  `SolidityNode` has been replaced by `FullNode --solidity`. CI testing scenarios([buildkite/interface-status-test-in-docker](https://buildkite.com/tronprotocol/interface-status-test-in-docker/builds/5916))need to be adapted accordingly.

framework/src/main/java/org/tron/program/FullNode.java
```diff
@@ -21,12 +21,16 @@ public class FullNode {
...
   public static void main(String[] args) {
     ExitManager.initExceptionHandler();
-    logger.info("Full node running.");
     Args.setParam(args, Constant.TESTNET_CONF);
     CommonParameter parameter = Args.getInstance();
 
     LogService.load(parameter.getLogbackPath());
 
+    if (parameter.isSolidityNode()) {
+      SolidityNode.run();
```
SeriousCoding789: how about add `logger.info("Solidity node running.");`. As FullNode logic also provide solidty API function with `/walletsolidity/...`

framework/src/main/java/org/tron/program/SolidityNode.java
```diff
@@ -55,25 +51,12 @@ public SolidityNode(Manager dbManager) {
...
-
-    logger.info("index switch is {}",
-        BooleanUtils.toStringOnOff(BooleanUtils
-            .toBoolean(parameter.getStorage().getIndexSwitch())));
-
     if (ObjectUtils.isEmpty(parameter.getTrustNodeAddr())) {
-      logger.error("Trust node is not set.");
-      return;
-    }
-    parameter.setSolidityNode(true);
```
SeriousCoding789, line 72: will remove this `parameter.setSolidityNode(true);` effect the respected http/rpc service start?

SeriousCoding789: make sure you check all related sub-class not depend this value

Sunny6889:
I have a question:
In existing FullNode with config `node.http.solidityEnable` with default is true,  HTTP API `/walletsolidity/...` will return solidity info. Same as`node.rpc.solidityEnable` for RPC request with `protocol.WalletSolidity`. 
It seems we don't need SolidityNode at all, it only confuses users. Thus I think we don't need to support SolidityNode related features.

Sunny6889:
In addition java-tron already have 5 HttpService and 2 RpcService (ignore the PBFT services)
- FullNodeHttpApiService
- SolidityNodeHttpApiService
- HttpApiOnSolidityService
- FullNodeJsonRpcHttpService
- JsonRpcServiceOnSolidity 
- RpcApiService 
- RpcApiServiceOnSolidity 
I would suggest we just upgrade FullNode clearly to separate the Solidity functionality through API.

framework/src/main/java/org/tron/program/FullNode.java
halibobo1205: here: 

Image

framework/src/main/java/org/tron/program/SolidityNode.java
halibobo1205: `parameter.setSolidityNode(true)` is replaced by `--solidity`.

Image

Sunny6889:
> @Sunny6889 Java-tron's support for SolidityNode is still quite valuable, particularly for accelerating historical synchronization. By configuring trusted nodes, SolidityNode can skip the computationally expensive signature verification process, resulting in roughly a 33% reduction in synchronization time.

@halibobo1205 For SolidityNode it only sync soliditied blocks? Could you share more about where the logic control it?

lvs0075 (approved):

PR #6399, opened 2025-07-11
feat(consensus): convert the local proposal expire time configuration to a chain parameter

lxcmyf:
**What does this PR do?**
Close [#767](https://github.com/tronprotocol/tips/issues/767)

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ This will mount the `output-directory` and `logs` directories on the host, the d
 
 [Tron Developers & SRs](https://discord.gg/hqKvyAM) is Tron's official Discord channel. Feel free to join this channel if you have any questions.
 
-[Core Devs Community](https://t.me/troncoredevscommunity) is the Telegram channel for java-tron community developers. If you want to contribute to java-tron, please join this channel.
+The [Core Devs Community](https://t.me/troncoredevscommunity) and [Tron Official Developer Group](https://t.me/TronOfficialDevelopersGroupEn) are Telegram channels specifically designed for java-tron community developers to engage in technical discussions.
 
 [tronprotocol/allcoredev](https://gitter.im/tronprotocol/allcoredev) is the official Gitter channel for developers.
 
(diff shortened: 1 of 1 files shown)


halibobo1205 (approved):

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1297,6 +1297,10 @@ public static void setParam(final Config config) {
         config.hasPath(Constant.COMMITTEE_ALLOW_TVM_BLOB) ? config
             .getInt(Constant.COMMITTEE_ALLOW_TVM_BLOB) : 0;
 
+    PARAMETER.proposalVotingWindow =
+        config.hasPath(Constant.COMMITTEE_PROPOSAL_VOTING_WINDOW) ? config
+            .getInt(Constant.COMMITTEE_PROPOSAL_VOTING_WINDOW) : 86400L;
+
```
317787106: You'd better restore it in method Args.clearParam.

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -724,6 +724,10 @@ public class CommonParameter {
   @Setter
   public long allowTvmBlob;
 
+  @Getter
+  @Setter
+  public long proposalVotingWindow;
```
317787106: Add comment as other: //committee parameter. proposal will be valid in this number of block.

framework/src/main/java/org/tron/core/config/args/Args.java
lxcmyf: Okay, it's only for unit test calls.

Sunny6889:
hi @lxcmyf I did not find the logic to pre-fill the DynamicPropertiesStore with the existing default maintenance value. There do have Args read from config, but I think that is only used for private test config.

lxcmyf:
> hi @lxcmyf I did not find the logic to pre-fill the DynamicPropertiesStore with the existing default maintenance value. There do have Args read from config, but I think that is only used for private test config.

@Sunny6889 As shown in the following code, the default value can refer to the historical writing method.
image

barbatos2011 (approved):

actuator/src/main/java/org/tron/core/actuator/ProposalCreateActuator.java
```diff
@@ -53,7 +53,8 @@ public boolean execute(Object result) throws ContractExeException {
 
       long currentMaintenanceTime =
           chainBaseManager.getDynamicPropertiesStore().getNextMaintenanceTime();
-      long now3 = now + CommonParameter.getInstance().getProposalExpireTime();
```
fyyhtx, line 56: Have you considered removing the code related to the deprecated configuration item "proposalExpireTime"?

lvs0075 (approved):

PR #6427, opened 2025-08-21
fix(test):  fix the ZksnarkInitTest

Federico2014:
**What does this PR do?**
Fix the ZksnarkInitTest

### DIFF
--- a/framework/src/test/java/org/tron/core/exception/TronErrorTest.java
+++ b/framework/src/test/java/org/tron/core/exception/TronErrorTest.java
@@ -9,10 +9,12 @@ import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import com.typesafe.config.ConfigObject;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.nio.file.Path;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.After;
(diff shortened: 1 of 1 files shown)


framework/src/test/java/org/tron/core/exception/TronErrorTest.java
```diff
@@ -57,7 +59,16 @@ public void testTronError() {
...
+    Field initializedField = ZksnarkInitService.class.getDeclaredField("initialized");
+    initializedField.setAccessible(true);
+    Field modifiersField = Field.class.getDeclaredField("modifiers");
+    modifiersField.setAccessible(true);
+    modifiersField.setInt(initializedField,
+        initializedField.getModifiers() & ~java.lang.reflect.Modifier.FINAL);
+    AtomicBoolean atomicBoolean = (AtomicBoolean) initializedField.get(null);
+    atomicBoolean.set(false);
+
     try (MockedStatic<JLibrustzcash> mock = mockStatic(JLibrustzcash.class)) {
```
halibobo1205, line 69: Is a try-finally statement needed here to restore the original value for `initialized`?

Federico2014: If `initialized` is not restored to its original value, it may cause `librustzcashInitZksnarkParams` to be initialized one more time, it has no other impact. However, `initialized` should still be restored to its original value.

halibobo1205 (approved):

PR #6441, opened 2025-09-08
fix(test): remove all junit dir and add timeout for gRPC call

317787106:
**What does this PR do?**

- remove all junit directory for two cases
- add TimeoutInterceptor for gRPC call.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/test/java/org/tron/core/event/BlockEventGetTest.java
+++ b/framework/src/test/java/org/tron/core/event/BlockEventGetTest.java
@@ -11,9 +11,10 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import lombok.extern.slf4j.Slf4j;
-import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -63,22 +64,30 @@ public class BlockEventGetTest extends BlockGenerate {
   protected Manager dbManager;
   long currentHeader = -1;
   private TronNetDelegate tronNetDelegate;
-  private TronApplicationContext context;
+  private static TronApplicationContext context;
 
 
   static LocalDateTime localDateTime = LocalDateTime.now();
   private long time = ZonedDateTime.of(localDateTime,
       ZoneId.systemDefault()).toInstant().toEpochMilli();
 
-  protected void initDbPath() throws IOException {
-    dbPath = temporaryFolder.newFolder().toString();
+
+  public static String dbPath() {
+    try {
+      return temporaryFolder.newFolder().toString();
+    } catch (IOException e) {
+      Assert.fail("create temp folder failed");
+    }
+    return null;
+  }
+
+  @BeforeClass
+  public static void init() {
+    Args.setParam(new String[] {"--output-directory", dbPath()}, Constant.TEST_CONF);
   }
 
   @Before
   public void before() throws IOException {
-    initDbPath();
-    logger.info("Full node running.");
-    Args.setParam(new String[] {"-d", dbPath}, Constant.TEST_CONF);
     Args.getInstance().setNodeListenPort(10000 + port.incrementAndGet());
 
     context = new TronApplicationContext(DefaultConfig.class);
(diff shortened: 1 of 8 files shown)


framework/src/test/java/org/tron/core/event/BlockEventGetTest.java
```diff
@@ -108,8 +117,10 @@ public void before() throws IOException {
     dps.saveAllowTvmShangHai(1);
   }
 
-  @After
-  public void after() throws IOException {
+  @AfterClass
+  public static void after() throws IOException {
+    context.destroy();
```
0xbigapple, line 121: Initializing context in @Before but destroying it in @AfterClass , may cause resource leaks ?

framework/src/test/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandlerTest.java
```diff
@@ -95,6 +109,7 @@ public void testProcessMessage() throws Exception {
 
   @After
   public void destroy() {
+    context.destroy();
```
0xbigapple, line 112: peer.setChannel() in @Before， need close？

0xbigapple: Args.setParam in BeforeClass ,but clearParam in After?

framework/src/test/java/org/tron/core/event/BlockEventGetTest.java
317787106: fix it

framework/src/test/java/org/tron/core/net/messagehandler/SyncBlockChainMsgHandlerTest.java
317787106: fix it

0xbigapple (approved):

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
```diff
@@ -159,12 +160,15 @@ public static void init() throws IOException {
 
     channelFull = ManagedChannelBuilder.forTarget(fullNode)
         .usePlaintext()
+        .intercept(new TimeoutInterceptor(5000))
```
waynercheung, line 163: what about using "withDeadline", such as 
```
blockingStubFull = WalletGrpc.newBlockingStub(channelFull)
    .withDeadline(Deadline.after(30, TimeUnit.SECONDS));
```

or 
```
WalletGrpc.newBlockingStub(channelFull)
    .withDeadlineAfter(30, TimeUnit.SECONDS);
```


No need to write a custom TimeoutInterceptor, reducing implementation effort.

317787106: withDeadlineAfter or withDeadline  means the total life of current channel since it was created, It's the wrong way to set timeout. Only the Interceptor is the right way to set timeout for any Connection instance. See also [trident](https://github.com/tronprotocol/trident/blob/2889428c889278b401d3acf1a428a7d48383e8f6/core/src/main/java/org/tron/trident/core/ApiWrapper.java#L219)

waynercheung: According to the gRPC Java API, both withDeadline(Deadline deadline) and withDeadlineAfter(long duration, TimeUnit unit) set a deadline for individual RPC calls made through a stub, not the lifetime of the ManagedChannel. For example, WalletGrpc.newBlockingStub(channelFull).withDeadlineAfter(30, TimeUnit.SECONDS) ensures that calls like createAssetIssue fail with a DEADLINE_EXCEEDED exception if they don’t complete within 30 seconds, without affecting the channel’s lifecycle.

If you need consistent timeouts across all RPC calls on a channel (e.g., all calls to fullNode), TimeoutInterceptor is more convenient.

317787106: Using Interceptor is more simple. Don't need set Deadline for every call.

waynercheung (approved):
LGTM

kuny0707 (approved):

halibobo1205 (approved):

lvs0075 (approved):

PR #6443, opened 2025-09-09
fet(event): optimize event service to obtain transaction information

zeusoo001:
**What does this PR do?**
Optimize event service to obtain transaction information.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/event/BlockEventGet.java
+++ b/framework/src/main/java/org/tron/core/services/event/BlockEventGet.java
@@ -88,18 +88,13 @@ public class BlockEventGet {
   }
 
   public SmartContractTrigger getContractTrigger(BlockCapsule block, long solidNum) {
-    TransactionRetCapsule result;
-    try {
-      result = manager.getChainBaseManager().getTransactionRetStore()
-        .getTransactionInfoByBlockNum(ByteArray.fromLong(block.getNum()));
-    } catch (BadItemException e) {
-      throw new RuntimeException(e);
-    }
+
+    GrpcAPI.TransactionInfoList list = manager.getTransactionInfoByBlockNum(block.getNum());
 
     SmartContractTrigger contractTrigger = new SmartContractTrigger();
     for (int i = 0; i < block.getTransactions().size(); i++) {
       Protocol.Transaction tx = block.getInstance().getTransactions(i);
-      Protocol.TransactionInfo txInfo = result.getInstance().getTransactioninfo(i);
+      Protocol.TransactionInfo txInfo = list.getTransactionInfo(i);
 
       List<ContractTrigger> triggers = parseLogs(tx, txInfo);
       for (ContractTrigger trigger : triggers) {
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/tron/core/services/event/HistoryEventService.java
```diff
@@ -67,7 +70,8 @@ private void syncEvent() {
         tmp++;
         endNum = manager.getDynamicPropertiesStore().getLatestSolidifiedBlockNum();
       }
-      initEventService(manager.getChainBaseManager().getBlockIdByNum(endNum));
+      long startNum = endNum == 0 ? 0 : endNum - 1;
```
0xbigapple, line 73: If now solidifiedBlockNum is 100,  but the `startSyncBlockNum` in config is 1000,  it still start writing events from 99?

xxo1shine: Will write from 100, and you need to configure `startSyncBlockNum` reasonably.

framework/src/main/java/org/tron/core/services/event/HistoryEventService.java
```diff
@@ -54,7 +54,10 @@ private void syncEvent() {
     try {
       long tmp = instance.getStartSyncBlockNum();
       long endNum = manager.getDynamicPropertiesStore().getLatestSolidifiedBlockNum();
-      while (tmp <= endNum) {
+      while (tmp < endNum) {
```
waynercheung, line 57: if tmp == endNum, then the while loop exists, and the startNum will be set to tmp - 1. For example, when tmp is 108, endNum is 109, and the while loop will be executed, then when tmp is 109 and endNum is 109, startNum is 109 -1 = 108, for the block 109, it will not be run the codes "realtimeEventService.flush(blockEvent, false); solidEventService.flush(blockEvent);" will not be executed, is that any problem?

framework/src/main/java/org/tron/core/services/event/BlockEventGet.java
```diff
@@ -88,18 +94,13 @@ public BlockEvent getBlockEvent(long blockNum) throws Exception {
...
-      throw new RuntimeException(e);
-    }
+
+    GrpcAPI.TransactionInfoList list = manager.getTransactionInfoByBlockNum(block.getNum());
 
     SmartContractTrigger contractTrigger = new SmartContractTrigger();
     for (int i = 0; i < block.getTransactions().size(); i++) {
       Protocol.Transaction tx = block.getInstance().getTransactions(i);
-      Protocol.TransactionInfo txInfo = result.getInstance().getTransactioninfo(i);
+      Protocol.TransactionInfo txInfo = list.getTransactionInfo(i);
```
waynercheung, line 103: if list get from getTransactionInfoByBlockNum is empty, or list.getTransactionInfoCount() != block.getTransactions().size(), is there any problem?

xxo1shine: No problem, this situation does not exist, otherwise the data is inconsistent.

PR #6446, opened 2025-09-10
feat(command): change newly added "--keystore" parameter to "--keystore-factory"

halibobo1205:
**What does this PR do?**
     Rename `--keystore` parameter to `--keystore-factory`, pre-PR: #6397.

The `--keystore` parameter name could be misleading, as it typically refers to encrypted key storage files or wallets, while this parameter actually runs a KeystoreFactory. The new name `--keystore-factory` makes the parameter's purpose more explicit and avoids potential confusion.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/common/src/main/java/org/tron/common/parameter/CommonParameter.java
+++ b/common/src/main/java/org/tron/common/parameter/CommonParameter.java
@@ -211,8 +211,8 @@ public class CommonParameter {
   //If you are running KeystoreFactory, this flag is set to true
   @Getter
   @Setter
-  @Parameter(names = {"--keystore"}, description = "running KeystoreFactory")
-  public boolean keystore = false;
+  @Parameter(names = {"--keystore-factory"}, description = "running KeystoreFactory")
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -211,7 +211,7 @@ public class CommonParameter {
   //If you are running KeystoreFactory, this flag is set to true
   @Getter
   @Setter
-  @Parameter(names = {"--keystore"}, description = "running KeystoreFactory")
+  @Parameter(names = {"--keystore-factory"}, description = "running KeystoreFactory")
   public boolean keystore = false;
```
0xbigapple: Can the variable in the code be changed together to avoid confusion?

halibobo1205: done

PR #6447, opened 2025-09-10
refactor(TransactionRegister): enhance actuator registration

halibobo1205:
**What does this PR do?**

- Add registration status check to prevent duplicate initialization
- Implement proper error handling with TronError exception throwing
- Optimize package scanning scope from "org.tron" to "org.tron.core.actuator"

   Optimize TransactionRegister to prevent duplicate CI registrations.
 ## first init) 
 ```
Before: Reflections took 375 ms to scan 11 urls, producing 525 keys and 3121 values
After:  Reflections took 188 ms to scan 3 urls, producing 4 keys and 78 values 
```
 ## repeat init
```
Before: Reflections took 165 ms to scan 11 urls, producing 525 keys and 3121 values
After:  Reflections took 24 ms to scan 3 urls, producing 4 keys and 78 values
```

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### CODE REVIEW

**Comment by lvs0075** [comment]
This feature has been included in the discussions for version 4.8.2：https://github.com/tronprotocol/java-tron/issues/6585#issue-4092779531

File: actuator/src/main/java/org/tron/core/utils/TransactionRegister.java [hunk 1]
```diff
@@ -1,24 +1,58 @@
...
   public static void registerActuator() {
-    Reflections reflections = new Reflections("org.tron");
-    Set<Class<? extends AbstractActuator>> subTypes = reflections
-        .getSubTypesOf(AbstractActuator.class);
-    for (Class _class : subTypes) {
-      try {
-        _class.newInstance();
-      } catch (Exception e) {
-        logger.error("{} contract actuator register fail!", _class, e);
+    if (REGISTERED.get()) {
```
waynercheung, line 17: `registerActuator()` is not just caching reflection work; it is establishing actuator registration state, and that state is configuration-dependent because `TransactionFactory.register()` filters by `actuator.whitelist`.

With this early return, once the first registration succeeds, the same JVM can no longer rebuild actuator registration state, even if the effective whitelist changes later. That turns a scan optimization into a lifecycle/semantics change.

For this PR, I would strongly prefer the simpler and safer approach:
- remove the global one-time `REGISTERED` short-circuit,
- keep the narrowed scan scope (`org.tron.core.actuator`),
- keep the fail-fast `TronError` behavior.

That preserves the low-risk performance improvement without introducing configuration-related registration issues.

framework/src/test/java/org/tron/core/utils/TransactionRegisterTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    TransactionRegister.registerActuator();
+    assertTrue("After first call, should be registered", TransactionRegister.isRegistered());
+
+    TransactionRegister.registerActuator();
+    assertTrue("After second call, should still be registered", TransactionRegister.isRegistered());
+  }
+
+  @Test
+  public void testRegistrationRunsExactlyOnce() {
```
waynercheung, line 89: Since actuator registration depends on `actuator.whitelist`, we need regression coverage for re-initialization with different effective configuration in the same JVM.

Please add a test that:
1. initializes registration with whitelist A,
2. changes to whitelist B (or empty whitelist),
3. calls `registerActuator()` again,
4. verifies that actuator resolution reflects the second configuration.

Without that coverage, this test suite would not catch stale registration state introduced by the one-time guard.

framework/src/test/java/org/tron/core/utils/TransactionRegisterTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+  @Test
+  public void testConcurrentAccessThreadSafe() throws InterruptedException {
+    final int threadCount = 5;
+    Thread[] threads = new Thread[threadCount];
+    final AtomicBoolean testPassed = new AtomicBoolean(true);
+
+    for (int i = 0; i < threadCount; i++) {
+      threads[i] = new Thread(() -> {
+        try {
```
waynercheung, line 57: This concurrent test only catches `Exception`, but `registerActuator()` now throws `TronError`, which extends `Error`, not `Exception`.

That means a child-thread failure may escape this catch block entirely, so the test can still pass even when registration actually fails.

Please catch `Throwable` here, or use an `ExecutorService`/`Future`-based pattern so failures from worker threads are propagated reliably back to the test thread.

waynercheung:
Thanks for the optimization work here. I agree with the direction of narrowing the Reflections scan scope and failing fast on actuator instantiation failures.

However, I’m requesting changes for one design issue and the related test gaps.

The main problem is the new process-wide `REGISTERED` guard in `TransactionRegister`. Actuator registration is configuration-dependent because `TransactionFactory.register()` filters by `actuator.whitelist`, so caching a global "already registered" state changes the lifecycle semantics, not just the reflection cost.

For this PR, I would recommend removing the global one-time guard and keeping the lower-risk improvements:
- scan only `org.tron.core.actuator`
- keep the fail-fast `TronError` behavior

Please also add regression coverage for re-initialization with different whitelist values in the same JVM, and fix the concurrent test so it correctly captures `TronError` from worker threads.

liuyifei001:
@halibobo1205 @waynercheung  `actuator.whitelist` should be deprecated or removed. If the whitelist is not empty, the node will fork. Since `registerActuator` is used to verify this whitelist, it would be better to remove this function entirely as part of the optimization.

actuator/src/main/java/org/tron/core/utils/TransactionRegister.java
halibobo1205: `actuator.whitelist` should be deprecated or removed, and it can not be changed during JVM run time, except for CI.

framework/src/test/java/org/tron/core/utils/TransactionRegisterTest.java
halibobo1205: Do not use this invalid configuration `actuator.whitelist`; it should be removed.

halibobo1205:
`actuator.whitelist` is a deprecated parameter that should be removed in a future version. When the whitelist is non-empty, the node will fork from the network, so it should never be used. I think this can be removed by another issue or pr @waynercheung @liuyifei001

framework/src/test/java/org/tron/core/utils/TransactionRegisterTest.java
halibobo1205: The concurrent test is designed to verify the thread-safety of the registration path. With the double-check locking in place, only one thread actually performs registration while others either early-return or wait. Even if TronError were thrown (which escapes catch (Exception)), the remaining threads would re-enter the synchronized block and retry. As long as any one thread succeeds, REGISTERED is set to true and the test passes correctly.

That said, I can change it to catch (Throwable) for clarity if you prefer, but it does not affect the test's correctness in practice.

liuyifei001 (approved):

waynercheung:
@halibobo1205  @liuyifei001 
Thanks for the clarification. If `actuator.whitelist` is no longer considered a supported runtime feature, then I’m OK with the `REGISTERED` guard here.

Please consider removing or hard-disabling `actuator.whitelist` in a follow-up PR so the code semantics match the intended project direction explicitly.

waynercheung (approved):
LGTM

PR #6449, opened 2025-09-12
api(CodeQL): fix error message exposure in http response

halibobo1205:
**What does this PR do?**
    Revert the change to HTTP error message handling, pre PR #6417
     The workaround broke clients depending on the original error output.
- Unit Tests
- Manual Testing

**Follow up**
     We may need to open a new issue to discuss how to resolve this warning going forward.
**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/http/solidity/GetTransactionByIdSolidityServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/solidity/GetTransactionByIdSolidityServlet.java
@@ -30,7 +30,12 @@ public class GetTransactionByIdSolidityServlet extends RateLimiterServlet {
       String input = request.getParameter("value");
       fillResponse(ByteString.copyFrom(ByteArray.fromHexString(input)), visible, response);
     } catch (Exception e) {
-      Util.processError(e, response);
+      logger.debug("Exception: {}", e.getMessage());
+      try {
+        response.getWriter().println(e.getMessage());
+      } catch (IOException ioe) {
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/tron/core/services/http/solidity/GetTransactionByIdSolidityServlet.java
```diff
@@ -41,7 +46,12 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
       JsonFormat.merge(params.getParams(), build, params.isVisible());
       fillResponse(build.build().getValue(), params.isVisible(), response);
     } catch (Exception e) {
-      Util.processError(e, response);
+      logger.debug("Exception: {}", e.getMessage());
+      try {
+        response.getWriter().println(e.getMessage());
+      } catch (IOException ioe) {
+        logger.debug("IOException: {}", ioe.getMessage());
+      }
```
317787106, line 54: Maybe it's not a good idea to modify the error message and remove Exception for every API. A better solution would be to update the implementation of Util.processError uniformly.

halibobo1205: @317787106 This is a legacy API; do not modify anything, even the error messages.

317787106: Get it

317787106 (approved):

0xbigapple (approved):

PR #6455, opened 2025-09-23
feat(exit): throw TronError if JDK does not support

halibobo1205:
**What does this PR do?**
    Replace the UnsupportedOperationException with TronError if the JDK does not support.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -1,6 +1,7 @@
 package org.tron.core.config.args;
 
 import static java.lang.System.exit;
+import static org.fusesource.jansi.Ansi.ansi;
 import static org.tron.common.math.Maths.max;
 import static org.tron.common.math.Maths.min;
 import static org.tron.core.Constant.ADD_PRE_FIX_BYTE_MAINNET;
@@ -45,6 +46,7 @@ import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.fusesource.jansi.AnsiConsole;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.tron.common.arch.Arch;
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -374,7 +376,14 @@ private static Map<String, String[]> getOptionGroup() {
...
-    Arch.throwIfUnsupportedJavaVersion();
+    try {
+      Arch.throwIfUnsupportedJavaVersion();
+    } catch (UnsupportedOperationException e) {
+      AnsiConsole.systemInstall();
+      System.out.println(ansi().fgRed().a(e.getMessage()).reset());
+      AnsiConsole.systemUninstall();
+      throw new TronError(e, TronError.ErrCode.JDK_VERSION);
+    }
     clearParam(); // reset all parameters to avoid the influence in test
```
317787106, line 390: The method name `throwIfUnsupportedJavaVersion` can be modified as `checkJdkCompatible` ?
What's the output of three lines related with AnsiConsole ?

platform/src/main/java/common/org/tron/common/arch/Arch.java
```diff
@@ -73,8 +73,8 @@ public static void throwIfUnsupportedJavaVersion() {
     if (isX86() && !isJava8()) {
       logger.info(withAll());
       throw new UnsupportedOperationException(String.format(
-          "Java %s is required for %s architecture. Detected version %s",
-          "1.8 ", getOsArch(), javaSpecificationVersion()));
+          "Java %s is required for %s architecture. Detected version %s", "1.8",
+          getOsArch(), javaSpecificationVersion()));
     }
```
317787106, line 78: Why not write Java 1.8 directly ?

framework/src/main/java/org/tron/core/config/args/Args.java
halibobo1205: `throwIfUnsupportedJavaVersion` is the more accurate and intention-revealing name because it clearly communicates that the method will throw an exception if the condition is not met.

This error will appear on the console to alert the user, rather than being quietly recorded in the log files.

image

platform/src/main/java/common/org/tron/common/arch/Arch.java
halibobo1205: I believe either style is fine.

framework/src/main/java/org/tron/core/config/args/Args.java
317787106: If you want to show this infomation in console, this AnsiConsole should be placed in ExitManager, so all TronError has the same action.

halibobo1205: To avoid confusion caused by silent execution when using -h or -v flags, errors are explicitly logged to the console in this context. Console output is not required for errors in other scenarios.

317787106: Get it.

PR #6463, opened 2025-10-13
feat(event): optimize event service thread close logic

zeusoo001:
**What does this PR do?**
Optimize event service thread close logic.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/event/HistoryEventService.java
+++ b/framework/src/main/java/org/tron/core/services/event/HistoryEventService.java
@@ -29,6 +29,8 @@ public class HistoryEventService {
   @Autowired
   private Manager manager;
 
+  private volatile boolean isClosed = false;
+
   private volatile Thread thread;
 
   public void init() {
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/tron/core/services/event/HistoryEventService.java
```diff
@@ -61,7 +64,7 @@ private void syncEvent() {
       long tmp = instance.getStartSyncBlockNum();
       long endNum = manager.getDynamicPropertiesStore().getLatestSolidifiedBlockNum();
       while (tmp < endNum) {
-        if (thread.isInterrupted()) {
+        if (thread.isInterrupted() || isClosed) {
```
SeriousCoding789, line 67: It is a good resource leakage fix. I wonder how you find it?

317787106: It may throw exception when the program closes:

Image

PR #6467, opened 2025-10-15
feat(jsonrpc): add "input" as field for transaction input

simbadMarino:
**What does this PR do?**
Adds "input" field for transaction input while keeping "data" for legacy and compatibility.

Impacted JSON-RPC API methods:
- eth_call
- eth_estimateGas
- buildTransaction
  

- Align TRON jsonrpc with [EVM new standard](https://github.com/ethereum/go-ethereum/pull/28078) . 
- Reduce development efforts for ETH/EVM developers porting their projects to TRON.
- General EVM tooling compatibility improvements

- Unit Tests
- Manual Testing

**Follow up**
- Probably will need to check further simple improvements like this to increase TRON compatibility with eth json RPC

**Extra details**
- Additional discussions within Foundry devs community: https://github.com/alloy-rs/alloy/issues/2371

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
@@ -537,4 +537,22 @@ public class JsonRpcApiUtil {
     random.nextBytes(uid);
     return ByteArray.toHexString(uid);
   }
+
+   /**
+   * Check if transaction data parameter is null or empty
+   * Supports both 'data' and 'input' parameters
+   */
+  public static boolean paramTransactionDataIsNull(String data, String input) {
+    return paramStringIsNull(data) && paramStringIsNull(input);
+  }
+  
+  /**
+   * Get transaction data with priority: input > data
+   */
+  public static String getTransactionData(String data, String input) {
+    if (!paramStringIsNull(input)) {
+      return input;
+    }
+    return data;
+  }
 }
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
@@ -43,6 +43,9 @@ public class CallArguments {
   private String data;
   @Getter
   @Setter
+  private String input; //Add input parameter to align with Ethereum https://github.com/ethereum/go-ethereum/pull/28078
+  @Getter
+  @Setter
   private String nonce; // not used
 
   /**
@@ -57,8 +60,11 @@ public class CallArguments {
       throw new JsonRpcInvalidRequestException("invalid json request");
     } else if (paramStringIsNull(to)) {
       // data is null
-      if (paramStringIsNull(data)) {
-        throw new JsonRpcInvalidRequestException("invalid json request");
+      if (paramStringIsNull(input)) {
+        if(paramStringIsNull(data))
+        {
+          throw new JsonRpcInvalidRequestException("invalid json request"); //If both input and data are bnull throw an error, take input as priority
+        }
       }
 
       contractType = ContractType.CreateSmartContract;
@@ -85,5 +91,27 @@ public class CallArguments {
 
   public long parseValue() throws JsonRpcInvalidParamsException {
     return parseQuantityValue(value);
+  }
+
+    /**
+   * Get the transaction data, supporting both 'data' and 'input' parameters
+   * for Ethereum compatibility. 'input' takes precedence if both are provided.
+   * 
+   * @return the transaction data/input
+   */
+  public String getTransactionData() {
+    // Prioritize 'input' for Ethereum compatibility
+    if (StringUtils.isNotEmpty(input)) {
+      return input;
+    }
+    return data;
+  }
+
+  /**
+   * Set transaction data, updating the appropriate field based on which was used
+   */
+  public void setTransactionData(String transactionData) {
+    // For backward compatibility, default to setting 'data'
+    this.data = transactionData;
   }
 }
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
@@ -43,10 +43,10 @@ public class CallArguments {
   private String data;
   @Getter
   @Setter
-  private String input; //Add input parameter to align with Ethereum https://github.com/ethereum/go-ethereum/pull/28078
+  private String nonce; // not used
   @Getter
   @Setter
-  private String nonce; // not used
+  private String input; //Add input parameter to align with Ethereum https://github.com/ethereum/go-ethereum/pull/28078
 
   /**
    * just support TransferContract, CreateSmartContract and TriggerSmartContract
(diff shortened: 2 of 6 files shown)


0xbigapple:
Good job! Maybe this pr can be added in the next version, along with some other related interface changes. The features for release 4.8.1 are already close to being ready,  and I think it may be too late to include this in this version.

lvs0075:
This feature has been included in the discussions for version 4.8.2

framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
```diff
@@ -57,8 +60,11 @@ public ContractType getContractType(Wallet wallet) throws JsonRpcInvalidRequestE
       throw new JsonRpcInvalidRequestException("invalid json request");
     } else if (paramStringIsNull(to)) {
```
waynercheung, line 61: The new `input` support is validated with `paramStringIsNull(...)`, but later resolved with `StringUtils.isNotEmpty(...)` in `getTransactionData()`.

Those are not equivalent in this codebase because `paramStringIsNull("0x") == true`, while `StringUtils.isNotEmpty("0x") == true`.

That creates an inconsistent flow where the request can be validated based on `data`, but then executed using `input`, silently discarding the actual calldata / bytecode.

  Concrete example:
  - `to == null`
  - `input == "0x"`
  - `data == "0x6080..."`

Please use one canonical transaction-data resolver and one emptiness definition everywhere.

framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
```diff
@@ -86,4 +92,26 @@ public ContractType getContractType(Wallet wallet) throws JsonRpcInvalidRequestE
...
     return parseQuantityValue(value);
   }
+
+    /**
+   * Get the transaction data, supporting both 'data' and 'input' parameters
+   * for Ethereum compatibility. 'input' takes precedence if both are provided.
+   * 
+   * @return the transaction data/input
+   */
+  public String getTransactionData() {
```
waynercheung, line 102: The `input > data` priority is correct, but the conflict case is still undefined.

If both `input` and `data` are provided with different values, the current implementation silently prefers `input`. That makes caller mistakes hard to detect and diverges from geth, which rejects the request explicitly.

Please add a shared conflict check for:
- both fields present
 - both non-empty under the project's own null/empty rules
- values not equal

and return a clear params error instead of silently swallowing one side.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -537,4 +537,22 @@ public static String generateFilterId() {
     random.nextBytes(uid);
     return ByteArray.toHexString(uid);
   }
+
+   /**
+   * Check if transaction data parameter is null or empty
+   * Supports both 'data' and 'input' parameters
+   */
+  public static boolean paramTransactionDataIsNull(String data, String input) {
```
waynercheung, line 545: These helpers(`paramTransactionDataIsNull` and `getTransactionData`) look like the intended canonical implementation, but they are not actually used by `CallArguments` / `BuildArguments`.

Right now the PR adds a shared utility and still keeps separate local implementations, and those implementations already differ on `"0x"` semantics. Please either:
  - make this utility the single source of truth, or
  - remove it

but avoid leaving the refactor half-done.

framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
```diff
@@ -86,4 +92,26 @@ public ContractType getContractType(Wallet wallet) throws JsonRpcInvalidRequestE
...
+    if (StringUtils.isNotEmpty(input)) {
+      return input;
+    }
+    return data;
+  }
+
+  /**
+   * Set transaction data, updating the appropriate field based on which was used
+   */
+  public void setTransactionData(String transactionData) {
```
waynercheung, line 113: `setTransactionData()` is currently unused, and its name is broader than its behavior: it only writes `data`, not `input`.

That is a maintainability trap for future callers. Please either remove it, or document very explicitly why it only updates `data`.

framework/src/test/java/org/tron/core/services/jsonrpc/CallArgumentsTest.java
```diff
@@ -28,7 +28,7 @@ public class CallArgumentsTest extends BaseTest {
   public void init() {
```
waynercheung, line 28: It has the same problem as `BuildArgumentsTest.java:39`

waynercheung (changes requested):
Thanks for the contribution. The direction is valuable and aligning TRON JSON-RPC with the Ethereum `input` field is the right goal, but I don’t think this is merge-ready yet.

1. The new `input` / `data` handling is internally inconsistent.
Validation currently uses `paramStringIsNull(...)`, while resolution uses `StringUtils.isNotEmpty(...)`.
In this codebase those are not equivalent, especially for `"0x"`, so the same request can be validated using one field and executed using the other.

2. Conflict behavior is still undefined.
When both `input` and `data` are provided with different values, the implementation silently prefers `input`. Please reject that case explicitly with a clear params error rather than dropping one side.

3. Test coverage is insufficient.
The updated tests only adapt constructor signatures and do not assert the new compatibility behavior. Please add explicit coverage for the supported combinations and the `"0x"` edge cases.

4. `setTransactionData()` is unused and misleading as currently implemented.
5. The new helpers in `JsonRpcApiUtil` are not wired in, so the refactor is left half-done.
6. There are a few minor style issues in `CallArguments` that should be cleaned up before merge.

**Process note**
This PR is opened from the fork’s `develop` branch and currently includes a merge commit from an internal fork PR. That is not a correctness issue by itself, but it makes review scope more mutable and the history noisier than necessary. For follow-up changes, please prefer a dedicated feature branch rebased onto the latest upstream `master` or `develop`, and keep the PR branch frozen during review.

simbadMarino:
Hi @waynercheung , thanks a lot for the detailed review and changes suggestions. I will proceed to with the edits as soon as possible!

waynercheung:
Hi @simbadMarino, thanks again for working on this! We noticed there haven't been any updates since your last reply. Just checking in - have you had a chance to start on the revisions, or have you run into any blockers we can help with?

We're keen to move this forward and are happy to assist if anything is unclear from the review comments. Could you give us a rough sense of your expected timeline for the changes?

If it turns out you're not able to get to this in the near future, please let us know - we may need to pick it up on our end to keep things on schedule. No pressure either way, just want to make sure we're aligned. Thanks!

simbadMarino:
> Hi @simbadMarino, thanks again for working on this! We noticed there haven't been any updates since your last reply. Just checking in - have you had a chance to start on the revisions, or have you run into any blockers we can help with?
> 
> We're keen to move this forward and are happy to assist if anything is unclear from the review comments. Could you give us a rough sense of your expected timeline for the changes?
> 
> If it turns out you're not able to get to this in the near future, please let us know - we may need to pick it up on our end to keep things on schedule. No pressure either way, just want to make sure we're aligned. Thanks!

Hi @waynercheung , apologies for the delays, I've been low on bandwidth recently, I inspected the review comments and even when the changes might not feel like a huge effort I indeed foresee some back and forth and iterations if I proceed by myself and prefer no to slow down or even compromise the introduction of this feature in the next release. 

Having said that, I consider i would be much more efficient for you guys to pick it up from here.

Let me know how to proceed with this PR. 

Thank you!

waynercheung:
Hi @simbadMarino , thanks for the update - we really appreciate you taking the time to review the comments.

Completely understood your situation. I will take it from here and push the remaining changes. This PR will be closed, but your contribution and the groundwork you've laid here will absolutely be carried forward.

Thanks again for kicking this off - it's genuinely appreciated!

PR #6495, opened 2025-12-19
fix(api): Optimize memory allocation for the filter interface of JSON-RPC

317787106:
**What does this PR do?**

- Optimize memory allocation for the filter interface of JSON-RPC, `eth_newFilter`, `eth_newBlockFilter`
- add config item `node.jsonrpc.maxBlockFilterNum`

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
+++ b/chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
@@ -423,6 +423,7 @@ public class SnapshotManager implements RevokingDatabase {
 
   private void deleteCheckpoint() {
     if(checkTmpStore == null) {
+      // only occurs in mock test. TODO fix test
       return;
     }
     try {
--- a/common/src/main/java/org/tron/common/parameter/CommonParameter.java
+++ b/common/src/main/java/org/tron/common/parameter/CommonParameter.java
@@ -527,6 +527,9 @@ public class CommonParameter {
   @Getter
   @Setter
   public int jsonRpcMaxSubTopics = 1000;
+  @Getter
+  @Setter
+  public int jsonRpcMaxBlockFilterNum = 50000;
 
   @Getter
   @Setter
--- a/framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
+++ b/framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
@@ -546,6 +546,7 @@ public class EventPluginLoader {
     }
     int queueSize = 0;
     if (eventListeners == null || eventListeners.isEmpty()) {
+      // only occurs in mock test. TODO fix test
       return false;
     }
     for (IPluginEventListener listener : eventListeners) {
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
@@ -7,6 +7,7 @@ import com.googlecode.jsonrpc4j.JsonRpcErrors;
 import com.googlecode.jsonrpc4j.JsonRpcMethod;
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
@@ -18,6 +19,7 @@ import org.tron.common.runtime.vm.DataWord;
 import org.tron.common.utils.ByteArray;
 import org.tron.core.exception.BadItemException;
 import org.tron.core.exception.ItemNotFoundException;
+import org.tron.core.exception.jsonrpc.JsonRpcExceedLimitException;
 import org.tron.core.exception.jsonrpc.JsonRpcInternalException;
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidParamsException;
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidRequestException;
@@ -29,6 +31,9 @@ import org.tron.core.services.jsonrpc.types.CallArguments;
 import org.tron.core.services.jsonrpc.types.TransactionReceipt;
 import org.tron.core.services.jsonrpc.types.TransactionResult;
 
+/**
+ * Error code refers to https://www.quicknode.com/docs/ethereum/error-references
+ */
 @Component
 public interface TronJsonRpc {
 
@@ -292,9 +297,10 @@ public interface TronJsonRpc {
 
   @JsonRpcMethod("eth_newBlockFilter")
   @JsonRpcErrors({
+      @JsonRpcError(exception = JsonRpcExceedLimitException.class, code = -32005, data = "{}"),
       @JsonRpcError(exception = JsonRpcMethodNotFoundException.class, code = -32601, data = "{}"),
   })
-  String newBlockFilter() throws JsonRpcMethodNotFoundException;
+  String newBlockFilter() throws JsonRpcExceedLimitException, JsonRpcMethodNotFoundException;
 
   @JsonRpcMethod("eth_uninstallFilter")
   @JsonRpcErrors({
@@ -472,5 +478,35 @@ public interface TronJsonRpc {
       }
       this.removed = removed;
     }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || this.getClass() != o.getClass()) {
+        return false;
+      }
+      LogFilterElement item = (LogFilterElement) o;
+      if (!Objects.equals(blockHash, item.blockHash)) {
+        return false;
+      }
+      if (!Objects.equals(transactionHash, item.transactionHash)) {
+        return false;
+      }
+      if (!Objects.equals(transactionIndex, item.transactionIndex)) {
+        return false;
+      }
+      if (!Objects.equals(logIndex, item.logIndex)) {
+        return false;
+      }
+      return removed == item.removed;
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(blockHash, transactionHash, transactionIndex, logIndex, removed);
+    }
+
   }
 }
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
@@ -11,6 +11,8 @@ import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.getTxID;
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.triggerCallContract;
 
 import com.alibaba.fastjson.JSON;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.GeneratedMessageV3;
 import java.io.Closeable;
@@ -25,10 +27,10 @@ import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.bouncycastle.util.encoders.Hex;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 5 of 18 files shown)


framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
```diff
@@ -472,5 +474,39 @@ public LogFilterElement(String blockHash, Long blockNum, String txId, Integer tx
...
+        return false;
+      }
+      if (!Objects.equals(logIndex, item.logIndex)) {
+        return false;
+      }
+      return removed == item.removed;
+    }
+
+    @Override
+    public int hashCode() {
```
bladehan1, line 507: Hashcodes can be duplicated.

For different logs of the same transaction, the hashcode of an element with `remove=true` in the previous log will be the same as the hashcode of the element in the next log.

317787106: logIndex is different, so the hashCode will not be same.

bladehan1: hashCode1:  logIndex=1 and remove=true 
hashCode2: logIndex=2 and remove=false
hashCode1==hashCode2

0xbigapple, line 507: `Objects.hash(blockHash, transactionHash, transactionIndex, logIndex, removed) ` is more simple?

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
```diff
@@ -23,6 +24,7 @@
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidRequestException;
 import org.tron.core.exception.jsonrpc.JsonRpcMethodNotFoundException;
 import org.tron.core.exception.jsonrpc.JsonRpcTooManyResultException;
+import org.tron.core.net.peer.Item;
```
0xbigapple: This import is not used.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
317787106: Good idea.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
317787106: Yes, it will be removed.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
bladehan1: hashCode1: logIndex=1 and remove=true
hashCode2: logIndex=2 and remove=false
hashCode1==hashCode2

317787106: Good suggestion. Use `Objects.hash(blockHash, transactionHash, transactionIndex, logIndex, removed)` instead of it.

bladehan1 (approved):

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -226,8 +236,17 @@ public static void handleLogsFilter(LogsFilterCapsule logsFilterCapsule) {
           LogMatch.matchBlock(logFilter, logsFilterCapsule.getBlockNumber(),
               logsFilterCapsule.getBlockHash(), logsFilterCapsule.getTxInfoList(),
               logsFilterCapsule.isRemoved());
-      if (CollectionUtils.isNotEmpty(elements)) {
-        logFilterAndResult.getResult().addAll(elements);
+
+      for (LogFilterElement element : elements) {
+        LogFilterElement newElement;
+        try {
+          //compare with hashcode() first, then with equals(). If not exist, put it.
```
waynercheung: change to "// compare"? add blank before 'compare'.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -226,8 +236,17 @@ public static void handleLogsFilter(LogsFilterCapsule logsFilterCapsule) {
...
+          newElement = logElementCache.get(element, () -> element);
+        } catch (ExecutionException e) {
+          logger.error("Getting/loading LogFilterElement from cache fails", e);//never happen
```
waynercheung: ‘e);//never happen’ -->  'e); // never happen'

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
```diff
@@ -423,6 +423,7 @@ public List<String> getCheckpointList() {
 
   private void deleteCheckpoint() {
     if(checkTmpStore == null) {
+      //only occurs in mock test. TODO fix test
```
waynercheung: '//only' --> '// only'?

framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
```diff
@@ -546,6 +546,7 @@ public boolean isBusy() {
     }
     int queueSize = 0;
     if (eventListeners == null || eventListeners.isEmpty()) {
+      //only occurs in mock test. TODO fix test
```
waynercheung: '//only' --> '// only'

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: fix

chainbase/src/main/java/org/tron/core/db2/core/SnapshotManager.java
317787106: fix

framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
317787106: fix

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -1529,6 +1548,7 @@ public static Object[] getFilterResult(String filterId, Map<String, BlockFilterA
 
   @Override
   public void close() throws IOException {
+    logElementCache.invalidateAll();
```
waynercheung, line 1576: Is it necessary to call invalidateAll()? 

The close() method is invoked when the application is about to stop completely. When the entire application (the JVM process) exits, all the memory it occupies, including all objects stored in the logElementCache, will be automatically reclaimed by the operating system. 

At the very last moment before the application shuts down, would manually clearing a memory cache that is about to vanish along with it (i.e., invalidateAll()) consume the final CPU cycles to perform a meaningless task?

halibobo1205 (approved):

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -527,6 +527,9 @@ public class CommonParameter {
   @Getter
   @Setter
   public int jsonRpcMaxSubTopics = 1000;
+  @Getter
+  @Setter
+  public int jsonRpcMaxFilterNum = 50000;
```
0xbigapple: Has this default value been tested, or is there any other reference for it?

317787106: Yes. It has been tested for mainnet. But nile should set it as 30000 or smaller.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: also clean blockHashCache is better ?

317787106: Yes, i will add it.

bladehan1 (approved):

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
```diff
@@ -292,9 +296,10 @@ String newFilter(FilterRequest fr) throws JsonRpcInvalidParamsException,
 
   @JsonRpcMethod("eth_newBlockFilter")
   @JsonRpcErrors({
+      @JsonRpcError(exception = JsonRpcInvalidRequestException.class, code = -32600, data = "{}"),
```
wink-link: According to [jsonrpc_specification](https://www.jsonrpc.org/specification), 
-32600 means "Invalid Request The JSON sent is not a valid Request object", but when the eth_newBlockFilter request exceeds the max num, the request itself is valid, so change to another code?

waynercheung: According to jsonrpc_specification,
-32600 means "Invalid Request The JSON sent is not a valid Request object", but when the eth_newBlockFilter request exceeds the max num, the request itself is valid, so change to another code?

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -181,16 +195,28 @@ public static void handleBLockFilter(BlockFilterCapsule blockFilterCapsule) {
...
         it.remove();
         continue;
       }
-      entry.getValue().getResult().add(ByteArray.toJsonHex(blockFilterCapsule.getBlockHash()));
+      String finalBlockHash = blockHash;
+      try {
+        // compare with hashcode() first, then with equals(). If not exist, put it.
+        blockHash = blockHashCache.get(blockHash, () -> finalBlockHash);
+      } catch (ExecutionException e) {
+        logger.error("Getting/loading blockHash from cache fails", e);// never happen
```
waynercheung: ';// never happen' --> '; // never happen'

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -226,8 +252,17 @@ public static void handleLogsFilter(LogsFilterCapsule logsFilterCapsule) {
...
-      if (CollectionUtils.isNotEmpty(elements)) {
-        logFilterAndResult.getResult().addAll(elements);
+
+      for (LogFilterElement element : elements) {
+        LogFilterElement newElement;
+        try {
+          // compare with hashcode() first, then with equals(). If not exist, put it.
+          newElement = logElementCache.get(element, () -> element);
+        } catch (ExecutionException e) {
+          logger.error("Getting/loading LogFilterElement from cache fails", e);// never happen
```
waynercheung: ';// never happen' --> '; // never happen'

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
317787106: OK, replace it with JsonRpcExceedLimitException.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -181,16 +196,28 @@ public static void handleBLockFilter(BlockFilterCapsule blockFilterCapsule) {
...
 
+    String blockHash = ByteArray.toJsonHex(blockFilterCapsule.getBlockHash());
     while (it.hasNext()) {
       Entry<String, BlockFilterAndResult> entry = it.next();
       if (entry.getValue().isExpire()) {
         it.remove();
         continue;
       }
-      entry.getValue().getResult().add(ByteArray.toJsonHex(blockFilterCapsule.getBlockHash()));
+      String finalBlockHash = blockHash;
```
waynercheung: what about moving cache lookup outside the loop to avoid redundant queries.

Since `originalBlockHash` is constant, querying the cache N times in the 
loop is unnecessary. Moving it outside:
- Reduces O(N) cache accesses to O(1)
- Improves performance with identical functionality
- Better error handling with fallback strategy


```

// Approach A: Optimal - Query the cache once outside the loop (Recommended)
final String originalBlockHash = ByteArray.toJsonHex(blockFilterCapsule.getBlockHash());

// Query the cache only once
String cachedBlockHash;
try {
    cachedBlockHash = blockHashCache.get(originalBlockHash, () -> originalBlockHash);
} catch (ExecutionException e) {
    logger.error("Getting/loading blockHash from cache failed", e);
    cachedBlockHash = originalBlockHash;  // Fallback: use the original value
}

// Use the cached result directly inside the loop
while (it.hasNext()) {
    Entry<String, BlockFilterAndResult> entry = it.next();
    if (entry.getValue().isExpire()) {
        it.remove();
        continue;
    }
    entry.getValue().getResult().add(cachedBlockHash);
}
```

317787106: It might put unnecessary blockHash into blockHashCache if filter map is empty. But it's a good suggestion

kuny0707 (approved):

PR #6500, opened 2025-12-30
func(vm): optimize selfdestruct restriction

yanghang8612:


actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java
```diff
@@ -936,8 +951,15 @@ public Pair<Boolean, byte[]> execute(byte[] rawData) {
       byte[] hash = Sha256Hash.hash(CommonParameter
           .getInstance().isECKeyCryptoEngine(), combine);
 
-      byte[][] signatures = extractBytesArray(
-          words, words[3].intValueSafe() / WORD_SIZE, rawData);
+      if (VMConfig.allowTvmSelfdestructRestriction()) {
+        int sigArraySize = words[words[3].intValueSafe() / WORD_SIZE].intValueSafe();
```
SeriousCoding789, line 955: This should be a general improvement?

yanghang8612: Yep, we enhance processing efficiency by moving the judgment to the front end.

SeriousCoding789: Then the PR title and description better adjust accordingly.

actuator/src/main/java/org/tron/core/vm/program/Memory.java
```diff
@@ -98,7 +98,7 @@ public void write(int address, byte[] data, int dataSize, boolean limited) {
 
   public void extendAndWrite(int address, int allocSize, byte[] data) {
     extend(address, allocSize);
-    write(address, data, data.length, false);
+    write(address, data, allocSize, false);
```
SeriousCoding789, line 101: Not sure whether below is better?
```
int writeSize = Math.min(allocSize, data.length);
write(address, data, writeSize, true); 
```

yanghang8612: Your suggestion does not align with the original intent of this method's design.

actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java
yanghang8612: There is no difference; the goal is to optimize the proposal content.

PR #6511, opened 2026-01-15
feat(vm): save in database the input and output of InternalTransactions

roseshahar:
**What does this PR do?**
Add input (`data`) and output of InternalTransactions (i.e. traces in EVM)

You can learn about the attributes of a trace [here](https://www.alchemy.com/docs/reference/what-are-evm-traces)

Since TRON doesn't save these specific fields in its database, indexers are unable to fetch very important actions that happen inside the network, like function calls and contract deployments via factories.

- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/program/Program.java
+++ b/actuator/src/main/java/org/tron/core/vm/program/Program.java
@@ -816,6 +816,11 @@ public class Program {
     // 4. CREATE THE CONTRACT OUT OF RETURN
     byte[] code = createResult.getHReturn();
 
+    // capture output (deployed code) on internal transaction for indexing
+    if (internalTx != null) {
+      internalTx.setOutput(code);
+    }
+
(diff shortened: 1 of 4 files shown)


bretk1989 (approved):

lvs0075:
This feature primarily involves changes to the storage protocol, so it requires thorough discussion within the community：https://github.com/tronprotocol/tips/issues/816#issue-3820903930.

halibobo1205:
Thanks for the contribution! The idea of exposing internal transaction input/output for indexers is valuable.

However, it looks like there are still some open questions in the related TIP discussion (https://github.com/tronprotocol/tips/issues/816) — notably the pending 200K-block storage validation and the community consensus on configuration approach.

Would you mind converting this PR to draft status for now? Once the outstanding items in the TIP discussion are resolved and a consensus is reached, it would be great to move this forward. This way the PR won't get stale review cycles in the meantime.

Happy to re-review once the TIP discussion concludes!

PR #6514, opened 2026-01-20
fix(crypto): throw exception if system shielded tx is not supported.

Federico2014:
**What does this PR do?**
Throw an exception if the system shielded tx is not supported to avoid network resource abuse.  

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -874,7 +874,7 @@ public class Manager {
 
     if (isShieldedTransaction(trx.getInstance()) && !chainBaseManager.getDynamicPropertiesStore()
         .supportShieldedTransaction()) {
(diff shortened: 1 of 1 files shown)


halibobo1205 (approved):

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -874,14 +874,13 @@ public boolean pushTransaction(final TransactionCapsule trx)
 
     if (isShieldedTransaction(trx.getInstance()) && !chainBaseManager.getDynamicPropertiesStore()
         .supportShieldedTransaction()) {
-      return false;
+      throw new ContractValidateException("System shielded transaction is not supported.");
```
waynercheung: what about using "ShieldedTransferContract is not supported"?

warku123 (approved):

kuny0707 (approved):

PR #6527, opened 2026-01-26
chore: add safety guard for negative cycle index

operagxsasha:
**What does this PR do?**

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/chainbase/src/main/java/org/tron/core/service/MortgageService.java
+++ b/chainbase/src/main/java/org/tron/core/service/MortgageService.java
@@ -213,9 +213,6 @@ public class MortgageService {
       beginCycle = oldEndCycle;
     }
     if (beginCycle < endCycle) {
-        if (beginCycle <= 0) {
-            return 0L;
-        }
     
         for (Pair<byte[], Long> vote : srAddresses) {
         


chainbase/src/main/java/org/tron/core/service/MortgageService.java
```diff
@@ -214,6 +214,10 @@ private long computeReward(long beginCycle, long endCycle, AccountCapsule accoun
     }
     if (beginCycle < endCycle) {
       for (Pair<byte[], Long>  vote : srAddresses) {
+        if (beginCycle == 0) {
```
bladehan1: Image
beginCycle will be start with 1

operagxsasha: Correct, this prevents beginCycle from being 0

vividctrlalt: I agree with @bladehan1's observation. After tracing the call sites of `computeReward(beginCycle, endCycle, accountCapsule)`:

1. In `withdrawReward()` (line 94): `beginCycle = delegationStore.getBeginCycle(address)`, and the delegation cycle starts from 1.
2. Before reaching line 215, `beginCycle` has already been validated (`beginCycle > currentCycle` returns early at line 98-100), and may have been incremented (`beginCycle += 1` at line 116).
3. In `queryReward()` (line 142): same pattern — `beginCycle` is loaded from store and validated before use.

So `beginCycle == 0` cannot occur in normal execution flow. Additionally, even as a defensive check, using `continue` inside the vote loop is incorrect — it skips individual votes rather than handling the invalid cycle at the method level. A proper guard would be placed before the for-loop, e.g., `if (beginCycle <= 0) return 0;`.

Also noting that the indentation of the added code is inconsistent with the surrounding code style.

chainbase/src/main/java/org/tron/core/service/MortgageService.java
```diff
@@ -214,6 +214,10 @@ private long computeReward(long beginCycle, long endCycle, AccountCapsule accoun
...
+       continue;
```
warku123: BTW, the indentation looks off this line. Please fix to match project style (4 spaces).

chainbase/src/main/java/org/tron/core/service/MortgageService.java
```diff
@@ -213,7 +213,12 @@ private long computeReward(long beginCycle, long endCycle, AccountCapsule accoun
       beginCycle = oldEndCycle;
     }
     if (beginCycle < endCycle) {
-      for (Pair<byte[], Long>  vote : srAddresses) {
+        if (beginCycle <= 0) {
```
warku123: As @bladehan1 said, this check may not be necessary. But if we still add it for safety, consider placing it at **line 200** (method entry) for fail-fast, rather than inside the nested `if` block. When `beginCycle <= 0`, we should return early before any cycle logic.

Also, please fix the **indentation on line 220** — it appears to use extra spaces (should align with 4-space standard).

operagxsasha: After reviewing call sites and existing validation, I agree that beginCycle <= 0 cannot occur in normal execution flow. Removing the defensive guard to preserve protocol invariants.

chainbase/src/main/java/org/tron/core/service/MortgageService.java
```diff
@@ -213,7 +213,9 @@ private long computeReward(long beginCycle, long endCycle, AccountCapsule accoun
       beginCycle = oldEndCycle;
     }
     if (beginCycle < endCycle) {
-      for (Pair<byte[], Long>  vote : srAddresses) {
+    
+        for (Pair<byte[], Long> vote : srAddresses) {
```
warku123, line 217: @operagxsasha 

The PR title mentions adding a "safety guard for negative cycle index", but the diff only shows formatting changes. 

Could you double-check if the intended logic changes are missing from this PR?

operagxsasha: Thanks for pointing this out. I'll double-check the intended logic change and update the PR if something is missing.

lvs0075:
At the moment, I can only see formatting changes. If there are no further updates, I will close this PR. Please feel free to submit another PR when you have a better idea.

Sunny6889:
Close it as mentioned above, please submit follow up in new PRs.

PR #6531, opened 2026-01-29
feat(event): optimize the event cache solid block ID update logic

zeusoo001:
**What does this PR do?**
Optimize the event cache solid block ID update logic.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/test/java/org/tron/core/event/BlockEventCacheTest.java
+++ b/framework/src/test/java/org/tron/core/event/BlockEventCacheTest.java
@@ -42,10 +42,10 @@ public class BlockEventCacheTest {
     BlockCapsule.BlockId blockId = new BlockCapsule.BlockId(getBlockId(), 2);
     event.setBlockId(blockId);
     event.setParentId(b1);
-    event.setSolidId(b1);
+    event.setSolidId(blockId);
     BlockEventCache.add(event);
     Assert.assertEquals(event, BlockEventCache.getHead());
-    Assert.assertEquals(b1, BlockEventCache.getSolidId());
+    Assert.assertEquals(blockId, BlockEventCache.getSolidId());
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/tron/core/services/event/BlockEventCache.java
```diff
@@ -66,7 +66,14 @@ public static void add(BlockEvent blockEvent) throws EventException {
     }
 
     if (blockEvent.getSolidId().getNum() > solidId.getNum()) {
-      solidId = blockEvent.getSolidId();
+      BlockCapsule.BlockId headBlockId = head.getBlockId();
+      if (blockEvent.getSolidId().getNum() <= headBlockId.getNum()) {
+        solidId = blockEvent.getSolidId();
+      } else if (blockEvent.getBlockId().equals(headBlockId)) {
```
bladehan1, line 72: From the perspective of the `blockEvent` event: 
`blockEvent.getBlockId().getNum()` >= `blockEvent.getSolidId().getNum()`
From the perspective of the `head` assignment:
`head.getBlockId().getNum()` >= `blockEvent.getBlockId().getNum()`

Therefore, it should always be true:
`headBlockId.getNum()>= blockEvent.getBlockId().getNum()>= blockEvent.getSolidId().getNum()`

That is,
`headBlockId.getNum()>= blockEvent.getSolidId().getNum()`

In any case, what situation might cause it to not be true?

xxo1shine: `blockEvent.getSolidId()` represents the current solidified block height. For example, if 218 blocks are currently synchronized, when loading the 100th block, the current solidified block height might be 200.

PR #6565, opened 2026-03-02
refactor(config): extract ConfigKey and remove testnet config

vividctrlalt:
## Summary

- Extract ~232 HOCON config key constants from `Constant.java` into a new `ConfigKey.java`, leaving only business constants in `Constant.java`
- Remove obsolete testnet configuration and `0xa0` address format
- Unify all config files and tests to use mainnet address format (`T-` Base58 / `41` hex)

## Changes

- **New `ConfigKey.java`**: All HOCON config path strings (e.g. `"node.rpc.port"`) moved here, organized by config section
- **Cleaned `Constant.java`**: Only ~32 business constants remain, grouped by domain (address, transaction, energy, proposal, etc.)
- **Updated `Args.java`**: References changed from `Constant.XXX` to `ConfigKey.XXX` for config keys; address prefix hardcoded to mainnet
- **Config files**: Removed testnet `addressPrefix` setting; `net.type` marked as deprecated
- **Test configs**: Converted all genesis/witness addresses from `27-` prefix (testnet) to `T-` prefix (mainnet) and `a0` hex to `41` hex
- **FreezeTest**: Fixed CREATE2 factory bytecode magic byte (`0xa0` → `0x41`) to match mainnet prefix
- **IstanbulTest**: Updated expected chain ID to match new genesis block hash

## Test plan

- [x] Full test suite: 2326 tests passed, 0 failures, 24 ignored


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -411,17 +415,11 @@ public static void setParam(final String[] args, final String confFileName) {
    */
   public static void setParam(final Config config) {
```
warku123, line 417: Hi @vividctrlalt, thanks for this refactoring PR! The separation of `ConfigKey` from `Constant` definitely improves code maintainability. 

I have a question about  compatibility:

This PR completely removes the `net.type` configuration support and hardcodes the mainnet address prefix (`0x41`). While I understand the goal to simplify the codebase, I'm wondering if we should consider users who might still have `net.type=testnet` in their existing config files.

**Would it be better to keep the config parsing but add a deprecation warning:**

```java
if (config.hasPath("net.type")) {
  logger.warn("net.type is deprecated since v4.x and will be ignored. "
      + "Mainnet address format (0x41) is now always used. "
      + "Please remove this option from your config file.");
}
```

What do you think?

vividctrlalt: Thanks for the review!

Regarding `net.type`: this config has always been misleading — it was never actually used in production. The only references were in test code. Since no real user depends on it, a deprecation warning would just add unnecessary noise. Removing it cleanly is the right approach.

Regarding the `ConfigKey` extraction: ~88% of the constants in `Constant.java` are HOCON config path strings used exclusively by `Args.java`. Keeping them in `Constant` pollutes a class that should only hold business constants. Extracting them into `ConfigKey` gives `Args` its own dedicated config key registry and keeps `Constant` clean

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -101,6 +101,7 @@ public class Args extends CommonParameter {
 
   public static void clearParam() {
     PARAMETER.shellConfFileName = "";
+    PARAMETER.configFilePath = "";
```
Sunny6889, line 104: The newly introduced `configFilePath` field seems redundant. Looking at where it's set in `Args.setParam()`:

```java
PARAMETER.setConfigFilePath(
    StringUtils.isNoneBlank(PARAMETER.shellConfFileName)
        ? PARAMETER.shellConfFileName : confFileName);
Config config = Configuration.getByFileName(PARAMETER.shellConfFileName, confFileName);
```

`configFilePath` is simply the resolved result of `shellConfFileName ?? confFileName` — no actual path construction or concatenation happens despite the name suggesting otherwise. It's then consumed only by `DynamicArgs`:

```java
// init()
configFile = new File(parameter.getConfigFilePath());
// reload()
Config config = Configuration.getByFileName(parameter.getConfigFilePath(), null);
```

This can be achieved without adding a new field. For example, just backfill `shellConfFileName` with the default when it's blank:

```java
if (StringUtils.isBlank(PARAMETER.shellConfFileName)) {
    PARAMETER.shellConfFileName = confFileName;
}
```

Then `DynamicArgs` can simply use `parameter.getShellConfFileName()` directly, and the new `@Getter @Setter` field on `CommonParameter` can be removed entirely.

If the intent is to keep a separate "resolved config name" field, consider at least renaming it to something more accurate like `resolvedConfFileName`, since the current name `configFilePath` implies a full filesystem path which is misleading.

framework/src/main/java/org/tron/core/config/args/Args.java
Sunny6889: Could this be a breaking change for anyone running a private TRON chain with net.type = testnet?

The removal of the `net.type` conditional means any private fork configured with `net.type = testnet` (using `0xa0` address prefix) would silently switch to `0x41` after upgrading, with no warning logged. While no known TRON network actually uses `0xa0`, it would be safer to log a deprecation warning instead of silently ignoring it:

```java
if (config.hasPath("net.type") 
    && "testnet".equalsIgnoreCase(config.getString("net.type"))) {
    logger.warn("net.type = testnet is deprecated and has no effect. "
        + "Address prefix defaults to mainnet (0x41).");
}
```

framework/src/main/java/org/tron/program/FullNode.java
```diff
@@ -21,7 +20,7 @@ public class FullNode {
    */
   public static void main(String[] args) {
     ExitManager.initExceptionHandler();
-    Args.setParam(args, Constant.NET_CONF);
+    Args.setParam(args, "config.conf");
```
bladehan1, line 23: Ideally, it should be a constant.

framework/src/test/java/org/tron/common/runtime/vm/VMContractTestBase.java
```diff
@@ -46,9 +46,8 @@ public class VMContractTestBase {
   MortgageService mortgageService;
 
   static {
-    // 27Ssb1WE8FArwJVRRb8Dwy3ssVGuLY8L3S1 (test.config)
-    WITNESS_SR1_ADDRESS =
-        Constant.ADD_PRE_FIX_STRING_TESTNET + "299F3DB80A24B20A254B89CE639D59132F157F13";
+    // TDmHUBuko2qhcKBCGGafu928hMRj1tX2RW (test.config)
+    WITNESS_SR1_ADDRESS = "41" + "299F3DB80A24B20A254B89CE639D59132F157F13";
```
bladehan1, line 50: Ideally, it should be a constant, mentioned as 41.

framework/src/test/java/org/tron/core/config/args/ArgsTest.java
```diff
@@ -59,7 +58,7 @@ public void destroy() {
   @Test
   public void get() {
     Args.setParam(new String[] {"-c", TestConstants.TEST_CONF, "--keystore-factory"},
-        Constant.NET_CONF);
+        "config.conf");
```
bladehan1, line 61: Ideally, it should be a constant.

framework/src/test/java/org/tron/core/exception/TronErrorTest.java
```diff
@@ -138,7 +137,7 @@ public void rateLimiterServletInitTest() {
   @Test
   public void shutdownBlockTimeInitTest() {
     Map<String, String> params = new HashMap<>();
-    params.put(Constant.NODE_SHUTDOWN_BLOCK_TIME, "0");
+    params.put("node.shutdown.BlockTime", "0");
```
bladehan1, line 140: Ideally, it should be a constant.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Good catch\! You're right that `configFilePath` is essentially just the resolved result of `shellConfFileName ?? confFileName`. Backfilling `shellConfFileName` directly is cleaner. Will fix this.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Thanks for the suggestion. However, no private chain could actually use `net.type = testnet` in practice — simply setting this config alone wouldn't make a functional testnet. It would require extensive changes to other configurations as well. Moreover, the codebase had already broken the semantics of this field by binding it to the meaningless `0xa0` address prefix purely for testing purposes.

So I still prefer removing it entirely. That said, adding a deprecation warning is a reasonable compromise — it's low cost and helps users understand that this config has no effect.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: I understand this variable represents the current configuration file path; this resolves the issue of whether it originates from the default value or a shell parameter.

After the node starts, reading the configuration file path only requires this variable, without needing additional checks.

framework/src/main/java/org/tron/program/FullNode.java
vividctrlalt: I'd prefer keeping it as a literal here. As the entry point of the application, explicitly specifying the config filename in `main()` is more readable — you can see at a glance what file the node loads on startup. This is the only place in the codebase that uses `"config.conf"`, so there's no duplication concern and no need to extract it into a constant.

framework/src/test/java/org/tron/core/exception/TronErrorTest.java
vividctrlalt: All config key constants have been moved to `ConfigKey` and made package-private, accessible only to `Args.java`. This is intentional — it prevents unrelated code from arbitrarily adding or referencing config keys. Test code cannot access these constants directly, and tests should not break the encapsulation of the main code. Given that test cases are short and focused, using inline config strings is clearer and more readable.

framework/src/test/java/org/tron/common/runtime/vm/VMContractTestBase.java
vividctrlalt: The `41` hex prefix is hardcoded throughout the entire codebase — it's the only valid address prefix and will never change. Using it directly in test code is straightforward and readable.

framework/src/test/java/org/tron/core/config/args/ArgsTest.java
vividctrlalt: Same reasoning as above — there's no need to introduce a string constant for `"config.conf"` just for test code. Using the filename directly is clearer and more readable. The previous approach of wrapping config filenames in constants is not a common practice and adds indirection without benefit.

framework/src/main/java/org/tron/core/config/args/DynamicArgs.java
```diff
@@ -52,36 +51,16 @@ public void init() {
...
-      confFilePath = parameter.getShellConfFileName();
-    } else  {
-      confFilePath = Constant.NET_CONF;
-    }
-
-    File confFile = new File(confFilePath);
-    if (!confFile.exists()) {
-      logger.warn("Configuration path is required! No such file {}", confFile);
-      return null;
+    long lastModifiedTime = configFile.lastModified();
```
Sunny6889, line 54: I think here configFile need a Null protection, since here `run()` is a public method, what if `parameter.isDynamicConfigEnable()` return false and here `run()` is called, it will return NullPointerException.
Another risk would be: while running the configuration file is accidentally deleted, what will happen?

framework/src/main/java/org/tron/core/config/args/DynamicArgs.java
```diff
@@ -52,36 +51,16 @@ public void init() {
...
+      lastModified = lastModifiedTime;
     }
-    return confFile;
   }
 
   public void reload() {
     logger.debug("Reloading ... ");
-    Config config = Configuration.getByFileName(parameter.getShellConfFileName(),
-        Constant.NET_CONF);
+    Config config = Configuration.getByFileName(parameter.getConfigFilePath(), null);
```
Sunny6889, line 63: please confirm is it really safe to remove the fallback `Constant.NET_CONF`?

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: @Sunny6889 Although your suggestion makes sense, but  this modification would cause test failures due to shared mutable state. The root cause of this issue lies in the parameter initialization logic. Please refer to #6567 for details. The redundant variable issue you mentioned will be addressed as part of the parameter initialization logic refactoring.

framework/src/main/java/org/tron/core/config/args/DynamicArgs.java
vividctrlalt: In the current code logic, this is fully guaranteed. The readability issue will be addressed in the parameter initialization refactoring.

framework/src/main/java/org/tron/core/config/args/Args.java
Sunny6889: ok we can keep this variable, but I suggest a better name such as `resolvedConfFileName` instead of `configFilePath` @vividctrlalt, as FilePath is a bit missleading.

framework/src/main/java/org/tron/core/config/args/DynamicArgs.java
vividctrlalt: 1. DynamicArgs is a Spring @Component — no external code calls run() directly. This is an internal class, not a public API.

2. Config file deleted at runtime — File.lastModified() returns 0L when the file doesn't exist and won't throw an exception. Since 0 > lastModified is false, reload() won't be triggered. No issue here.

Sunny6889: but the test code is actually call this `run()`

kuny0707 (approved):

framework/src/main/java/org/tron/core/config/args/DynamicArgs.java
vividctrlalt: Minimizing over-defensive checks is a good practice. The original null checks could actually hide bugs. Unnecessary null checks mislead readers into thinking null is a possible state. As for the test case, it just needs to be properly initialized.

vividctrlalt: DynamicArgs is an internal @Component where run() is only invoked after init() guarantees configFile is initialized.

PR #6569, opened 2026-03-04
refactor(config): extract CLIParameter and restructure Args init flow

### LINKED ISSUE #6567: CLI flags silently overridden by config file for 13 parameters
#### Software Versions

```
Version : develop branch (commit 13b68eac0)
```

#### Expected behaviour

CLI flags should take priority over config file values. This is the standard convention (`CLI > config file > default`) and is already correctly implemented for 9 other parameters (e.g. `--storage-db-directory`, `--trust-node`) using the `Optional` pattern.

```bash
java -jar FullNode.jar --rpc-thread 16
# Expected: rpcThreadNum = 16
```

#### Actual behaviour

For 13 parameters, `Args.setParam(Config)` unconditionally overwrites the CLI value with the config file value. The CLI flags are silently ignored.

```bash
# config.conf contains: node.rpc.thread = 8
java -jar FullNode.jar --rpc-thread 16
# Actual: rpcThreadNum = 8 (CLI value lost)
```

Root cause — unconditional assignment in `Args.setParam()`:

```java
PARAMETER.rpcThreadNum = config.hasPath(ConfigKey.NODE_RPC_THREAD)
    ? config.getInt(ConfigKey.NODE_RPC_THREAD)
    : Runtime.getRuntime().availableProcessors() / 2;
// JCommander-set value of 16 is overwritten to 8
```

#### Frequency

100% — every time a config file key exists for any of the 13 affected parameters.

1. Set `node.rpc.thread = 8` in `config.conf`
2. Start node with `java -jar FullNode.jar -c config.conf --rpc-thread 16`
3. Observe that `rpcThreadNum` is 8, not 16

Any of the following 13 parameters can reproduce the same issue:

| # | CLI Flag | Field | Config Key |
|---|----------|-------|------------|
| 1 | `--support-constant` | `supportConstant` | `vm.supportConstant` |
| 2 | `--max-energy-limit-for-constant` | `maxEnergyLimitForConstant` | `vm.maxEnergyLimitForConstant` |
| 3 | `--lru-cache-size` | `lruCacheSize` | `vm.lruCacheSize` |
| 4 | `--min-time-ratio` | `minTimeRatio` | `vm.minTimeRatio` |
| 5 | `--max-time-ratio` | `maxTimeRatio` | `vm.maxTimeRatio` |
| 6 | `--long-running-time` | `longRunningTime` | `vm.longRunningTime` |
| 7 | `--save-internaltx` | `saveInternalTx` | `vm.saveInternalTx` |
| 8 | `--save-featured-internaltx` | `saveFeaturedInternalTx` | `vm.saveFeaturedInternalTx` |
| 9 | `--max-connect-number` | `maxHttpConnectNumber` | `node.maxHttpConnectNumber` |
| 10 | `--rpc-thread` | `rpcThreadNum` | `node.rpc.thread` |
| 11 | `--solidity-thread` | `solidityThreads` | `node.solidity.threads` |
| 12 | `--validate-sign-thread` | `validateSignThreadNum` | `node.validateSignThreadNum` |
| 13 | `--history-balance-lookup` | `historyBalanceLookup` | `storage.balance.history.lookup` |

#### Backtrace

Not applicable — no crash or exception. The bug is silent incorrect behavior.

---

Below is the full findings of all 34 `@Parameter` CLI flags in `CommonParameter` and additional findings.

<details>
<summary>full findings: all 34 CLI parameters by category</summary>

### Category 1: Startup parameters (6) — No issue

Only control startup flow, no config file overlap.

| CLI Flag | Field | Description |
|----------|-------|-------------|
| `-c`, `--config` | `shellConfFileName` | Config file path |
| `-d`, `--output-directory` | `outputDirectory` | Data directory |
| `--log-config` | `logbackPath` | Log config file |
| `-h`, `--help` | `help` | Print help and exit |
| `-v`, `--version` | `version` | Print version and exit |
| `--password` | `password` | Witness password |

### Category 2: CLI-only parameters (10) — No issue

Only set by CLI, `setParam(Config)` does not override.

| CLI Flag | Field | Description |
|----------|-------|-------------|
| `-w`, `--witness` | `witness` | Witness mode |
| `-p`, `--private-key` | `privateKey` | Witness private key |
| `--witness-address` | `witnessAddress` | Witness address |
| `--debug` | `debug` | TVM debug mode |
| `--fast-forward` | `fastForward` | Fast forward mode |
| `--solidity` | `solidityNode` | Solidity node mode |
| `--keystore-factory` | `keystoreFactory` | KeystoreFactory mode |
| `--p2p-disable` | `p2pDisable` | Disable P2P module |
| `--es` | `eventSubscribe` | Enable event subscription |
| `--contract-parse-enable` | `contractParseEnable` | Contract parse switch |

### Category 3: CLI + Config, CLI priority (9) — Correct

Uses `Optional` pattern, CLI value takes priority when present.

| CLI Flag | Field | Config Key |
|----------|-------|------------|
| `--storage-db-directory` | `storageDbDirectory` | `storage.db.directory` |
| `--storage-db-engine` | `storageDbEngine` | `storage.db.engine` |
| `--storage-db-synchronous` | `storageDbSynchronous` | `storage.db.synchronous` |
| `--storage-index-directory` | `storageIndexDirectory` | `storage.index.directory` |
| `--storage-index-switch` | `storageIndexSwitch` | `storage.index.switch` |
| `--storage-transactionHistory-switch` | `storageTransactionHistorySwitch` | `storage.transactionHistory.switch` |
| `--trust-node` | `trustNodeAddr` | `node.trustNode` |
| `--save-cancel-all-unfreeze-v2-details` | `saveCancelAllUnfreezeV2Details` | `vm.saveCancelAllUnfreezeV2Details` |
| `--seed-nodes` | `seedNodes` | `seed.node.ip.list` |

### Category 4: CLI + Config, config overwrites CLI (13) — BUG

See "Steps to reproduce" section above for full list.

</details>

<details>
<summary>Other issues found during investigation</summary>

### `clearParam()` misses 9 fields

`clearParam()` is used for test state reset, but the following fields are not cleared, which may cause test pollution:

- `maxEnergyLimitForConstant`, `lruCacheSize`
- `allowDelegateOptimization`, `unfreezeDelayDays`
- `allowDynamicEnergy`, `dynamicEnergyThreshold`, `dynamicEnergyIncreaseFactor`, `dynamicEnergyMaxFactor`
- `saveCancelAllUnfreezeV2Details`

### Inconsistent default value patterns

Two patterns coexist in `Args.setParam()`:

- **Pattern A** (ternary): default value explicit — `config.hasPath(key) ? config.get(key) : defaultValue`
- **Pattern B** (if-block, no else): default value hidden in `CommonParameter` field declaration

Mixing both increases maintenance burden. Suggest unifying to Pattern A.

### `CommonParameter` overloaded

780 lines, ~150 fields, mixing CLI parameters, config file values, and runtime objects. Long-term suggestion: split into domain-specific classes (`RpcConfig`, `StorageConfig`, `VmConfig`, etc.).

</details>

<details>
<summary>Suggested improvement roadmap</summary>

### Industry comparison

| Project | CLI/Config Overlap | Strategy | Priority Rule |
|---------|-------------------|----------|---------------|
| **Nginx** | Nearly none | CLI for startup ops only; all business config in file | CLI > config |
| **Hadoop** | Full overlap | All items overridable via `-D key=value`; multi-layer XML merge | CLI > site XML > default XML |
| **Kubernetes** | Heavy → deprecating CLI | Migrating to config file; CLI flags marked deprecated | CLI > config file |

java-tron's current situation is most similar to early Kubernetes.

### Short-term (current release): Fix the priority bug

```java
if (PARAMETER.rpcThreadNum <= 0) {
    PARAMETER.rpcThreadNum = config.hasPath(ConfigKey.NODE_RPC_THREAD)
        ? config.getInt(ConfigKey.NODE_RPC_THREAD)
        : Runtime.getRuntime().availableProcessors() / 2;
}
```

### Mid-term (next release): Deprecate overlapping CLI flags

```java
@Deprecated  // Use node.rpc.thread in config.conf instead
@Parameter(names = {"--rpc-thread"}, description = "[DEPRECATED] Num of gRPC thread")
public int rpcThreadNum;
```

### Long-term (future release): Converge to Nginx model

Only keep ~8 startup-class CLI parameters (`-c`, `-d`, `-w`, `-p`, `-h`, `-v`, `--password`, `--log-config`). All other runtime parameters managed exclusively by config file.

</details>

vividctrlalt:
## Summary
- Extract CLI `@Parameter` annotations from `CommonParameter` into a new `CLIParameter` class, separating CLI parsing from runtime config state
- Restructure `Args.setParam()` into a clear 4-step flow: parse CLI → `applyConfigParams` → `applyCLIParams` → `initLocalWitnesses`
- Refactor `WitnessInitializer` into three static methods (`initFromCLIPrivateKey`, `initFromCFGPrivateKey`, `initFromKeystore`), with routing logic moved to `Args.initLocalWitnesses`
- Remove `password`, `privateKey`, `witnessAddress` from `CommonParameter` — sensitive credentials are now passed directly as method parameters and never stored in global state
- Fixes #6567
## Test plan
- [x] `./gradlew :framework:test --tests "org.tron.core.config.args.*"` — all parameter tests pass
- [x] `./gradlew :framework:test` — full framework tests pass


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1292,6 +1008,157 @@ public static void setParam(final Config config) {
...
+      PARAMETER.saveCancelAllUnfreezeV2Details = cmd.saveCancelAllUnfreezeV2Details;
+    }
+    if (assigned.containsKey("--long-running-time")) {
+      PARAMETER.longRunningTime = cmd.longRunningTime;
+    }
+    if (assigned.containsKey("--max-connect-number")) {
+      PARAMETER.maxHttpConnectNumber = cmd.maxHttpConnectNumber;
+    }
+    if (assigned.containsKey("--storage-db-directory")) {
+      PARAMETER.storageDbDirectory = cmd.storageDbDirectory;
```
bladehan1: In applyCLIParams(), 7 storage-related parameters are written only to intermediate fields on PARAMETER, not to the PARAMETER.storage object that is actually used at runtime. The approach in applyConfigParams() is correct (writing via PARAMETER.storage.setXxx()), and applyCLIParams() should be consistent with it.

Current code in applyCLIParams:

PARAMETER.storageDbDirectory = cmd.storageDbDirectory;

Should be changed to write directly to PARAMETER.storage, consistent with applyConfigParams:

PARAMETER.storage.setDbDirectory(cmd.storageDbDirectory);

Please check the relevant parameters.

vividctrlalt: Thanks for catching this!

The existing tests did not cover the scenario where CLI storage parameters override config file values, which is why this bug went undetected after the refactor. Two test cases have been added to cover both the CLI-overrides-config and config-defaults-without-CLI scenarios.

Fixed in c4cb0e15a. Please continue reviewing, thank you!

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1793,5 +1637,120 @@ public String getOutputDirectory() {
...
+    for (ParameterDescription parameterDescription : parameterDescriptionList) {
+      String parameterName = parameterDescription.getParameterized().getName();
+      stringParameterDescriptionMap.put(parameterName, parameterDescription);
+    }
+
+    StringBuilder helpStr = new StringBuilder();
+    helpStr.append("Name:\n\tFullNode - the java-tron command line interface\n");
+    String programName = Strings.isNullOrEmpty(jCommander.getProgramName()) ? "FullNode.jar" :
+        jCommander.getProgramName();
+    helpStr.append(String.format("%nUsage: java -jar %s [options] [seedNode <seedNode> ...]%n",
```
warku123, line 1678: @vividctrlalt, here I noticed that `seedNode` was defined as a JCommander Main Parameter (without names attribute). But in `applyCLIParams()`, I didn't find where it's being copied to `PARAMETER.seedNodes`.

I might have missed it, but I couldn't find the handling in applyCLIParams(). Should we add something like:

```
if (!cmd.seedNodes.isEmpty()) {
    PARAMETER.seedNodes = cmd.seedNodes;
}
```

Apologies if I misunderstood the code flow. Please let me know if I'm missing something! Thanks for your patience.

vividctrlalt:
@warku123 Good catch! You're right — `seedNodes` (JCommander main parameter) was not being copied to `PARAMETER.seedNode` in `applyCLIParams()`. 

Fixed in 9c2c55bda:
- Inlined `loadSeeds` so it only reads from config
- Added `seedNodes` handling in `applyCLIParams()` to override config values when CLI args are provided

Thanks for the careful review!

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1292,6 +986,164 @@ public static void setParam(final Config config) {
...
+      PARAMETER.outputDirectory = cmd.outputDirectory;
+    }
+    if (assigned.containsKey("--witness")) {
+      PARAMETER.witness = cmd.witness;
+    }
+    if (assigned.containsKey("--support-constant")) {
+      PARAMETER.supportConstant = cmd.supportConstant;
+    }
+    if (assigned.containsKey("--max-energy-limit-for-constant")) {
+      PARAMETER.maxEnergyLimitForConstant = cmd.maxEnergyLimitForConstant;
```
waynercheung, line 1011: We should also has the value check as the `applyConfigParams`(previously `applyConfigParams`) does:
`PARAMETER.maxEnergyLimitForConstant = max(3_000_000L, cmd.maxEnergyLimitForConstant, true);`

vividctrlalt: Thanks for the review! A couple of notes on this:

1. This is actually a pre-existing issue, not something introduced by this refactor. The original code before the refactor also directly assigns `cmd.maxEnergyLimitForConstant` without the `max()` check — so this is outside the scope of the current refactoring PR.

2. The long-term direction is to move towards a fully configuration-file-driven approach (similar to Nginx), eliminating CLI parameter overrides entirely. These CLI parameters are expected to be deprecated in the future, so investing in additional validation for them may not be worthwhile.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1292,6 +986,164 @@ public static void setParam(final Config config) {
...
+    if (assigned.containsKey("--p2p-disable")) {
+      PARAMETER.p2pDisable = cmd.p2pDisable;
+    }
+    if (assigned.containsKey("--history-balance-lookup")) {
+      PARAMETER.historyBalanceLookup = cmd.historyBalanceLookup;
+    }
+    if (assigned.containsKey("--log-config")) {
+      PARAMETER.logbackPath = cmd.logbackPath;
+    }
+    if (!cmd.seedNodes.isEmpty()) {
```
waynercheung, line 1094: Inconsistent "explicit CLI override" detection for `seedNodes`: `applyCLIParams()` uses `isAssigned()` for most options, but seed nodes are applied whenever cmd.seedNodes is non-empty. With the current positional mapping, this might be fine, but it breaks the principle "only explicitly passed options override config". 
If you keep seed nodes positional, please document it clearly; otherwise, prefer making it a named option and gate it via isAssigned.

Beside, in the [#6567](https://github.com/tronprotocol/java-tron/issues/6567#issuecomment-4036611382) , @halibobo1205  has mentioned this related problem.

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
```diff
@@ -0,0 +1,156 @@
...
+  public boolean keystoreFactory;
+
+  @Parameter(names = {"--fast-forward"})
+  public boolean fastForward;
+
+  @Parameter(names = {"--es"}, description = "Start event subscribe server")
+  public boolean eventSubscribe;
+
+  @Parameter(names = {"--p2p-disable"}, description = "Switch for p2p module initialization. "
+      + "(defalut: false)", arity = 1)
```
waynercheung: "defalut" --> "default"

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
```diff
@@ -0,0 +1,156 @@
...
+  public String storageTransactionHistorySwitch;
+
+  @Parameter(names = {"--contract-parse-enable"}, description = "Switch for contract parses in "
+      + "java-tron. (default: true)")
+  public String contractParseEnable;
+
+  // ── Runtime parameters ──────────────────────────
+
+  @Parameter(names = {"--support-constant"}, description = "Support constant calling for TVM. "
+      + "(defalut: false)")
```
waynercheung: "defalut" --> "default"

framework/src/main/java/org/tron/core/config/args/ConfigKey.java
```diff
@@ -10,7 +10,7 @@ private ConfigKey() {
   }
 
   // local witness
-  public static final String LOCAL_WITNESS = "localwitness";
+  public static final String LOCAL_WITNESS = "localwitness"; //private key
```
waynercheung: "//private key" --> "// private key"

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1292,6 +986,164 @@ public static void setParam(final Config config) {
...
+        return;
+      }
+    }
+
+    // no private key source configured
+    throw new TronError("This is a witness node, but localWitnesses is null",
+        TronError.ErrCode.WITNESS_INIT);
+  }
+
+  public static void clearParam() {
```
waynercheung, line 1145: `CommonParameter.reset()` is marked `@VisibleForTesting` and documented as "Test-only", yet `Args.clearParam()` calls it unconditionally.

What about also adding `@VisibleForTesting` for `clearParam()`?

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1292,6 +986,164 @@ public static void setParam(final Config config) {
     logConfig();
   }
 
+  /**
+   * Apply CLI parameters that were explicitly passed.
+   * Only assigned parameters override Config values.
+   */
+  private static void applyCLIParams(CLIParameter cmd, JCommander jc) {
+    Map<String, ParameterDescription> assigned = jc.getParameters().stream()
```
waynercheung: `assigned` is collected into a `Map<String, ParameterDescription>` but only `containsKey()` is used. Consider collecting into a `Set<String>` instead to avoid unused values and reduce the chance of Collectors.toMap throwing if keys ever collide.

```
Set<String> assigned = jc.getParameters().stream()
    .filter(ParameterDescription::isAssigned)
    .map(ParameterDescription::getLongestName)
    .collect(Collectors.toSet());

// Then use contains() instead of containsKey()
if (assigned.contains("--verbose")) { ... }
```

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -23,119 +23,61 @@
...
-  public static CommonParameter PARAMETER = new CommonParameter();
+  protected static CommonParameter PARAMETER = new CommonParameter();
+
+  // Runtime chain state: set by VMConfig.initVmHardFork()
+  // when the energy-limit governance proposal is activated.
+  // Legacy: should belong to VMConfig, not here.
   @Setter
   public static boolean ENERGY_LIMIT_HARD_FORK = false;
+
+  // ── Startup parameters ────────────────────────
```
waynercheung: using "// -- Startup parameters --" instead of Unicode.

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -23,119 +23,61 @@
...
-  @Parameter(names = {"-d", "--output-directory"},
-      description = "Data directory for the databases (default:output-directory)")
   public String outputDirectory = "output-directory";
   @Getter
-  @Parameter(names = {"--log-config"}, description = "Logback config file")
   public String logbackPath = "";
-  @Getter
-  @Parameter(names = {"-h", "--help"}, help = true, description = "Show help message")
-  public boolean help = false;
+  // ── Flags (CLI + Config) ──────────────────────
```
waynercheung: --> // --Flags (CLI + Config) --

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -23,119 +23,61 @@
...
-  @Parameter(names = {"--storage-index-switch"}, description = "Storage index switch.(on or off)")
-  public String storageIndexSwitch = "";
-  @Parameter(names = {"--storage-transactionHistory-switch"},
-      description = "Storage transaction history switch.(on or off)")
-  public String storageTransactionHistorySwitch = "";
-  @Getter
-  @Parameter(names = {"--fast-forward"})
+  @Getter
   public boolean fastForward = false;
+  // ── Network / P2P ─────────────────────────────
```
waynercheung: using "// -- Network / P2P --"

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -197,26 +139,27 @@ public class CommonParameter {
...
 
-  //If you are running KeystoreFactory, this flag is set to true
+  // If you are running KeystoreFactory,
+  // this flag is set to true
   @Getter
   @Setter
-  @Parameter(names = {"--keystore-factory"}, description = "running KeystoreFactory")
   public boolean keystoreFactory = false;
 
+  // ── RPC / HTTP ────────────────────────────────
```
waynercheung: using "// -- RPC / HTTP --"

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -284,42 +222,44 @@ public class CommonParameter {
...
   @Getter
   @Setter
-  public long proposalExpireTime; // (ms)
+  public long proposalExpireTime;
   @Getter
   @Setter
-  public int checkFrozenTime; // for test only
+  public int checkFrozenTime; // clearParam: 1
+
+  // ── Committee parameters ──────────────────────
```
waynercheung: using "// -- Committee parameters --"

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -284,42 +222,44 @@ public class CommonParameter {
...
   @Getter
   @Setter
-  public long allowTvmSolidity059; //committee parameter
+  public long allowTvmSolidity059;
   @Getter
   @Setter
-  public long forbidTransferToContract; //committee parameter
+  public long forbidTransferToContract;
 
+  // ── Netty ─────────────────────────────────────
```
waynercheung: using "// -- Netty --"

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -383,42 +322,34 @@ public class CommonParameter {
...
-  @Parameter(names = {"--es"}, description = "Start event subscribe server")
   public boolean eventSubscribe = false;
   @Getter
   @Setter
-  public long trxExpirationTimeInMilliseconds; // (ms)
-  @Parameter(names = {"-v", "--version"}, description = "Output code version", help = true)
-  public boolean version;
+  public long trxExpirationTimeInMilliseconds;
+
+  // ── Shielded / ZK ─────────────────────────────
```
waynercheung: using "// -- Shielded / ZK --"

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * CLI parameter definitions parsed by JCommander.
+ * Fields here have NO default values — defaults live in CommonParameter.
+ * JCommander only populates fields that are explicitly passed on the
+ * command line.
+ */
+@NoArgsConstructor
+public class CLIParameter {
+
+  // ── Startup parameters ──────────────────────────
```
waynercheung: using "// -- Startup parameters --"

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Parameter(names = {"--storage-transactionHistory-switch"},
+      description = "Storage transaction history switch.(on or off)")
+  public String storageTransactionHistorySwitch;
+
+  @Parameter(names = {"--contract-parse-enable"}, description = "Switch for contract parses in "
+      + "java-tron. (default: true)")
+  public String contractParseEnable;
+
+  // ── Runtime parameters ──────────────────────────
```
waynercheung: using "// --  Runtime parameters --"

tranvandung89 (commented):
TAtCrLeZKwcUFbB1mEQwBaZ3kNAMKRrF1u　ok đấy

framework/src/main/java/org/tron/core/config/args/Args.java
waynercheung: Thanks for the context. Agree it may have existed before, but this refactor changes the precedence model to make CLI a first-class “highest priority” layer again (fixing #6567). Once CLI overrides are honored, lack of CLI-side validation becomes more user-visible and can lead to invalid runtime state (e.g. values below the safety floor) that config-path already protects against.

Even if CLI is planned to be deprecated, we still need it to be safe and consistent while it exists. The change here is low-cost: apply the same guardrail as config (max(3_000_000L, …)), so behavior is consistent across layers and we don’t regress safety/expectations.

If you prefer not to add validation now, could we at least (1) document the constraint in the CLI help, and (2) add a test to lock the desired behavior once CLI overrides are enabled? Otherwise we risk reintroducing "config is safe but CLI can bypass safety checks" inconsistencies.

vividctrlalt: Thanks for the detailed reasoning. I understand the concern about CLI bypassing config safety checks.

However, this behavior is **pre-existing** — the original code on `develop` also directly assigns CLI values without `max()` guardrails. This refactor's principle is to **restructure without expanding scope**; otherwise the refactor will never converge.

Also note that this PR targets the `develop` branch, not a release branch — there's no immediate production risk here.

That said, I agree it's a valid improvement. I'd prefer to address it as a **follow-up PR** focused specifically on CLI validation, which would be easier to review and test in isolation. Would that work for you?

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: This is **pre-existing behavior** — the original `loadSeeds()` on `develop` uses the exact same `seedNodes != null && !seedNodes.isEmpty()` check (line 1473 on develop). The `seedNodes` field is a JCommander positional parameter (`@Parameter(description=...)`) which doesn't support `isAssigned()` the same way named parameters do.

The refactor principle here is to **not expand scope** — if we start fixing every pre-existing inconsistency in this PR, the refactor will never converge. This PR restructures the init flow; it does not change individual parameter semantics.

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
vividctrlalt: Thanks for the thorough style review! Will fix all in the next commit:
- `defalut` → `default` (CLIParameter.java:60, :97)
- `//private key` → `// private key` (ConfigKey.java:13)
- Unicode section separators → ASCII `// -- Section --` format (CommonParameter.java & CLIParameter.java)
- `assigned` Map → Set (Args.java:994)
- Add `@VisibleForTesting` to `clearParam()` (Args.java:1133)

vividctrlalt:
### Fix: `getCommitIdAbbrev()` NPE — `catch (Exception)` is unacceptable

PR #6574 changed `catch (IOException)` to `catch (Exception)` to suppress an NPE here. **This is absolutely the wrong approach.** `catch (Exception)` is not a fix — it is sweeping bugs under the rug. It blindly swallows NPE, ClassCastException, SecurityException, and every other runtime error, making real problems invisible. This is a well-known anti-pattern flagged by SonarQube (S2221), Checkstyle (IllegalCatch), and every major Java coding standard. **It must not appear in production code.**

**Root cause:** `getResourceAsStream("git.properties")` returns `null` in CI (debian11/rockylinux) where gradle-git-properties skips generation due to incomplete `.git` metadata. `properties.load(null)` then throws NPE.

**Fix (2334eb3):** Explicit null check, consistent with `printVersion()` in the same file:
```java
if (in == null) {
    logger.warn("git.properties not found on classpath");
    return "";
}
```

**If you need to widen `catch (IOException)` to `catch (Exception)`, you are not fixing the problem — you are hiding it.**

kuny0707 (approved):

PR #6574, opened 2026-03-10
ci: add multi-platform build tests and system-test workflows

0xbigapple:
## Summary

This PR enhances the CI pipeline by introducing multi-platform build tests, Checkstyle and system tests workflows.

### 1. Multi-platform build tests
Add build tests across multiple platforms and architectures:

- **Ubuntu** — x86_64 + JDK 8 and aarch64 + JDK 17
- **macOS** — x86_64 + JDK 8 and aarch64 + JDK 17
- **Debian** — Debian 11 build test using the [`eclipse-temurin:8-jdk`](https://hub.docker.com/_/eclipse-temurin) base image.

### 2. Rocky Linux build test
Since **CentOS 7 and CentOS 8 images are not supported in GitHub-hosted runners**, as described in the  
[GitHub-hosted runners documentation](https://docs.github.com/en/actions/reference/runners/github-hosted-runners), a **Rocky Linux** build test has been added as a compatible replacement for **CentOS 8**.

### 3. Checkstyle workflow
Add a workflow to perform **Checkstyle** code style checks.

### 4. System test workflow
Add a workflow to execute a subset of daily build test cases from https://github.com/tronprotocol/system-test.

The workflow performs the following steps:

- Start a network with **one node including two SRs**
- Test core APIs
- Deploy smart contracts
- Trigger smart contract execution

### 5. Upgrade CodeQL from v3 to v4


.github/workflows/pr-check.yml
```diff
@@ -3,7 +3,7 @@ name: PR Check
 on:
   pull_request:
     branches: [ 'develop', 'release_**' ]
-    types: [ opened, edited, synchronize, reopened ]
+    types: [ opened, synchronize, reopened ]
```
Sunny6889: Removing `edited` from the trigger types means pr-lint won't re-run when the PR title or description is modified after opening. A contributor could pass the lint check initially, then edit the title to a non-conventional format before merging.

If the goal is to avoid unnecessary build/test runs on description edits, consider keeping `edited` in the workflow trigger and adding `if: github.event.action != 'edited' `to the build/test jobs so they skip edit-triggered runs while pr-lint still re-validates on every edit.

.github/workflows/pr-check.yml
```diff
@@ -197,13 +170,115 @@ jobs:
...
+            /github/home/.gradle/caches
+            /github/home/.gradle/wrapper
+          key: ${{ runner.os }}-rockylinux-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-rockylinux-gradle-
+
+      - name: Prepare checkstyle config copy
+        run: |
+          set -euxo pipefail
+          cp -f config/checkstyle/checkStyle.xml config/checkstyle/checkStyleAll.xml || true
```
Sunny6889: why here `|| true `？also the same question for other two places

.github/workflows/system-test.yml
```diff
@@ -0,0 +1,71 @@
+name: System Test
+
+on:
+  pull_request:
+    branches: [ 'develop', 'release_**' ]
```
halibobo1205, line 7: Is there a reason the master branch isn’t considered?

.github/workflows/pr-check.yml
liuyifei001: Good suggestion, I will optimize it.

.github/workflows/pr-check.yml
liuyifei001: Because this step begins with `set -euxo pipelinefail`, any command returning a non-zero status will immediately terminate the script. `true` ensures that the script will continue execution even if `checkStyle.xml` does not exist (`cp` fails). it can delete here.

`./gradlew --stop || true` It cannot be removed. If there is no running daemon, `--stop` will return a non-zero value, which needs to be considered to prevent step failure.

.github/workflows/system-test.yml
liuyifei001: Good suggestion, I will add it later.

.github/workflows/pr-check.yml
halibobo1205, line 125:
```
checkstyle {
    toolVersion = "${versions.checkstyle}"
    configFile = file("config/checkstyle/checkStyleAll.xml")
    // no failure condition configured
}
```
Since checkStyleAll.xml sets `<property name="severity" value="warning"/>`, all violations are reported as warnings. Gradle's Checkstyle plugin only fails the task on error-level violations, so the task always succeeds regardless of how many violations exist. The CI job is effectively a no-op — violations never block PR merges.
## Two Fix Options
### Option A: Change severity to error in checkStyleAll.xml
Violations are promoted to error level, causing the Gradle task to fail by default (ignoreFailures=false).
```xml
<property name="severity" value="error"/>
```
### Option B: Add maxWarnings = 0 in Gradle

```
checkstyle {
    toolVersion = "${versions.checkstyle}"
    configFile = file("config/checkstyle/checkStyleAll.xml")
    maxWarnings = 0
}
```

liuyifei001: If we fix it as suggested above, the behavior may differ from the previous implementation. But we can try.

framework/build.gradle
```diff
@@ -69,6 +69,7 @@ check.dependsOn 'lint'
 checkstyle {
     toolVersion = "${versions.checkstyle}"
     configFile = file("config/checkstyle/checkStyleAll.xml")
+    maxWarnings = 0
```
halibobo1205, line 72: `plugins` need this, too.

.github/workflows/pr-check.yml
```diff
@@ -163,20 +129,30 @@ jobs:
...
           - java: '17'
             runner: ubuntu-24.04-arm
+            os-name: ubuntu
+            arch: aarch64
+          - java: '8'
+            runner: macos-26-intel
+            os-name: macos
+            arch: x86_64
+          - java: '17'
+            runner: macos-latest
```
warku123: Just a minor thought: since `macos-latest` is a floating alias that could shift to a newer version over time, would it make sense to pin it to `macos-26` to stay consistent with macos-26-intel? 

It looks like `macos-26` is already generally available (actions/runner-images#13739). 
If there's a specific reason for using `macos-latest` here, I'd love to learn more about it!

liuyifei001: Is there available image for macos-26 aarch64 ? I will fix it as macos-26.

warku123: Yes, according to actions/runner-images#13739, macos-26 points to Apple Silicon (aarch64) by default.

Thanks for the quick fix!

liuyifei001 (approved):

halibobo1205 (approved):

.github/workflows/pr-check.yml
```diff
@@ -1,6 +1,8 @@
 name: PR Check
 
 on:
+  push:
```
waynercheung, line 4: This may be a **bug**： the **pr-lint will crash on push events**.

This workflow(line 15) triggers on both push (line 4-5) and pull_request (line 6-8). 

The script accesses `context.payload.pull_request.title`, but on push events `context.payload.pull_request` is **undefined**, which throws a TypeError and marks this job as failure. Since downstream jobs (`build, docker-build-*`) depend on `pr-lint` via `needs`, they will all be skipped on push events.

Suggest splitting into two workflows:
- pr-check.yml: triggered by `pull_request` only, includes `pr-lint + checkstyle + builds`.
- post-merge.yml: triggered by `push` only, includes `checkstyle + builds (no pr-lint)`.

This avoids the need for conditional `if / always()` workarounds on the needs chain.

Sunny6889 (approved):

.github/workflows/pr-check.yml
liuyifei001: Thanks for you issue. I fix it as this: skip pr-lint when push; When the PR lint job succeeds or is skipped, the jobs depending on it can still run.

.github/workflows/pr-check.yml
```diff
@@ -12,6 +14,7 @@ concurrency:
 jobs:
   pr-lint:
     name: PR Lint
+    if: github.event_name == 'pull_request'
```
waynercheung, line 17: The fix works: `adding if: github.event_name == 'pull_request'` to pr-lint and `!failure()` to downstream jobs correctly handles the push/PR scenarios.

One remaining issue: `!failure()` also returns true when jobs are cancelled. If a user manually cancels the workflow, build jobs will still start. Suggest adding `!cancelled()`:
` if: github.event.action != 'edited' && !failure() && !cancelled()`

liuyifei001: Users cannot manually cancel only the PR lint job. When the PR is updated, the running PR lint job will be canceled and all jobs will be triggered again.

.github/workflows/pr-check.yml
```diff
@@ -197,13 +174,109 @@ jobs:
...
           restore-keys: ${{ runner.os }}-${{ matrix.arch }}-gradle-
 
-      - name: Run tests
-        run: ./gradlew test
+      - name: Build
+        run: ./gradlew clean build --no-daemon
 
-      - name: Upload test reports
-        if: failure()
+  docker-build-rockylinux:
```
waynercheung, line 181: I think you can add `timeout-minutes: 60` as the `system-test` job.
Besides, `build` and `docker-build-debian11` can also add `timeout-minutes`.

.github/workflows/system-test.yml
```diff
@@ -0,0 +1,87 @@
...
+    timeout-minutes: 60
+
+    steps:
+      - name: Set up JDK 8
+        uses: actions/setup-java@v4
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+
+      - name: Clone system-test
```
waynercheung, line 27: Using raw git clone + git checkout has a few drawbacks:
1. The release_workflow branch is hardcoded — if it's deleted or renamed, CI breaks with no clear error.
2. No GitHub token auth — will fail if the repo ever becomes private.
3. No shallow clone optimization.

Suggest using actions/checkout instead:
```
  - name: Clone system-test
    uses: actions/checkout@v4
    with:
      repository: tronprotocol/system-test
      ref: release_workflow
      path: system-test
```

.github/workflows/pr-check.yml
waynercheung: Thanks for the clarification. You're right that the concurrency group handles the PR update case well — the entire old run is cancelled and a new one starts fresh.

The `!cancelled()` is more about the manual "Cancel workflow run" case from the Actions UI. In that scenario, `!failure()` alone would briefly allow pending jobs to start before GitHub cancels them again. **The practical impact is minimal** — at most a few seconds of wasted runner time.

Adding `!cancelled() `is a **best practice for correctness,** but I agree it's not a blocking issue. Feel free to address it or leave as-is.

.github/workflows/system-test.yml
```diff
@@ -0,0 +1,87 @@
...
+
+          echo "FullNode failed to start within $((MAX_ATTEMPTS * INTERVAL)) seconds."
+          echo "=== FullNode log (last 50 lines) ==="
+          tail -50 fullnode.log || true
+          exit 1
+
+      - name: Run system tests
+        working-directory: system-test
+        run: |
+          cp solcDIR/solc-linux-0.8.6 solcDIR/solc
```
waynercheung, line 82: If the `system-test` repo updates the `solc` version or changes the directory structure, this cp will fail and may produce confusing downstream test errors.

Suggest adding an existence check:
```
  if [ ! -f solcDIR/solc-linux-0.8.6 ]; then
    echo "ERROR: solc binary not found at solcDIR/solc-linux-0.8.6"
    exit 1
  fi
  cp solcDIR/solc-linux-0.8.6 solcDIR/solc
```

.github/workflows/pr-check.yml
waynercheung, line 118: Cache key glob patterns are inconsistent — `checkstyle` and `build`(line 174) use `**/*.gradle`, while `rockylinux`(line 213) and `debian11`(line 274) use `**/*.gradle*.` Both work the same for this project, but suggest standardizing to one pattern for maintainability.

.github/workflows/pr-check.yml
liuyifei001: Add it.

.github/workflows/system-test.yml
liuyifei001: Optimize it.

.github/workflows/system-test.yml
liuyifei001: Add it.

.github/workflows/pr-check.yml
```diff
@@ -93,53 +96,17 @@ jobs:
...
-          key: ${{ runner.os }}-${{ matrix.arch }}-gradle-${{ hashFiles('**/*.gradle', '**/gradle-wrapper.properties') }}
-          restore-keys: ${{ runner.os }}-${{ matrix.arch }}-gradle-
-
-      - name: Build
-        run: ./gradlew clean build -x test
-
   checkstyle:
     name: Checkstyle
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
waynercheung, line 101: Checkstyle is pure static analysis and architecture-independent.
Why do you use `ARM` instead of `ubuntu-latest (x86)`?

liuyifei001: Use a pinned version instead of a mutable version.

.github/workflows/pr-check.yml
liuyifei001: It doesn't matter much. If needed, I will change it next time.

.github/workflows/pr-check.yml
```diff
@@ -93,53 +96,17 @@ jobs:
...
-          path: |
-            ~/.gradle/caches
-            ~/.gradle/wrapper
-          key: ${{ runner.os }}-${{ matrix.arch }}-gradle-${{ hashFiles('**/*.gradle', '**/gradle-wrapper.properties') }}
-          restore-keys: ${{ runner.os }}-${{ matrix.arch }}-gradle-
-
-      - name: Build
-        run: ./gradlew clean build -x test
-
   checkstyle:
```
waynercheung, line 99: Nit (non-blocking): Currently checkstyle is only enabled for `framework` and `plugins`. It might be worth considering extending it to other modules (`actuator`, `chainbase`, `common`, etc.) in a follow-up PR to enforce consistent code style across the whole project.

liuyifei001: Good idea. The current purpose is to migrate it from buildkite to workflow. Might extends it in another PR.

.github/workflows/pr-check.yml
liuyifei001: Thanks for your feedback. There is no strong need to add `!cancelled()` now. If a similar scenario occurs during execution, we can add it later.

waynercheung (approved):
LGTM

kuny0707 (approved):

PR #6575, opened 2026-03-11
docs: fix shieldedTransaction typos in comments

operagxsasha:
Fix typos in comments: shiededTransaction - shieldedTransaction


vividctrlalt:
This PR cannot be merged — it accidentally **renames** `protocol/src/main/protos/api/api.proto` to `protocol/src/main/protos/api/end for shiededTransaction`, which would break the build.

This was likely caused by pasting text into the filename field in GitHub's web editor.

Please redo the PR with only the two comment fixes, without the file rename.

PR #6580, opened 2026-03-16
fix(config): deprecate CLI params and add guardrails

vividctrlalt:
## Summary
Follow-up to #6569 review feedback:

- **Deprecate 23 CLI parameters** that can be replaced by config-file options, marking them with `@Deprecated` in `CLIParameter` and logging runtime warnings via reflection when used
- **Add `max(3_000_000L, ...)` guardrail** for `--max-energy-limit-for-constant` CLI path, consistent with config-file validation (addresses @waynercheung's review in #6569)
- **Document `seedNodes` positional parameter** behavior with comment explaining why `isEmpty()` is used instead of `isAssigned()`, plus deprecation warning

### Retained CLI params (11)
`-c`, `-d`, `--log-config`, `-h`, `-v`, `-w`, `-p`, `--witness-address`, `--password`, `--keystore-factory`, `--debug`

### Deprecated CLI params (23)
All storage, VM, network/thread, and misc runtime parameters — these should be configured via config file.

## Test plan
- [x] `./gradlew compileJava` — all modules compile successfully
- [x] All CI checks passed on fork PR (vividctrlalt/java-tron#11)
- [ ] `./gradlew :framework:test --tests "org.tron.core.config.args.*"` — parameter tests pass
- [ ] `./gradlew :framework:test` — full framework tests pass

## Related
- Fixes review comments from #6569
- Addresses @waynercheung's feedback on CLI validation consistency
- Addresses @halibobo1205's note on seedNodes positional parameter (#6567)


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1091,7 +1107,12 @@ private static void applyCLIParams(CLIParameter cmd, JCommander jc) {
     if (assigned.contains("--log-config")) {
       PARAMETER.logbackPath = cmd.logbackPath;
     }
+    // seedNodes is a JCommander positional (main) parameter, which does not support
+    // isAssigned(). An empty-check is used instead — this is safe because the default
+    // is an empty list, so non-empty means the user explicitly passed values on CLI.
     if (!cmd.seedNodes.isEmpty()) {
+      logger.warn("Positional seed-node arguments are deprecated. "
```
3for, line 1159: Since `seedNodes` is deprecated, the corresponding help info should also be removed, such as: [helpStr.append(String.format("%nUsage: java -jar %s [options] [seedNode <seedNode> ...]%n", programName));](https://github.com/vividctrlalt/java-tron/blob/fix/pr6569-review-followup/framework/src/main/java/org/tron/core/config/args/Args.java#L1699)

vividctrlalt: Thanks for the feedback! Since `seedNodes` is deprecated but not yet removed, users may still want to use it, so I kept the Usage line but added a deprecation note right below it. Also noticed that `--es` and `--solidity` in the help output are deprecated too (marked in this PR), so I added `(deprecated)` hints to all deprecated options in the help text. See 84027da.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -998,6 +998,21 @@ private static void applyCLIParams(CLIParameter cmd, JCommander jc) {
...
+          try {
+            return CLIParameter.class.getDeclaredField(pd.getParameterized().getName())
+                .isAnnotationPresent(Deprecated.class);
+          } catch (NoSuchFieldException e) {
+            return false;
+          }
+        })
+        .forEach(pd -> logger.warn(
+            "CLI option '{}' is deprecated and will be removed in a future release. "
+                + "Please use the corresponding config-file option instead.",
```
warku123: Consider enriching the warning message with the specific config-file key mapping so users know exactly how to migrate.

And if we implement this, it's necessary to add a static map like this:
```Java
private static final Map<String, String> DEPRECATED_CLI_TO_CONFIG = Map.ofEntries(
    Map.entry("--storage-db-directory", "storage.db.directory"),
    Map.entry("--storage-db-engine", "storage.db.engine"),
    Map.entry("--storage-db-synchronous", "storage.db.synchronous"),
    // add remaining params...
);
```

And change this log like this:
```Java
   .forEach(pd -> {
      String cliOption = pd.getLongestName();
      String configKey = DEPRECATED_CLI_TO_CONFIG.get(cliOption);
      String suggestion = configKey != null 
          ? String.format(" Please use config key '%s' instead.", configKey)
          : " Please use the corresponding config-file option instead.";
      logger.warn("CLI option '{}' is deprecated and will be removed in a future release.{}",
          cliOption, suggestion);
    });
```

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -1008,7 +1023,8 @@ private static void applyCLIParams(CLIParameter cmd, JCommander jc) {
       PARAMETER.supportConstant = cmd.supportConstant;
     }
     if (assigned.contains("--max-energy-limit-for-constant")) {
-      PARAMETER.maxEnergyLimitForConstant = cmd.maxEnergyLimitForConstant;
+      PARAMETER.maxEnergyLimitForConstant = max(3_000_000L,
```
warku123: I noticed that `3_000_000L` is already defined as `ENERGY_LIMIT_IN_CONSTANT_TX` in Constant.java:
```java
// In common/src/main/java/org/tron/core/Constant.java
public static final long ENERGY_LIMIT_IN_CONSTANT_TX = 3_000_000L;
```

Would it make sense to reuse this constant here instead of the inline literal?

framework/src/main/java/org/tron/core/config/args/Args.java
3for: Some deprecated CLI parameters (e.g., `--solidity`, `--p2p-disable`, `--es`, `--fast-forward`) have no equivalent "config-file option" in `.conf`, making the message “Please use the corresponding config-file option instead” misleading.

vividctrlalt: Good suggestion! Added a `DEPRECATED_CLI_TO_CONFIG` map that provides specific config-key hints in deprecation warnings.

For params with a config equivalent:
`CLI option '--storage-db-engine' is deprecated and will be removed in a future release. Please use config key 'storage.db.engine' instead.`

For params without a config equivalent (`--solidity`, `--p2p-disable`):
`CLI option '--solidity' is deprecated and will be removed in a future release.`

This also addresses @3for's concern about misleading messages. See 5b0e37be7.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Good catch! Replaced both occurrences of `3_000_000L` with `ENERGY_LIMIT_IN_CONSTANT_TX` from `Constant.java`. See 5b0e37be7.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Valid point. Params like `--solidity` and `--p2p-disable` have no config-file equivalent — they are runtime mode switches being phased out entirely. The warning now omits the "use config-file option" suggestion for these params and simply states they will be removed. See 5b0e37be7.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -83,6 +84,41 @@
...
 
+  /**
+   * Maps deprecated CLI option names to their config-file equivalents.
+   * Options not in this map have no config equivalent and are being removed entirely.
+   */
+  private static final Map<String, String> DEPRECATED_CLI_TO_CONFIG;
+
+  static {
+    Map<String, String> m = new HashMap<>();
+    m.put("--storage-db-directory", "storage.db.directory");
```
liuyifei001, line 95: This variable `private static final String DB_DIRECTORY_CONFIG_KEY = "storage.db.directory";` has already been defined and does not need to be redefined; otherwise, it will be difficult to maintain. It should be set to public.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -83,6 +84,41 @@
...
+    m.put("--save-featured-internaltx", "vm.saveFeaturedInternalTx");
+    m.put("--save-cancel-all-unfreeze-v2-details", "vm.saveCancelAllUnfreezeV2Details");
+    m.put("--long-running-time", "vm.longRunningTime");
+    m.put("--max-connect-number", "node.maxHttpConnectNumber");
+    m.put("--rpc-thread", "node.rpc.thread");
+    m.put("--solidity-thread", "node.solidity.threads");
+    m.put("--validate-sign-thread", "node.validateSignThreadNum");
+    m.put("--trust-node", "node.trustNode");
+    m.put("--history-balance-lookup", "storage.balance.history.lookup");
+    m.put("--es", "event.subscribe");
```
3for: `--es` and `--fast-forward` are both boolean flags with **false** default value. They are used to control branching logic in business flows within [Manager](https://github.com/vividctrlalt/java-tron/blob/fix/pr6569-review-followup/framework/src/main/java/org/tron/core/db/Manager.java#L564)、[AdvService](https://github.com/vividctrlalt/java-tron/blob/fix/pr6569-review-followup/framework/src/main/java/org/tron/core/net/service/adv/AdvService.java#L117)、[BlockMsgHandler](https://github.com/vividctrlalt/java-tron/blob/fix/pr6569-review-followup/framework/src/main/java/org/tron/core/net/messagehandler/BlockMsgHandler.java#L76).

In the current implementation, their mapped configurations (`event.subscribe` and `node.fastForward`) cannot substitute for the original behavior controlled by these flags. So, the current configuration mapping is inconsistent with the actual code logic.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Thanks for the review! These string literals in the map are user-facing hint text shown in deprecation warnings, not config-parsing keys used in code logic. While `Storage.java` does define similar constants (e.g. `DB_DIRECTORY_CONFIG_KEY`), they are intentionally `private` — and I'd prefer not to change their visibility just to serve a deprecated-params map that will eventually be removed along with the CLI params themselves.

Additionally, the config key constants should be further modularized by domain in the future (storage, vm, node, etc.), so coupling them across classes now would work against that direction.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: You're right. `--es` now maps to a proper `event.subscribe.enable` config key.

For `--fast-forward`, this is a specialized node runtime mode (not a general-purpose option), so it should be deprecated without providing a config equivalent. Ideally it should be built as a separate dedicated jar. Removed it from the mapping. See 9d6027aa5.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -661,6 +696,9 @@ public static void applyConfigParams(
     PARAMETER.eventFilter =
         config.hasPath(ConfigKey.EVENT_SUBSCRIBE_FILTER) ? getEventFilter(config) : null;
 
+    PARAMETER.eventSubscribe = config.hasPath(ConfigKey.EVENT_SUBSCRIBE_ENABLE)
+        && config.getBoolean(ConfigKey.EVENT_SUBSCRIBE_ENABLE);
```
3for, line 700: It would be helpful to add test cases for the newly introduced `event.subscribe.enable` configuration, specifically to verify that it is properly overridden by the CLI flag `--es`, similar to [`testCliOverridesStorageConfig`](https://github.com/vividctrlalt/java-tron/blob/fix/pr6569-review-followup/framework/src/test/java/org/tron/core/config/args/ArgsTest.java#L298).

vividctrlalt: Added two test cases in `ArgsTest`:
- `testEventSubscribeFromConfig` — verifies `event.subscribe.enable = true` is read from config
- `testCliEsOverridesConfig` — verifies CLI `--es` takes priority over config value

See 5bfa482ec.

3for (approved):
LGTM now.

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
```diff
@@ -44,63 +47,78 @@ public class CLIParameter {
   @Parameter(names = {"--password"}, description = "password")
   public String password;
 
+  @Deprecated
   @Parameter(names = {"--solidity"}, description = "running a solidity node for java tron")
   public boolean solidityNode;
```
liuyifei001, line 51: SolidityNode.jar is no longer generated during the build process, but the `--solidity` option is still in use, particularly in synchronization scenarios. Since this command-line parameter has no equivalent configuration item in the config file, it should not be marked as deprecated.

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
```diff
@@ -44,63 +47,78 @@ public class CLIParameter {
...
   public boolean fastForward;
 
+  @Deprecated
   @Parameter(names = {"--es"}, description = "Start event subscribe server")
   public boolean eventSubscribe;
 
+  @Deprecated
   @Parameter(names = {"--p2p-disable"}, description = "Switch for p2p module initialization. "
       + "(default: false)", arity = 1)
   public boolean p2pDisable;
```
liuyifei001, line 66: The `--p2p-disable` command-line option also has no equivalent configuration item in the config file and remains useful in certain scenarios. For example, when used together with shutdown, it allows querying the state information of a specific block. Therefore, it should not be marked as deprecated.

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
vividctrlalt: Agreed. `--solidity` is a runtime mode switch that determines an entirely different execution path (`SolidityNode.start()` vs `FullNode`), not a configuration parameter. It has no config-file equivalent and shouldn't have one — it's choosing *which program to run*, not *how to configure it*. Will remove the deprecated annotation.

framework/src/main/java/org/tron/core/config/args/CLIParameter.java
vividctrlalt: Agreed. `--p2p-disable` is also a runtime mode switch — it starts the node in offline/local-query mode without P2P networking. This is a distinct execution mode, not a tunable config parameter. Will remove the deprecated annotation as well.

liuyifei001 (approved):

kuny0707 (approved):

PR #6581, opened 2026-03-16
test: add dual DB engine test coverage with RocksDB support

vividctrlalt:
## Summary

java-tron supports both LevelDB and RocksDB, but the test suite only ran with the default engine (LevelDB). This PR adds dual DB engine test coverage so the same business tests can run under both engines, and fixes several related issues.

### What changed

- **Centralize test config**: Introduce `TestEnv` class with config file constants (`TEST_CONF`, `NET_CONF`, `MAINNET_CONF`, etc.) and utility methods, replacing 200+ hardcoded config strings across test files
- **Dual DB engine override**: Add `withDbEngineOverride()` to `BaseTest` and `TestEnv`, allowing Gradle to inject `--storage-db-engine ROCKSDB` via system property `tron.test.db.engine`
- **Parameterized DB tests**: Create `DbDataSourceImplTest` with `@RunWith(Parameterized.class)` to run common DB operations (put/get, batch, reset, iterator, etc.) against both LevelDB and RocksDB, eliminating ~400 lines of duplicate test code from `LevelDbDataSourceImplTest` and `RocksDbDataSourceImplTest`
- **`BaseMethodTest`**: New base class for tests requiring per-method Spring context isolation (fresh context per `@Test` method)
- **ARM64 support**: Auto-skip LevelDB tests on ARM64 via `assumeLevelDbAvailable()`; Gradle ARM64 builds auto-inject RocksDB engine
- **Gradle `testWithRocksDb` task**: New CI task that reruns the full test suite with RocksDB engine
- **CI workflow**: Add `testWithRocksDb` step to `pr-check.yml` for x86_64 builds
- **ARM64 engine override test**: `Arm64EngineOverrideTest` verifies ARM64 silently forces RocksDB even when config says LevelDB
- **Flaky test fixes**: `TrieTest` (missing test body), `ConditionallyStopTest` (Spring context leak), `NeedBeanCondition` (null-safety), `NativeMessageQueue` (missing shutdown hook)

### Why

1. RocksDB is the production engine on ARM64 and increasingly on x86 — tests should cover it
2. Duplicate LevelDB/RocksDB test files were a maintenance burden (~900 lines of near-identical code)
3. Hardcoded config strings across 200+ files made refactoring error-prone

## Test plan

- [x] `./gradlew :framework:test` — all existing tests pass
- [x] `./gradlew :framework:testWithRocksDb` — business tests pass with RocksDB engine
- [x] `./gradlew :framework:test --tests "org.tron.common.storage.DbDataSourceImplTest"` — parameterized tests run both engines
- [x] `./gradlew :framework:test --tests "org.tron.common.storage.Arm64EngineOverrideTest"` — ARM64 mock tests pass
- [x] Checkstyle clean
- [x] CI all green on 6 platform/JDK combinations (debian11, rockylinux, ubuntu x86, ubuntu aarch64, macOS aarch64, macOS x86)


framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -134,6 +134,18 @@ public static void setParam(final String[] args, final String confFileName) {
     initLocalWitnesses(config, cmd);
   }
 
+  /**
+   * Validate final configuration after all sources (defaults, config, CLI) are applied.
+   */
+  public static void validateConfig() {
```
3for, line 140: `Args.validateConfig()` is executed before `KeystoreFactory.start()`. With the default `config.conf` setting `db.engine = "LEVELDB"`, running in `--keystore-factory` mode (e.g., when only generating or importing a keystore on ARM64) causes the program to exit early due to a database engine validation failure. However, this execution path does not require the database to be opened at all.
It would be preferable to defer this validation until the code path that actually initializes the database, or to skip the validation entirely for no-DB modes such as `--keystore-factory`.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -134,6 +134,18 @@ public static void setParam(final String[] args, final String confFileName) {
...
+    if (Arch.isArm64() && !"ROCKSDB".equals(PARAMETER.storage.getDbEngine())) {
```
3for, line 141: Change `"ROCKSDB".equals(PARAMETER.storage.getDbEngine())` to `"ROCKSDB".equals(PARAMETER.storage.getDbEngine()).toUpperCase()`, so that values like `rocksdb` or `RocksDB` are also supported.

3for:
The statement “`./gradlew :framework:testWithRocksDb — business tests pass with RocksDB engine`” is not accurate.

Although `testWithRocksDb` initially switches the test environment to `ROCKSDB`, the global engine configuration is later reset to `LEVELDB` during test execution. As a result, the tests assumed to be executed in RocksDB engine are **not running under a pure RocksDB environment**. Instead, they operate in a mixed state where the context is initialized with RocksDB, but the global engine configuration is reverted to LevelDB. This weakens the intended RocksDB coverage and may potentially mask real issues.

To illustrate this, I added a log in the `fork()` test of `ManagerTest.java` right after `Args.setParam(new String[]{}, TestEnv.TEST_CONF);`:

```java
// added variable
String engineBeforeReset = Args.getInstance().getStorage().getDbEngine();
Args.setParam(new String[]{}, TestEnv.TEST_CONF);
// added log
logger.info("[testWithRocksDb-demo] requested={}, beforeReset={}, afterReset={}",
    System.getProperty("tron.test.db.engine"),
    engineBeforeReset,
    Args.getInstance().getStorage().getDbEngine());
```

Running:

```
./gradlew :framework:testWithRocksDb --tests org.tron.core.db.ManagerTest.fork --info
```

produces:

```
18:17:57.889 INFO [o.t.c.d.ManagerTest] 
[testWithRocksDb-demo] requested=ROCKSDB, beforeReset=ROCKSDB, afterReset=LEVELDB
```

This confirms that although RocksDB is requested and initially active, the global engine configuration is subsequently reverted to LevelDB during the test execution.

PR #6586, opened 2026-03-18
test(framework): add dual DB engine coverage with flaky test fixes

### LINKED ISSUE #6587: [Discussion] Should aarch64 throw an exception when db.engine=LEVELDB?
## Background

`Storage.java:177` currently handles aarch64 by silently falling back to RocksDB:

```java
if (Arch.isArm64()) {
  logger.warn("Arm64 architecture detected, using RocksDB as db engine, ignore config.");
  return ROCKS_DB_ENGINE;
}
```

`config.conf` file defaults to `db.engine = "LEVELDB"` — shared between x86 and aarch64

And the default config for x86 is `DEFAULT_DB_ENGINE = "LEVELDB"` 

Aarch64 does not support LevelDB. As a result, aarch64 nodes running the default config always hit this fallback, rolling back to RocksDB, and print only one warning log.

## Problem

`checkOrInitEngine` provides a helpful error on x86: when the configured engine mismatches the existing data directory, the error message points directly to the cause.

This breaks down on aarch64. Consider a node migrated from x86 (with a LevelDB data directory) to arm64: both the config and the on-disk data are LevelDB, so they appear consistent to `checkOrInitEngine`. The silent engine switch to RocksDB then causes it to throw:

`Cannot open LEVELDB database with ROCKSDB engine`

Operators see a database error with no indication that the real cause is an architecture/engine incompatibility. The error chain is broken precisely because the fallback happens before `checkOrInitEngine` can reason about it correctly.

## Two constraints prevent simply changing the defaults:

* `config.conf` defaults to `db.engine = "LEVELDB"` — shared between x86 and aarch64, so this cannot be changed  
* `DEFAULT_DB_ENGINE = "LEVELDB"` — changing this would affect x86 users

Therefore, it's not possible to simply throw an exception when using aarch64 \+ LEVELDB; otherwise, all arm64 users using the default configuration will fail to start.

## Open for discussion

* **Targeted exception on data directory mismatch** — throw a clear "architecture does not support the configured engine" exception only when an existing LevelDB data directory is detected (hooking into or extending `checkOrInitEngine` logic), while continuing to fall back silently for fresh nodes with no existing data.  
* **Other approaches** — are there alternative ways to surface a clear error to operators without changing the default config or breaking the happy path?  
* **Status quo** — is the current behavior acceptable given the constraints?

vividctrlalt:
### New test infrastructure
- Extract shared DB test logic into `DbDataSourceImplTest`, `BaseMethodTest`, `TestConstants`
- Extend `BaseTest` to support DB engine override via Typesafe Config system property
- Add `testWithRocksDb` Gradle task; CI runs it on x86_64 after default build

### Flaky test fixes (20+ classes)
- VM: FreezeTest, FreezeV2Test, VoteTest, InternalTransactionCallTest
- DB: ChainbaseTest, CheckpointV2Test, SnapshotImplTest, SnapshotManagerTest, SnapshotRootTest, RevokingDbWithCacheNewValueTest
- Core: ManagerTest, TransactionExpireTest, ForkControllerTest, ShieldedReceiveTest
- Services: MetricsApiServiceTest, PeerStatusCheckTest, SyncServiceTest, ComputeRewardTest, ConditionallyStopTest
- Config: ArgsTest, DynamicArgsTest
- New: TrieTest

### Production code (minimal)
- Add `Args.validateConfig()` for ARM64 + LevelDB rejection at startup
- Call `validateConfig()` in `FullNode.main()` (production entry only, tests unaffected)
- Remove unused `ROCKS_DB_ENGINE` constant in `Storage.java`

## Test plan
- [ ] All existing tests pass with default (LevelDB) engine
- [ ] `./gradlew :framework:testWithRocksDb` passes on x86_64
- [ ] ARM64 builds exclude LevelDB tests and use RocksDB
- [ ] CI pr-check workflow runs dual-engine tests on x86_64

### DIFF
--- a/common/src/main/java/org/tron/core/config/args/Storage.java
+++ b/common/src/main/java/org/tron/core/config/args/Storage.java
@@ -29,7 +29,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.iq80.leveldb.CompressionType;
 import org.iq80.leveldb.Options;
-import org.tron.common.arch.Arch;
 import org.tron.common.cache.CacheStrategies;
 import org.tron.common.cache.CacheType;
 import org.tron.common.utils.DbOptionalsUtils;
@@ -90,7 +89,6 @@ public class Storage {
    * Default values of directory
    */
   private static final String DEFAULT_DB_ENGINE = "LEVELDB";
-  private static final String ROCKS_DB_ENGINE = "ROCKSDB";
   private static final boolean DEFAULT_DB_SYNC = false;
   private static final boolean DEFAULT_EVENT_SUBSCRIBE_CONTRACT_PARSE = true;
   private static final String DEFAULT_DB_DIRECTORY = "database";
@@ -175,10 +173,6 @@ public class Storage {
   private final Map<String, Sha256Hash> dbRoots = Maps.newConcurrentMap();
 
   public static String getDbEngineFromConfig(final Config config) {
-    if (Arch.isArm64()) {
-      logger.warn("Arm64 architecture detected, using RocksDB as db engine, ignore config.");
-      return ROCKS_DB_ENGINE;
-    }
     return config.hasPath(DB_ENGINE_CONFIG_KEY)
         ? config.getString(DB_ENGINE_CONFIG_KEY) : DEFAULT_DB_ENGINE;
   }
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -130,15 +130,19 @@ public class Args extends CommonParameter {
     // 3. CLI overrides Config (highest priority)
     applyCLIParams(cmd, jc);
 
-    // 4. Init witness (depends on CLI witness flag)
+    // 4. Apply platform constraints (e.g. ARM64 forces RocksDB)
+    applyPlatformConstraints();
+
+    // 5. Init witness (depends on CLI witness flag)
     initLocalWitnesses(config, cmd);
   }
 
   /**
-   * Validate final configuration after all sources (defaults, config, CLI) are applied.
-   * ARM64 only supports RocksDB; db.engine config is ignored with a warning.
+   * Apply platform-specific constraints after all config sources are resolved.
+   * ARM64 does not support LevelDB (native JNI library unavailable),
+   * so db.engine is forced to RocksDB regardless of config or CLI settings.
    */
-  public static void validateConfig() {
+  private static void applyPlatformConstraints() {
     if (Arch.isArm64()
         && !Constant.ROCKSDB.equalsIgnoreCase(PARAMETER.storage.getDbEngine())) {
       logger.warn("ARM64 only supports RocksDB, ignoring db.engine='{}'",
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -801,11 +801,8 @@ public class Args extends CommonParameter {
             .getInt(ConfigKey.COMMITTEE_ALLOW_OPTIMIZED_RETURN_VALUE_OF_CHAIN_ID) : 0;
 
     initBackupProperty(config);
-    if (Constant.ROCKSDB.equalsIgnoreCase(CommonParameter
-        .getInstance().getStorage().getDbEngine())) {
-      initRocksDbBackupProperty(config);
-      initRocksDbSettings(config);
-    }
+    initRocksDbBackupProperty(config);
+    initRocksDbSettings(config);
 
     PARAMETER.actuatorSet =
         config.hasPath(ConfigKey.ACTUATOR_WHITELIST)
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -135,6 +135,18 @@ public class Args extends CommonParameter {
   }
 
   /**
+   * Validate final configuration after all sources (defaults, config, CLI) are applied.
+   */
+  public static void validateConfig() {
+    if (Arch.isArm64() && !"ROCKSDB".equals(PARAMETER.storage.getDbEngine())) {
+      throw new TronError(
+          "ARM64 architecture only supports RocksDB. Current engine: "
+              + PARAMETER.storage.getDbEngine(),
+          TronError.ErrCode.PARAMETER_INIT);
+    }
+  }
+
+  /**
    * Apply parameters from config file.
    */
   public static void applyConfigParams(
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -136,13 +136,14 @@ public class Args extends CommonParameter {
 
   /**
    * Validate final configuration after all sources (defaults, config, CLI) are applied.
+   * ARM64 only supports RocksDB; db.engine config is ignored with a warning.
    */
   public static void validateConfig() {
-    if (Arch.isArm64() && !"ROCKSDB".equals(PARAMETER.storage.getDbEngine())) {
-      throw new TronError(
-          "ARM64 architecture only supports RocksDB. Current engine: "
-              + PARAMETER.storage.getDbEngine(),
-          TronError.ErrCode.PARAMETER_INIT);
+    if (Arch.isArm64()
+        && !Constant.ROCKSDB.equalsIgnoreCase(PARAMETER.storage.getDbEngine())) {
+      logger.warn("ARM64 only supports RocksDB, ignoring db.engine='{}'",
+          PARAMETER.storage.getDbEngine());
+      PARAMETER.storage.setDbEngine(Constant.ROCKSDB);
     }
   }
 
--- a/framework/src/main/java/org/tron/program/FullNode.java
+++ b/framework/src/main/java/org/tron/program/FullNode.java
@@ -24,7 +24,6 @@ public class FullNode {
     ExitManager.initExceptionHandler();
     checkJdkVersion();
     Args.setParam(args, "config.conf");
-    Args.validateConfig();
     CommonParameter parameter = Args.getInstance();
 
     LogService.load(parameter.getLogbackPath());
--- a/framework/src/main/java/org/tron/program/FullNode.java
+++ b/framework/src/main/java/org/tron/program/FullNode.java
@@ -24,6 +24,7 @@ public class FullNode {
     ExitManager.initExceptionHandler();
     checkJdkVersion();
     Args.setParam(args, "config.conf");
+    Args.validateConfig();
     CommonParameter parameter = Args.getInstance();
 
     LogService.load(parameter.getLogbackPath());
--- a/framework/src/test/java/org/tron/common/BaseMethodTest.java
+++ b/framework/src/test/java/org/tron/common/BaseMethodTest.java
@@ -1,0 +1,98 @@
+package org.tron.common;
+
+import java.io.IOException;
+import java.util.Arrays;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
+import org.tron.common.application.Application;
+import org.tron.common.application.ApplicationFactory;
+import org.tron.common.application.TronApplicationContext;
+import org.tron.core.ChainBaseManager;
+import org.tron.core.config.DefaultConfig;
+import org.tron.core.config.args.Args;
+import org.tron.core.db.Manager;
+
+/**
+ * Base class for tests that need a fresh Spring context per test method.
+ *
+ * Each @Test method gets its own TronApplicationContext, created in @Before
+ * and destroyed in @After, ensuring full isolation between tests.
+ *
+ * Subclasses can customize behavior by overriding hook methods:
+ *   extraArgs()      — additional CLI args (e.g. "--debug")
+ *   configFile()     — config file (default: config-test.conf)
+ *   beforeContext()  — runs after Args.setParam, before Spring context creation
+ *   afterInit()      — runs after Spring context is ready (e.g. get extra beans)
+ *   beforeDestroy()  — runs before context shutdown (e.g. close resources)
+ *
+ * Use this when:
+ *   - methods modify database state that would affect other methods
+ *   - methods need different CommonParameter settings (e.g. setP2pDisable)
+ *   - methods need beforeContext() to configure state before Spring starts
+ *
+ * If methods are read-only and don't interfere with each other, use BaseTest instead.
+ * Tests that don't need Spring (e.g. pure unit tests) should NOT extend either base class.
+ */
+@Slf4j
+public abstract class BaseMethodTest {
+
+  @Rule
+  public final TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  protected TronApplicationContext context;
+  protected Application appT;
+  protected Manager dbManager;
+  protected ChainBaseManager chainBaseManager;
+
+  protected String[] extraArgs() {
+    return new String[0];
+  }
+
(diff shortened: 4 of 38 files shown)


common/src/main/java/org/tron/core/config/args/Storage.java
```diff
@@ -175,10 +173,6 @@ public class Storage {
   private final Map<String, Sha256Hash> dbRoots = Maps.newConcurrentMap();
 
   public static String getDbEngineFromConfig(final Config config) {
-    if (Arch.isArm64()) {
```
3for, line 178: This change and `validateConfig` introduce a regression: on ARM64, starting the node with the default `config.conf` (where `db.engine` is set to `LEVELDB`) now results in a startup failure.

The issue can be reproduced as follows:

```bash
$ ./gradlew :framework:buildFullNodeJar
…..
Building for architecture: aarch64, Java version: 17
…..

$ java -jar framework/build/libs/FullNode.jar
$ tail -f logs/tron.log
…..
15:17:21.876 ERROR [main] [Exit](ExitManager.java:49) Shutting down with code: PARAMETER_INIT(1), reason: ARM64 architecture only supports RocksDB. Current engine: LEVELDB
```

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -134,6 +134,18 @@ public static void setParam(final String[] args, final String confFileName) {
     initLocalWitnesses(config, cmd);
   }
 
+  /**
+   * Validate final configuration after all sources (defaults, config, CLI) are applied.
+   */
+  public static void validateConfig() {
+    if (Arch.isArm64() && !"ROCKSDB".equals(PARAMETER.storage.getDbEngine())) {
```
3for: Change `"ROCKSDB".equals(PARAMETER.storage.getDbEngine())` to `"ROCKSDB".equals(PARAMETER.storage.getDbEngine()).toUpperCase()` or `Constant.ROCKSDB.equalsIgnoreCase(PARAMETER.storage.getDbEngine())`, so that values like `rocksdb` or `RocksDB` are also supported, just like [https://github.com/vividctrlalt/java-tron/blob/test/dual-db-engine-coverage/framework/src/main/java/org/tron/core/config/args/Args.java#L799](https://github.com/vividctrlalt/java-tron/blob/test/dual-db-engine-coverage/framework/src/main/java/org/tron/core/config/args/Args.java#L799).

framework/src/test/java/org/tron/core/tire/TrieTest.java
```diff
@@ -119,6 +120,30 @@ public void test2() {
...
+   *    Merkle Trie invariant.
+   *
+   * Production impact: low. AccountStateCallBack uses TrieImpl to build per-block
+   * account state tries. Duplicate put(key, sameValue) is unlikely in production
+   * because account state changes between transactions (balance, nonce, etc.).
+   *
+   * Proper fix: TrieImpl.insert() should short-circuit when the existing value
+   * equals the new value, avoiding unnecessary invalidate(). See TrieImpl:188-192.
+   */
+  @Ignore("TrieImpl bug: root hash depends on insertion order with duplicate key-value puts")
```
3for, line 146: Why was `@Ignore` used instead of addressing the issue in `TrieImpl.insert()`?

common/src/main/java/org/tron/core/config/args/Storage.java
halibobo1205: Agreed. I've opened an issue to continue the discussion: #6587

3for:
Running `./gradlew :framework:testWithRocksDb` on ARM64 causes `testLevelDb` to fail. It would be better to add `assumeLevelDbAvailable();` at the beginning of `testLevelDb` to skip this test when LevelDB is not available on the platform.

```
org.tron.core.db.DBIteratorTest > testLevelDb FAILED
    java.lang.UnsatisfiedLinkError: Could not load library. Reasons: [java.lang.LinkageError: Unable to load library leveldbjni64-1.8, java.lang.LinkageError: Unable to load library leveldbjni-1.8, java.lang.LinkageError: Unable to load library leveldbjni, java.lang.LinkageError: Unable to load library from /var/folders/p5/bmmxtx5x14j_6znwywzj09w40000gq/T/libleveldbjni-1.8-7667426779319649693.jnilib, java.lang.LinkageError: Unable to load library from /Users/***/.hawtjni/leveldbjni/libleveldbjni-1.8-2478008178187835858.jnilib]
        at org.fusesource.hawtjni.runtime.Library.doLoad(Library.java:204)
        at org.fusesource.hawtjni.runtime.Library.load(Library.java:156)
        at org.fusesource.leveldbjni.JniDBFactory.<clinit>(JniDBFactory.java:48)
        at org.tron.core.db.DBIteratorTest.testLevelDb(DBIteratorTest.java:35)

2315 tests completed, 6 failed, 28 skipped

> Task :framework:testWithRocksDb FAILED

```

vividctrlalt:
Thanks @3for for catching this! Fixed in d57d110c7 — added `assumeLevelDbAvailable()` to `DBIteratorTest.testLevelDb` so it skips on ARM64 where LevelDB JNI is unavailable.

framework/src/test/java/org/tron/core/tire/TrieTest.java
vividctrlalt: Good question! This is not a test bug — it's a bug in the production code `TrieImpl.insert()`. When inserting a duplicate key with the same value, the root hash changes depending on insertion order, which is incorrect.

The comment on `@Ignore` explains the root cause: `TrieImpl.insert()` should short-circuit when the existing value equals the new value, avoiding an unnecessary `invalidate()` call (see `TrieImpl:188-192`).

Since the scope of this PR is test refactoring (dual DB engine coverage), fixing production code in `TrieImpl` is out of scope here. The `@Ignore` annotation documents the known issue so it can be tracked and fixed separately. Also worth noting that this trie code is not yet enabled in production, so the impact is limited for now.

common/src/main/java/org/tron/core/config/args/Storage.java
vividctrlalt: I believe fail-fast is the correct behavior here. If a user runs on ARM64 with `db.engine=LEVELDB`, the node **should** refuse to start with a clear error message, rather than silently falling back to RocksDB or deferring the failure to runtime.

The principle is: **no hidden behavior**. If you're running on ARM64, you must explicitly configure a supported engine. A silent fallback would mask a misconfiguration and could lead to confusion later (e.g., user expects LevelDB but data is actually stored in RocksDB).

That said, I understand the concern about the default `config.conf` shipping with `LEVELDB`. This is really a config-level issue — the default config should either be platform-aware or the documentation should clearly state that ARM64 users need to set `db.engine=ROCKSDB`. Happy to discuss further in #6587.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Good catch! Will fix to use `Constant.ROCKSDB.equalsIgnoreCase()` for consistency.

common/src/main/java/org/tron/core/config/args/Storage.java
vividctrlalt: To add more context: the whole point of this series of changes is to **eliminate silent behaviors in FullNode**. Silently swapping the DB engine in production is exactly the kind of hidden behavior we're trying to remove — it should be strictly prohibited.

If an ARM64 node has `db.engine=LEVELDB` in `config.conf`, that's a **configuration management problem**, not something the application should paper over at runtime. Mixing config-management concerns into production code leads to harder-to-debug issues and undermines the principle of explicit configuration.

The correct fix belongs in the deployment/config layer — ARM64 environments should ship with the right `config.conf` from the start.

framework/src/main/java/org/tron/program/FullNode.java
```diff
@@ -24,6 +24,7 @@ public static void main(String[] args) {
     ExitManager.initExceptionHandler();
     checkJdkVersion();
     Args.setParam(args, "config.conf");
+    Args.validateConfig();
```
3for: 1. **ARM64 normalization happens too late**
   On ARM64, `dbEngine = LEVELDB` is corrected to `ROCKSDB` in `validateConfig()` *after* `setParam()`.
   However, `applyConfigParams` already runs before that and checks the engine for RocksDB initialization. Since the engine is still `LEVELDB` at that point, RocksDB backup/init logic is skipped.

2. **Should run after CLI override**
   `applyConfigParams` runs before `applyCLIParams`, so `.conf` can be overridden by `--storage-db-engine`.
   The ARM64 normalization should be applied **after both config and CLI are resolved**.

3. **Method name is misleading**
   `validateConfig` mutates config instead of just validating.
   Consider renaming it to `normalizeConfig` or `finalizeConfig`.

```
    .....
    Args.setParam(args, "config.conf");
    Args.validateConfig(); // normalizes dbEngine to ROCKSDB on ARM64
    ......
}
```
```
public static void setParam(final String[] args, final String confFileName) {
    .....
    // 2. Config overrides defaults
    applyConfigParams(config);

    // 3. CLI overrides Config (highest priority)
    applyCLIParams(cmd, jc);

    // 4. Init witness (depends on CLI witness flag)
    initLocalWitnesses(config, cmd);
  }
```
```
 public static void applyConfigParams(
      final Config config) {
    ......
    initBackupProperty(config);
    if (Constant.ROCKSDB.equalsIgnoreCase(CommonParameter
        .getInstance().getStorage().getDbEngine())) {
      // dbEngine is still "LEVELDB" here; RocksDB path is never reached
      initRocksDbBackupProperty(config);
      initRocksDbSettings(config);
    }
}
```

framework/src/test/java/org/tron/common/BaseMethodTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @After
+  public final void destroyContext() {
+    beforeDestroy();
+    if (appT != null) {
+      appT.shutdown();
+    }
+    if (context != null) {
+      context.close();
+    }
```
3for: When `appT != null`, there is a potential **double shutdown** issue: the same Spring singleton bean may be shut down twice.

Specifically, `appT.shutdown()` is called explicitly, and then `context.close()` triggers another shutdown via [TronApplicationContext](https://github.com/vividctrlalt/java-tron/blob/test/dual-db-engine-coverage/framework/src/main/java/org/tron/common/application/TronApplicationContext.java#L27-L31), where `appT.shutdown()` is invoked again.

```java
public final void destroyContext() {
    beforeDestroy();
    if (appT != null) {
      appT.shutdown(); // first shutdown
    }
    if (context != null) {
      context.close(); // triggers second shutdown
    }
    Args.clearParam();
}
```

framework/src/main/java/org/tron/program/FullNode.java
vividctrlalt: Good catch on all three points! Fixed in abb52f6c3:

1. **Timing** — moved the ARM64 engine override into `setParam()`, after `applyCLIParams()` and before `initLocalWitnesses()`. Now `applyConfigParams()` still reads the original config value, but the engine is corrected before any downstream code relies on it.
2. **CLI override** — since `applyPlatformConstraints()` runs after `applyCLIParams()`, both config and CLI sources are resolved before the ARM64 check.
3. **Naming** — renamed `validateConfig()` → `applyPlatformConstraints()` to accurately reflect that it mutates config rather than just validating. Also changed to `private` since it's now called internally within `setParam()`.

framework/src/test/java/org/tron/common/BaseMethodTest.java
vividctrlalt: Good catch! Fixed in abb52f6c3 — removed the explicit `appT.shutdown()` call. `context.close()` already triggers `appT.shutdown()` via `TronApplicationContext.destroyBean()`, so the explicit call was redundant.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -130,10 +130,27 @@ public static void setParam(final String[] args, final String confFileName) {
     // 3. CLI overrides Config (highest priority)
     applyCLIParams(cmd, jc);
 
-    // 4. Init witness (depends on CLI witness flag)
+    // 4. Apply platform constraints (e.g. ARM64 forces RocksDB)
+    applyPlatformConstraints();
+
+    // 5. Init witness (depends on CLI witness flag)
     initLocalWitnesses(config, cmd);
```
3for, line 137: In **Step 2 (`applyConfigParams`)**, there is logic that depends on `dbEngine`. However, in **Step 4 (`applyPlatformConstraints`)**, `dbEngine` may still be modified.

This creates an ordering issue. It would be better to **move the backup initialization logic that depends on `dbEngine` out of `applyConfigParams` and execute it after Step 4**, so it operates on the final effective engine.

Alternatively, if there is a better approach, it would be worth considering.

That said, this change would impact existing tests related to `applyConfigParams`, since it effectively moves both:

* backup-related initialization, and
* `logConfig()` (final config logging)

out of `applyConfigParams`.

Suggested adjustment:
```java
// Insert into `setParam` AFTER Step 4
// Move out from `applyConfigParams`
initBackupProperty(config);
if (Constant.ROCKSDB.equalsIgnoreCase(
    CommonParameter.getInstance().getStorage().getDbEngine())) {
  initRocksDbBackupProperty(config);
  initRocksDbSettings(config);
}

// Move out from `applyConfigParams`
logConfig(); // log final effective configuration

// Insert into `setParam` BEFORE Step 5
```

It would also be helpful to add tests to verify that, on ARM64, even if `LEVELDB` is configured:

* the effective `dbEngine` is ultimately normalized to `ROCKSDB`, and
* `initRocksDbBackupProperty` and `initRocksDbSettings` are both correctly executed.

3for:
Based on the latest discussion in [[Discussion] Should aarch64 throw an exception when db.engine=LEVELDB? #6587](https://github.com/tronprotocol/java-tron/issues/6587), we still need a more robust and user-friendly way to handle mismatches between the **existing data directory** and the **final effective `dbEngine`**.

framework/src/main/java/org/tron/core/config/args/Args.java
vividctrlalt: Good point. The real issue is that `applyConfigParams` shouldn't conditionally skip reading config values based on the current engine. Config parsing should be unconditional — **if it's in the config file, read it**. The engine selection only matters at runtime when the values are actually used, not at config parsing time.

Fixed in d70c1c4df — removed the `if (ROCKSDB)` guard around `initRocksDbBackupProperty` and `initRocksDbSettings`. These methods only store config values in memory with no side effects, so reading them unconditionally is safe and eliminates the timing dependency entirely.

vividctrlalt:
Thanks for the detailed scenario matrix! The "Bad" cases (ARM64 + existing LevelDB data dir) are expected — this is a data migration scenario, not something the engine override should silently handle. Users in this situation need to run `Toolkit.jar db convert` to migrate their data from LevelDB to RocksDB before starting on ARM64.

Improving the error message in `checkOrInitEngine` to surface this guidance (e.g., mentioning `db convert`) would be a good follow-up, but that's better tracked in #6587 to keep this PR focused on test coverage.

framework/src/main/java/org/tron/core/config/args/Args.java
3for: “Unconditionally reading and parsing configuration values into memory” is a great approach—simpler and more elegant than the modification I suggested.

3for:
> Thanks for the detailed scenario matrix! The "Bad" cases (ARM64 + existing LevelDB data dir) are expected — this is a data migration scenario, not something the engine override should silently handle. Users in this situation need to run `Toolkit.jar db convert` to migrate their data from LevelDB to RocksDB before starting on ARM64.
> 
> Improving the error message in `checkOrInitEngine` to surface this guidance (e.g., mentioning `db convert`) would be a good follow-up, but that's better tracked in #6587 to keep this PR focused on test coverage.

Agreed — this issue is out of scope for this PR.

3for (approved):
LGTM now.

kuny0707 (approved):

PR #6593, opened 2026-03-20
fix(jsonrpc): add blockNumOrHashOrTag length check 

Sunny6889:
**What does this PR do?**
- Reliable Fixes: Added input validation for JSON-RPC block identifiers and hex parsing to reject null/overlong inputs, reducing risk of malformed-request abuse. Add ThreadLocal safety.
- Chores: Centralised rate limiter default values and lowered P2P rate limiter defaults (global and per-IP) for safer defaults.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -409,12 +416,6 @@ public String getTrxBalance(String address, String blockNumOrTag)
       }
       return ByteArray.toJsonHex(balance);
     } else {
```
Sunny6889, line 418: delete the below `ByteArray.hexToBigInteger(blockNumOrTag);` as it is actually no need to do this extra computation, since it will throw errors anyway.

vividctrlalt:
Suggestion: consider squashing the commits before merging. There are currently 26 commits, many of which are incremental fixes (`fix lint`, `fix merge`, `fix review`, `fix format`, etc.). Squashing into 1-2 meaningful commits would keep the git history clean and make it easier to understand the change in the future.

For example:
```bash
git rebase -i develop
# squash all into one commit:
# fix(jsonrpc): add blockNumOrHashOrTag length check
```

PR #6611, opened 2026-03-30
feat(vm): implement TIP-7823

ouy95917:
This PR aims to implement [TIP-7823](https://github.com/tronprotocol/tips/issues/826), and for this purpose, the `allowTvmOsaka` proposal has been added.


framework/src/test/java/org/tron/common/runtime/vm/AllowTvmOsakaTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.Assert;
+import org.junit.Test;
+import org.tron.common.utils.ByteUtil;
+import org.tron.core.vm.PrecompiledContracts;
+import org.tron.core.vm.config.ConfigLoader;
+import org.tron.core.vm.config.VMConfig;
+
+@Slf4j
+public class AllowTvmOsakaTest extends VMTestBase {
```
CodeNinjaEvan, line 13: The current test only covers two cases: all-zero lengths (valid) and an oversized baseLen (invalid). Please add the following cases:

baseLen == 1024 → should succeed (boundary value)
baseLen == 1025 → should fail (just over the limit)
Oversized expLen only → should fail
Oversized modLen only → should fail

All limits exceeded while allowTvmOsaka is disabled → should succeed (no restriction applied)

ouy95917: Thanks for the review! All suggested test cases have been added in the latest commit:

  - baseLen == 1024 → passes (boundary value)
  - baseLen == 1025 → fails (just over the limit)
  - Oversized expLen only → fails
  - Oversized modLen only → fails
  - All limits exceeded with allowTvmOsaka disabled → passes (no restriction applied)

  Also added a helper method toLenBytes(int) for cleaner ABI-encoded length construction. All tests pass locally.

actuator/src/main/java/org/tron/core/vm/PrecompiledContracts.java
```diff
@@ -660,6 +662,11 @@ public Pair<Boolean, byte[]> execute(byte[] data) {
       int expLen = parseLen(data, 1);
       int modLen = parseLen(data, 2);
 
+      if (VMConfig.allowTvmOsaka()
+          && (baseLen > UPPER_BOUND || expLen > UPPER_BOUND || modLen > UPPER_BOUND)) {
+        return Pair.of(false, EMPTY_BYTE_ARRAY);
+      }
```
CodeNinjaEvan, line 668: Hey @ouy95917, just a quick curiosity question here — is there a specific calculation or reasoning behind why baseLen, expLen, and modLen are each capped at 1024 bytes specifically? I'm wondering whether this was driven by cryptographic practicality (e.g., covering RSA-8192 as the largest real-world key size), or some other consideration from the EIP authors. Any context you can share — or a pointer to the relevant EIP discussion thread — would be greatly appreciated. Thanks!

ouy95917: The 1024-byte (8192-bit) limit was mainly driven by cryptographic practicality — it covers RSA keys up to 8192 bits, which is the largest size used in practice. Ethereum's on-chain analysis (~7 years of mainnet data) confirmed the max observed input was only 513 bytes, so 1024 gives plenty of headroom.

There's a more detailed discussion on this in the TIP issue: https://github.com/tronprotocol/tips/issues/826 — covers the rationale, error semantics, security background, etc. Worth a read if you're interested.

SecretCipher7 (approved):

CodeNinjaEvan (approved):
LGTM

lvs0075 (approved):

CodeNinjaEvan:
Relates to tronprotocol/tips#826

PR #6613, opened 2026-03-31
refactor(vm): optimize opcode energy cost calculation with `BigInteger`

yanghang8612:
## Summary

- Use `BigInteger` instead of `DataWord` for memory size computation in energy cost calculation to improve precision
- Add `allowTvmOsaka` proposal config in `VMConfig` for future activation control
- Add unit tests for the optimized calculation

### DIFF
--- a/framework/src/test/java/org/tron/common/runtime/vm/VoteWitnessCost3Test.java
+++ b/framework/src/test/java/org/tron/common/runtime/vm/VoteWitnessCost3Test.java
@@ -124,8 +124,8 @@ public class VoteWitnessCost3Test extends BaseTest {
     // witness array zero, amount array non-zero
     Program program = mockProgram(0, 0, 64, 1, 0);
     long cost = EnergyCost.getVoteWitnessCost3(program);
-    // witnessMemNeeded = 0 (size is zero)
-    // amountArraySize = 1 * 32 + 32 = 64, memNeeded = 64 + 64 = 128
+    // witnessArraySize = 0 * 32 + 32 = 32, witnessMemNeeded = 0 + 32 = 32
+    // amountArraySize = 1 * 32 + 32 = 64, amountMemNeeded = 64 + 64 = 128
     // memWords = 128 / 32 = 4
     // memEnergy = 3 * 4 + 4 * 4 / 512 = 12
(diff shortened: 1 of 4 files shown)


framework/src/test/java/org/tron/common/runtime/vm/VoteWitnessCost3Test.java
```diff
@@ -0,0 +1,244 @@
...
+    // memEnergy = 3 * 1 + 1 * 1 / 512 = 3
+    assertEquals(30003, cost);
+  }
+
+  @Test
+  public void testZeroLengthOneArray() {
+    // witness array zero, amount array non-zero
+    Program program = mockProgram(0, 0, 64, 1, 0);
+    long cost = EnergyCost.getVoteWitnessCost3(program);
+    // witnessMemNeeded = 0 (size is zero)
```
CodeNinjaEvan: The unit tests in VoteWitnessCost3Test are well-structured — especially appreciate the overflow boundary coverage in testLargeArrayLengthOverflow and testLargeOffsetOverflow, and the consistency check against cost2 in testConsistentWithCost2ForSmallValues. One minor note: the comment in testZeroLengthOneArray states witnessMemNeeded = 0, but cost3 computes size = 0 * 32 + 32 = 32, which is non-zero — worth clarifying if this is an intentional behavior change from cost2.

CodeNinjaEvan (approved):

framework/src/test/java/org/tron/common/runtime/vm/VoteWitnessCost3Test.java
yanghang8612: Good catch! The comment was indeed inaccurate — in cost3, even when `arrayLength` is 0, the size is computed as `0 * 32 + 32 = 32` (accounting for the dynamic array length prefix), so `witnessMemNeeded` is 32, not 0. This is an intentional behavior difference from cost2, where a zero-length `DataWord` after `mul` stays zero and `memNeeded` returns 0. Fixed the comment in the latest push. Thanks for the review!

CodeNinjaEvan (approved):
LGTM

lvs0075 (approved):

PR #6614, opened 2026-03-31
test(framework): consolidate and stabilize CredentialsTest

3for:
**What does this PR do?**

This PR cleans up and consolidates `CredentialsTest` coverage in the `framework` module.

- removes the duplicate test under the incorrect package path `org.tron.keystroe`
- keeps the test in the correct package `org.tron.keystore`
- replaces random key generation with deterministic mocked `SignInterface` fixtures
- adds stronger assertions for `Credentials.create(...)`, `equals(...)`, and `hashCode()`

The previous test setup had two problems:

1. test coverage was split across two locations, including a typo package path
2. some assertions depended on random key generation, which made the test intent less explicit and less stable

This change makes the test deterministic, easier to understand, and better aligned with the current `Credentials` contract:
- `address` is derived from `getAddress()`
- equality depends on both `cryptoEngine` and `address`
- equal objects must have the same `hashCode`

- Unit Tests
  - `./gradlew framework:test --tests org.tron.keystore.CredentialsTest`

**Follow up**

No follow-up is required.

**Extra details**

This is a test-only change. No production logic, protocol behavior, or runtime code path is modified.


Federico2014 (approved):

framework/src/test/java/org/tron/keystore/CredentialsTest.java
```diff
@@ -1,25 +1,36 @@
...
 import org.tron.common.crypto.sm2.SM2;
 import org.tron.common.utils.ByteUtil;
+import org.tron.common.utils.StringUtil;
 
-@Slf4j
-public class CredentialsTest extends TestCase {
+public class CredentialsTest {
+
+  private static final byte[] ADDRESS_1 = ByteUtil.hexToBytes(
+      "410102030405060708090a0b0c0d0e0f1011121314");
```
CodeNinjaEvan, line 14: Nice work on the fixture improvement! Just a quick question out of curiosity — the old "TQhZ7...".getBytes() returns 34 UTF-8 bytes rather than a proper 21-byte address, so switching to ByteUtil.hexToBytes("410102...") definitely improves semantic correctness. That said, does this change have any practical impact on the test outcomes, or is it more of a correctness/readability improvement on the fixture side? Just want to make sure I'm not missing anything.

3for: @CodeNinjaEvan Yes, this is mainly a fixture correctness improvement rather than an intended behavioral change.

`SignInterface.getAddress()` is expected to return raw address bytes, so using a canonical 21-byte TRON address fixture is a better fit for the API contract. The previous Base58 string UTF-8 bytes could still make the test pass because `Credentials.create(...)` currently just feeds the returned bytes into `encode58Check(...)` without validating the address shape.

So it would be better to keep the new fixture as-is. It makes the test semantically correct, and it doesn’t materially change the current behavior under test.

CodeNinjaEvan: > @CodeNinjaEvan Yes, this is mainly a fixture correctness improvement rather than an intended behavioral change.
> 
> `SignInterface.getAddress()` is expected to return raw address bytes, so using a canonical 21-byte TRON address fixture is a better fit for the API contract. The previous Base58 string UTF-8 bytes could still make the test pass because `Credentials.create(...)` currently just feeds the returned bytes into `encode58Check(...)` without validating the address shape.
> 
> So it would be better to keep the new fixture as-is. It makes the test semantically correct, and it doesn’t materially change the current behavior under test.

Thanks for clarifying! Makes total sense — keeping the new fixture as-is.

framework/src/test/java/org/tron/keystore/CredentialsTest.java
```diff
@@ -28,21 +39,43 @@ public void testCreateFromSM2() {
...
+    Credentials sameAddressDifferentEngineCredential = Credentials.create(signInterface2);
+    Credentials differentCredential = Credentials.create(signInterface3);
+
+    Assert.assertFalse(aObject.equals(credential));
+    Assert.assertFalse(credential.equals(aObject));
+    Assert.assertFalse(credential.equals(null));
+    Assert.assertEquals(credential, sameCredential);
+    Assert.assertEquals("Equal credentials must have the same hashCode",
+        credential.hashCode(), sameCredential.hashCode());
+    Assert.assertNotEquals(credential, sameAddressDifferentEngineCredential);
```
lxcmyf, line 77: One semantic question: this test now locks in that two `Credentials` objects with the same address but different `SignInterface` instances are not equal. Is that the intended long-term contract?

framework/src/test/java/org/tron/keystore/CredentialsTest.java
```diff
@@ -28,21 +39,43 @@ public void testCreateFromSM2() {
       Credentials.create(SM2.fromNodeId(ByteUtil.hexToBytes("fffffffffff"
           + "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
           + "fffffffffffffffffffffffffffffffffffffff")));
+      Assert.fail("Expected IllegalArgumentException");
```
lxcmyf, line 42: Non-blocking: could we also add a happy-path test for `Credentials.create(SM2)`? Right now the SM2 overload is only covered through the invalid-input case.

framework/src/test/java/org/tron/keystore/CredentialsTest.java
3for: @lxcmyf Yes, this is intentional. The existing `Credentials.equals()` implementation already includes both `cryptoEngine` and `address` in its equality check, and `hashCode()` is consistent with this behavior.

A `Credentials` object represents a specific signing identity. Even if two instances share the same address, a different `SignInterface` (i.e., `cryptoEngine`) implies a different signing capability, so they should not be considered equal.

This test does not introduce new behavior; it simply documents and verifies the contract already defined in the `public boolean equals(Object o)` method of `Credentials.java`.

framework/src/test/java/org/tron/keystore/CredentialsTest.java
3for: @lxcmyf Good suggestion. The SM2 happy-path coverage is indeed thin here. However, this PR is focused on stabilizing the existing `CredentialsTest` fixtures. Adding new SM2 coverage would require additional crypto setup, so it’s better handled in a separate follow-up PR. I’ll track it for future work.

bladehan1 (approved):

lxcmyf (approved):

PR #6618, opened 2026-04-01
fix(jsonrpc): add blockNumOrHashOrTag length check 

SeriousCoding789:
**What does this PR do?**
- Reliable Fixes: Added input validation for JSON-RPC block identifiers and hex parsing to reject null/overlong inputs, reducing risk of malformed-request abuse. Add ThreadLocal safety.
- Chores: Centralised rate limiter default values and lowered API rate limiter defaults (global and per-IP) for safer defaults.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 </h4>
 
 <p align="center">
-  <a href="https://gitter.im/tronprotocol/allcoredev"><img src="https://img.shields.io/gitter/room/tronprotocol/java-tron.svg"></a>
+  <a href="https://discord.gg/hqKvyAM"><img src="https://img.shields.io/badge/chat-on%20discord-7289da.svg"></a>
   <a href="https://codecov.io/gh/tronprotocol/java-tron"><img src="https://codecov.io/gh/tronprotocol/java-tron/branch/develop/graph/badge.svg" /></a>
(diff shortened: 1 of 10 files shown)


vividctrlalt:
A few concerns:

**1. Rate limiter defaults reduced 50x — breaking change**

Global QPS 50000→1000, IP QPS 10000→1000. Nodes running default config will drop to 1/50th throughput after upgrading with no warning. This should at minimum be called out in release notes so operators can explicitly set the old values if needed.

**2. Length-only validation is weaker than removed hex validation**

The removed `hexToBigInteger` try-catch blocks validated hex format and gave `BLOCK_NUM_ERROR`. The new `validateBlockNumOrHashOrTag` only checks `length > 512` — a 500-char string of non-hex characters passes the new check but would have been caught before. Also `ethGetBlockByHash` (line 324) is missing the validation call while `ethGetBlockTransactionCountByHash` (line 300) has it.

**3. volatile on Chainbase.head is insufficient**

`head()` reads the `head` field multiple times during snapshot chain traversal without synchronization. If `setHead()` runs between reads, the method can traverse a stale chain. volatile ensures visibility but not atomicity of compound read operations — a local variable assignment at method entry (`Snapshot h = head`) or synchronization on the read path would be needed.

PR #6621, opened 2026-04-01
fix(db): resolve resource leakage in CheckPointV2Store.close()

warku123:
**What does this PR do?**

Fix resource leakage in `CheckPointV2Store.close()` method by properly handling exception scenarios during resource cleanup.

- Split the try-catch block to ensure parent resources are released even when `writeOptions.close()` throws exception
- Change `dbSource.closeDB()` to `super.close()` for proper inheritance chain cleanup
- Add separate exception handling for writeOptions and parent resources with detailed log messages

Previously, if `writeOptions.close()` threw an exception, the subsequent `dbSource.closeDB()` (now `super.close()`) would not be executed, causing resource leakage. This is particularly problematic in test environments where multiple database instances are created and destroyed.

- Unit Tests (added `testCloseResourceLeakage()` to verify `super.close()` is always called)
- Manual Testing

**Follow up**

N/A

**Extra details**

- Fix checkstyle violations (import order and line length)
- Add Javadoc annotations for test class and methods


chainbase/src/main/java/org/tron/core/store/CheckPointV2Store.java
```diff
@@ -70,9 +70,13 @@ public void close() {
     logger.debug("******** Begin to close {}. ********", getName());
     try {
       writeOptions.close();
```
halibobo1205, line 73: Nice fix overall — this looks solid and addresses the issue well 👍

Just a couple of minor points that might be worth considering:

* **Duplicate logs**: The subclass `close()` logs *"Begin to close"* / *"End to close"*, and `super.close()` logs the same messages again (with different levels — `debug` vs `info`). This could result in duplicated log entries for a single close operation, which might be a bit confusing when debugging.

* **Redundant catch**: Since the parent `close()` already handles all exceptions internally and does not rethrow, the `try { super.close(); } catch (Exception e) { ... }` in the subclass will never actually catch anything. It might be safe to simplify this part.

warku123: @halibobo1205 Thanks for the thorough review! That's an excellent points:

1. **Duplicate logs**: You're absolutely right — both subclass and parent class log the same messages. I'll remove the debug logs from the subclass and let the parent class handle all the logging.

2. **Redundant catch**: Good catch! Since `super.close()` handles all exceptions internally without re-throwing, the extra try-catch in the subclass is indeed unnecessary. I'll simplify it to just call `super.close()` directly.

I'll update the PR shortly with these changes.

vividctrlalt:
The fix direction is right — splitting try-catch to ensure `dbSource.closeDB()` runs. But there are a few issues:

**1. Parent class has the same bug (P1)**

`TronDatabase.close()` still has `writeOptions.close()` and `dbSource.closeDB()` in the same try block. `super.close()` doesn't guarantee `dbSource.closeDB()` runs if parent's `writeOptions.close()` throws. The fix moves the problem from child to parent. Should split `TronDatabase.close()` the same way:

```java
try {
    writeOptions.close();
} catch (Exception e) {
    logger.warn("Failed to close writeOptions in {}.", getName(), e);
}
try {
    dbSource.closeDB();
} catch (Exception e) {
    logger.warn("Failed to close dbSource in {}.", getName(), e);
}
```

**2. Test doesn't test the bug (P1)**

The fix is about exception handling, but the test never makes `writeOptions.close()` throw. It only verifies happy-path invocation. Need a test that mocks `writeOptions.close()` to throw, then verifies `dbSource.closeDB()` still executes.

**3. `dbSource.closeDB()` never verified (P1)**

The whole point is ensuring `dbSource` gets closed, but the test only checks `writeOptions.close()` was called. Should add `verify(dbSource).closeDB()`.

**4. Reflection-based test is fragile (P2)**

Using `getDeclaredField("writeOptions")` + `setAccessible(true)` + spy tests implementation details. Breaks if the field is renamed. Consider testing at the behavioral level instead.

**5. Double close in finally block (P2)**

`store.close()` is called at line 88, then again in the finally block at line 98 on the already-closed store. Then restores spy state on a dead object — this cleanup is meaningless.

**6. Vacuous assertions (P2)**

`assertNotNull(spyParentWriteOptions)` on line 83 — this spy was just created two lines above, it can never be null. Same for the other two asserts.

**7. Debug logging removed (P3)**

The child's Begin/End close logging was removed. Parent still logs at INFO level, so this is minor, but child-specific close failures are less visible now.

PR #6602, opened 2026-03-25
feat(metrics): add block transaction count histogram for empty block monitoring

ToXMon:
## Summary
Implements [Issue #6590](https://github.com/tronprotocol/java-tron/issues/6590) — adds Prometheus metrics for empty block detection and SR set change monitoring:

- `tron:block_transaction_count` — Histogram tracking distribution of transaction counts per block (enables empty block detection via bucket `le="0.0"`)
- `tron:sr_set_change_total` — Counter tracking SR set changes with `added`/`removed` labels

### New Metrics
| Metric | Type | Labels | Description |
|--------|------|--------|-------------|
| `tron:block_transaction_count` | Histogram | `miner` | Distribution of tx counts per block (empty blocks: `le="0.0"`) |
| `tron:sr_set_change_total` | Counter | `witness`, `change_type` | SR set changes (added/removed) |

### Key Implementation Details
- **Histogram buckets**: `[0, 10, 50, 100, 200, 500, 1000, 2000, 5000, 10000]`
- **Empty block query**: `tron:block_transaction_count_bucket{le="0.0"}`
- **Empty block ratio**: `rate(tron:block_transaction_count_bucket{le="0.0"}[1h]) / rate(tron:block_transaction_count_count[1h])`

### Files Modified
1. `common/src/main/java/org/tron/common/prometheus/MetricKeys.java` — removed BLOCK_EMPTY counter, added BLOCK_TRANSACTION_COUNT histogram
2. `common/src/main/java/org/tron/common/prometheus/MetricsCounter.java` — removed BLOCK_EMPTY registration
3. `common/src/main/java/org/tron/common/prometheus/MetricsHistogram.java` — added overloaded init() for custom buckets
4. `framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java` — histogramObserve() for all blocks, keeps SR counter
5. `framework/src/test/java/org/tron/core/metrics/prometheus/PrometheusApiServiceTest.java` — updated tests for histogram

## Testing
- `testMetric()` — PASSED
- `testEmptyBlockMetric()` — PASSED  
- `testSrSetChangeMetric()` — PASSED

## Advantages Over Simple Counter
- **Rich insights**: Tracks full distribution of tx counts, not just empty blocks
- **Flexible queries**: Operators can query percentiles, trends, specific ranges
- **Backward compatible**: Still enables empty block monitoring via `le="0.0"` bucket

## Backward Compatibility
[...]

PR #6624, opened 2026-04-01
feat(metrics): add block transaction count and SR set change monitoring

### LINKED ISSUE #6590: [Feature] Add Prometheus metrics for empty blocks and SR set changes
## Background

Currently, java-tron's Prometheus metrics provide comprehensive monitoring for transactions, blocks, P2P connections, and consensus status. However, there are two important operational aspects that lack visibility:

1. **Empty Block Detection**: There is no metric to track blocks that contain no transactions. Monitoring empty blocks is crucial for identifying potential network issues, misconfigured witness nodes, or degraded transaction throughput.

2. **Super Representative (SR) Set Changes**: The active witness set changes during maintenance periods when new SRs are elected or removed. Currently, there is no way to track these changes through metrics, making it difficult to monitor consensus layer dynamics and SR rotation events.

## Problem Statement & Rationale

- Node operators cannot identify which witness nodes are producing empty blocks, making it difficult to diagnose network health issues
- There is no real-time visibility into SR set changes

**What are the use cases?**

1. Identify witnesses producing excessive empty blocks
2. Monitor SR election results and consensus transitions

## Proposed Solution

Add two new Prometheus counters to track empty blocks and SR set changes:

- `tron:block_empty_total` - Tracks empty blocks
- `tron:sr_set_change_total` - Tracks SR set changes (add/remove)

## Testing Strategy

**Test Scenarios**
1. **Unit Test**: Add unit test about two counters.
2. **Manual Test**: Build a private chain and test the case include produce empty and SR set change.

## Scope of Impact

- [x] Other: Metrics/Monitoring system

**Backward Compatibility**

Fully backward compatible. Existing metrics remain unchanged. New metrics are opt-in via Prometheus scraping.

## Implementation

The implementation involves:

1. **Metric Registration**: Add new counter definitions in `MetricKeys.java`, `MetricLabels.java`, and `MetricsCounter.java`

2. **BlockChainMetricManager Modifications**: In `BlockChainMetricManager.java`:
     - Add empty block detection logic
     - Add SR set change detection logic

3. **Testing**: Extend `PrometheusApiServiceTest` to cover new metrics functionality

- [x] Yes, I can implement this feature

**Estimated Complexity**

- [x] Medium (moderate changes)

**References**

- [Prometheus Counter Best Practices](https://prometheus.io/docs/practices/instrumentation/#use-labels)

warku123:
**What does this PR do?**

Add two new metrics enhancements for better network monitoring:

1. **Block Transaction Count Histogram**
   - Replace `tron:block_empty_total` counter with `tron:block_transaction_count` histogram
   - Track transaction count distribution per block with buckets: [0, 10, 50, 100, 200, 500, 1000, 2000, 5000, 10000]
   - Empty blocks can be monitored via `le=0.0` bucket

2. **SR Set Change Monitoring**
   - Add `tron:sr_set_change` counter to track SR membership changes
   - Detect SR additions and removals at each maintenance time interval
   - Labels: `action` (add/remove), `witness` (SR address)

**Changes:**
- Add overloaded `init()` method in `MetricsHistogram` for custom buckets
- Add `BLOCK_TRANSACTION_COUNT` histogram metric
- Add `SR_SET_CHANGE` counter with `SR_ADD`/`SR_REMOVE` labels
- Implement SR set change detection logic in `BlockChainMetricManager`
- Update `PrometheusApiServiceTest` with comprehensive test coverage

- **Transaction histogram** provides richer insights than simple counter for network analysis while maintaining empty block detection capability
- **SR set change monitoring** enables tracking of network consensus participant changes for governance analysis

Compared to #6602, this PR provides a cleaner implementation focused solely on metrics enhancement without additional complexity.

- Unit Tests (updated `PrometheusApiServiceTest`)
- Manual Testing

**Follow up**

N/A

**Extra details**

- Closes #6590
- Related to #6602 (alternative cleaner implementation)


common/src/main/java/org/tron/common/prometheus/MetricKeys.java
```diff
@@ -14,6 +14,7 @@ public static class Counter {
     public static final String TXS = "tron:txs";
     public static final String MINER = "tron:miner";
     public static final String BLOCK_FORK = "tron:block_fork";
+    public static final String SR_SET_CHANGE = "tron:sr_set_change";
```
halibobo1205, line 17: The name `tron:sr_set_change` can be misleading — it sounds like it tracks changes to the entire SR set, when it actually monitors rotation of the top 27 block-producing SRs (Active Witnesses).

Suggested alternatives for better clarity:

- `tron:active_sr_change`

- `tron:active_witness_change`

- `tron:top27_sr_change`

This way, operators can immediately understand what the metric represents when viewing Prometheus dashboards.

common/src/main/java/org/tron/common/prometheus/MetricsCounter.java
```diff
@@ -14,6 +14,7 @@ class MetricsCounter {
     init(MetricKeys.Counter.TXS, "tron  txs  info .", "type", "detail");
     init(MetricKeys.Counter.MINER, "tron  miner info .", "miner", "type");
     init(MetricKeys.Counter.BLOCK_FORK, "tron  block fork info .", "type");
+    init(MetricKeys.Counter.SR_SET_CHANGE, "tron sr set change .", "action", "witness");
```
halibobo1205, line 17: The `witness` label uses addresses. Although the Active SR pool is relatively limited (27 slots), as SRs rotate in and out over time, label values will `accumulate monotonically` — they only grow, never shrink. This is a known Prometheus anti-pattern.

**Practical impact assessment**:

- TRON's SR candidate pool is not infinite, and rotation frequency is low, so it won't explode in the short term

- However, as a `long-running node metric`, months or years of operation could still accumulate a significant number of time series

**Possible approaches**:

- If the address dimension is truly needed, keep the current design, but document this characteristic

- Or remove the `witness` label entirely, keeping only `action` (add/remove), and output address details via logs instead. Follow the principle that `Prometheus should be used for alerting and awareness, while logs are for precise investigation and root-cause analysis`.

common/src/main/java/org/tron/common/prometheus/MetricKeys.java
Little-Peony: I'd like to keep `tron:sr_set_change` as-is.

Per our [official documentation](https://tronprotocol.github.io/documentation-en/mechanism-algorithm/sr/#how-to-become-a-super-representative), "Super Representative" is formally defined as the top 27 candidates by vote count — the block-producing set is precisely what "SR" means in the TRON protocol. The broader candidate pool is referred to as "SR candidates" or "SR partners", not "SR".

Given this, `tron:sr_set_change` is unambiguous: it tracks changes in the SR set, which by protocol definition is already scoped to the top 27. Adding qualifiers like `active_` or `top27_` would be redundant with the established terminology and could itself introduce confusion for operators already familiar with TRON's nomenclature.

framework/src/test/java/org/tron/core/metrics/prometheus/PrometheusApiServiceTest.java
```diff
@@ -38,6 +40,9 @@
 
 @Slf4j(topic = "metric")
 public class PrometheusApiServiceTest extends BaseTest {
+
+  private static final String MINER_LABEL = "miner";
```
Little-Peony: The `MINER_LABEL` constant is already defined in `MetricsHistogram` (as a `private static final`). Redefining it here creates a second source of truth — if the label name changes, both files need updating.

Consider either exposing the constant with package/protected visibility in `MetricsHistogram`, or extracting it to a shared location (e.g., `MetricLabels`), so the test can reference it directly instead of duplicating the string literal.

framework/src/test/java/org/tron/core/metrics/prometheus/PrometheusApiServiceTest.java
```diff
@@ -80,6 +85,56 @@ protected void check() throws Exception {
...
+      Assert.assertEquals("Each new witness should have 1 SR_ADD record", 1,
+          srAddCount.intValue());
+      
+      // Collect empty blocks count from histogram bucket
+      Double witnessEmptyBlock = CollectorRegistry.defaultRegistry.getSampleValue(
+          "tron:block_transaction_count_bucket",
+          new String[] {MINER_LABEL, "le"}, new String[] {witnessBase58, "0.0"});
+      Assert.assertNotNull("Empty block bucket should exist for witness: " + witnessBase58,
+          witnessEmptyBlock);
+      totalNewWitnessEmptyBlocks += witnessEmptyBlock;
```
Little-Peony, line 132: `totalNewWitnessEmptyBlocks` represents a block count, which is inherently an integer. Declaring it as `double` just to accommodate Prometheus's `Double` return type leaks the storage detail into the accumulator's type. The same applies to `witnessEmptyBlock` — histogram bucket counts are always whole numbers.

Suggest:
```java
int totalNewWitnessEmptyBlocks = 0;
// ...
int witnessEmptyBlock = CollectorRegistry.defaultRegistry
    .getSampleValue(...).intValue();
Assert.assertNotNull(...);
totalNewWitnessEmptyBlocks += witnessEmptyBlock;
// ...
Assert.assertEquals(blocks, totalNewWitnessEmptyBlocks);
```

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -169,6 +175,46 @@ public void applyBlock(BlockCapsule block) {
...
+          MetricLabels.Counter.SR_ADD, StringUtil.encode58Check(Hex.decode(address)));
+    }
+    for (String address : removed) {
+      Metrics.counterInc(MetricKeys.Counter.SR_SET_CHANGE, 1,
+          MetricLabels.Counter.SR_REMOVE, StringUtil.encode58Check(Hex.decode(address)));
+    }
+    if (!added.isEmpty() || !removed.isEmpty()) {
+      lastActiveWitnesses.clear();
+      lastActiveWitnesses.addAll(currentWitnesses);
+    }
```
Little-Peony, line 228: `lastActiveWitnesses` stores witness addresses as hex strings, but the only reason to decode them later is to re-encode as base58 for the Prometheus label. This round-trip — `ByteString → hex → byte[] → base58` — adds unnecessary complexity.

Since base58 strings are equally suitable for set difference operations (`added.removeAll(...)`), the hex intermediate format serves no purpose. Storing base58 directly eliminates the `Hex.decode()` call in `recordSrSetChange`:

```java
// Instead of:
.map(w -> Hex.toHexString(w.toByteArray()))

// Use:
.map(w -> StringUtil.encode58Check(w.toByteArray()))
```

Then in `recordSrSetChange`, `address` can be used directly as the label value without decoding.

common/src/main/java/org/tron/common/prometheus/MetricsCounter.java
warku123: Thanks for raising the high-cardinality concern — it's a valid Prometheus best practice to watch for.

  However, I believe the witness label is acceptable here given TRON's specific constraints:

  1. Bounded candidate pool: The SR candidate pool is finite. Even over years of operation, the total number of distinct addresses that have ever held an SR slot has a practical ceiling — far from the
  unbounded growth typical of high-cardinality anti-patterns.
  2. Low rotation frequency: SR set changes only occur at maintenance intervals (every 6 hours), so new label values are introduced very infrequently.
  3. Diagnostic value: The primary purpose of this metric is to identify which SR was added or removed. Dropping the witness label would reduce the metric to just "something changed", requiring operators to
  cross-reference logs — losing the at-a-glance visibility in Grafana dashboards that makes this metric useful in the first place.

  In summary, the worst-case time series count is roughly candidate_pool_size × 2 (add/remove), which stays well within Prometheus's comfortable operating range.

  That said, I agree it's worth documenting this characteristic. Could you suggest what form you'd prefer — a code comment at the metric registration site, a note in the PR description, or something else?

halibobo1205: prefer to `a code comment at the metric registration site`.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -169,6 +175,47 @@ public void applyBlock(BlockCapsule block) {
       Metrics.counterInc(MetricKeys.Counter.TXS, block.getTransactions().size(),
           MetricLabels.Counter.TXS_SUCCESS, MetricLabels.Counter.TXS_SUCCESS);
     }
+    // Record transaction count distribution for all blocks (including empty blocks)
+    int txCount = block.getTransactions().size();
```
halibobo1205: L174, L175, and L179 can use txCount.

halibobo1205:
Two questions regarding observability adoption of the new metrics:

1. **PromQL query guide**: Could you provide some common query examples for the new `tron:block_transaction_count` histogram and `tron:sr_set_change` counter? For instance:
   - How to query empty block rate via the `le="0.0"` bucket
   - How to calculate P50/P99 transaction counts over a given time window
   - How to query SR membership changes from the most recent maintenance cycle

2. **Grafana dashboard examples**: Are there any accompanying Grafana dashboard JSONs or screenshots? If so, it would be helpful to include them in the PR description or documentation, so node operators can import and use them directly, reducing the onboarding effort.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -169,6 +175,47 @@ public void applyBlock(BlockCapsule block) {
...
+    Metrics.histogramObserve(MetricKeys.Histogram.BLOCK_TRANSACTION_COUNT, txCount,
```
lxcmyf: `metricsService.applyBlock(block)` runs before `dbManager.pushBlock(block)` in `TronNetDelegate.processBlock()`, so this histogram is updated for blocks that are later rejected or only live on a fork. For an empty-block monitoring metric, that produces false positives per witness and lets invalid peer traffic skew the distribution. This should be recorded only after the block has been applied successfully.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -169,6 +175,47 @@ public void applyBlock(BlockCapsule block) {
...
+        StringUtil.encode58Check(address));
+
+    // SR set change detection
+    long nextMaintenanceTime = dbManager.getDynamicPropertiesStore().getNextMaintenanceTime();
```
lxcmyf: The SR-set change detection is reading `nextMaintenanceTime` and `getActiveWitnesses()` before the current block is applied. In the real pipeline those values are updated during `dbManager.pushBlock()` / `consensus.applyBlock()`, so on the maintenance block this branch still sees the old schedule and old maintenance timestamp. That makes `tron:sr_set_change` show up one block late, and the added test masks the problem by mutating both stores manually before generating the next block.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
warku123: Good catch. I actually noticed this ordering issue during my investigation — `metricsService.applyBlock()` is called before `dbManager.pushBlock()`, meaning all metrics in `BlockChainMetricManager.applyBlock()` are recorded for blocks that may later be rejected.

However, this is not specific to the histogram I added — the existing metrics in the same method (e.g., `MINER_LATENCY`, `TXS`, `NET_LATENCY`, duplicate witness detection) all share the same problem. I chose to keep the same pattern to maintain architectural consistency and avoid expanding the scope of this PR.

If you agree, I'd like to open a separate issue to move `metricsService.applyBlock()` after `dbManager.pushBlock()` succeeds in `TronNetDelegate.processBlock()`, which would fix the timing problem for all metrics at once.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
warku123: Same root cause as above — since `metricsService.applyBlock()` runs before `dbManager.pushBlock()`, the maintenance time and active witness list haven't been updated yet at the point of reading.

In fact, this is precisely why the SR-set change detection relies on `nextMaintenanceTime` change as the trigger — because `getActiveWitnesses()` still returns the old schedule at this point, comparing against the previous `nextMaintenanceTime` is the only way to detect that a maintenance cycle has occurred and capture the witness list transition.

This would be addressed together in that follow-up issue: once `metricsService.applyBlock()` is moved after a successful `pushBlock()`, the SR-set change detection will naturally read the up-to-date state without needing this workaround.

warku123:
@halibobo1205 Thanks for the great suggestion! Here are the PromQL examples and Grafana dashboard screenshots:

### 1. Common PromQL Queries

**Empty block rate** (sliding window):
```promql
rate(tron:block_transaction_count_bucket{le="0.0"}[$__rate_interval])
/ ignoring(le)
rate(tron:block_transaction_count_count[$__rate_interval])
```

**Per-SR empty block count** (last 1 hour):
```promql
round(increase(tron:block_transaction_count_bucket{le="0.0"}[1h]))
```

**Average transactions per block** (sliding window):
```promql
rate(tron:block_transaction_count_sum[$__rate_interval])
/ rate(tron:block_transaction_count_count[$__rate_interval])
```

**Transaction count distribution** (non-cumulative buckets, e.g. 0-10):
```promql
round(increase(tron:block_transaction_count_bucket{le="10.0"}[$__rate_interval]))
- ignoring(le)
round(increase(tron:block_transaction_count_bucket{le="0.0"}[$__rate_interval]))
```

**SR set change events**:
```promql
tron:sr_set_change_total
```

**SR set change total by action**:
```promql
sum(tron:sr_set_change_total) by (action)
```

### 2. Grafana Dashboard Screenshots

Tested on a 2-node private net (1 witness + 1 fullnode with metrics enabled), scraped by Prometheus every 3s. The dashboard contains two sections:

- **Block Transaction**: Empty block rate, per-SR empty block count, average transactions per block, and transaction count distribution
- **SR Set Change**: SR set change event table and cumulative totals by action (add/remove)

截屏2026-04-09 16 33 52

截屏2026-04-09 16 34 00

The full Grafana dashboard JSON is also available if anyone wants to import it directly.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -164,11 +170,52 @@ public void applyBlock(BlockCapsule block) {
...
+      MetricsUtil.meterMark(MetricsKey.BLOCKCHAIN_TPS, txCount);
+      Metrics.counterInc(MetricKeys.Counter.TXS, txCount,
           MetricLabels.Counter.TXS_SUCCESS, MetricLabels.Counter.TXS_SUCCESS);
     }
+    // Record transaction count distribution for all blocks (including empty blocks)
+    Metrics.histogramObserve(MetricKeys.Histogram.BLOCK_TRANSACTION_COUNT, txCount,
+        StringUtil.encode58Check(address));
+
+    // SR set change detection
+    long nextMaintenanceTime = dbManager.getDynamicPropertiesStore().getNextMaintenanceTime();
```
halibobo1205: The SR set change detection logic in `BlockChainMetricManager.applyBlock()` could be simplified by moving it into `MaintenanceManager.doMaintenance()`, where `currentWits` (before) and `newWits` (after) are already computed at lines 132-148. This would:

1. Eliminate the need for `lastActiveWitnesses`, `lastNextMaintenanceTime` fields and per-block `nextMaintenanceTime` checks
2. Avoid duplicating the maintenance-time detection that `MaintenanceManager` already handles
3. Reduce to just a few `Metrics.counterInc` calls at the existing diff site

halibobo1205:
The new metrics are well-documented in the PR description. For node operators and dashboard maintainers who may not track individual PRs, it might be worth considering:

1. Adding a `METRICS_CHANGELOG.md` to the repo to track metric additions, modifications, and deprecations — future PRs can maintain it as well
2. Including a metrics change entry in release notes
3. Filing a follow-up issue/TODO to update the [official metrics documentation](https://tronprotocol.github.io/documentation-en/using_javatron/metrics/) and the [Grafana dashboard template](https://grafana.com/grafana/dashboards/16567-java-tron-server/)

This ensures metric changes have a centralized tracking point for operators to discover and configure.

halibobo1205:
Two concerns about the histogram choice:

1. **Empty block → block number correlation**: When an empty block alert fires, operators typically need to know *which* block was empty to investigate (network issue? SR misbehavior?). Prometheus metrics can't carry block numbers, so this requires either a log line at the observation site or a separate mechanism. Is there an existing log that operators can correlate against? If not, it might be worth adding one alongside the metric.

2. **Histogram cost vs. benefit**: Compared to a simple `tron:block_empty_total` counter, the histogram introduces ~13× the time series per SR (10 buckets + sum/count). The absolute cost is modest, but it's only worth it if transaction distribution analysis (P50/P99, large-block ratio, etc.) is an actual use case. Could you share a concrete scenario where the distribution buckets would be queried? If empty block alerting is the primary driver, a counter + correlated log would be simpler and cheaper.

warku123:
> Two concerns about the histogram choice:
> 
> 1. **Empty block → block number correlation**: When an empty block alert fires, operators typically need to know _which_ block was empty to investigate (network issue? SR misbehavior?). Prometheus metrics can't carry block numbers, so this requires either a log line at the observation site or a separate mechanism. Is there an existing log that operators can correlate against? If not, it might be worth adding one alongside the metric.
> 2. **Histogram cost vs. benefit**: Compared to a simple `tron:block_empty_total` counter, the histogram introduces ~13× the time series per SR (10 buckets + sum/count). The absolute cost is modest, but it's only worth it if transaction distribution analysis (P50/P99, large-block ratio, etc.) is an actual use case. Could you share a concrete scenario where the distribution buckets would be queried? If empty block alerting is the primary driver, a counter + correlated log would be simpler and cheaper.

Regarding the empty block → block number correlation: there is an existing log that operators can use. In `TronNetDelegate.processBlock()`, every successfully applied block emits:

```
logger.info("Success process block {}", blockId.getString());
```

`blockId.getString()` includes the block number, so when an empty-block alert fires, operators can look up the log around that time window to identify the specific blocks. Since empty blocks occur at a predictable rate (one per slot when no transactions are pending), the correlation is straightforward.

That said, if a more targeted signal is preferred, we could add a `log.debug` at the histogram observation site that only fires for empty blocks:

```java
if (txCount == 0) {
  logger.debug("Empty block observed: num={}, miner={}", block.getNum(), witnessAddress);
}
```

Happy to add this if it would improve operator experience.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
warku123: Moving the SR-set change detection into `MaintenanceManager.doMaintenance()` is an interesting suggestion, but I'd prefer to keep all metric-related logic centralized in `BlockChainMetricManager`. Currently `MaintenanceManager` has zero metric calls — mixing metric concerns into it would scatter observability code across unrelated components, making it harder for future maintainers to find and update metrics in one place.

The `lastActiveWitnesses` / `lastNextMaintenanceTime` fields are admittedly a bit of boilerplate, but they exist precisely because `metricsService.applyBlock()` runs before `dbManager.pushBlock()` — once the follow-up issue moves that call to after a successful push, the detection can be simplified while still staying within `BlockChainMetricManager`.

warku123:
> The new metrics are well-documented in the PR description. For node operators and dashboard maintainers who may not track individual PRs, it might be worth considering:
> 
> 1. Adding a `METRICS_CHANGELOG.md` to the repo to track metric additions, modifications, and deprecations — future PRs can maintain it as well
> 2. Including a metrics change entry in release notes
> 3. Filing a follow-up issue/TODO to update the [official metrics documentation](https://tronprotocol.github.io/documentation-en/using_javatron/metrics/) and the [Grafana dashboard template](https://grafana.com/grafana/dashboards/16567-java-tron-server/)
> 
> This ensures metric changes have a centralized tracking point for operators to discover and configure.

All three suggestions are well-taken:

- `METRICS_CHANGELOG.md`: Agreed — I'll follow up with a separate PR to introduce this file and add an entry for these two metrics. Keeping it as a dedicated PR makes it easier for the team to discuss the format and adopt it as a standard going forward.

- Release notes: I can provide a draft entry that maintainers can incorporate directly when cutting the release. Something like:

  - New metrics: `tron:block_transaction_count` (histogram, per-SR transaction distribution) and `tron:sr_set_change` (counter, active SR set rotation events)

- Official documentation & Grafana template: Will open a follow-up PR to track updates to the [Grafana dashboard template](https://github.com/tronprotocol/tron-docker/tree/develop/metric_monitor/grafana_dashboard).

bladehan1 (commented):
Review with 2 findings (P1). See inline comments.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -42,6 +45,9 @@ public class BlockChainMetricManager {
   private long failProcessBlockNum = 0;
   @Setter
   private String failProcessBlockReason = "";
+  private final Set<String> lastActiveWitnesses = ConcurrentHashMap.newKeySet();
```
bladehan1: **[P1] Two issues with `lastActiveWitnesses` and `lastNextMaintenanceTime`**

**1) False thread-safety guarantee from `ConcurrentHashMap.newKeySet()`**

`applyBlock()` is only called from `TronNetDelegate.processBlock()`, which holds the `dbManager` lock — a single-threaded path. More importantly, the compound operations in `recordSrSetChange()` (read → diff → clear → addAll) are **not atomic** even with a concurrent set. If concurrent access were ever needed, this implementation would silently produce incorrect diffs.

Suggest replacing with a plain `HashSet`:
```java
// Only accessed from block processing thread
private final Set<String> lastActiveWitnesses = new HashSet<>();
```

**2) In-memory state is not rolled back during fork switching**

`lastActiveWitnesses` and `lastNextMaintenanceTime` are pure in-memory state that does **not** participate in the revoke/undo mechanism.

During `Manager.switchFork()`:
1. `eraseBlock()` rolls back the old chain — DB state (`nextMaintenanceTime`, `activeWitnesses`) is reverted via `revokingStore.fastPop()`
2. `Manager.applyBlock()` applies new-chain blocks — DB state is rebuilt
3. But this internal `applyBlock` does **not** go through `TronNetDelegate.processBlock()`, so `metricsService.applyBlock()` is never called and the in-memory state is **not updated**

**Scenario — fork across a maintenance boundary:**
- Old chain passed maintenance: `lastNextMaintenanceTime = T1`, `lastActiveWitnesses = {SR_old}`
- Fork switches back to a pre-maintenance chain: DB now has `nextMaintenanceTime = T0 < T1`
- On the next block, metrics sees `nextMaintenanceTime(T0) != lastNextMaintenanceTime(T1)` → triggers SR change detection
- But it compares `lastActiveWitnesses` (stale, from old chain) against the current store (new chain) → **phantom SR changes** or **missed real changes**

Suggest adding a rollback guard:
```java
} else if (nextMaintenanceTime != lastNextMaintenanceTime) {
    if (nextMaintenanceTime < lastNextMaintenanceTime) {
        // Fork rollback detected — reinitialize instead of diffing
        lastActiveWitnesses.clear();
        lastActiveWitnesses.addAll(...current witnesses...);
        lastNextMaintenanceTime = nextMaintenanceTime;
        return;
    }
    // normal maintenance transition — diff and record
    ...
}
```

bladehan1 (commented):
Review: 2 findings on `lastActiveWitnesses` / `lastNextMaintenanceTime` state management.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
bladehan1: On second thought, adding a rollback guard here over-complicates the monitoring layer. The core issue is that SR set change detection is **business logic monitoring** — it depends on chain state transitions (maintenance boundaries, fork switching), which makes it fundamentally unsuitable for an in-process metric that cannot participate in the revoke mechanism.

A simpler and more robust approach would be to **move SR set change detection off-chain**: an external monitor reads the `/metrics` endpoint or listens to block events, compares consecutive maintenance periods' active witness lists, and emits its own alerts. This completely decouples monitoring from chain state management and avoids the fork/rollback problem entirely.

For the histogram (`BLOCK_TRANSACTION_COUNT`), in-chain observation is fine — it's a stateless per-block metric with no accumulated state to go stale. The issue is specific to `SR_SET_CHANGE` because it maintains cross-block state (`lastActiveWitnesses`, `lastNextMaintenanceTime`) that the chain's undo mechanism cannot reach.

warku123: Thanks for the thorough analysis! Both points are valid — I've addressed them in the latest commit (7fdb3c5):

**1) `ConcurrentHashMap.newKeySet()` → `HashSet`**

Agreed. Since `applyBlock()` is only called from `TronNetDelegate.processBlock()` under `synchronized (blockLock)`, using a concurrent set adds no real safety and gives a misleading impression. Replaced with a plain `HashSet` with a comment clarifying the single-threaded access pattern.

**2) Fork rollback guard**

Added a check for `nextMaintenanceTime < lastNextMaintenanceTime` — since maintenance time only increases monotonically under normal conditions, a decrease is a reliable signal that a fork rollback has occurred. In that case, we reinitialize the in-memory state from the current DB instead of diffing against stale data:

```java
if (nextMaintenanceTime < lastNextMaintenanceTime) {
    // Fork rollback detected — reinitialize instead of diffing
    lastActiveWitnesses.clear();
    lastActiveWitnesses.addAll(currentWitnesses);
} else {
    recordSrSetChange(currentWitnesses);
}
lastNextMaintenanceTime = nextMaintenanceTime;
```

Regarding your second comment about moving SR set change detection off-chain entirely — I understand the concern that cross-block state cannot participate in the revoke mechanism, but I believe the rollback guard above addresses the root issue with minimal complexity. Removing the in-process metric would lose the at-a-glance visibility in Grafana dashboards that makes it useful for operators, and pushing the diff logic to an external monitor adds operational overhead (deploying/maintaining a separate component) for an edge case that is already handled. I'd prefer to keep it in-chain with this guard in place.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -164,11 +171,58 @@ public void applyBlock(BlockCapsule block) {
...
+    } else if (nextMaintenanceTime != lastNextMaintenanceTime) {
+      Set<String> currentWitnesses = chainBaseManager.getWitnessScheduleStore().getActiveWitnesses()
+          .stream()
+          .map(w -> StringUtil.encode58Check(w.toByteArray()))
+          .collect(Collectors.toSet());
+      if (nextMaintenanceTime < lastNextMaintenanceTime) {
+        // Fork rollback detected — reinitialize instead of diffing
+        lastActiveWitnesses.clear();
+        lastActiveWitnesses.addAll(currentWitnesses);
+      } else {
```
lvs0075: If a fork rollback does occur, wouldn’t the error message already have been written to the metrics?

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -164,11 +171,58 @@ public void applyBlock(BlockCapsule block) {
...
-      MetricsUtil.meterMark(MetricsKey.BLOCKCHAIN_TPS, block.getTransactions().size());
-      Metrics.counterInc(MetricKeys.Counter.TXS, block.getTransactions().size(),
+    int txCount = block.getTransactions().size();
+    if (txCount > 0) {
+      MetricsUtil.meterMark(MetricsKey.BLOCKCHAIN_TPS, txCount);
+      Metrics.counterInc(MetricKeys.Counter.TXS, txCount,
           MetricLabels.Counter.TXS_SUCCESS, MetricLabels.Counter.TXS_SUCCESS);
     }
+    // Record transaction count distribution for all blocks (including empty blocks)
+    Metrics.histogramObserve(MetricKeys.Histogram.BLOCK_TRANSACTION_COUNT, txCount,
```
halibobo1205: [SHOULD] If the primary goal is empty block monitoring, a simple Counter or Gauge would suffice. A Histogram is better suited when transaction count distribution analysis is actually needed — it can always be upgraded later.

warku123: Actually, we have disscussed this question in the issue https://github.com/tronprotocol/java-tron/issues/6590, and finally decide to implement `Histogram` instead of the simple `Counter`

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
warku123: @bladehan1 @lvs0075 You are absolutely right: the rollback guard can only re-calibrate the in-memory baseline for subsequent blocks; **it cannot retract Prometheus counters that have already been incremented**. This is an inherent limitation because in-process metrics do not participate in the chain's revoking mechanism.

That said, this scenario — a fork rollback that crosses a maintenance boundary — is **extremely rare in practice** because TRON's maintenance interval is 6 hours and deep reorganizations of that magnitude are uncommon on the mainnet.

From an observability perspective, the recorded add/remove events are not necessarily "false positives" either: they **reflect real SR set transitions that occurred on the abandoned chain at the time**. When a fork rollback does cross a maintenance boundary, the chain genuinely experienced a different SR set during that (later invalidated) period. The metric captures that transient state change, even though the chain eventually switched to another fork.

In other words, `SR_SET_CHANGE` behaves like an **event log of witnessed transitions** rather than a guaranteed final-state snapshot. For operators, seeing an add followed by a remove (or vice versa) after a rollback is an indicator that a reorganization crossed the maintenance boundary.

This is a semantic trade-off. If the project prefers metrics to strictly mirror only the final canonical chain state, the only robust solution is to move SR change detection off-chain as @bladehan1 suggested. Otherwise, we can accept this limitation and document it clearly. Please let me know which direction the team prefers.

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
halibobo1205: The SR set change detection logic (reading `WitnessScheduleStore`, stream mapping, `encode58Check `encoding, `HashSet` diff) runs on every block regardless of whether metrics are enabled. The `Metrics.enabled()` guard inside `counterInc()/histogramObserve()` only prevents the final write — the upstream computation still executes unconditionally.

Suggest wrapping the entire SR detection block and the histogram observation (including the `encode58Check` call) with a `Metrics.enabled()` check to avoid unnecessary computation when metrics are disabled.

warku123: Good catch. Wrapped the entire block (histogram observation + SR set change detection) with a `Metrics.enabled()` guard in commit 7791944. This avoids the DB reads, `encode58Check` encoding, stream mapping, and HashSet diff on every block when Prometheus is disabled.

Note that the existing metrics in the same method (e.g., `MINER_LATENCY` histogram + `encode58Check` at L151-152) have the same issue, but I'll leave those untouched to keep this PR scoped to the new metrics only.

halibobo1205 (approved):

framework/src/main/java/org/tron/core/metrics/blockchain/BlockChainMetricManager.java
```diff
@@ -164,11 +164,17 @@ public void applyBlock(BlockCapsule block) {
...
-    if (block.getTransactions().size() > 0) {
-      MetricsUtil.meterMark(MetricsKey.BLOCKCHAIN_TPS, block.getTransactions().size());
-      Metrics.counterInc(MetricKeys.Counter.TXS, block.getTransactions().size(),
+    int txCount = block.getTransactions().size();
+    if (txCount > 0) {
+      MetricsUtil.meterMark(MetricsKey.BLOCKCHAIN_TPS, txCount);
+      Metrics.counterInc(MetricKeys.Counter.TXS, txCount,
           MetricLabels.Counter.TXS_SUCCESS, MetricLabels.Counter.TXS_SUCCESS);
     }
+    if (Metrics.enabled()) {
```
SeriousCoding789: I think this `if (Metrics.enabled()) {` is no long needed, or put it at first line of this function?

warku123: Thanks for the review. Two reasons I'd like to keep this guard where it is:

1. **Not redundant** — the histogram label `StringUtil.encode58Check(address)` runs a sha256 checksum, and Java evaluates it before `histogramObserve` is even entered. The internal `Metrics.enabled()` check inside `observe()` only short-circuits the Prometheus call, not the argument evaluation. Wrapping the call site is what actually saves the per-block sha256 work when metrics are disabled.

2. **Can't move to the first line** — the earlier part of `applyBlock` maintains `witnessInfo` / `dupWitnessBlockNum`, which are read by `getDupWitness()` to populate `BlockChainInfo`. That's non-metric state and must run regardless of `Metrics.enabled()`. An early return at the top of the function would skip the duplicate-witness bookkeeping.

So the guard is intentionally local to the histogram call.

common/src/main/java/org/tron/common/prometheus/MetricsCounter.java
warku123: @halibobo1205 Added a comment at the `SR_SET_CHANGE` constant definition in `MetricKeys.java` (commit 96bf315) explaining the cardinality rationale:

> witness label: bounded cardinality -- SR candidate pool is finite, rotation is infrequent (at most once per maintenance interval); kept for at-a-glance SR identification in dashboards rather than requiring log cross-referencing.

Sunny6889 (approved):

halibobo1205 (approved):
TODO
1. Adding a METRICS_CHANGELOG.md to the repo to track metric additions, modifications, and deprecations — future PRs can maintain it as well
2. Including a metrics change entry in release notes
3. Filing a follow-up issue/TODO to update the [official metrics documentation](https://tronprotocol.github.io/documentation-en/using_javatron/metrics/) and the [Grafana dashboard template](https://grafana.com/grafana/dashboards/16567-java-tron-server/)

PR #6625, opened 2026-04-01
refactor(toolkit): move keystore-factory to plugins module

Federico2014:
**What does this PR do?**

  Moves the keystore-factory functionality from the toolkit to the plugins module, and relocates keystore utility classes to the crypto module for better modularity.

  This refactoring improves the project structure by:
  1. Moving `KeystoreFactory.java` from `framework/src/main/java/org/tron/program` to `plugins/src/main/java/common/org/tron/plugins` - aligns with the plugins-based architecture
  2. Relocating keystore utility classes (`Credentials.java`, `Wallet.java`, `WalletFile.java`, `WalletUtils.java`) from `framework` to `crypto` module - better separation of concerns
  3. Updating all import statements across the codebase to reflect the new locations

  **Changes:**
  - 27 files changed, 119 insertions(+), 77 deletions(-)
  - Keystore classes moved to `crypto/src/main/java/org/tron/common/crypto/keystore/`
  - KeystoreFactory moved to `plugins` module
  - Updated imports in Wallet.java, Args.java, FullNode.java, and test files
  - Added plugins/README.md documentation

  - Unit Tests
  - Manual Testing

  **Follow up**

  None - this is a pure refactoring with no behavioral changes.

  **Extra details**

  This change is part of the ongoing effort to modularize the java-tron codebase and improve separation between core framework components and plugin-based features.


vividctrlalt:
The PR description says "pure refactoring with no behavioral changes", but there are two behavioral changes:

**1. Hardcoded `isECKeyCryptoEngine` breaks SM2 support**

`Wallet.decrypt()` and `WalletUtils.generateNewWalletFile()` replaced `Args.getInstance().isECKeyCryptoEngine()` with hardcoded `true`. But `KeystoreFactory` in plugins still reads the config via `CommonParameter.getInstance().isECKeyCryptoEngine()`. This creates a round-trip failure: an SM2 wallet generated by KeystoreFactory cannot be decrypted by `Wallet.decrypt()`, because decrypt always assumes ECKey.

If crypto module can't depend on Args, the boolean should be passed as a parameter rather than hardcoded.

**2. `--keystore-factory` CLI flag removed without deprecation**

Operators using `java -jar FullNode.jar --keystore-factory` will get an unrecognized option error after upgrading. The migration path (use `Toolkit.jar keystore-factory` instead) is not documented. Also, `CLIParameter.keystoreFactory` field is still declared but never read — dead code.

PR #6627, opened 2026-04-01
refactor(crypto): remove SM2/SM3 crypto algorithm support

Federico2014:
**What does this PR do?**
  This PR removes SM2/SM3 cryptographic algorithm support from the java-tron codebase:
  - Deletes SM2.java and SM2Signer.java implementation files
  - Removes cryptoEngine configuration option from CommonParameter and Args
  - Simplifies Sha256Hash to SHA-256 only (removes SM3 support)
  - Removes isECKeyCryptoEngine parameter from SignUtils methods
  - Updates 70+ files to use simplified crypto APIs

 SM2/SM3 algorithms are not used in the TRON network, which relies exclusively on ECKey (SECP256K1) for cryptographic operations. This cleanup removes technical debt and simplifies the crypto module by making ECKey the explicit, only supported crypto engine.

  - Build verification: `./gradlew build -x test` passed successfully
  - Checkstyle verification on framework and plugins modules

  **Follow up**
  None required. This is a one-time cleanup with no functional changes.

  **Extra details**
  - 75 files changed: 207 insertions(+), 2596 deletions(-)
  - 4 files deleted entirely


vividctrlalt:
The benefit of this change is minimal — `isECKeyCryptoEngine` is always `true` and has no impact on runtime behavior. Removing it only saves one parameter per call site.

However, the cost is significant: 75 files changed across core modules for a purely cosmetic cleanup. The SM2/SM3 support was designed as a crypto extension point and may still have value as a reserved abstraction. I'd suggest leaving it as-is.

PR #6631, opened 2026-04-01
feat(protocol): add protoLint check for enum validation

### LINKED ISSUE #6515: AccountCapsule` and `Wallet` frozenV2 construction logic
> Issue reproduced on current `develop` branch.

---

#### Expected behaviour

For the API `walletsolidity/getaccount`, the returned `frozenV2` field should:

1. Contain **only entries with explicit `type`**
2. Have **correct `amount` bound to each `type`**
3. Be returned in a **stable and logical order**, for example:
   - `ENERGY`
   - `TRON_POWER`

Expected example:

```json
"frozenV2": [
  {
    "type": "ENERGY",
    "amount": 20400000000
  },
  {
    "type": "TRON_POWER",
    "amount": 20000000000
  }
]
Actual behaviour

The walletsolidity/getaccount API returns an unexpected frozenV2 structure:

"frozenV2": [
  {
    "amount": 20000000000
  },
  {
    "type": "ENERGY",
    "amount": 20400000000
  },
  {
    "type": "TRON_POWER"
  }
]


Problems observed:

The first entry has no type field

The TRON_POWER entry has no amount

The order of entries is inconsistent and does not match logical resource grouping

amount values appear to be mis-associated across entries

This makes it difficult for API consumers to reliably parse frozenV2.

Frequency

100% (always reproducible for this account state)

Steps to reproduce the behaviour

Run a FullNode / SolidityNode using the develop branch

Ensure an account has:

Stake V2 frozen balances for ENERGY and TRON_POWER

Call the API:

walletsolidity/getaccount


Observe the returned frozenV2 field

Full response example
{
  "account_name": "TestD",
  "address": "TBcBVYnotCt6CtRyBiwnaALjMyoypXH8Wj",
  "balance": 999934031076964,
  "net_usage": 247,
  "latest_opration_time": 1702994064000,
  "free_net_usage": 1403,
  "latest_consume_time": 1702994064000,
  "latest_consume_free_time": 1702991967000,
  "net_window_size": 28271000,
  "net_window_optimized": true,
  "account_resource": {
    "latest_consume_time_for_energy": 1702993065000,
    "energy_window_size": 28800000,
    "energy_window_optimized": true
  },
  "frozenV2": [
    {
      "amount": 20000000000
    },
    {
      "type": "ENERGY",
      "amount": 20400000000
    },
    {
      "type": "TRON_POWER"
    }
  ]
}

Backtrace
N/A

Additional notes

This issue appears to be related to how frozenV2 entries are constructed and ordered
when converting account data to RPC response objects (likely in Wallet RPC layer),
rather than an issue with the underlying AccountCapsule storage itself.

Explicit type assignment and deterministic ordering of frozenV2 entries
would greatly improve API consistency.

### PR DESCRIPTION
**Pull request by 0xbigapple**
**What does this PR do?**

   - Enforce `UNKNOWN_` prefix for Enum index 0 to ensure JSON serialization compatibility.
   - Included a legacyEnums Set  to make sure  the CI passes for old code.
   - Bound to the standard check task (`check.dependsOn protoLint`) which run automatically during CI and local full builds. Also support run manually with ` ./gradlew :protocol:protoLint` or `./gradlew :protocol:check`.
   - Add multi-platform buf (v1.61.0) checksums to verification-metadata.xml.

close https://github.com/tronprotocol/java-tron/issues/6515

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

- FAIL OUTPUT:

```
> Task :protocol:protoLint FAILED
🔍 Generating Proto AST image using buf CLI...

❌ [Protocol Design Violation] The following enums violate the java-tron API evolution standard:
  - [core/contract/common.proto] Enum "ResourceCode" has index 0: "BANDWIDTH". It MUST start with "UNKNOWN_".
  - [core/Tron.proto] Enum "Permission.PermissionType" has index 0: "Owner". It MUST start with "UNKNOWN_".

```

- SUCCESS OUTPUT:

```
> Task :protocol:protoLint
🔍 Generating Proto AST image using buf CLI...
✅ Proto Enum validation passed!
```

### CODE REVIEW

**Comment by vividctrlalt** [comment]
Nice work — this solves a real problem and the implementation is solid. A few suggestions:

**Should fix:**

1. **`org.gradle.internal.os.OperatingSystem`** (protoLint.gradle:15) — This is a Gradle internal API with no stability guarantee across major versions. The root `build.gradle` already uses `System.getProperty("os.name")` for platform detection, which is a stable public API. Suggest using the same approach here.

2. **buf stderr not captured** (protoLint.gradle:91-98) — When buf fails, the error message only shows the exit code. All diagnostic output (proto syntax errors, compilation failures) is lost. Adding `standardOutput`/`errorOutput` capture to the `exec` block would save developers from having to rerun buf manually to debug.

**Worth considering:**

3. **Incremental build inputs incomplete** (protoLint.gradle:43-44) — `inputs` doesn't include the buf version or `build/extracted-include-protos/` directory. Upgrading buf or changing external proto dependencies may not invalidate the task cache. Adding `inputs.property('bufVersion', bufVersion)` would fix the version case.

4. **Legacy whitelist maintenance** — The 21-entry whitelist is hardcoded with no documented update process. A new contributor whose enum fails validation won't know this list exists. A comment pointing to the issue or a note in CONTRIBUTING.md would help.

5. **Enum vs message detection** (protoLint.gradle:133) — `definition.value != null` as the discriminator works but is undocumented. A brief comment explaining why this is correct in buf's JSON AST format would help future maintainers.

6. **`check.dependsOn protoLint`** (protoLint.gradle:167) — This runs proto validation on every `gradle build`, including local dev builds that don't touch proto files. Consider `onlyIf { providers.environmentVariable('CI').present }` or scoping it to proto-change detection so local iteration stays fast.

7. **v1beta1 config format** (protoLint.gradle:89) — The inline buf config uses a beta format. A static `buf.yaml` file in the protocol/ directory would be more forward-compatible and auto-detected by buf.

**Comment by 0xbigapple** [comment]
@vividctrlalt Thanks for the detailed review! I really appreciate the feedback and the time you took to look into this. I've considered the points you raised and would like to share my thoughts for discussion:

### Regarding the "Should fix" items:

1. **`OperatingSystem`**: I understand the concern about using a Gradle internal API. However, `org.gradle.internal.os.OperatingSystem` is already being used in the root `build.gradle` and `framework/build.gradle`. I leaned towards keeping it here to maintain a consistent approach to platform detection throughout the project.
2. **buf stderr not captured**: By default, Gradle's `exec` block pipes all `stdout` and `stderr` directly to the console, so any diagnostic details (like syntax errors) from `buf` are actually visible in the build logs. The reason I set `ignoreExitValue = true` was to prevent Gradle from immediately throwing its generic "non-zero exit value" error. This allows the script to catch the failure and throw a more descriptive `GradleException` with a helpful hint ("Ensure your .proto files are valid"), while still preserving the full `buf` output in the console.

### Regarding the "Worth considering" items:

3. **Incremental build inputs**: Since `bufVersion` is defined directly inside `protoLint.gradle`, and the script itself is registered as an input (`inputs.file('protoLint.gradle')`), any change to the version will modify the script and correctly invalidate the cache. As for external protos, since this linting is strictly focused on our internal `java-tron` enum standards, not triggering a rerun on external dependency changes seems like the right boundary.
4. **Legacy whitelist maintenance**: I've included a link to the original design issue (#6515) in the header comments at the top of the file. Hopefully, this provides the necessary historical context for any new contributor who needs to understand the standard.
5. **Enum vs message detection**: Great suggestion! **I have added the comments** in the lasted commit.
6. **`check.dependsOn protoLint`**: Thanks to the `inputs` and `outputs` configuration, Gradle's incremental build (`UP-TO-DATE`) feature will skip this task entirely if no proto files or the script itself have been modified. Therefore, it shouldn't add any overhead to everyday local dev builds.
7. **v1beta1 config format**: I agree that a static `buf.yaml` is the standard approach. I kept the Buf config inline to avoid introducing an extra config file for a narrowly scoped rule. Maybe this should be revisited if we later upgrade Buf or standardize Buf usage more broadly in the repository.

I hope these explanations clarify the design choices! Let me know what you think about these points, and I'm happy to discuss further if you feel strongly about any of them.

**Comment by vividctrlalt** [comment]
Thanks for the thorough responses — your points on #2 (stderr default passthrough) and #3 (script-as-input covers version changes) are correct. I agree with all your explanations.

Two suggestions going forward:

1. **Internal API (`org.gradle.internal.os.OperatingSystem`)** — Keeping it consistent with the rest of the project makes sense for now. Suggest a separate PR in the future to unify platform detection across all Gradle scripts to public APIs, reducing risk if Gradle drops this internal class in a future major version.

2. **Extract legacy enum whitelist to a separate file** — The 21-entry whitelist in the Gradle script works, but as it grows or changes, a dedicated file (e.g. `protocol/legacy-enums.txt`) would be easier to maintain and review. Could be done in this PR if straightforward, otherwise as a follow-up.

**Review by RomanReohord69 — APPROVED** [review]

File: protocol/protoLint.gradle [hunk 1]
```diff
@@ -0,0 +1,169 @@
...
+task protoLint {
+    group = "verification"
+    description = "Validate Protobuf Enums using buf generated JSON AST. Enforces 'UNKNOWN_' prefix for index 0 to ensure JSON serialization backward compatibility."
+    
+    // Explicitly depend on:
+    // 1. extractIncludeProto: ensure external protos are available in build/extracted-include-protos/main.
+    // 2. generateProto: fix Gradle implicit dependency warning due to output directory overlap.
+    dependsOn 'extractIncludeProto', 'generateProto'
+
+    // Incremental build support: Only run if proto files or the script itself changes
```
waynercheung: `protoLint` does not only depend on `src/main/protos` and `protoLint.gradle`; it also depends on the extracted include protos consumed by buf below via `build/extracted-include-protos/main`.

`dependsOn 'extractIncludeProto'` only guarantees execution order. It does not make the extracted include directory part of `protoLint`'s input snapshot, so Gradle can still consider this task `UP-TO-DATE` when those files change.

I reproduced this locally on the current PR head:
1. Run `./gradlew :protocol:protoLint`
2. Touch a file under `protocol/build/extracted-include-protos/main`
3. Run the same command again

`:protocol:protoLint` is still reported as `UP-TO-DATE`.

Even if the lint rule only validates repo-owned enums, these include protos still participate in buf image generation and import resolution, so changes there can affect task correctness (including whether buf succeeds at all). Please add the extracted include directory as an input, ideally wired from the actual `extractIncludeProto` output rather than a hardcoded path.

protocol/protoLint.gradle
```diff
@@ -0,0 +1,169 @@
...
+        ].collect { it.toString() } as Set
+
+        // 2. Build JSON AST Image using buf CLI
+        def imageDir = file("${buildDir}/tmp/buf")
+        def imageFile = file("${imageDir}/proto-ast.json")
+        imageDir.mkdirs()
+        
+        println "🔍 Generating Proto AST image using buf CLI..."
+        
+        def bufConfig = '{"version":"v1beta1","build":{"roots":["src/main/protos","build/extracted-include-protos/main"]}}'
```
waynercheung: This hardcodes `build/extracted-include-protos/main` instead of deriving it from `buildDir` or from the actual `extractIncludeProto` task output.

It works with the current layout, but it couples this task to a specific plugin output path. Since `protoLint` already depends on `extractIncludeProto`, could we source the include root from that task or at least build it from `buildDir`?

protocol/protoLint.gradle
```diff
@@ -0,0 +1,169 @@
...
+    doLast {
+        def bufExe = configurations.bufTool.singleFile
+        if (!bufExe.exists() || !bufExe.canExecute()) {
+            bufExe.setExecutable(true)
+        }
+
+        // 1. Legacy Whitelist
+        // Contains enums that existed before the 'UNKNOWN_' standard was enforced.
+        // Format: "filename.proto:EnumName" or "filename.proto:MessageName.EnumName"
+        def legacyEnums = [
```
waynercheung, line 68: Optional follow-up: I agree with the earlier suggestion (from @vividctrlalt) to move this whitelist into a dedicated data file if that is straightforward in this PR, or track it as a follow-up otherwise.

The current list is still manageable, but once it grows further it will be easier to review and maintain as data (for example `protocol/legacy-enums.txt`) rather than embedded script logic.

protocol/protoLint.gradle
0xbigapple: You're right. Even though protoLint only validates our own enums, buf physically reads the extracted include protosfor import resolution and image generation. If those files change and the task is incorrectly cached as UP-TO-DATE, we'd get a false success. A task's declared inputs should reflect everything it actually reads — not just what it logically validates.

Fixed. The include directory is now declared as an input, and the path is derived from extractIncludeProto.destDir rather than hardcoded:

```
   def extractTask = tasks.named('extractIncludeProto').get()
   def includeProtoDir = extractTask.destDir.get().asFile
   def includeProtoDirRel = projectDir.toPath().relativize(includeProtoDir.toPath()).toString()
   inputs.dir(includeProtoDir)
```

protocol/protoLint.gradle
0xbigapple: Fixed. The path is now derived from `extractIncludeProto.destDir` and the buf config is built via `JsonBuilder` instead of string interpolation, which also eliminates the backslash escaping issue on Windows.

protocol/protoLint.gradle
0xbigapple: Thanks for the suggestion. This list is not expected to grow — it exists solely for historical enums that predate the `UNKNOWN_` convention. All new enums must follow the standard and should never be whitelisted. Given that, I think keeping it inline is sufficient.

protocol/protoLint.gradle
```diff
@@ -0,0 +1,178 @@
...
+// Depend on the buf executable published on Maven Central
+dependencies {
+    bufTool "build.buf:buf:${bufVersion}:${platform}-${machine}@exe"
+}
+
+task protoLint {
+    group = "verification"
+    description = "Validate Protobuf Enums using buf generated JSON AST. Enforces 'UNKNOWN_' prefix for index 0 to ensure JSON serialization backward compatibility."
+    
+    // Explicitly depend on:
```
waynercheung, line 38: Nit: this comment still mentions the concrete `build/extracted-include-protos/main` path, but the implementation now derives the include root from `extractIncludeProto.destDir`.

Suggested wording:
```
// Explicitly depend on:
// 1. extractIncludeProto: ensure external protos are extracted before buf runs.
//    The include root is derived from that task's actual output below.
// 2. generateProto: fix Gradle implicit dependency warning due to output directory overlap.
```

halibobo1205 (approved):

protocol/protoLint.gradle
0xbigapple: Thanks for catching this! Updated the comment in the latest commit.

waynercheung (approved):
LGTM

protocol/protoLint.gradle
```diff
@@ -0,0 +1,179 @@
...
+
+        // 4. Report Results
+        if (!errors.isEmpty()) {
+            println "\n❌ [Protocol Design Violation] The following enums violate the java-tron API evolution standard (Issue #6515):"
+            errors.each { println "  - $it" }
+            throw new GradleException("Proto Enum validation failed. See above for details.")
+        } else {
+            println "✅ Proto Enum validation passed!"
+            // Update marker file for Gradle incremental build cache
+            markerFile.text = "Success"
```
317787106, line 174: Suggestion: keep silent on successful compilation, consistent with other modules. Only output logs when errors occur.

lvs0075 (approved):
LGTM

protocol/protoLint.gradle
0xbigapple: Thanks for the suggestion! This is a reasonable point. I understand the benefits of the "silent on success" pattern. However, I'd prefer to keep the output in this case:
  1. User Experience for a New Validation Step
    - New developers (or in CI) see the output once when they first clone or when proto files change, visible output confirms it's running.
  2. Minimal Console Noise
    - Just 2 log lines—very light overhead
    - Once cached, subsequent gradle build calls skip protoLint entirely with UP-TO-DATE

PR #6635, opened 2026-04-02
feat(ci): add coverage check workflow and auto reviewer assignment

bladehan1:
### What does this PR do?

- Introduces a `base vs PR` coverage comparison gate in PR CI to prevent unnoticed regression of overall coverage.

- Aggregates multi-module `jacocoTestReport.xml` using `madrapps/jacoco-report` (without using the Gradle aggregation plugin).

- Adds/adjusts the coverage task pipeline:
  - `coverage-base`: generates base coverage using `pull_request.base.sha`
  - `coverage-pr` (reuses Debian11 unit test results): generates PR coverage
  - `coverage-gate`: performs gate comparison and outputs results

- Gate rules:
  - `changed files coverage > 60%`
  - `pr_overall - base_overall >= -0.1%`

- Standardizes coverage execution environment to `Debian11 + JDK8 (x86_64)`, executed in parallel to avoid increasing wall-clock time significantly.

- Disables automatic PR comments (`update-comment: false`) to reduce noise, while retaining logs and step summary outputs.

- Removes the `macos` x86 runner (`macos-26-intel`) execution item to reduce the impact of occasional failures in this environment on CI stability.

- Adds auto reviewer assignment workflow (`pr-reviewer.yml`) that assigns reviewers based on PR title scope (conventional commit format), replacing CODEOWNERS.

- Disables Gradle build cache (`--no-build-cache`) selectively for coverage jobs to ensure accurate coverage data.

### Why are these changes required?

- Passing a static threshold does not guarantee that the current changes did not introduce regression.

- Compared to maintaining historical baseline storage (artifact/JSON state), base-vs-PR comparison under the same conditions is lighter, more practical, and has lower maintenance cost.

- Avoids using Gradle aggregation to prevent potential conflicts in multi-module class/exec merging; instead, uses Action-side XML aggregation for better stability.

- Auto reviewer assignment ensures consistent, scope-based review coverage without relying on CODEOWNERS.

### `--no-build-cache` Strategy

**Problem:** GitHub Actions `actions/cache` persists `~/.gradle/caches/` (including `build-cache-1/`) across workflow runs. When the Gradle build cache is restored, test tasks are marked as `FROM-CACHE` and skipped entirely — no tests execute, and no `.exec` files are generated. This was observed in [run #23942322694](https://github.com/tronprotocol/java-tron/actions/runs/23942322694) where debian11, rockylinux, and coverage-base completed in ~1 min instead of 20+ min.

**Root cause:** Gradle build cache keys are computed from task inputs (source code, JDK version, JVM args, dependencies). When inputs are identical across runs, cached test outputs (including `.exec` files) are reused. While the cached data is technically valid for identical inputs, it means tests are not actually executed and coverage data comes from a previous run.

**Solution — selective `--no-build-cache`:**

| Job | `--no-build-cache` | Reason |
|-----|-------------------|--------|
| macos26 | ✗ | Verification only — build cache speeds up CI, no coverage needed |
| ubuntu24 | ✗ | Verification only — same as above |
| rockylinux | ✗ | Verification only — same as above |
| **debian11** | ✓ | Must generate fresh `.exec` files for PR coverage |
| **coverage-base** | ✓ | Must generate fresh `.exec` files for base coverage |

**Cache isolation:** `coverage-base` uses an independent cache key prefix (`coverage-base-x86_64-gradle-`) to prevent cross-contamination with `debian11`'s dependency cache.

**Note:** `--no-build-cache` only disables Gradle's task output cache. Dependency download cache (`modules-2/`) and Gradle wrapper cache remain active, so dependency resolution is still fast.

### Testing

- Unit Tests
- Manual Testing
- Coverage gate scenario validation (4 cases):
  - overall `< 60%`
  - overall `> 60%`
  - delta `< -0.1%`
  - delta `>= -0.1%`
- Validation points:
  - changed files gate pass/fail
  - delta gate pass/fail
  - output includes `Changed/PR/Base/Delta` and is readable
- Auto reviewer workflow validated on fork repository PR
- Verified `--no-build-cache` fix: confirmed test tasks execute fully (not `FROM-CACHE`) and `.exec` files are freshly generated

**Note:** `-0.1` means `-0.1%` (not `-1%`)

### Auto Reviewer Behavior

Reviewer assignment follows a "first-assign-only" policy — the workflow only assigns reviewers when no reviewers are currently requested on the PR. Once reviewers are assigned (either automatically or manually), subsequent title edits or reopens will not trigger reassignment. This design preserves manually added reviewers and avoids unintended accumulation from title scope changes.

### Follow Up

Future optimization: refine trigger conditions from `paths-ignore` to whitelist-based (`paths` / `paths-filter`) to reduce unnecessary runs.

### Extra Details

- Key workflow changes: `.github/workflows/pr-build.yml`, `.github/workflows/pr-reviewer.yml`
- References:
  - JaCoCo: https://www.jacoco.org/jacoco/trunk/index.html
  - JaCoCo Docs: https://www.jacoco.org/jacoco/trunk/doc/
  - madrapps/jacoco-report: https://github.com/madrapps/jacoco-report


Sunny6889 (approved):

.github/workflows/pr-cancel.yml
```diff
@@ -37,7 +37,11 @@ jobs:
                 );
 
                 for (const run of runs) {
-                  const isTargetPr = !run.pull_requests?.length || run.pull_requests.some((pr) => pr.number === prNumber);
+                  if (!run) {
+                    continue;
+                  }
+                  const prs = Array.isArray(run.pull_requests) ? run.pull_requests : [];
+                  const isTargetPr = prs.length === 0 || prs.some((pr) => pr.number === prNumber);
```
halibobo1205, line 44: Is this change intended to fix the following error?
```
Run actions/github-script@v8
TypeError: Cannot read properties of undefined (reading 'pull_requests')
Error: Unhandled error: TypeError: Cannot read properties of undefined (reading 'pull_requests')
at eval (eval at callAsyncFunction (/home/runner/work/_actions/actions/github-script/v8/dist/index.js:36187:16), <anonymous>:23:31)
at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
at async main (/home/runner/work/_actions/actions/github-script/v8/dist/index.js:36285:20)
```

.github/workflows/pr-build.yml
```diff
@@ -181,3 +174,232 @@ jobs:
...
+
+      - name: Upload PR coverage reports
+        uses: actions/upload-artifact@v6
+        with:
+          name: jacoco-coverage-pr
+          path: |
+            **/build/reports/jacoco/test/jacocoTestReport.xml
+          if-no-files-found: error
+
+  coverage-base:
```
halibobo1205, line 189: The `coverage-base` job runs a full `clean build` + `testWithRocksDb` + `jacocoTestReport` on the base branch for every PR. The Gradle cache (`actions/cache@v4`) only caches downloaded dependencies (JARs from Maven Central, etc.), **not** the compiled output (`build/` directory), so the compilation and test execution times are not reduced.

This effectively **doubles the CI time and resource consumption** per PR. Consider optimizing by caching the base branch coverage report (e.g., uploading it to an artifact or cache after each merge to `develop`), so that PRs can compare against the cached baseline instead of rebuilding the base branch from scratch every time.

.github/workflows/pr-cancel.yml
bladehan1: Yes, this PR fixes that error. The `pr-cancel.yml` now includes null handling for `pull_requests` to prevent the `TypeError`.

.github/workflows/pr-build.yml
bladehan1: The CI wall-clock time does **not** double — `coverage-base` runs as a parallel job alongside the existing build/test jobs, so the overall pipeline duration remains roughly the same.

The resource consumption does increase by one additional job, but it's not a full doubling since it only runs unit tests + coverage on a single environment (Debian11 + JDK8).

Regarding caching the base branch coverage report:
1. This would require building a **commit-level coverage storage system** (e.g., upload/download artifacts keyed by commit SHA), and we'd also need to handle **concurrency issues** (multiple PRs merging to develop simultaneously, cache invalidation, etc.).
2. Given that the current approach only adds one parallel job for unit test coverage without increasing wall-clock time, the simplicity trade-off is worthwhile. We can revisit this optimization if resource consumption becomes a concern.

.github/workflows/pr-build.yml
```diff
@@ -177,7 +170,236 @@ jobs:
...
+          if [ -z "$BASE_XMLS" ] || [ -z "$PR_XMLS" ]; then
+            echo "Missing jacocoTestReport.xml files for base or PR."
+            exit 1
+          fi
+          echo "base_xmls=$BASE_XMLS" >> "$GITHUB_OUTPUT"
+          echo "pr_xmls=$PR_XMLS" >> "$GITHUB_OUTPUT"
+
+      - name: Aggregate base coverage
+        id: jacoco-base
+        uses: madrapps/jacoco-report@v1.7.2
```
halibobo1205, line 284: If `madrapps/jacoco-report` is replacing Codecov, suggest:

1. Delete `codecov.yml` in this PR to avoid confusion
2. Remove the `Codecov` badge from `README` — since coverage data is no longer being uploaded, https://app.codecov.io/gh/tronprotocol/java-tron will never update again, keeping a stale badge seems pointless

bladehan1: Done in 4be3316:

1. Removed the Codecov badge from `README.md`
2. Added a `DEPRECATED` header to `codecov.yml` — retained the file for reference rather than deleting it outright, so the historical configuration remains visible. It can be safely deleted in a follow-up cleanup.

halibobo1205 (approved):

aleksola (approved):

PR #6656, opened 2026-04-08
feat(vm): implement TIP-7939 CLZ opcode

yanghang8612:
## Summary

Implement TIP-7939 CLZ opcode (`0x1e`), gated behind the existing `allowTvmOsaka` flag.

- New opcode: pops 1 value, pushes the number of leading zero bits (`256` if zero)
- Gas cost: 5 (`LOW_TIER`, matching MUL)
- Added `TRON_V1_5` version with `appendOsakaOperations` to support Osaka opcodes
- Test cases cover all TIP vectors, energy cost verification, `[128, 255]` unsigned-cast cases, and rejection before Osaka activation
- Scope: java-tron node-side opcode support only; compiler/tooling emission and `osaka` version signaling are expected to be handled separately

Related: https://github.com/tronprotocol/tips/issues/838

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/OperationActions.java
+++ b/actuator/src/main/java/org/tron/core/vm/OperationActions.java
@@ -2,6 +2,7 @@ package org.tron.core.vm;
 
 import static org.tron.common.crypto.Hash.sha3;
 import static org.tron.common.utils.ByteUtil.EMPTY_BYTE_ARRAY;
+import static org.tron.common.utils.ByteUtil.numberOfLeadingZeros;
 
 import java.math.BigInteger;
 import java.util.ArrayList;
@@ -284,6 +285,17 @@ public class OperationActions {
 
     final DataWord result = word2.shiftRightSigned(word1);
     program.stackPush(result);
+    program.step();
+  }
+
+  public static void clzAction(Program program) {
+    DataWord word = program.stackPop();
+    int clz = numberOfLeadingZeros(word.getData());
+    if (clz == 256) {
+      program.stackPush(new DataWord(256));
+    } else {
+      program.stackPush(DataWord.of((byte) clz));
+    }
     program.step();
   }
 
--- a/framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
+++ b/framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
@@ -77,6 +77,7 @@ public class OperationsTest extends BaseTest {
     VMConfig.initAllowTvmIstanbul(0);
     VMConfig.initAllowTvmLondon(0);
     VMConfig.initAllowTvmCompatibleEvm(0);
+    VMConfig.initAllowTvmOsaka(0);
   }
 
   @Test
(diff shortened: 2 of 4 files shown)


actuator/src/main/java/org/tron/core/vm/OperationActions.java
```diff
@@ -287,6 +288,17 @@ public static void sarAction(Program program) {
     program.step();
   }
 
+  public static void clzAction(Program program) {
+    DataWord word = program.stackPop();
+    int clz = numberOfLeadingZeros(word.getData());
+    if (clz == 256) {
+      program.stackPush(new DataWord(256));
```
lily309, line 295: Why not define 256 as a static constant, so that whenever `word` is 0, it is pushed directly onto the stack?

yanghang8612: Even if a `256` DataWord constant is defined, it still needs to be cloned when pushed onto the stack. The overhead involved is nearly equivalent to creating a new instance from scratch, so the constant approach was not adopted.

actuator/src/main/java/org/tron/core/vm/OperationActions.java
```diff
@@ -287,6 +288,17 @@ public static void sarAction(Program program) {
...
+    } else {
+      program.stackPush(DataWord.of((byte) clz));
```
CodeNinjaEvan, line 297: Nice work overall! One minor suggestion: since DataWord.of(byte num) directly assigns bb[31] = num, casting clz to (byte) when it's in [128, 255] introduces a subtle signed/unsigned ambiguity. Would it be worth simplifying both branches into one using new DataWord(int), which handles the full range cleanly?

actuator/src/main/java/org/tron/core/vm/OperationActions.java
lily309: Okay, got it. Good job.

actuator/src/main/java/org/tron/core/vm/OperationActions.java
yanghang8612: Thanks for the review! The `(byte)` cast here is safe — `DataWord.of(byte)` assigns the raw bit pattern to `bb[31]`, and `DataWord.value()` reads it back as unsigned via `new BigInteger(1, data)`. This is consistent with how byte arrays are used throughout the codebase (e.g., `DataWord.of((byte) 0xFF)` correctly represents 255). The split avoids the double `ByteBuffer.allocate` overhead in `DataWord(int)` for the common non-zero path.

framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
```diff
@@ -904,6 +904,81 @@ public void testPush0() throws ContractValidateException {
...
+    // CLZ(1) = 255
+    val = new byte[32];
+    val[31] = 0x01;
+    op = buildCLZBytecode(val);
+    program = new Program(op, op, invoke, interTrx);
+    testOperations(program);
+    Assert.assertEquals(new DataWord(255), program.getStack().pop());
+
+    // Verify energy cost = LOW_TIER(5) + PUSH32 cost(3) = 8
+    Assert.assertEquals(8, program.getResult().getEnergyUsed());
```
alan-eth, line 968: The test coverage is really solid — the chosen vectors (0, 1, 255, 256, boundary bits) cover the most critical scenarios nicely! 🎯

One thing that might be worth adding: a couple of test cases where the CLZ result falls in the 128-255 range (e.g., `byte[16]=0x01` → CLZ=128, `byte[30]=0x01` → CLZ=247). This would explicitly exercise the `(byte)` cast path for values above 127, giving extra confidence in the signed/unsigned handling. Would make the already-great test suite even more bulletproof!

framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
```diff
@@ -904,6 +904,81 @@ public void testPush0() throws ContractValidateException {
...
+
+    VMConfig.initAllowTvmOsaka(0);
```
alan-eth, line 970: Nice work on properly resetting `VMConfig.initAllowTvmOsaka(0)` in cleanup — shows good testing discipline!

Would it be worth adding a small negative test to verify that CLZ is correctly rejected when `allowTvmOsaka` is disabled? This would serve as a safety net to ensure the feature gate can never be accidentally bypassed. Just a suggestion to make the gating even more robust 🙂

framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
yanghang8612: Thanks — good coverage suggestion. Adding two vectors in the `[128, 255]` band so the `(byte)` cast path is exercised explicitly: `byte[16] = 0x01 → CLZ = 128` and `byte[30] = 0x01 → CLZ = 247`. Between those and the existing 0 / 1 / 255 / 256 edges, the cast handling on values where the sign bit flips is now directly pinned by an assertion rather than being inferred from `DataWord.value()`'s `new BigInteger(1, data)` round-trip.

framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
yanghang8612: Agreed, worth pinning the gate as a hard invariant. Adding a `CLZ before Osaka activation` case that runs with `allowTvmOsaka = 0` and asserts the program halts with the same error TVM raises for any unknown opcode. Keeps the gate from quietly disappearing under a future op-table refactor.

Both this and the `[128, 255]` vectors above will go in together.

CodeNinjaEvan (approved):
LGTM

Federico2014 (approved):

aiden3885 (approved):

CodeNinjaEvan (approved):
LGTM

PR #6658, opened 2026-04-08
feat(http): add SizeLimitHandler to enforce request body size limit

### LINKED ISSUE #6604: [Feature] Unified HTTP Request Body Size Limit
## Problem Statement

　　The current HTTP service layer of java-tron lacks a unified pre-ingress limit for request body size. The issues include: lack of early rejection; validation only occurs after the full request body is read (post-read validation); risk of memory amplification, where oversized requests consume memory even if ultimately rejected; scattered validation logic (e.g., JSON-RPC and broadcasthex use Util.checkBodySize, leading to inconsistent implementation and easy omission); inconsistency between HTTP and gRPC behavior (gRPC already has maxInboundMessageSize, while HTTP lacks an equivalent mechanism).

## Proposed Solution

　　Introduce a unified HTTP body pre-ingress limit at the Jetty layer to enable early rejection, align with gRPC limit strategies, and provide independent configuration capabilities for HTTP and JSON-RPC.

### Specification

　　Core mechanism: Introduce a SizeLimitHandler at the top of the Jetty handler chain, performing size checks during streaming (streaming enforcement). Requests exceeding the limit are rejected immediately without entering the application layer.
　　Unified configuration: Introduce node.http.maxMessageSize and node.jsonrpc.maxMessageSize, which apply independently to HTTP and JSON-RPC respectively; gRPC continues to use node.rpc.maxMessageSize, and the three are independent of each other; the default values of the two new configuration items are set to 4 times node.rpc.maxMessageSize, as HTTP request bodies are generally 2–6 times larger than JSON-RPC, and the average is used here.
　　Application layer adjustment: Remove body size validation responsibility from the application layer; retain Util.checkBodySize as deprecated for compatibility.

**API Changes** (if applicable)

　　Requests with `Content-Length` exceeding the limit return **413 Payload Too Large** before servlet dispatch. For chunked transfer (no `Content-Length`), the behavior depends on the servlet exception handling chain:

| Request type | HTTP Servlets (132) | JSON-RPC Servlet |
|---|---|---|
| `Content-Length` exceeds limit | **413** (Jetty rejects before dispatch) | **413** (same) |
| Chunked, exceeds limit | **200 + error JSON** `{"Error":"BadMessageException"}` | **200 + empty body** |

　　Root cause: `SizeLimitHandler` throws `BadMessageException` (RuntimeException) during streaming body read. HTTP servlets catch it in their own `catch(Exception) → Util.processError()` and write error JSON. `JsonRpcServlet` delegates to jsonrpc4j, which internally catches exceptions during request parsing and returns an empty response.

　　**OOM protection is effective in all cases** — the body read is always truncated at the configured limit regardless of the response status code. Returning 200 + empty body for malformed/oversized chunked JSON-RPC requests is acceptable: it does not affect normal requests, increases attacker difficulty, and the alternative — pre-reading the request body to trigger the size limit (catch the exception to report the error), then wrapping the already-read body into an `HttpServletRequestWrapper` to continue the normal request flow — adds significant complexity for marginal benefit.

**Configuration Changes** (if applicable)
　　New configuration items: node.http.maxMessageSize, node.jsonrpc.maxMessageSize; node.rpc.maxMessageSize remains unchanged and is only used for gRPC.

## Scope of Impact

**Breaking Changes**
　　Requests with `Content-Length` exceeding the configured protocol limits (node.http.maxMessageSize or node.jsonrpc.maxMessageSize) will be rejected early (413), and calls relying on large request bodies will no longer be available; mitigation: adjust configurations or split requests.

**Backward Compatibility**
　　Behavior is controlled independently per protocol configuration, allowing flexible operational tuning; Util.checkBodySize is temporarily retained (deprecated).

## Implementation

　　Implementation: Construct SizeLimitHandler in HttpService.start() or its subclass methods. Select the corresponding limit based on request path or entry type (HTTP uses node.http.maxMessageSize, JSON-RPC uses node.jsonrpc.maxMessageSize), and insert it at the very beginning of the handler chain; it must execute before servlet/filter, enforce streaming processing to avoid full buffering, and intercept at an early connection stage.

　　Yes.

**Estimated Complexity**
　　Low → Medium, changes are concentrated at the HTTP ingress layer, involving multi-entry limit selection logic with controllable risk.

## Testing Strategy

**Test Scenarios**
　　Oversized requests: HTTP > node.http.maxMessageSize or JSON-RPC > node.jsonrpc.maxMessageSize → return 413, do not enter application layer, no significant memory growth; normal requests are  unaffected; JSON-RPC is independently controlled by node.jsonrpc.maxMessageSize; stress test with continuous large requests to verify stable memory usage without amplification.

　　Streaming-based enforcement avoids full loading, reduces memory allocation compared to application-layer validation, lowers GC pressure, and improves DoS resistance.

## Alternatives Considered (Optional)

　　Per-endpoint application-layer validation: high redundancy, easy to miss, cannot prevent memory amplification; Servlet Filter: executed too late, cannot intercept at the connection layer; single unified configuration: cannot satisfy independent governance requirements across different protocols.

　　Essentially ingress hardening: validate after read → reject before read, scattered logic → centralized control, application-layer fallback → container-layer protection; also enables independent governance boundaries across multiple protocols, improving DoS resistance and reducing operational and ecosystem communication costs.

**Related Issues/PRs**
　　PR: #6658

bladehan1:
## What does this PR do?

Add Jetty `SizeLimitHandler` at the server handler level to enforce request body size limits for all HTTP and JSON-RPC endpoints, preventing OOM denial-of-service from oversized payloads.

### Changes

- Introduce `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize` as independent, configurable size limits
- Default: 4 MB, consistent with gRPC defaults
- Support human-readable size values (e.g. `4m`, `128MB`) via HOCON `getMemorySize()` for all three size configs
- Wire `SizeLimitHandler` into `HttpService.initContextHandler()` as the outermost handler
- Each `HttpService` subclass (4 HTTP + 3 JSON-RPC) sets `maxRequestSize` from the corresponding config getter
- Initialize `maxRequestSize` with a safe 4MB default to prevent silent reject-all if a future subclass omits the assignment
- Deprecate `Util.checkBodySize()` — updated to use `httpMaxMessageSize`, retained as fallback for backward compatibility
- Existing `node.rpc.maxMessageSize` now also supports human-readable sizes (backward compatible — bare integers still treated as bytes)

Previously, HTTP request body size was only validated at the application layer (`Util.checkBodySize()`), which reads the entire body into memory before checking. The JSON-RPC interface had no size validation at all. This allows an attacker to send arbitrarily large payloads, causing OOM and denial of service.

Moving the limit to the Jetty handler chain provides:
1. **OOM protection** — oversized payloads are never fully buffered into memory
2. **Streaming enforcement** — limits are enforced during read, not after full buffering
3. **Unified coverage** — all HTTP and JSON-RPC endpoints are protected by a single mechanism
4. **Independent tuning** — operators can configure HTTP and JSON-RPC limits separately

### Scope and known limitations

This PR's primary goal is **OOM protection**, not uniform HTTP 413 responses.

**Chunked transfer behavior differs by servlet type** due to a pre-existing exception handling design in the servlet chain:

| Request type | HTTP Servlets (132) | JSON-RPC Servlet |
|---|---|---|
| `Content-Length` exceeds limit | **413** (Jetty rejects before dispatch) | **413** (same) |
| Chunked / no `Content-Length`, exceeds limit | **200 + error JSON** `{"Error":"BadMessageException"}` | **200 + empty body** |

Root cause: `SizeLimitHandler` truncates body read at the limit and throws `BadMessageException` (RuntimeException) during streaming.
- **HTTP servlets** have their own `catch(Exception)` → `Util.processError()` writes error JSON to the response body, so the client sees `200 + {"Error":"..."}`.
- **JsonRpcServlet** delegates to jsonrpc4j, which internally catches exceptions during request parsing and returns an empty response → client sees `200 + empty body`.

This is a **pre-existing behavior** of the jsonrpc4j library, not introduced by this PR. Any exception during request body parsing produces the same 200 empty body result, with or without SizeLimitHandler. OOM protection is effective in all cases — the body read is truncated regardless of the response status code. Returning 200 + empty body for malformed/oversized chunked requests is acceptable: it does not affect normal requests, increases attacker difficulty, and the alternative — pre-reading the request body to trigger the size limit (catch the exception to report the error), then wrapping the already-read body into an `HttpServletRequestWrapper` to continue the normal request flow — adds significant complexity for marginal benefit.

Closes #6604

## Migration

This PR makes `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize` independent of `node.rpc.maxMessageSize`. Previously, HTTP servlets enforced their body limit through `Util.checkBodySize()`, which read the gRPC limit (`node.rpc.maxMessageSize`); JSON-RPC had no body-size validation at all.

**Behavior change:**

| Limit              | Before                                              | After                                       |
| ------------------ | --------------------------------------------------- | ------------------------------------------- |
| HTTP body          | bounded by `node.rpc.maxMessageSize` (via `Util.checkBodySize`)  | `node.http.maxMessageSize`, default 4 MB    |
| JSON-RPC body      | not validated                                       | `node.jsonrpc.maxMessageSize`, default 4 MB |
| gRPC message       | `node.rpc.maxMessageSize`, default 4 MB             | unchanged                                   |

**Affected deployments:** those that previously set `node.rpc.maxMessageSize` above 4 MB (for example `8m`) and relied on the same limit applying to HTTP bodies. After upgrading, HTTP and JSON-RPC fall back to the new 4 MB default unless the matching key is set explicitly. The HTTP fallback was intentionally not preserved through `node.rpc.maxMessageSize` so that the new "independent per-protocol limit" contract stays clean.

**Required action for affected operators:**

```hocon
node {
  rpc.maxMessageSize     = 8m   # gRPC, unchanged
  http.maxMessageSize    = 8m   # NEW: set explicitly to preserve prior HTTP body limit
  jsonrpc.maxMessageSize = 8m   # NEW: opt in to enable JSON-RPC body validation at the desired size
}
```

Deployments that did not customize `node.rpc.maxMessageSize` are unaffected (all three default to 4 MB).

## This PR has been tested by

- `SizeLimitHandlerTest` (10 tests): boundary, independent limits, UTF-8 byte counting, chunked transfer, zero-limit, checkBodySize consistency
- `JsonrpcServiceTest.testJsonRpcSizeLimitIntegration`: real JSON-RPC integration test covering normal passthrough, Content-Length oversized (413), and chunked oversized (200 empty body)
- `ArgsTest` (5 new tests): human-readable sizes (KB/MB/GB × binary/SI), raw integer backward compatibility, zero-value, error paths (exceeds int max, negative, invalid unit, non-numeric)
- `UtilTest`: checkBodySize uses httpMaxMessageSize

## Follow-up

- [x] ~~Fix chunked oversized JSON-RPC to return proper error~~ — Won't fix: exception is swallowed inside jsonrpc4j, not in our servlet code. Fixing requires pre-reading the request body to trigger the size limit (catch the exception to report the error), then wrapping the already-read body into an `HttpServletRequestWrapper` to continue the normal request flow — significant test complexity for marginal benefit. Current behavior (200 + empty body) is acceptable.
- [ ] Remove `Util.checkBodySize()` callers once SizeLimitHandler is stable

## Changed files

| Component | Changes |
|---|---|
| `HttpService` | Add `maxRequestSize` field (default 4MB), wire `SizeLimitHandler` in `initContextHandler()` |
| `Args` / `ConfigKey` / `CommonParameter` | Parse `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize`; refactor all three to use `getMemorySize()`; extract `parseMaxMessageSize` private helper to dedupe range-check / cast logic; comment legacy `maxMessageSize` field as the gRPC counterpart |
| 7 service subclasses | Set `maxRequestSize` from protocol-specific config getter |
| `Util.checkBodySize()` | Mark `@Deprecated`, switch to `httpMaxMessageSize` |
| `config.conf` | Add documented config examples for HTTP, RPC, JSON-RPC size limits with zero-value behavior |
| `SizeLimitHandlerTest` | **New**: 10 tests covering HTTP limits, boundary, UTF-8, chunked, independence, zero-limit, checkBodySize consistency |
| `JsonrpcServiceTest` | **New**: `testJsonRpcSizeLimitIntegration` — real JSON-RPC integration test |
| `ArgsTest` | **New**: 5 tests for human-readable parsing (KB/MB/GB), error paths |
| `UtilTest` | **New**: checkBodySize consistency test |

## Follow up
**NOTE: human-readable size support is revert by https://github.com/tronprotocol/java-tron/pull/6790**


xxo1shine:
@bladehan1 One observation on the config validation in `Args.java`: the guard currently rejects `<= 0` values with `TronError`, but there is no upper-bound check. An operator who accidentally sets `node.http.maxMessageSize = 2147483647` would silently run with a 2 GB limit. Adding a reasonable maximum (e.g. 128 MB) with an explicit warn-or-throw would make misconfiguration more visible.

bladehan1:
@xxo1shine 
Thanks for the review. I think that when users manually configure values, it's unnecessary to set all boundaries except for error-prone values. Especially for values ​​that are obviously likely to have problems, setting boundaries is unnecessary.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -327,6 +327,7 @@ public static Transaction packTransaction(String strTransaction, boolean selfTyp
     }
   }
 
+  @Deprecated
```
lxcmyf, line 330: Marking this helper as deprecated does not make the new HTTP limit effective yet. `PostParams.getPostParams()` and many servlets still call `Util.checkBodySize()`, and that method is still enforcing `parameter.getMaxMessageSize()` (the gRPC limit), not `httpMaxMessageSize`. So any request whose body is `> node.rpc.maxMessageSize` but `<= node.http.maxMessageSize` will pass Jetty and then still be rejected in the servlet layer, which means the new independent HTTP setting is not actually honored for a large part of the API surface.

framework/src/main/java/org/tron/common/application/HttpService.java
```diff
@@ -63,7 +66,9 @@ protected void initServer() {
   protected ServletContextHandler initContextHandler() {
     ServletContextHandler context = new ServletContextHandler(ServletContextHandler.SESSIONS);
     context.setContextPath(this.contextPath);
-    this.apiServer.setHandler(context);
+    SizeLimitHandler sizeLimitHandler = new SizeLimitHandler(this.maxRequestSize, -1);
```
lxcmyf, line 80: This moves oversized-request handling in front of every servlet, so those requests never reach `RateLimiterServlet.service()` / `Util.processError()`. Today the HTTP APIs consistently set `application/json` and serialize failures through `Util.printErrorMsg(...)`; after this change an over-limit body gets Jetty's default 413 response instead. That is a client-visible behavior change for existing callers, and the new test only checks status codes so it would not catch the response-format regression.

framework/src/main/java/org/tron/core/services/http/Util.java
bladehan1: Good catch, thanks. `checkBodySize()` has been updated to use `parameter.getHttpMaxMessageSize()` instead of `parameter.getMaxMessageSize()`, so the servlet-layer fallback now honors the independent HTTP limit. See the latest push.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Thanks for flagging the response-format difference. I did a detailed comparison:

**Before (checkBodySize rejection):**
- HTTP status: **200 OK** (processError() does not set status code)
- Body: `{"Error":"java.lang.Exception : body size is too big, the limit is 4194304"}`

**After (SizeLimitHandler rejection):**
- HTTP status: **413 Payload Too Large**
- Body: Jetty default error page (non-JSON)

The format is indeed not fully compatible, but I believe this is acceptable:

1. **The existing behavior is itself incorrect** — returning `200 OK` with an error JSON body violates HTTP semantics. Clients that only check status codes would incorrectly treat the oversized request as successful. The new `413` is the proper HTTP response for this scenario.

2. **The trigger threshold is very high** — default is 4 MB. Normal API requests are far below this. Only abnormal or malicious payloads hit this limit, so the impact surface is negligible for legitimate clients.

3. **413 is a standard HTTP status code** — all HTTP client libraries handle it correctly. Clients already need to handle non-JSON infrastructure errors (e.g., Jetty 503, proxy 502/504).

4. **The new layer provides a real security benefit** — `SizeLimitHandler` rejects during streaming, before the body is fully buffered into memory, which is the core OOM protection. Falling back to application-layer formatting would defeat this purpose.

waynercheung, line 80: `SizeLimitHandler` only guarantees a pre-servlet 413 when `Content-Length` is known. For chunked / unknown-length requests, enforcement happens while downstream code is reading the request body. In the current codebase many HTTP handlers catch broad `Exception` around `request.getReader()`, and `RateLimiterServlet` also catches unexpected `Exception`, so the streaming over-limit exception can be absorbed before Jetty turns it into a 413. That means the PR still does not prove the “uniform 413” behavior described in the issue for requests without `Content-Length`.

framework/src/test/java/org/tron/common/jetty/SizeLimitHandlerTest.java
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  // -- HTTP service body-size tests -------------------------------------------
+
+  @Test
+  public void testHttpBodyWithinLimit() throws Exception {
+    Assert.assertEquals(200, post(httpServerUri, new StringEntity("small body")));
+  }
+
+  @Test
```
waynercheung, line 155: these tests all use `StringEntity`, so they exercise the fixed-length / `Content-Length` path only. They do not cover chunked or unknown-length requests, where `SizeLimitHandler` enforces during body reads instead of before servlet dispatch. Since the existing servlet chain has broad `catch (Exception)` blocks, we need at least one end-to-end test for the streaming path against a real HTTP servlet and the JSON-RPC servlet, and it should assert that an oversized request still surfaces as HTTP 413

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -526,6 +474,20 @@ public static void applyConfigParams(
     PARAMETER.maxMessageSize = config.hasPath(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE)
         ? config.getInt(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE) : GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
+    int defaultHttpMaxMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
+    PARAMETER.httpMaxMessageSize = config.hasPath(ConfigKey.NODE_HTTP_MAX_MESSAGE_SIZE)
+        ? config.getInt(ConfigKey.NODE_HTTP_MAX_MESSAGE_SIZE) : defaultHttpMaxMessageSize;
+    if (PARAMETER.httpMaxMessageSize <= 0) {
```
waynercheung: The new config validation rejects non-positive values, but extremely large limits are still accepted silently. I don’t think a hard cap is mandatory, but at least warning on suspiciously large values would make misconfiguration easier to catch.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -526,6 +474,20 @@ public static void applyConfigParams(
     PARAMETER.maxMessageSize = config.hasPath(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE)
         ? config.getInt(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE) : GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
+    int defaultHttpMaxMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
```
waynercheung: These two new operator-facing config keys should also be documented in config.conf. Right now node.http.maxMessageSize and node.jsonrpc.maxMessageSize are parsed here, but they are not exposed in the sample config alongside the existing documented knobs, which makes discovery and rollout harder for operators. Please add commented examples and a short description under node.http and node.jsonrpc.

framework/src/main/java/org/tron/common/application/HttpService.java
waynercheung, line 80: Can we fail fast here if maxRequestSize was never initialized? This field currently relies on every HttpService subclass remembering to assign it, and the Java default of 0 turns into an accidental zero-byte limit. An explicit validation here would be safer than silently inheriting a bad default.

framework/src/main/java/org/tron/core/services/http/Util.java
0xbigapple: Can `SizeLimitHandler` reliably cover all request-body paths? If the answer is yes, I think it would be better to remove the remaining `Util.checkBodySize()` calls entirely instead of keeping two different enforcement layers. Right now the two checks are not measuring exactly the same thing: `SizeLimitHandler` enforces on the raw HTTP request-body bytes seen by Jetty, while `Util.checkBodySize()` runs after the body has already been decoded into a `String` and then measures `body.getBytes().length`. In edge cases such as line-ending normalization or non-UTF-8 request encodings, that value may not be strictly identical to the byte count used by `SizeLimitHandler`, which means we could end up with small discrepancies where Jetty accepts a request but the servlet-layer fallback still rejects it. If `SizeLimitHandler` is the intended authoritative mechanism, having only one source of truth seems safer and easier to reason about.

framework/src/main/java/org/tron/common/application/HttpService.java
0xbigapple: Oversized requests will now be rejected before they enter the existing servlet/filter pipeline. That means they will no longer go through the current application-side interceptor / rate-limit / metrics path. This may be perfectly fine, but it seems worth documenting or at least acknowledging so everyone is aware of the observability and control-path change.

bladehan1: Good point. I've verified the behavior with tests and here's what actually happens:

**With `Content-Length`**: `SizeLimitHandler` rejects *before* servlet dispatch → clean **413**, servlet is never invoked. This covers the vast majority of real-world HTTP clients.

**Chunked (no `Content-Length`)**: `SizeLimitHandler` installs a `LimitInterceptor` on `HttpInput`. When the accumulated bytes exceed the limit during `getReader()`/`getInputStream()`, it throws `BadMessageException`. However, as you identified, the broad `catch (Exception)` in `RateLimiterServlet` (line 119) and individual servlets absorbs this, so the client sees **200 + error JSON** instead of 413.

**The OOM protection goal is still met**: the body read is truncated at the limit — the full oversized payload is never buffered into memory, which is the core security objective of this PR.

The 413-vs-200 discrepancy for chunked requests is a consequence of the existing broad `catch (Exception)` pattern in the servlet chain, not a `SizeLimitHandler` deficiency. Tightening those catch blocks is a worthwhile follow-up but belongs in a separate PR to limit blast radius.

I'll add chunked transfer tests to document and assert this behavior.

framework/src/test/java/org/tron/common/jetty/SizeLimitHandlerTest.java
bladehan1: Agreed. I've added the following tests using `InputStreamEntity(stream, -1)` to force chunked `Transfer-Encoding`:

- `testChunkedBodyWithinLimit` — chunked request within the limit → **200**
- `testChunkedBodyExceedsLimit` — chunked oversized request hitting a `BroadCatchServlet` (simulates real servlet `catch(Exception)` pattern) → **200 + error JSON** (not 413, because the broad catch absorbs the `BadMessageException`)

Also replaced `EchoServlet` with `BroadCatchServlet` across all tests to mirror the real servlet chain behavior. See latest push.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Addressed by switching from `config.getInt()` to `config.getMemorySize().toBytes()` (Typesafe Config's built-in memory size parser). This change eliminates the int overflow concern entirely:

- **`getMemorySize()`** returns `long` via `ConfigMemorySize.toBytes()`, so values like `10737418240` (10GB) no longer silently overflow — they are stored accurately.
- **For gRPC `maxMessageSize`**: explicitly validated `<= Integer.MAX_VALUE` since the gRPC API (`maxInboundMessageSize`) requires `int`. Values exceeding this fail fast at startup.
- **For HTTP / JSON-RPC**: field type widened to `long`, matching Jetty `SizeLimitHandler(long)`. No artificial upper bound needed.
- **Human-readable sizes** are now supported (`4m`, `128MB`, etc.), which makes misconfiguration much less likely — writing `4m` vs `4194304` is self-documenting and harder to get wrong.

See latest push.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Done. Added commented examples in `config.conf` for all three size parameters:

```hocon
node.http {
    # The maximum request body size for HTTP API, default 4M (4194304 bytes).
    # Supports human-readable sizes: 4m, 4MB, 4194304. Must be positive.
    # maxMessageSize = 4m
}

node.rpc {
    # The maximum message size allowed to be received on the server, default 4M (4194304 bytes).
    # Supports human-readable sizes: 4m, 4MB, 4194304. Must be positive.
    # maxMessageSize = 4m
}

node.jsonrpc {
    # The maximum request body size for JSON-RPC API, default 4M (4194304 bytes).
    # Supports human-readable sizes: 4m, 4MB, 4194304. Must be positive.
    # maxMessageSize = 4m
}
```

See latest push.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Thanks for the suggestion. After consideration, I don't think a fail-fast check is needed here:

1. **0 is a valid value** — after commit `c64d3b2`, zero is explicitly allowed for all three size parameters. `SizeLimitHandler(0, -1)` rejects all non-empty request bodies, which is a legitimate operator choice (e.g., disabling an HTTP endpoint while keeping the port open for health checks).

2. **All 7 subclasses already assign the field** — every `HttpService` subclass sets `maxRequestSize` from `Args.getInstance().getHttpMaxMessageSize()` or `getJsonRpcMaxMessageSize()` before `initContextHandler()` runs. The config-level validation (non-negative check in `Args.java`) ensures the value is always sane.

3. **Operator responsibility** — the size limit is an operational config. Adding a runtime guard for an uninitialized field would be defending against a developer coding error (forgetting to set the field in a future subclass), not a misconfiguration. That's better caught by code review than runtime checks.

So the current design relies on config validation + operator intent, which I think is the right layer for this.

framework/src/main/java/org/tron/core/services/http/Util.java
bladehan1: Good analysis. I agree that long-term, a single enforcement layer is cleaner. Here's a detailed breakdown of the two layers:

**Are they actually measuring the same thing?**

I added tests in commit `5e3eed0` to verify this empirically:

| Scenario | SizeLimitHandler (wire bytes) | checkBodySize (`body.getBytes().length`) | Result |
|----------|------|------|------|
| ASCII JSON | N | N | **Identical** (`testWireBytesMatchCheckBodySizeForAsciiJson`) |
| UTF-8 JSON with CJK | N | N | **Identical** (`testWireBytesMatchCheckBodySizeForUtf8Json`) |
| Body with `\r\n` line endings | N | N − k (k = number of `\r` stripped) | **checkBodySize ≤ wire** (`testCheckBodySizeSafeDirectionWithNewlines`) |

For TRON's JSON API (UTF-8, mostly ASCII), the two measurements are **strongly consistent**. The only divergence comes from line-ending normalization via `lines().collect()`, and it's in a **safe direction** — `checkBodySize` sees fewer bytes, so it never rejects what `SizeLimitHandler` accepts for the same limit value.

Even in the unlikely case of minor discrepancies, operators can adjust the threshold up or down to achieve their desired limit — the behavior is predictable, not erratic.

**Why keep both for now?**

Removing `checkBodySize()` in this PR would touch 100+ servlet files (direct calls + `PostParams.getPostParams()`). I'd prefer to keep this PR focused on adding `SizeLimitHandler` and handle the removal in a follow-up:

1. **This PR** — add `SizeLimitHandler` as the authoritative first layer, mark `checkBodySize()` as `@Deprecated`
2. **Follow-up PR** — remove `checkBodySize()` and all call sites

This follows the standard deprecate-then-remove pattern and keeps each PR reviewable.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Good point. Acknowledged — oversized requests rejected by `SizeLimitHandler` (413 before servlet dispatch) will bypass `RateLimiterServlet` and application-level metrics/logging. This is the expected trade-off: rejecting bad traffic early saves resources, but those requests won't appear in application-layer observability.

For chunked requests specifically, the broad `catch(Exception)` in servlets absorbs the `BadMessageException`, so those DO go through the servlet pipeline and are visible in the existing error path.

If Jetty-level 413 observability becomes important, a future enhancement could add a Jetty `Handler` or access log filter to count rejected requests — but that's outside this PR's scope.

waynercheung: @bladehan1  Thanks for validating the streaming path. I think we should separate plain HTTP servlets from JSON-RPC here, because the real control flow is different.

For the plain HTTP servlet endpoints, the current inconsistency looks fixable in this PR with limited blast radius.
The over-limit exception is still coming from Jetty's `SizeLimitHandler`; it is only being normalized into 200 by the broad catch chain. A concrete fix would be to special-case request-size exceptions in `Util.processError(...)` and/or `RateLimiterServlet`: if the exception (or any cause) is a `BadMessageException` with code 413, bypass the generic application-error formatting and rethrow / `sendError(413)` instead.

For JSON-RPC, we need to clarify the contract first. Fixed-length oversized requests are already rejected before `JsonRpcServlet` / jsonrpc4j is invoked, so they currently come back as transport-level 413 rather than HTTP 200. For chunked requests, the limit is still enforced by Jetty while jsonrpc4j reads `req.getInputStream()`, but jsonrpc4j catches `Throwable` internally and this project's `HttpStatusCodeProvider` hardcodes HTTP 200. So the current synthetic `BroadCatchServlet` test does not prove the real JSON-RPC behavior.

If the intended contract is uniform 413 for oversized requests, I think HTTP should be fixed in this PR, and JSON-RPC needs an explicit integration fix in the jsonrpc4j layer (for example a custom `JsonRpcServer.handle(...)` that special-cases request-size exceptions and sends/rethrows 413 instead of letting them fall into the generic jsonrpc4j error path). If that is intentionally out of scope, then the PR description and issue contract should be narrowed accordingly.

framework/src/test/java/org/tron/common/jetty/SizeLimitHandlerTest.java
```diff
@@ -0,0 +1,364 @@
...
+   * Chunked oversized body hitting a servlet with broad catch(Exception).
+   *
+   * <p>SizeLimitHandler's LimitInterceptor throws BadMessageException during
+   * streaming read, but the servlet's catch(Exception) absorbs it and returns
+   * 200 + error JSON instead of 413. This matches real TRON servlet behavior.
+   *
+   * <p>OOM protection still works: the body read is truncated at the limit.
+   */
+  @Test
+  public void testChunkedBodyExceedsLimit() throws Exception {
```
waynercheung, line 207: By asserting `200 + error JSON` here, this test is codifying behavior that still contradicts the issue/PR description of uniform HTTP 413 for oversized requests. If the intended scope is now “OOM protection with a known response-format limitation on chunked requests”, please update the PR text and issue contract in the same change; otherwise this should remain an identified gap rather than a passing acceptance test.

framework/src/main/java/org/tron/core/config/args/Args.java
waynercheung: The `getMemorySize()` + `long` change addresses the original overflow concern itself, and widening the HTTP / JSON-RPC fields to `long` is the right direction.

What is still missing is direct coverage for the new parsing contract: human-readable values such as `4m` / `128MB`, zero values now being accepted, and the fail-fast path when `node.rpc.maxMessageSize` exceeds `Integer.MAX_VALUE`.
Since this change materially alters config parsing semantics, please add targeted `Args` tests before merge.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -523,8 +471,30 @@ public static void applyConfigParams(
             ? config.getLong(ConfigKey.NODE_RPC_MAX_CONNECTION_AGE_IN_MILLIS)
             : Long.MAX_VALUE;
 
-    PARAMETER.maxMessageSize = config.hasPath(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE)
-        ? config.getInt(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE) : GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
+    long rpcMaxMessageSize = config.hasPath(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE)
```
waynercheung, line 474: Please add targeted `Args` coverage for the new parsing semantics here: human-readable sizes (`4m`, `128MB`), zero-valued limits, and startup failure when `node.rpc.maxMessageSize > Integer.MAX_VALUE`. This logic changed materially and should have direct tests.

framework/src/main/resources/config.conf
```diff
@@ -223,6 +223,10 @@ node {
     solidityPort = 8091
     PBFTEnable = true
     PBFTPort = 8092
+
+    # The maximum request body size for HTTP API, default 4M (4194304 bytes).
```
waynercheung, line 227: Now that zero is intentionally allowed, please document its concrete behavior here. “Must be non-negative” is not enough: operators may reasonably read `0` as “unlimited” or “disabled”, while `SizeLimitHandler(0, -1)` actually means “reject any non-empty request body”. That semantic needs to be explicit in the sample config.

framework/src/main/java/org/tron/common/application/HttpService.java
waynercheung: I don't think allowing operator-configured `0` removes the need for an explicit initialization guard here. The  concern is not that `0` is invalid; it's that an intentional zero-byte limit and an uninitialized field are now  indistinguishable, and the failure mode is a silently broken API.

Since the code now explicitly treats `0` as "reject every non-empty body", relying on Java's default zero is even riskier than before. I would still prefer making the initialization contract explicit here: constructor/abstract getter, or at least a sentinel value that makes "forgot to set `maxRequestSize` in a future subclass" fail fast rather than inheriting 0-byte behavior.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Added comprehensive `Args` tests in the latest push:

**`testMaxMessageSizeHumanReadable`** — covers all three memory tiers with both binary and SI units:

| Tier | Binary (1024-based) | SI (1000-based) |
|------|---------------------|-----------------|
| KB   | `k`, `K`, `Ki`, `KiB` | `kB` |
| MB   | `m`, `M`, `Mi`, `MiB` | `MB` |
| GB   | `g`, `Gi`, `GiB`      | `GB` |

Plus: raw integer backward compatibility, zero-value acceptance.

**Error path tests** — each with `Assert.fail()` + message assertion:

| Test | Input | Exception | Asserted message |
|------|-------|-----------|-----------------|
| `testRpcMaxMessageSizeExceedsIntMax` | `3g` | `TronError` | `"node.rpc.maxMessageSize must be non-negative"` |
| `testMaxMessageSizeNegativeValue` | `-4m` | `IllegalArgumentException` | `"negative"` |
| `testMaxMessageSizeInvalidUnit` | `4x` | `ConfigException.BadValue` | `"Could not parse size-in-bytes unit"` |
| `testMaxMessageSizeNonNumeric` | `abc` | `ConfigException.BadValue` | `"No number in size-in-bytes value"` |

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Covered in the same push — see reply above on the sibling comment.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Agreed that HTTP and JSON-RPC have different control flows for chunked requests. After consideration, I'm narrowing this PR's scope to **OOM protection** — ensuring oversized payloads are never fully buffered into memory, regardless of the HTTP status code returned.

**What this PR guarantees:**
- `Content-Length` present → clean 413 before servlet dispatch (both HTTP and JSON-RPC)
- Chunked / no `Content-Length` → body read is truncated at the limit, OOM prevented; status code depends on the servlet's exception handling

**What belongs in follow-up PRs:**
1. HTTP chunked 413 consistency — special-case `BadMessageException` in `RateLimiterServlet` / `Util.processError`
2. JSON-RPC chunked 413 — requires changes in the jsonrpc4j integration layer

I'll update the PR description to reflect this scoped contract.

framework/src/test/java/org/tron/common/jetty/SizeLimitHandlerTest.java
bladehan1: You're right — asserting 200 here was documenting a known limitation, not an intended contract. I've updated the PR description to clarify that this PR's scope is OOM protection (streaming truncation), not uniform 413. The test stays as-is to document the actual chunked behavior, and the 413 consistency fix for HTTP chunked requests will be a follow-up PR.

framework/src/main/resources/config.conf
bladehan1: Done. Added explicit documentation in config.conf for all three maxMessageSize entries:

```
Setting to 0 rejects all non-empty request bodies (not "unlimited").
```

See latest push.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Addressed. \`maxRequestSize\` now has an explicit default:

\`\`\`java
protected long maxRequestSize = 4 * 1024 * 1024; // 4MB
\`\`\`

A future subclass that forgets the assignment gets a 4MB limit instead of silently rejecting everything. This makes the uninitialized case distinguishable from an intentional \`0\` configured by the operator.

See latest push.

waynercheung:
[SHOULD] With the updated PR scope, I no longer see the chunked `413` gap itself as a contract mismatch, since the description now explicitly scopes this PR to OOM protection rather than uniform HTTP status behavior. The remaining gap is test realism: `SizeLimitHandlerTest` still does not exercise the production JSON-RPC stack, because the `/jsonrpc` path is backed by a synthetic `BroadCatchServlet`, not the real `JsonRpcServlet -> jsonrpc4j` chain. Since the PR still claims protection for JSON-RPC endpoints, I strongly recommend either adding one real JSON-RPC integration case or making the test naming/comments explicit that this is validating generic `HttpInput` interception rather than end-to-end JSON-RPC behavior.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -327,10 +327,12 @@ public static Transaction packTransaction(String strTransaction, boolean selfTyp
...
   public static void checkBodySize(String body) throws Exception {
     CommonParameter parameter = Args.getInstance();
-    if (body.getBytes().length > parameter.getMaxMessageSize()) {
-      throw new Exception("body size is too big, the limit is " + parameter.getMaxMessageSize());
+    if (body.getBytes().length > parameter.getHttpMaxMessageSize()) {
```
waynercheung, line 333: [SHOULD] Even as a deprecated fallback, `checkBodySize()` is still using `body.getBytes()` with the platform default charset. That makes the fallback behavior environment-dependent and can drift from the actual wire-byte accounting. Please use an explicit charset (for example UTF-8, or ideally the request charset when available) so this deprecated path stays deterministic until it is removed.

framework/src/main/java/org/tron/common/application/HttpService.java
```diff
@@ -29,6 +30,8 @@ public abstract class HttpService extends AbstractService {
 
   protected String contextPath;
 
+  protected long maxRequestSize = 4 * 1024 * 1024; // 4MB
```
waynercheung, line 34: [NIT] This default would be better expressed via a shared constant such as `GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE` instead of hardcoding `4 * 1024 * 1024`. The current form duplicates the default semantics in a second place and creates avoidable drift risk later.

waynercheung:
[NIT] `SizeLimitHandlerTest` relies heavily on banner-style ruler comments such as `// -- ... --------` throughout the file. This style adds visual noise without structural value, and is harder to maintain - new tests may not fit neatly into an existing section, and the dashes need manual alignment. The usual best practice here is to rely on descriptive test names, minimal section comments, and helper methods / smaller test classes for grouping, rather than heavy visual separators.

Note also that one comment has already drifted from the implementation: the Javadoc on `testChunkedBodyWithinLimit` still reads "succeed (EchoServlet)", but the servlet under test is now `BroadCatchServlet`. Please update that reference while cleaning up the comments.

framework/src/main/java/org/tron/core/services/http/Util.java
bladehan1: `checkBodySize()` is already marked `@Deprecated` and scheduled for removal in the follow-up PR. The `body.getBytes()` charset issue is a pre-existing concern in the original code, not introduced by this PR. I'd prefer not to layer further changes onto a deprecated path — the removal PR will eliminate this entirely.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Intentional — `HttpService` is in the `common` module and should not depend on `GrpcUtil` (which is gRPC-specific). This default is solely a safety net for uninitialized subclasses, not a general configuration default. The actual runtime value always comes from `Args.getInstance().getHttpMaxMessageSize()`, which references `GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE` at the config layer. Keeping the constant inline here avoids coupling the HTTP base class to gRPC.

bladehan1:
> banner-style ruler comments / stale EchoServlet Javadoc

Done. Removed all banner-style ruler comments and fixed the stale `EchoServlet` reference in `testChunkedBodyWithinLimit` Javadoc. See latest push.

bladehan1:
> JSON-RPC test realism / BroadCatchServlet does not exercise the real jsonrpc4j chain

Agreed — the PR inserts SizeLimitHandler on the JSON-RPC path, so we should verify no regression on the real stack.

**Added** `testJsonRpcSizeLimitIntegration()` in `JsonrpcServiceTest`, using the Spring-injected `FullNodeJsonRpcHttpService` (real `JsonRpcServlet` + jsonrpc4j). Covers three scenarios with explicit expected results:

1. **Normal JSON-RPC request** through SizeLimitHandler → `200` with valid `"result"` (no regression)
2. **Content-Length oversized** → `413` before servlet dispatch
3. **Chunked oversized** → `200` with empty body

Scenario 3 deserves explanation: `BadMessageException` (RuntimeException) thrown during body read propagates through jsonrpc4j → out of `JsonRpcServlet.doPost()` (no try/catch) → caught by `RateLimiterServlet.catch(Exception unexpected)` which logs but does not write a response. Jetty defaults to 200 empty body. **OOM protection is still effective** — the body read is truncated at the limit.

The silent swallowing in `RateLimiterServlet` is a pre-existing design issue (affects all RuntimeExceptions from any servlet, not just BadMessageException). Fixing it impacts 100+ servlets and belongs in a dedicated PR.

**Also cleaned up** `SizeLimitHandlerTest`: removed the 3 redundant `testJsonRpcBody*` tests (BroadCatchServlet cannot represent the real jsonrpc4j chain), renamed `TestJsonRpcService` → `SecondHttpService`, kept only `testTwoServicesHaveIndependentLimits` for validating per-service limit isolation.

See latest push.

waynercheung:
Re-checking the current PR head (`6f0a0f0`), I still do not see the changes described in the reply on this branch yet.

Specifically:
- `SizeLimitHandlerTest` still contains the banner-style section comments, and the stale `EchoServlet` reference is still present.
- I do not see the new `testJsonRpcSizeLimitIntegration()` in `JsonrpcServiceTest`.
- The synthetic `/jsonrpc` tests in `SizeLimitHandlerTest` are still present.
- `Util.checkBodySize()` is still using `body.getBytes()` with the platform default charset.

If there is a newer push on another branch or commit, please point me to it. Otherwise I think these threads should stay open until the code is actually updated in this PR.

bladehan1:
@waynercheung 

The push was a little late, but you can view it now.
As mentioned above, `Util.checkBodySize()` no longer adds `body.getBytes`.
Another issue has been updated.

framework/src/main/java/org/tron/core/services/http/Util.java
waynercheung: Understood. Since this is a pre-existing issue on a deprecated fallback path and not something introduced by this PR, I’m okay not expanding the change scope here. I do not consider this a blocker for the current PR.

That said, the issue still exists while `checkBodySize()` remains live, so I’d like it to stay tracked in the follow-up cleanup/removal work rather than being treated as fully resolved.

framework/src/main/java/org/tron/common/application/HttpService.java
waynercheung: Makes sense. Since this field is only acting as a local safety-net default in the HTTP base class, and the actual operator-facing default still comes from the config layer, I’m okay with keeping the inline 4MB value here to avoid introducing an unnecessary dependency on gRPC-specific constants.

framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
```diff
@@ -1099,4 +1103,74 @@ public void testWeb3ClientVersion() {
       Assert.fail();
     }
   }
+
+  /**
+   * Verifies SizeLimitHandler integration with the real JsonRpcServlet + jsonrpc4j stack.
+   *
+   * Covers: normal request no regression, Content-Length oversized 413,
+   * and chunked oversized handled gracefully (body truncated, 200 + empty body
```
waynercheung, line 1111: [NIT] After tracing the actual path more closely, this comment is not just stronger than what the test proves - the mechanism is also inaccurate. For chunked / unknown-length requests, the over-limit exception is raised while jsonrpc4j is reading the request body, and `JsonRpcServer.handle(...)` catches it internally before it can propagate back to `RateLimiterServlet`. Since `JsonRpcServlet` configures the JSON-RPC `HttpStatusCodeProvider` to always return 200, the current client-visible behavior becomes `200 + empty body`. Please reword this comment/assertion accordingly.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Thanks for the detailed breakdown. After further investigation, I need to correct the root cause attribution for JSON-RPC chunked requests:

**Correction: the exception is swallowed by jsonrpc4j, not `RateLimiterServlet`.**

The `BroadCatchServlet` in `SizeLimitHandlerTest` did not accurately model the JSON-RPC path — you were right to flag this. I've since added `JsonrpcServiceTest.testJsonRpcSizeLimitIntegration`, which hits the real `JsonRpcServlet` → jsonrpc4j stack and confirms: jsonrpc4j internally catches `Throwable` during request parsing and returns an empty response (200 + empty body). `RateLimiterServlet` is not involved in the exception handling for this path.

**Scoping decision:**

- **HTTP chunked 413**: special-casing `BadMessageException` in `RateLimiterServlet` / `Util.processError` is feasible but touches the shared servlet error path. Deferring to a follow-up PR to limit blast radius.
- **JSON-RPC chunked 413**: the fix would require pre-reading the request body to trigger the size limit, catching the exception to report the error, then wrapping the already-read body into an `HttpServletRequestWrapper` to continue the normal jsonrpc4j flow. This adds significant complexity and test surface for marginal benefit.

**Why 200 + empty body is acceptable for JSON-RPC chunked:**
- Does not affect normal requests (only triggered by oversized chunked payloads)
- OOM protection is effective — body read is truncated at the limit regardless
- Increases attacker difficulty — no useful error information is leaked
- The alternative fix path (pre-read → catch → `HttpServletRequestWrapper`) is disproportionate to the benefit

PR description and follow-up list have been updated accordingly.

framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
bladehan1: Agreed, good catch. The comment and assertion messages have been corrected in the latest push (`7df020e`):

- Javadoc: `"because jsonrpc4j absorbs the BadMessageException"` (was `RateLimiterServlet`)
- Inline comment: `"absorbed by jsonrpc4j catch(Exception)"` (was `RateLimiterServlet catch(Exception)`)
- Assertion message: `"(jsonrpc4j absorbs BadMessageException)"` (was `RateLimiterServlet absorbs BadMessageException`)

The mechanism is as you described: `SizeLimitHandler` throws `BadMessageException` during body read → `JsonRpcServer.handle(...)` catches it internally → `HttpStatusCodeProvider` returns 200 → client sees `200 + empty body`.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -523,8 +471,30 @@ public static void applyConfigParams(
...
+        ? config.getMemorySize(ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE).toBytes()
+        : GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
+    if (rpcMaxMessageSize < 0 || rpcMaxMessageSize > Integer.MAX_VALUE) {
+      throw new TronError("node.rpc.maxMessageSize must be non-negative and <= "
+          + Integer.MAX_VALUE + ", got: " + rpcMaxMessageSize, PARAMETER_INIT);
+    }
```
Sunny6889, line 480:
```
 if (PARAMETER.httpMaxMessageSize <= 0) {                                 
       throw new TronError("node.http.maxMessageSize must be > 0, got: "
           + PARAMETER.httpMaxMessageSize, PARAMETER_INIT);                   
}   
```

framework/src/test/java/org/tron/core/services/http/UtilTest.java
```diff
@@ -129,6 +129,32 @@ public void testPackTransactionWithInvalidType() {
...
+      String withinHttpLimit = new String(new char[150]).replace('\0', 'a');
+      // should pass: 150 < httpMaxMessageSize(200), even though > maxMessageSize(100)
+      Util.checkBodySize(withinHttpLimit);
+
+      String exceedsHttpLimit = new String(new char[201]).replace('\0', 'b');
+      try {
+        Util.checkBodySize(exceedsHttpLimit);
+        Assert.fail("expected exception for body exceeding httpMaxMessageSize");
+      } catch (Exception e) {
+        Assert.assertTrue(e.getMessage().contains("200"));
```
Sunny6889, line 150: Suggest use 
```
Assert.assertThrows(Exception.class,
        () -> Util.checkBodySize(exceedsHttpLimit));
Assert.assertTrue(e.getMessage().contains("200"));
```
to remove the try/catch block

framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
```diff
@@ -1099,4 +1103,72 @@ public void testWeb3ClientVersion() {
...
+   * because jsonrpc4j absorbs the BadMessageException).
+   */
+  @Test
+  public void testJsonRpcSizeLimitIntegration() {
+    long testLimit = 1024;
+    try {
+      long originalLimit = fullNodeJsonRpcHttpService.getMaxRequestSize();
```
waynercheung, line 1118: [NIT] `originalLimit` is restored only in the inner `finally`, which is reached only after `fullNodeJsonRpcHttpService.start()` succeeds. If `start()` throws, this shared bean keeps the overridden limit for the rest of the class. Please move the `setMaxRequestSize(originalLimit)` restore into the outer `finally` so the test  always leaves the service in its original state.

waynercheung:
[NIT] A few of the newly added test comments use decorative non-ASCII punctuation such as `→` and `—`. This is not a correctness issue, and Google Java Style does allow Unicode where it improves readability, but these particular symbols are decorative rather than semantic. For consistency with the rest of the Java codebase, I’d prefer plain ASCII punctuation (`->`, `-`) in comments. The UTF-8 test literals themselves (for example `一` / `测试地址`) are of course justified and should stay.

framework/src/main/java/org/tron/core/config/args/Args.java
waynercheung: I don’t think we should change this to `<= 0` in this PR. The current implementation, config docs, and tests all explicitly treat `0` as a valid value meaning “reject all non-empty request bodies”, not as an invalid configuration.
Changing that now would be a behavior change rather than a local validation cleanup, and would require updating the documented semantics and the related tests as well.

waynercheung:
[DISCUSS] `node.rpc.maxMessageSize` already has a stronger fail-fast guard because the gRPC API ultimately requires an `int` (`<= Integer.MAX_VALUE`). By contrast, `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize` are currently accepted as any non-negative `long`. I’m not convinced they should be forced to the same `Integer.MAX_VALUE` cap, but do we want some operator guardrail here as well (for example a startup warning or a documented recommended upper bound for suspiciously large values)? Very large HTTP / JSON-RPC limits materially weaken the intended OOM protection, and values above `Integer.MAX_VALUE` also exceed the effective range of the deprecated `checkBodySize()` fallback.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Agreed, keeping 0 as valid

framework/src/test/java/org/tron/core/services/http/UtilTest.java
bladehan1: Done. Applied `Assert.assertThrows` in `UtilTest`, and extended the same fix to all 4 similar patterns in `ArgsTest`.

framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
bladehan1: Fixed. Hoisted `originalLimit` before the outer `try` and moved the restore into the outer `finally` — safe even if `start()` throws.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Agree with @waynercheung — `0` is intentionally valid (reject-all semantics). Keeping the current `< 0` guard.

bladehan1:
Done. Replaced all `→` and `—` with `->` and `-` in newly added test comments across `SizeLimitHandlerTest` and `JsonrpcServiceTest`.

bladehan1:
> [DISCUSS] `node.rpc.maxMessageSize` already has a stronger fail-fast guard because the gRPC API ultimately requires an `int` (`<= Integer.MAX_VALUE`). By contrast, `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize` are currently accepted as any non-negative `long`. I’m not convinced they should be forced to the same `Integer.MAX_VALUE` cap, but do we want some operator guardrail here as well (for example a startup warning or a documented recommended upper bound for suspiciously large values)? Very large HTTP / JSON-RPC limits materially weaken the intended OOM protection, and val



Valid concern. For this PR the focus is OOM baseline protection — adding an upper-bound guardrail here would expand scope and potentially delay the core fix. A startup warning for suspiciously large values (e.g. > 128 MB) is worth a follow-up, but I'd prefer to land this first.

waynercheung:
> For this PR the focus is OOM baseline protection — adding an upper-bound guardrail here would expand scope and potentially delay the core fix.

 [SHOULD] I agree this is not the same kind of hard API constraint as gRPC, but I still think the config layer should fail fast for HTTP / JSON-RPC values above Integer.MAX_VALUE in this PR.

Reason: although Jetty accepts `long`, the current java-tron request-processing path does not realistically support arbitrarily large bodies. Many downstream paths still materialize request bodies into `String` / `byte[]`, and once `httpMaxMessageSize > Integer.MAX_VALUE`, the deprecated `checkBodySize()` fallback also becomes ineffective.

Since this PR is already touching the shared maxMessageSize parsing/validation block, aligning the effective supported range here would be a small change with good operational safety benefits.

bladehan1:
Thanks for the detailed follow-up. After weighing the trade-off I'd like to respectfully hold the current position, but with a clearer principled reason than the scope-only framing from my earlier reply.

**Failure mode at the edge is loud, not silent.** When `node.http.maxMessageSize` / `node.jsonrpc.maxMessageSize` exceeds `Integer.MAX_VALUE`, the downstream paths you cited (`byte[]` / `String` materialization, the deprecated `checkBodySize()` fallback) fail with explicit `Throwable`s — `NegativeArraySizeException`, `OutOfMemoryError`, or the existing explicit size check — none of which silently truncate a body or return a wrong `200`. Operators hit a stack trace, not a subtle data bug.

**This is the same contract JVM `-Xmx`, Kafka `message.max.bytes`, Postgres `shared_buffers`, nginx `client_max_body_size` offer**: the config layer accepts any syntactically valid value, and operators who push the edge own the loud consequences. None of those systems pre-reject values that "might be too large for downstream buffers", because that would turn the config contract into a mirror of current implementation internals — and freeze those internals as part of the public contract.

**Our stance on config validation going forward**: the config layer validates syntax and business policy; it does not pre-enforce implementation-side capacity bounds unless the failure mode at the edge would be *silent* (silent truncation, silent 200, silent availability regression). Loud failures — even if the message isn't elegant — meet the baseline operator-trust contract. Adding an `Integer.MAX_VALUE` cap here would couple the config contract to "we currently materialize bodies", which is an implementation state we'd like to be free to change later (e.g. full end-to-end streaming) without a config-layer migration.

If evidence surfaces that a path *does* fail silently at long-overflow — e.g. a handler that casts `long` to `int` and continues processing truncated data — that's a different category of bug and I'd treat it as a must-fix at the implementation layer, not a config-layer patch. Happy to track that as a follow-up issue.

The `rpc.maxMessageSize` cap is a separate case: it's driven by a hard gRPC API constraint (`int`-typed), not by our own materialization choice, so that cap remains.

bladehan1:
added an explicit hard cap. `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize` now reject values > `Integer.MAX_VALUE` at startup with `TronError`, matching the existing `node.rpc.maxMessageSize` rule. Rationale: downstream body materialization (String / byte[] / StringBuilder) is int-bounded, so config beyond that produces an ambiguous silent mismatch — fail-fast at startup is safer than a runtime NegativeArraySizeException or a truncation surprise.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Follow-up: added an explicit hard cap. `node.http.maxMessageSize` and `node.jsonrpc.maxMessageSize` now reject values > `Integer.MAX_VALUE` at startup with `TronError`, matching the existing `node.rpc.maxMessageSize` rule. Rationale: downstream body materialization (String / byte[] / StringBuilder) is int-bounded, so config beyond that produces an ambiguous silent mismatch — fail-fast at startup is safer than a runtime NegativeArraySizeException or a truncation surprise.

Covered by new `testHttpMaxMessageSizeExceedsIntMax` and `testJsonRpcMaxMessageSizeExceedsIntMax` in `ArgsTest`, mirroring `testRpcMaxMessageSizeExceedsIntMax`.

framework/src/main/java/org/tron/core/config/args/Args.java
bladehan1: Follow-up: the three `maxMessageSize` docblocks in `config.conf` (http / rpc / jsonrpc) are now consistent and state the full constraint:

    # Must be non-negative and <= 2147483647 (Integer.MAX_VALUE, ~2 GiB).

Reflects the hard cap just added to the http / jsonrpc validators (previously only rpc had it).

lxcmyf (approved):

waynercheung (commented):
[DISCUSS] One compatibility point is worth calling out explicitly in the PR description / release notes: before this change, many existing HTTP servlet paths were effectively bounded by `node.rpc.maxMessageSize` via `Util.checkBodySize()`. After introducing independent `node.http.maxMessageSize`, deployments that previously increased only `node.rpc.maxMessageSize` (for example to `8m`) will fall back to the new 4 MB HTTP default unless they set `node.http.maxMessageSize` explicitly.
 
I'm not suggesting we restore `rpcMaxMessageSize` as the HTTP fallback, because that would blur the new "independent per-protocol config" contract again. But the migration impact should be documented clearly so operators do not get a silent behavior change on upgrade.
 
Suggest adding a short **Migration** subsection to the PR description, for example:
 
> If your deployment previously customized `node.rpc.maxMessageSize` (for example `= 8m`), HTTP endpoints were implicitly bounded by that value through `Util.checkBodySize()`. After this change, set `node.http.maxMessageSize` (and `node.jsonrpc.maxMessageSize` if applicable) explicitly if you want to preserve the previous effective limit.
>
> Default deployments (all three at 4 MB) are unaffected.
>
> Note: JSON-RPC endpoints previously had no unified pre-ingress size limit, so `node.jsonrpc.maxMessageSize = 4m` is a new enforcement point. Operators should review client request peak sizes before upgrade.

bladehan1:
> [DISCUSS] One compatibility point is worth calling out explicitly in the PR description / release notes: before this change, many existing HTTP servlet paths were effectively bounded by `node.rpc.maxMessageSize` via `Util.checkBodySize()`. After introducing independent `node.http.maxMessageSize`, deployments that previously increased only `node.rpc.maxMessageSize` (for example to `8m`) will fall back to the new 4 MB HTTP default unless they set `node.http.maxMessageSize` explicitly.

Agreed, this is a real silent behavior change on upgrade and should not stay implicit. I've added a `## Migration` subsection to the PR description covering:

- Before/after table for HTTP, JSON-RPC, gRPC body limits
- Which deployments are affected (those that bumped `node.rpc.maxMessageSize` above 4 MB and relied on it covering HTTP)
- A concrete `node { ... }` snippet showing the explicit keys to set to preserve prior behavior
- An explicit note that the HTTP fallback through `node.rpc.maxMessageSize` was intentionally not preserved, in order to keep the new "independent per-protocol limit" contract clean — matching your point that restoring the fallback would blur it

I deliberately did not add a migration note to `config.conf` itself: config files document current behavior, not transition history, and accumulating historical notes there ages poorly. Migration docs belong in PR description / release notes, which is where this lives now.

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -216,6 +216,12 @@ public class CommonParameter {
   public int maxMessageSize;
```
SeriousCoding789, line 235: [NIT] Please add a comments to state `maxMessageSize` refere to rpc max message size. To make it easy to understand comparing with the below `httpMaxMessageSize`  `jsonRpcMaxMessageSize`

Sunny6889 (approved):

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -523,8 +523,34 @@ public static void applyConfigParams(
...
+    PARAMETER.jsonRpcMaxMessageSize = config.hasPath(ConfigKey.NODE_JSONRPC_MAX_MESSAGE_SIZE)
+        ? config.getMemorySize(ConfigKey.NODE_JSONRPC_MAX_MESSAGE_SIZE).toBytes()
+        : defaultMaxMessageSize;
+    if (PARAMETER.jsonRpcMaxMessageSize < 0
+        || PARAMETER.jsonRpcMaxMessageSize > Integer.MAX_VALUE) {
+      throw new TronError("node.jsonrpc.maxMessageSize must be non-negative and <= "
+          + Integer.MAX_VALUE + ", got: "
+          + PARAMETER.jsonRpcMaxMessageSize, PARAMETER_INIT);
+    }
```
SeriousCoding789, line 454: [NIT] The three maxMessageSize parsing blocks in applyConfigParams() are structurally identical — read from config (with fallback), range-check, cast to int, assign. Having them inlined one after another in a 500+ line method makes it easy to miss a subtle difference between the three.
  
Consider extracting a small private helper:
```
private static int parseMaxMessageSize(Config config, String key, long defaultBytes) {
       long value = config.hasPath(key)
           ? config.getMemorySize(key).toBytes()
           : defaultBytes;
       if (value < 0 || value > Integer.MAX_VALUE) {
           throw new TronError(key + " must be non-negative and <= "
               + Integer.MAX_VALUE + ", got: " + value, PARAMETER_INIT);
       }
       return (int) value;
 }
```
   Then the three assignment lines become:
  
```
   long defaultSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
   PARAMETER.maxMessageSize     = parseMaxMessageSize(config, NODE_RPC_MAX_MESSAGE_SIZE,     defaultSize);
   PARAMETER.httpMaxMessageSize = parseMaxMessageSize(config, NODE_HTTP_MAX_MESSAGE_SIZE,    defaultSize);
   PARAMETER.jsonRpcMaxMessageSize = parseMaxMessageSize(config, NODE_JSONRPC_MAX_MESSAGE_SIZE, defaultSize);
```
This removes ~20 lines of near-duplicate code, makes the error messages consistent, and makes it trivial to add a fourth protocol limit in the future without copy-paste risk.

bladehan1:
Both addressed — pushed as a single follow-up commit.

**1. Legacy naming comment on `maxMessageSize`**

Added a brief two-line comment on `CommonParameter.maxMessageSize` clarifying that the legacy field refers to the gRPC limit, alongside the explicit `httpMaxMessageSize` / `jsonRpcMaxMessageSize` siblings. Kept it as a plain `//` comment (not Javadoc) to match the rest of the field block, which is uncommented. A rename to `rpcMaxMessageSize` would be the cleaner long-term fix, but it's a `public` field with cross-module callers, so out of scope here.

**2. Extracted `parseMaxMessageSize` helper**

Net effect: ~16 lines removed across the three blocks, error messages now guaranteed consistent (the previous string-concat blocks had subtly diverged on line breaking), and adding a future protocol limit becomes a one-liner.

One adjustment to your suggested signature: the helper returns `long` rather than `int`, because `httpMaxMessageSize` and `jsonRpcMaxMessageSize` are typed `long` (matching Jetty's `SizeLimitHandler(long, long)` expectation). The rpc assignment retains an explicit `(int)` cast at the call site to keep the gRPC `int32` hard constraint visible — same idea as the cap rationale we discussed earlier.

Resulting call sites:

```java
long defaultMaxMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
PARAMETER.maxMessageSize = (int) parseMaxMessageSize(
    config, ConfigKey.NODE_RPC_MAX_MESSAGE_SIZE, defaultMaxMessageSize);
PARAMETER.httpMaxMessageSize = parseMaxMessageSize(
    config, ConfigKey.NODE_HTTP_MAX_MESSAGE_SIZE, defaultMaxMessageSize);
PARAMETER.jsonRpcMaxMessageSize = parseMaxMessageSize(
    config, ConfigKey.NODE_JSONRPC_MAX_MESSAGE_SIZE, defaultMaxMessageSize);
```

`ArgsTest` (including the error-path cases for each of the three keys) passes unchanged.

bladehan1:
Resolved conflict with the develop branch.
Configuration parameter settings.

Sunny6889 (approved):

0xbigapple (approved):

waynercheung (approved):
LGTM

framework/src/main/java/org/tron/core/services/http/Util.java
ahouramasoud1369-alt, line 1: Java

PR #6662, opened 2026-04-09
chore: fix outdated content across docs/ and root-level docs

Little-Peony:
## Summary

Documentation cleanup. No production code or scripts modified.

### docs/
- **implement-a-customized-actuator** (EN + ZH): fix proto field number, missing imports, test lifecycle annotations, and references to non-existent APIs
- **modular-deployment** (EN + ZH): remove stale CMS GC flags (removed in JDK 14), drop Windows-specific startup note, remove deprecated Solidity Node port
- **modular-introduction** (EN + ZH): update module count from 8 → 9 (add `platform`) and revise module descriptions

### Root-level docs
- **CONTRIBUTING.md**: replace Travis CI badge/references → GitHub Actions; fix commit title length guidance
- **build.md**, **run.md**: removed (content superseded by README)
- **quickstart.md**: remove deprecated Solidity Node references


warku123 (approved):

docs/modular-introduction-en.md
```diff
@@ -16,7 +16,7 @@ The aim of java-tron modularization is to enable developers to easily build a de
 
 ![modular-structure](https://github.com/tronprotocol/java-tron/blob/develop/docs/images/module.png)
 
-A modularized java-tron consists of six modules: framework, protocol, common, chainbase, consensus and actuator. The function of each module is elaborated below.
+A modularized java-tron consists of eight modules: framework, protocol, common, chainbase, consensus, actuator, crypto and plugins. The function of each module is elaborated below.
```
halibobo1205: The module count should be nine, not eight — `platform` is missing. From `settings.gradle`:
```
framework, chainbase, protocol, actuator, consensus, common, crypto, plugins, platform
```

Please add a description for the `platform` module and update the count in both EN and ZH files.

build.md
```diff
@@ -1,81 +1,54 @@
 # How to Build
```
halibobo1205: After the rewrite, both build.md and run.md are mostly thin wrappers that point back to the README. Maintaining multiple files that say "see README" adds indirection for readers and maintenance burden (they can easily drift out of sync again — which is how they got stale in the first place).

Suggestion: consider removing build.md and run.md entirely, and moving the unique content (IntelliJ IDEA setup steps, jitpack dependency snippet) into the README or a CONTRIBUTING.md section. This would be a net reduction in maintenance surface.

317787106:
Suggest to delete build.md, gen.sh, run.md and some unused file totally, specify the alternative document in README.md. There are too many reduant files in root level.

halibobo1205:
The PR title and commits are scoped to docs:, but the changeset also modifies shell scripts (start.sh, ver.sh, gen.sh) with functional bug fixes (e.g., undefined $darwin variable, tautological condition, CMS→G1GC). Consider either splitting into two PRs (docs vs. scripts) or updating the title/scope to reflect the actual changes, e.g., docs, chore(scripts): ....

ver.sh
```diff
@@ -24,7 +23,7 @@ then
 fi
 
 
-versionPath="src/main/java/org/tron/program/Version.java"
+versionPath="framework/src/main/java/org/tron/program/Version.java"
```
halibobo1205, line 27: `ver.sh` has two invocation paths:
 1. **Direct execution** from the repo root: `./ver.sh` — the new path `framework/src/main/java/...` resolves correctly.
 2. **Via Gradle**: `./gradlew version` — the task is defined in `framework/build.gradle` with `commandLine 'bash', '-c', '../ver.sh'`, so the working directory is `framework/`. The new path resolves to `framework/framework/src/main/java/.../Version.java`, which **does not exist**:

```
 warning: could not open directory 'framework/framework/src/main/java/org/tron/program/': No such file or directory
fatal: pathspec 'framework/src/main/java/org/tron/program/Version.java' did not match any files
 ```

 The original relative path `src/main/java/...` was actually correct for the Gradle invocation.

Suggestion: add `cd "$(dirname "$0")"` at the top of `ver.sh` to anchor the working directory to the script's location (i.e., the repo root), so both invocation paths work correctly.

File: docs/modular-introduction-en.md (hunk 1 above)
Comment (Little-Peony) [reply docs/modular-introduction-en.md]: You're right — `platform` was missing. Fixed: updated the count to nine and added a description for `platform` in both EN and ZH files.

File: build.md (hunk 2 above)
Comment (Little-Peony) [reply build.md]: Good point. `build.md`, `run.md` and `gen.sh` are removed in the updated commit.

Regarding the unique content in `build.md` (IntelliJ IDEA setup steps and jitpack dependency snippet) — these are intentionally not moved into the README. A project README should stay high-level; IDE-specific setup and library dependency configuration are too granular and better suited for the [official documentation site](https://tronprotocol.github.io/documentation-en/). Adding them to README would make it harder to maintain and drift from the same problem we're trying to fix.

**Comment by Little-Peony** [comment]
Good point. `build.md`, `run.md` and `gen.sh` are removed in the updated commit.

File: ver.sh (hunk 3 above)
Comment (Little-Peony) [reply ver.sh:27]: Good catch. Added `cd "$(dirname "$0")"` at the top of `ver.sh` to anchor the working directory to the script's location, so both `./ver.sh` and `./gradlew version` resolve the path correctly.

File: docs/modular-deployment-en.md [hunk 4]
```diff
@@ -1,8 +1,8 @@
 # How to deploy java-tron after modularization
 
-After modularization, java-tron is launched via shell script instead of typing command: `java -jar FullNode.jar`.
+After modularization, the recommended way to launch java-tron is via the shell script generated in `bin/`. The classic `java -jar FullNode.jar` command is still fully supported as an alternative.
 
-*`java -jar FullNode.jar` still works, but will be deprecated in future*.
+> **Supported platforms**: Linux and macOS. Windows is not supported.
```
warku123, line 5: The Gradle build (`framework/build.gradle`) still generates Windows start scripts via `windowsStartScriptGenerator`, and `gradlew.bat` exists in the repo. Stating "Windows is not supported" here seems inconsistent with the actual build artifacts. Should we clarify that `start.sh` itself is Unix-only, but the `bin/*.bat` scripts generated by Gradle are still available for Windows?

SeriousCoding789: The windowsStartScriptGenerator configuration in build.gradle  (and gradle/windowsStartScript.txt) are indeed inconsistent with the intended direction of dropping Windows support. Cleaning those up will be done in a another PR — keeping this one focused on the documentation fix.

halibobo1205 (approved):
LGTM

warku123 (approved):

Sunny6889 (dismissed):

Sunny6889 (commented):
@317787106 please help do further review

317787106 (approved):

PR #6668, opened 2026-04-09
refactor(jsonrpc): centralize block selector parsing into JsonRpcApiUtil

0xbigapple:
**What does this PR do?**
This PR centralizes JSON-RPC block tag and block number resolution in `JsonRpcApiUtil` so block selectors are parsed consistently across the JSON-RPC layer.

It also removes `Wallet`'s reverse dependency on JSON-RPC code, eliminates the `-1 means latest` sentinel, and simplifies the block range resolution logic in `LogFilterWrapper`.

- Added `JsonRpcApiUtil.isBlockTag`, `parseBlockTag`, and `parseBlockNumber`, and moved block tag constants and related errors, including `safe`, into `JsonRpcApiUtil`.
- Updated `TronJsonRpcImpl` to delegate block selector parsing to `JsonRpcApiUtil`; `eth_getBlock*` paths now use shared block resolution logic.
- Centralized the existing `latest`-only validation for state-read APIs such as `getTrxBalance`, `getStorageAt`, `getABIOfSmartContract`, and `eth_call`.
- Removed JSON-RPC block parsing helpers from `Wallet`, replaced the `-1` sentinel behavior, and added `getHeadBlockNum()`.
- Simplified the branching logic in `LogFilterWrapper` by making block range handling strategy-based and easier to reason about.
- Added tests for block tag parsing and log filter strategy behavior.

Block selector parsing is currently duplicated across `Wallet`, `JsonRpcApiUtil`, and `TronJsonRpcImpl`, with slightly different responsibilities and control flow in each layer.

This creates several problems:
- duplicated and inconsistent block tag resolution logic
- an undesirable reverse dependency from core `Wallet` code to the JSON-RPC layer
- ambiguous `-1` sentinel handling that forces callers to reinterpret results differently
- complex block range branching in `LogFilterWrapper` that is harder to maintain and verify

By centralizing parsing in `JsonRpcApiUtil`, this PR makes the behavior easier to reason about, reduces coupling, and preserves existing JSON-RPC semantics while removing fragile internal conventions.

- Unit Tests
- Manual Testing

**Follow up**
None.

**Extra details**
- This PR is intended as a refactor and cleanup. All JSON-RPC endpoints return identical results for valid inputs.
Block-number hex parsing in the `eth_*` block-by-number endpoints now uses `BigInteger.longValueExact()` instead of `longValue()`: inputs that overflow `Long.MAX_VALUE` now return `invalid block number` instead of being silently truncated to their low 64 bits. Valid block numbers are unaffected.


framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -515,20 +518,52 @@ public static long parseEnergyFee(long timestamp, String energyPriceHistory) {
...
+
+  public static long parseBlockTag(String tag, Wallet wallet)
       throws JsonRpcInvalidParamsException {
-    if (PENDING_STR.equalsIgnoreCase(blockNumOrTag)) {
-      throw new JsonRpcInvalidParamsException(TAG_PENDING_SUPPORT_ERROR);
+    if (LATEST_STR.equalsIgnoreCase(tag)) {
+      // Use getNowBlock() to fetch the latest persisted block directly from blockStore,
+      // avoiding a race with updateDynamicProperties that updates latestBlockHeaderNumber
+      // before the block is written to blockStore/blockIndexStore.
+      Block block = wallet.getNowBlock();
```
bladehan1: **[P2 — Performance]** `parseBlockTag("latest")` loads a full `Block` from blockStore; consider using `DynamicPropertiesStore` cache instead.

`wallet.getNowBlock()` → `blockStore.getBlockByLatestNum(1)` deserializes a full `Block` protobuf (including all transactions) just to extract the block number.

`Wallet.getHeadBlockNum()` (which this PR already added at line 708) reads from `DynamicPropertiesStore.getLatestBlockHeaderNumber()` — a single in-memory snapshot cache lookup with zero I/O:

```
Wallet.getHeadBlockNum()
  → ChainBaseManager.getHeadBlockNum()
    → DynamicPropertiesStore.getLatestBlockHeaderNumber()
      → getUnchecked(LATEST_BLOCK_HEADER_NUMBER)  // snapshot cache
```

Currently `parseBlockTag("latest")` is rarely hit because callers bypass it, but as a public utility method its performance contract matters for future callers.

**Suggestion:** replace the `getNowBlock()` block-number extraction with `wallet.getHeadBlockNum()`. The race condition concern (block not yet in blockStore) is valid for paths needing the full Block object (like `getBlockByNumOrTag`), but `parseBlockTag` only returns the number — `DynamicPropertiesStore` is the authoritative source for that.

0xbigapple: Thanks for the review — the performance observation is spot on. `getNowBlock()` does deserialize a full `Block` protobuf (including all transactions), which is significantly heavier than `DynamicPropertiesStore.getLatestBlockHeaderNumber()`. However, the use of `getNowBlock()` here is intentional. `parseBlockTag` is a public utility method, and its contract should guarantee that the returned block number can be immediately used for subsequent queries. If the caller must understand the `DynamicPropertiesStore` vs `blockStore` timing gap to use it safely, the abstraction is leaky. 
You're right that this path is currently not hit for `"latest"`, but as a public utility method, its correctness contract matters for future callers. The cost of deserializing one block is acceptable compared to the subtle bug this prevents.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
waynercheung: [MUST] This still uses the old `ByteArray.hexToBigInteger(...).longValue()` conversion, which can silently truncate oversized block numbers. At minimum, this should be `longValueExact()` so out-of-range inputs fail as `invalid block number` instead of being reinterpreted. That would fix the correctness issue here without expanding the scope of this PR.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
waynercheung: [SHOULD] I think `parseBlockTag("latest")` should resolve via `wallet.getHeadBlockNum()` rather than `getNowBlock()`.

The current write ordering is `latestBlockHeaderNumber` -> `blockStore` -> `blockIndexStore`. Reading from `blockStore` does narrow the race relative to `DynamicPropertiesStore`, but neither approach guarantees that `blockIndexStore` is ready for the returned number, which is what subsequent `getBlockByNum(n)`-style lookups depend on.

Since neither approach guarantees that stronger contract, and this helper only returns a block number, I’d prefer the cheaper metadata source here and reserve `getNowBlock()` for block-returning paths such as `getBlockByNumOrTag("latest")`. If the narrower `blockStore` -> `blockIndexStore` window is intentional, a regression test demonstrating that intended contract would help.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -345,11 +342,49 @@ private byte[] hashToByteArray(String hash) throws JsonRpcInvalidParamsException
...
   private Block getBlockByJsonHash(String blockHash) throws JsonRpcInvalidParamsException {
     byte[] bHash = hashToByteArray(blockHash);
     return wallet.getBlockById(ByteString.copyFrom(bHash));
   }
 
+  private Block getBlockByNumOrTag(String blockNumOrTag) throws JsonRpcInvalidParamsException {
+    long blockNum;
+    if (JsonRpcApiUtil.isBlockTag(blockNumOrTag)) {
+      if (LATEST_STR.equalsIgnoreCase(blockNumOrTag)) {
+        // Read the head block directly from blockStore to avoid the window where
```
waynercheung: [NIT] Wording is a bit misleading here. Since this branch itself uses `getNowBlock()`, the phrase "avoid the window where getNowBlock() already sees the new head" reads backwards.

Maybe reword it to describe the actual race more directly: `latestBlockHeaderNumber` may already be updated while `blockIndexStore` is not yet written, so `getBlockByNum(n)` can still miss the head block.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
0xbigapple: Convinced. The key insight is that `parseBlockTag` only returns a block number, and all downstream callers go through `blockIndexStore` to fetch the actual block. `getNowBlock() `guarantees the block exists in `blockStore`, but callers never read `blockStore` directly — they go through `blockIndexStore`, which neither approach can guarantee. 
So the "correctness contract" I argued for earlier doesn't hold: the narrower blockStore→blockIndexStore window doesn't help callers in practice. Will switch to `wallet.getHeadBlockNum()`.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: Fixed.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: Fixed.

framework/src/main/java/org/tron/core/services/jsonrpc/filters/LogFilterWrapper.java
```diff
@@ -50,39 +52,50 @@ public LogFilterWrapper(FilterRequest fr, long currentMaxBlockNum, Wallet wallet
...
+        // toBlock = MAX_VALUE). If concrete: bounded query with fromBlock = min(toBlock,
+        // currentMaxBlockNum).
+        if (LATEST_STR.equalsIgnoreCase(fr.getToBlock())) {
           toBlockSrc = Long.MAX_VALUE;
+        } else {
+          toBlockSrc = JsonRpcApiUtil.parseBlockNumber(fr.getToBlock(), wallet);
         }
+        fromBlockSrc = min(toBlockSrc, currentMaxBlockNum);
+
+      } else if (!fromEmpty && toEmpty) {
```
waynercheung: [NIT] The `!toEmpty` / `!fromEmpty` checks are redundant here because the previous branches already exclude those cases, so static analysis reports them as always true when reached. This can be simplified to `else if (fromEmpty)` and `else if (toEmpty)`.

waynercheung:
[NIT] A few comments in this PR use the Unicode arrow `→`. This is readable, but I’d prefer plain ASCII `->` in source comments/Javadoc for consistency with the rest of the codebase and fewer tooling/font portability issues.

framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
```diff
@@ -491,23 +519,23 @@ public void testGetTrxBalance() {
...
+      Assert.assertEquals(TAG_NOT_SUPPORT_ERROR,
           e.getMessage());
     }
 
     try {
       tronJsonRpc.getTrxBalance("", "finalized");
       Assert.fail("Expected to be thrown");
     } catch (Exception e) {
-      Assert.assertEquals("TAG [earliest | pending | finalized] not supported",
+      Assert.assertEquals(TAG_NOT_SUPPORT_ERROR,
```
alan-eth, line 538: The `parseBlockTag("safe")` test at line 458 is a nice addition — `safe` was previously unhandled!

nit: Since `safe` is now part of `isBlockTag` and the error message was updated to include it, would it be worth adding at least one test case exercising `safe` through a state-read endpoint like `getTrxBalance`? The existing tests cover `earliest`, `pending`, and `finalized` through each endpoint but not `safe`:

```java
try {
  tronJsonRpc.getTrxBalance("", "safe");
  Assert.fail("Expected to be thrown");
} catch (Exception e) {
  Assert.assertEquals(TAG_NOT_SUPPORT_ERROR, e.getMessage());
}
```

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -345,11 +342,49 @@ private byte[] hashToByteArray(String hash) throws JsonRpcInvalidParamsException
     return bHash;
   }
 
+  /**
+   * Reject any block selector that is not "latest".
+   * Accepts "latest" silently; throws for other tags, numeric blocks, or invalid input.
+   */
+  private void requireLatestBlockTag(String blockNumOrTag)
```
alan-eth, line 349: Really clean extraction — the old code had the same tag-check + hex-parse + quantity-reject sequence copy-pasted across `getTrxBalance`, `getStorageAt`, `getABIOfSmartContract`, and `getCall`. Pulling it into `requireLatestBlockTag` with clear error stratification (tag error → number error → quantity error) is a significant readability win. 👏

framework/src/main/java/org/tron/core/services/jsonrpc/filters/LogFilterWrapper.java
0xbigapple: Good catch, updated.

0xbigapple:
> [NIT] A few comments in this PR use the Unicode arrow `→`. This is readable, but I’d prefer plain ASCII `->` in source comments/Javadoc for consistency with the rest of the codebase and fewer tooling/font portability issues.

Good catch, fixed.

framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
0xbigapple: Good point. Added a "safe" case to `testGetTrxBalance`, `testGetStorageAt`, `testGetABIOfSmartContract`, and `testGetCall`, mirroring the existing `earliest/pending/finalized `assertions.

alan-eth: Thanks for the thorough coverage — adding `safe` across all four endpoints is exactly the right call. Looks great!

waynercheung (commented):
[NIT] A few minor cleanups that can be done in a follow-up or squash:

1. `LogFilterWrapper.java:27-28` - there is still a double blank line between `toBlock` and the constructor; this should be a single blank line.
2. `LogFilterWrapperStrategyTest.java:164` - `filter` is unused in `testStrategy4_LatestGreaterThanSmallBlock_Throws`; the call can be just `createFilter(...)` since the exception is expected before assignment.
3. `LogFilterWrapperStrategyTest.java:19-22,126` - comments still use the Unicode arrow `→` while `LogFilterWrapper.java` was already switched to ASCII `->`; it would be better to align them for consistency.

framework/src/test/java/org/tron/core/services/jsonrpc/LogFilterWrapperStrategyTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.tron.core.Wallet;
+import org.tron.core.exception.jsonrpc.JsonRpcInvalidParamsException;
+import org.tron.core.services.jsonrpc.TronJsonRpc.FilterRequest;
+import org.tron.core.services.jsonrpc.filters.LogFilterWrapper;
+
+/**
+ * Verify LogFilterWrapper strategies match develop branch behavior.
+ *
+ * Four filter strategies based on parameter emptiness (develop branch semantics):
+ * - Strategy 1: Both fromBlock and toBlock are empty → (currentMaxBlockNum, Long.MAX_VALUE)
```
waynercheung: replacing '→' with '->'

0xbigapple:
> [NIT] A few minor cleanups that can be done in a follow-up or squash:
> 
> 1. `LogFilterWrapper.java:27-28` - there is still a double blank line between `toBlock` and the constructor; this should be a single blank line.
> 2. `LogFilterWrapperStrategyTest.java:164` - `filter` is unused in `testStrategy4_LatestGreaterThanSmallBlock_Throws`; the call can be just `createFilter(...)` since the exception is expected before assignment.
> 3. `LogFilterWrapperStrategyTest.java:19-22,126` - comments still use the Unicode arrow `→` while `LogFilterWrapper.java` was already switched to ASCII `->`; it would be better to align them for consistency.

Good catches, all three fixed.

framework/src/test/java/org/tron/core/services/jsonrpc/LogFilterWrapperStrategyTest.java
0xbigapple: Fixed

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -345,11 +342,49 @@ private byte[] hashToByteArray(String hash) throws JsonRpcInvalidParamsException
...
+      if (LATEST_STR.equalsIgnoreCase(blockNumOrTag)) {
+        // Return the head block directly from blockStore, bypassing blockIndexStore
+        // which may not yet be written when latestBlockHeaderNumber is already updated.
+        return wallet.getNowBlock();
+      }
+      return wallet.getBlockByNum(JsonRpcApiUtil.parseBlockTag(blockNumOrTag, wallet));
+    }
+    try {
+      blockNum = ByteArray.hexToBigInteger(blockNumOrTag).longValueExact();
+    } catch (Exception e) {
```
warku123, line 382: [SHOULD] No length guard before `hexToBigInteger`. Java's `BigInteger(String, int)` is O(n²), so a crafted oversized hex string (e.g. `"0x" + "a".repeat(10_000_000)`) can stall the thread and exhaust CPU. A `long` fits in at most 16 hex digits — rejecting inputs longer than 18 chars is sufficient:

```java
if (blockNumOrTag.length() > 18) {
    throw new JsonRpcInvalidParamsException(BLOCK_NUM_ERROR);
}
```

Same guard is needed in `requireLatestBlockTag`.

warku123: To add: the decimal path is actually worse. `ByteArray.hexToBigInteger` falls through to `new BigInteger(input, 10)` for inputs without a `0x` prefix, which is **O(n²)** in Java — versus O(n) for hex. With Jetty's default ~200 KB body limit, a caller can send a ~200,000-character decimal string and monopolize the thread far longer than the hex case. The length guard should cover both paths.

0xbigapple: Thanks for flagging — valid concern for both the hex and decimal paths. This refactor PR is scoped to behavior-preserving changes only, so I'd prefer not to expand it with a new input guard. The length-guard fix will land in a follow-up: https://github.com/tronprotocol/java-tron/pull/6693.

bladehan1 (approved):

framework/src/test/java/org/tron/core/services/jsonrpc/LogFilterWrapperStrategyTest.java
```diff
@@ -0,0 +1,170 @@
...
+    when(mockWallet.getHeadBlockNum()).thenReturn(CURRENT_MAX_BLOCK);
+    when(mockWallet.getSolidBlockNum()).thenReturn(CURRENT_MAX_BLOCK - 100);
+  }
+
+  private LogFilterWrapper createFilter(String fromBlock, String toBlock) throws Exception {
+    FilterRequest request = new FilterRequest(fromBlock, toBlock, null, null, null);
+    return new LogFilterWrapper(request, CURRENT_MAX_BLOCK, mockWallet, false);
+  }
+
+  // ============ Strategy 1: Both empty ============
```
waynercheung: [SHOULD] Avoid banner-style comments (`// ========= X =========`). They are visual noise and usually signal that the test is doing too much. Prefer splitting into separate @Test methods with descriptive names, or using parameterized tests. A single-line comment like `// KB tier` is sufficient if you must group within one method.

0xbigapple: Fixed.

317787106 (approved):

waynercheung (approved):
LGTM

PR #6671, opened 2026-04-09
feat(jsonrpc): add blockTimestamp to logs and receipts

### LINKED ISSUE #6617: [Feature] Add blockTimestamp to JSON-RPC log objects to improve efficiency
## Problem Statement

Currently, JSON-RPC log objects (such as those returned by `eth_getLogs` or within transaction receipts) do not include the `blockTimestamp`. Developers often need the block timestamp associated with an event/log to properly order, index, or display the events in their applications without having to make an additional `eth_getBlockByHash` or `eth_getBlockByNumber` call for every single log.

Furthermore, adding this field helps in aligning the TRON JSON-RPC implementation with the Ethereum `execution-apis` specification (specifically EIP PR [#639](https://github.com/ethereum/execution-apis/pull/639)), which introduced `blockTimestamp` to log objects.

## Proposed Solution

Add the `blockTimestamp` field to the log objects returned by JSON-RPC endpoints. This enhancement reduces the number of RPC calls needed by clients, improving efficiency and developer experience, while keeping TRON's Ethereum-compatible RPC interfaces aligned with the broader ecosystem standards.

### Specification

**API Changes**
JSON-RPC responses containing log objects will now include a new field. This directly affects the return values of the following JSON-RPC endpoints:
- `eth_getLogs`
- `eth_getFilterLogs`
- `eth_getFilterChanges`
- `eth_getTransactionReceipt`
- `eth_getBlockReceipts`

The new field added is:
- `blockTimestamp`: The timestamp of the block containing the log (as a hex string).

**Configuration Changes**
None.

**Protocol Changes** (if applicable)
None.

## Scope of Impact

- [x] API/RPC

**Breaking Changes**
No breaking changes. This only adds a new field to the response JSON.

**Backward Compatibility**
Fully backward compatible. Existing fields remain unchanged.

## Implementation

Yes, the implementation involves updating the `TransactionReceipt` and `LogMatch` classes in the `org.tron.core.services.jsonrpc` package to include and populate the `blockTimestamp` field when converting core protocol logs to JSON-RPC responses.

Yes

**Estimated Complexity**
Low (minor changes)

## Testing Strategy

The feature should be tested by modifying the existing JSON-RPC unit tests to verify that the `blockTimestamp` field is correctly populated in the returned JSON objects for both event queries and transaction receipt queries.

**Test Scenarios**
- `LogMatchExactlyTest`: Verify event logs contain the correct `blockTimestamp`.
- `TransactionReceiptTest`: Verify logs within transaction receipts contain the correct `blockTimestamp`.

The block timestamp is already available or easily accessible when constructing the log objects, so the performance impact is negligible.

0xbigapple:
**What does this PR do?**
Adds the `blockTimestamp` field to the log objects returned by JSON-RPC event and receipt query interfaces, aligning the TRON JSON-RPC implementation with the Ethereum execution-apis specification. close https://github.com/tronprotocol/java-tron/issues/6617

Specifically, it updates:
1. `LogFilterElement` (used in `eth_getLogs`, `eth_getFilterLogs`, `eth_getFilterChanges(about log/event filter)`) to include `blockTimestamp`.
2. `TransactionLog` within `TransactionReceipt` (used in `eth_getTransactionReceipt`, `eth_getBlockReceipts`) to include `blockTimestamp`.

The timestamp is extracted from the corresponding `TransactionInfo`'s `blockTimeStamp`, divided by 1000 to convert to seconds, and returned as a hexadecimal string (e.g., `"0x3e8"`).

To maintain compatibility with the evolving Ethereum JSON-RPC API standards. Adding the block generation timestamp directly to the log objects enables developers and dApps to process event streams with precise time context without needing to perform secondary `eth_getBlockByHash/Number` queries, significantly reducing network overhead and improving client efficiency.

- Unit Tests:
  - Added assertions in `LogMatchExactlyTest.java` to verify `blockTimestamp` in `LogFilterElement`.
  - Added assertions in `TransactionReceiptTest.java` to verify `blockTimestamp` in `TransactionReceipt.TransactionLog`.
- Manual Testing: N/A

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
@@ -466,7 +466,7 @@ public interface TronJsonRpc {
 
     public LogFilterElement(String blockHash, Long blockNum, String txId, Integer txIndex,
         String contractAddress, List<DataWord> topicList, String logData, int logIdx,
-        boolean removed, long blockTimestamp) {
+        boolean removed, long blockTimestampMs) {
       logIndex = ByteArray.toJsonHex(logIdx);
       this.blockNumber = blockNum == null ? null : ByteArray.toJsonHex(blockNum);
       this.blockHash = blockHash == null ? null : ByteArray.toJsonHex(blockHash);
@@ -479,7 +479,7 @@ public interface TronJsonRpc {
         topics[i] = ByteArray.toJsonHex(topicList.get(i).getData());
       }
       this.removed = removed;
-      this.blockTimestamp = ByteArray.toJsonHex(blockTimestamp / 1000);
+      this.blockTimestamp = ByteArray.toJsonHex(blockTimestampMs / 1000);
     }
 
     @Override
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
@@ -466,7 +466,7 @@ public interface TronJsonRpc {
 
     public LogFilterElement(String blockHash, Long blockNum, String txId, Integer txIndex,
         String contractAddress, List<DataWord> topicList, String logData, int logIdx,
-        boolean removed, Long blockTimestamp) {
+        boolean removed, long blockTimestamp) {
       logIndex = ByteArray.toJsonHex(logIdx);
       this.blockNumber = blockNum == null ? null : ByteArray.toJsonHex(blockNum);
       this.blockHash = blockHash == null ? null : ByteArray.toJsonHex(blockHash);
@@ -479,8 +479,7 @@ public interface TronJsonRpc {
         topics[i] = ByteArray.toJsonHex(topicList.get(i).getData());
       }
       this.removed = removed;
-      this.blockTimestamp = blockTimestamp == null ? null
-          : ByteArray.toJsonHex(blockTimestamp / 1000);
+      this.blockTimestamp = ByteArray.toJsonHex(blockTimestamp / 1000);
     }
 
     @Override
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
@@ -461,10 +461,12 @@ public interface TronJsonRpc {
     private final String[] topics;
     @Getter
     private final boolean removed;
+    @Getter
+    private final String blockTimestamp;
 
     public LogFilterElement(String blockHash, Long blockNum, String txId, Integer txIndex,
         String contractAddress, List<DataWord> topicList, String logData, int logIdx,
-        boolean removed) {
+        boolean removed, Long blockTimestamp) {
       logIndex = ByteArray.toJsonHex(logIdx);
       this.blockNumber = blockNum == null ? null : ByteArray.toJsonHex(blockNum);
       this.blockHash = blockHash == null ? null : ByteArray.toJsonHex(blockHash);
@@ -477,6 +479,8 @@ public interface TronJsonRpc {
         topics[i] = ByteArray.toJsonHex(topicList.get(i).getData());
       }
       this.removed = removed;
+      this.blockTimestamp = blockTimestamp == null ? null
+          : ByteArray.toJsonHex(blockTimestamp / 1000);
     }
 
     @Override
@@ -500,12 +504,16 @@ public interface TronJsonRpc {
       if (!Objects.equals(logIndex, item.logIndex)) {
         return false;
       }
-      return removed == item.removed;
+      if (removed != item.removed) {
+        return false;
+      }
+      return Objects.equals(blockTimestamp, item.blockTimestamp);
     }
 
     @Override
     public int hashCode() {
-      return Objects.hash(blockHash, transactionHash, transactionIndex, logIndex, removed);
+      return Objects.hash(blockHash, transactionHash, transactionIndex,
+          logIndex, removed, blockTimestamp);
     }
 
   }
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
@@ -109,7 +109,7 @@ public class TransactionReceipt {
 
     // Set logs
     List<TransactionLog> logList = new ArrayList<>();
-    String blockTimestamp = ByteArray.toJsonHex(txInfo.getBlockTimeStamp() / 1000);
+    String blockTimestamp = ByteArray.toJsonHex(blockCapsule.getTimeStamp() / 1000);
     for (int logIndex = 0; logIndex < txInfo.getLogCount(); logIndex++) {
       TransactionInfo.Log log = txInfo.getLogList().get(logIndex);
       TransactionLog transactionLog = new TransactionLog();
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
@@ -35,6 +35,7 @@ public class TransactionReceipt {
     private String data;
     private String[] topics;
     private boolean removed = false;
+    private String blockTimestamp;
 
     public TransactionLog() {}
   }
@@ -108,6 +109,7 @@ public class TransactionReceipt {
 
     // Set logs
     List<TransactionLog> logList = new ArrayList<>();
+    String blockTimestamp = ByteArray.toJsonHex(txInfo.getBlockTimeStamp() / 1000);
     for (int logIndex = 0; logIndex < txInfo.getLogCount(); logIndex++) {
       TransactionInfo.Log log = txInfo.getLogList().get(logIndex);
       TransactionLog transactionLog = new TransactionLog();
@@ -116,6 +118,7 @@ public class TransactionReceipt {
       transactionLog.setTransactionIndex(this.transactionIndex);
       transactionLog.setBlockHash(this.blockHash);
       transactionLog.setBlockNumber(this.blockNumber);
+      transactionLog.setBlockTimestamp(blockTimestamp);
 
       byte[] addressByte = convertToTronAddress(log.getAddress().toByteArray());
       transactionLog.setAddress(ByteArray.toJsonHexAddress(addressByte));
--- a/framework/src/test/java/org/tron/core/jsonrpc/LogMatchExactlyTest.java
+++ b/framework/src/test/java/org/tron/core/jsonrpc/LogMatchExactlyTest.java
@@ -37,6 +37,7 @@ public class LogMatchExactlyTest {
     LogInfo logInfo = new LogInfo(address, topics, data);
     logList.add(LogInfo.buildLog(logInfo));
     builder.addAllLog(logList);
+    builder.setBlockTimeStamp(1000000L);
 
     return builder.build();
   }
@@ -230,6 +231,8 @@ public class LogMatchExactlyTest {
       LogFilterElement logFilterElement1 = elementList.get(0);
       LogFilterElement logFilterElement2 = elementList2.get(0);
 
+      Assert.assertEquals("0x3e8", logFilterElement1.getBlockTimestamp());
+      Assert.assertEquals("0x3e8", logFilterElement2.getBlockTimestamp());
       Assert.assertEquals(logFilterElement1.hashCode(), logFilterElement2.hashCode());
       Assert.assertEquals(logFilterElement1, logFilterElement2);
 
--- a/framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java
+++ b/framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java
@@ -54,8 +54,11 @@ public class TransactionReceiptTest extends BaseTest {
 
     Protocol.Block block = Protocol.Block.newBuilder().setBlockHeader(
         Protocol.BlockHeader.newBuilder().setRawData(
-            Protocol.BlockHeader.raw.newBuilder().setNumber(1))).addTransactions(
-        transaction).build();
+            Protocol.BlockHeader.raw.newBuilder()
+                .setNumber(1)
+                .setTimestamp(1000000L)))
+        .addTransactions(transaction)
+        .build();
 
     BlockCapsule blockCapsule = new BlockCapsule(block);
     long energyFee = wallet.getEnergyFee(blockCapsule.getTimeStamp());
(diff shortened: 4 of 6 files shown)


framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
```diff
@@ -108,6 +109,7 @@ public TransactionReceipt(
 
     // Set logs
     List<TransactionLog> logList = new ArrayList<>();
+    String blockTimestamp = ByteArray.toJsonHex(txInfo.getBlockTimeStamp() / 1000);
```
xxo1shine: Why do you need to convert timestamp to this format `ByteArray.toJsonHex(txInfo.getBlockTimeStamp() / 1000)`

waynercheung: The conversion to `ByteArray.toJsonHex(... / 1000)` is correct for the JSON-RPC surface.

`TransactionInfo.blockTimeStamp` is stored in TRON internal milliseconds, while the Ethereum-compatible JSON-RPC layer uses hex-encoded Unix seconds for timestamp quantities. This is also consistent with the existing block-level `timestamp` field in `BlockResult`, which serializes `blockCapsule.getTimeStamp() / 1000` via `ByteArray.toJsonHex(...)`.

My concern here is the source rather than the format: this constructor already has `blockCapsule`, so the canonical value for a receipt log's `blockTimestamp` should come from the containing block. If `txInfo.getBlockTimeStamp()` is unset/default in older or partially populated records, this can serialize `0x0` even though the block has a real timestamp. Please switch this to `blockCapsule.getTimeStamp()`, or at least add an explicit fallback when `txInfo.getBlockTimeStamp() == 0`.

framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java
```diff
@@ -90,6 +91,7 @@ public void testTransactionReceipt() throws JsonRpcInternalException {
     Assert.assertEquals(transactionReceipt.getLogs()[0].getBlockNumber(), "0x1");
     Assert.assertEquals(transactionReceipt.getLogs()[0].getTransactionHash(), "0x31");
     Assert.assertEquals(transactionReceipt.getLogs()[0].getTransactionIndex(), "0x0");
+    Assert.assertEquals(transactionReceipt.getLogs()[0].getBlockTimestamp(), "0x3e8");
```
waynercheung: Good to assert the concrete value here.

However, this is still a constructor-level unit test, while the feature request is specifically about the public JSON-RPC response shape. Please also add an API-facing assertion in `JsonrpcServiceTest` that `eth_getTransactionReceipt` and `eth_getBlockReceipts` expose `blockTimestamp` with the expected hex-seconds value on the serialized response. The existing receipt equality check is useful, but it can still miss a false positive if both paths shape the same incomplete object.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchExactlyTest.java
```diff
@@ -230,6 +231,8 @@ public void testMatchBlock() {
       LogFilterElement logFilterElement1 = elementList.get(0);
       LogFilterElement logFilterElement2 = elementList2.get(0);
 
+      Assert.assertEquals("0x3e8", logFilterElement1.getBlockTimestamp());
```
waynercheung, line 234: This is the right kind of value assertion, but it only verifies the construction helper.

Since the compatibility change is on the external JSON-RPC response model, we should also add endpoint-level coverage for `eth_getLogs`, `eth_getFilterLogs`, and log-type `eth_getFilterChanges` to prove that the new field is actually present on the API boundary, not just on the intermediate object.

framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionReceipt.java
0xbigapple: Thanks for the review. I've traced the data path and checked on-chain history to better understand the concern — sharing my findings here for discussion.

**Data source:**

The `transactionInfo` in the `TransactionReceipt` constructor comes from `Wallet.getTransactionInfoById()`, which reads from `TransactionRetStore` (keyed by block number) or falls back to `TransactionHistoryStore` (keyed by tx ID). Both return `TransactionInfo` protobuf objects where `blockTimeStamp` is field 4.

**On-chain verification:**

I queried early mainnet data (block 1,000,000, ~July 2018) via TronGrid:

`https://api.trongrid.io/wallet/gettransactioninfobyid?value=6937ca320cec3fdb15e421da211994067d6766195e68dc7026a37597852e166b`

The response includes `blockTimeStamp: 1532895918000`, suggesting the field has been populated since the early days of mainnet.

**Question on snapshot scenarios:**

That said, I'd like to understand if your concern is related to something like #6401, where a snapshot-synced node returned `0x0` for `blockNumber` and `blockHash` on older transactions — indicating snapshot data can have incomplete fields. If `blockTimeStamp` could be similarly affected, the impact would go beyond this constructor. For example, `LogMatch.matchBlock()` directly passes `transactionInfo.getBlockTimeStamp()` to `LogFilterElement` for `eth_getLogs`, and that code path has no `blockCapsule` available as a fallback:

```java
new LogFilterElement(blockHash, blockNum, ..., transactionInfo.getBlockTimeStamp());
```

My initial thinking is that snapshot data integrity issues are better resolved at the snapshot level rather than through code-level workarounds. But I may be missing other scenarios — if there are cases beyond snapshots where `blockTimeStamp` could be unset, I'd like to understand the full picture so we can decide on the right approach together.

Looking forward to your thoughts.

waynercheung: @0xbigapple  Thanks for tracing the storage path and checking historical mainnet data. That does increase confidence that `TransactionInfo.blockTimeStamp` is normally populated.

My concern here is still source-of-truth selection rather than format or historical coverage. On this code path we already have `blockCapsule`, and we already use it as the authoritative source for other block-bound values such as `blockHash`, `blockNumber`, and the `block timestamp` used to derive `effectiveGasPrice`. In that context, using `txInfo.getBlockTimeStamp()` for **blockTimestamp** is unnecessarily weaker and inconsistent. Since the block is already loaded here, I still recommend using `blockCapsule.getTimeStamp()` directly, or at least falling back to it when `txInfo.getBlockTimeStamp() == 0`. I would treat the LogMatch path separately, since that path does not currently have the block object available.

0xbigapple: OK— from a source-of-truth consistency perspective, that makes sense. I've updated the `TransactionReceipt` constructor to use `blockCapsule.getTimeStamp()` directly, aligning it with `blockHash`, `blockNumber`.

framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java
0xbigapple: Agreed. I've added an explicit `blockTimestamp` assertion in `testGetBlockReceipts`, verifying the field value from both `eth_getBlockReceipts` and `eth_getTransactionReceipt` responses.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchExactlyTest.java
0xbigapple: Good point. The existing `getLogs`-related tests in `JsonrpcServiceTest` only covered error paths (block range exceeded, max topics exceeded) without any positive-path field validation. I've added a new `testGetLogs` that covers the positive path for `eth_getLogs`, including field-level assertions for `blockNumber`, `blockHash`, `logIndex`, `removed`, `topics`,`data`, and `blockTimestamp`. 
As for `eth_getFilterLogs` and `eth_getFilterChanges`, they share the same underlying
  `getLogsByLogFilterWrapper()` → `LogMatch.matchBlock()` path with `eth_getLogs`, differing only in how the filter is sourced. `testGetLogs` already covers the shared code path, so no additional tests are needed.

framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java
```diff
@@ -32,6 +32,7 @@ public void testTransactionReceipt() throws JsonRpcInternalException {
     Protocol.TransactionInfo transactionInfo = Protocol.TransactionInfo.newBuilder()
         .setId(ByteString.copyFrom("1".getBytes()))
         .setContractAddress(ByteString.copyFrom("address1".getBytes()))
+        .setBlockTimeStamp(1000000L)
         .setReceipt(Protocol.ResourceReceipt.newBuilder()
```
warku123, line 36: **framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java:35**

[Nit] Since `TransactionReceipt` now derives `blockTimestamp` from `blockCapsule.getTimeStamp()` rather than `txInfo.getBlockTimeStamp()`, this `.setBlockTimeStamp(1000000L)` on the `TransactionInfo` builder is no longer exercised by the code under test. Keeping it here is slightly misleading — it could give future maintainers the impression that the log timestamp still comes from `TransactionInfo`, and it might mask a regression if the source were accidentally switched back later. Would it make sense to remove this line for clarity?

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
```diff
@@ -477,6 +479,8 @@ public LogFilterElement(String blockHash, Long blockNum, String txId, Integer tx
         topics[i] = ByteArray.toJsonHex(topicList.get(i).getData());
       }
       this.removed = removed;
+      this.blockTimestamp = blockTimestamp == null ? null
```
bladehan1: **Null check unreachable from current call site**

The `blockTimestamp == null` branch is unreachable: the only call site (`LogMatch.matchBlock`) passes `transactionInfo.getBlockTimeStamp()`, which returns a primitive `long` (protobuf default 0), auto-boxed to a non-null `Long`. For blocks with an unset timestamp, this produces `"0x0"` rather than `null`.

**Suggestion:** Consider changing the parameter type to `long` (primitive) to match the actual contract, or keep `Long` and add a `blockTimestamp <= 0` check if `null` output is desired for unset timestamps.

bladehan1 (commented):
See review, comments

framework/src/test/java/org/tron/core/services/jsonrpc/TransactionReceiptTest.java
0xbigapple: Thanks for the review. In production, `txInfo.blockTimeStamp `is always set from the block header, so they're always consistent. The test sets both to the same value to reflect this real-world behavior.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
0xbigapple: Thanks for the suggestion. Changed the parameter type to primitive long and removed the null check.

waynercheung:
[SHOULD] The implementation looks good and I’m comfortable approving it. One follow-up I’d still recommend is adding direct API-level coverage for `eth_getFilterLogs` and `eth_getFilterChanges` in `JsonrpcServiceTest`, so the test surface fully matches the public APIs listed in the PR scope.

0xbigapple:
> [SHOULD] The implementation looks good and I’m comfortable approving it. One follow-up I’d still recommend is adding direct API-level coverage for `eth_getFilterLogs` and `eth_getFilterChanges` in `JsonrpcServiceTest`, so the test surface fully matches the public APIs listed in the PR scope.

Thanks for the follow-up. Since `eth_getFilterLogs` and `eth_getFilterChanges` share the same underlying code path as this change, the positive-path test on `eth_getLogs` already covers the shared logic including `blockTimestamp` validation. Adding full API-level tests for those two endpoints is beyond the scope of this change, as they didn't have tests before this PR either. That said, I think it would be great to have a dedicated tracking issue to cover API-level tests for all JSON-RPC endpoints holistically, not just these two.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
```diff
@@ -477,6 +479,7 @@ public LogFilterElement(String blockHash, Long blockNum, String txId, Integer tx
         topics[i] = ByteArray.toJsonHex(topicList.get(i).getData());
       }
       this.removed = removed;
+      this.blockTimestamp = ByteArray.toJsonHex(blockTimestamp / 1000);
```
317787106: They have the same varible name `blockTimestamp`, suggest rename them to distinguish.  Such as `String blockTimestamp` -> `String blockTimestampHex`

0xbigapple: Thanks for the suggestion. I'd like to discuss this one before changing it.

The `String blockTimestamp` field is the serialized JSON property exposed to clients. Since it's annotated with `@Getter`, Jackson uses the field name directly as the response property name. Renaming it to `blockTimestampHex` would change the public JSON-RPC response shape from:

```json
{ "blockTimestamp": "0x..." }
```

to:

```json
{ "blockTimestampHex": "0x..." }
```

That breaks Ethereum JSON-RPC compatibility, which is exactly what this PR is trying to align with. Preserving the name would require adding `@JsonProperty("blockTimestamp")` just to satisfy an internal naming concern, which I'd prefer to avoid.

The current code also follows the standard Java `this.x = x` constructor pattern — `blockHash` and `removed` in the same constructor use the same field/parameter name.

If the concern is distinguishing the raw `long` (milliseconds) from the hex `String` (seconds), I think it's more accurate to rename the **parameter** to reflect its unit, e.g. `long blockTimestampMs`. Happy to make that change if you think it adds clarity. Let me know your thoughts.

317787106: Using `long blockTimestampMs` is OK.

waynercheung:
> That said, I think it would be great to have a dedicated tracking issue to cover API-level tests for all JSON-RPC endpoints holistically, not just these two.

@0xbigapple  Thanks, that makes sense to me overall.

I agree this does not need to block the current PR. For `eth_getFilterLogs`, the existing `eth_getLogs` positive-path  test gives reasonable confidence because they share the same query path. My only nuance is that `eth_getFilterChanges` does not fully share the same API flow, since it goes through filter creation / queueing / pop semantics, even though it does reuse the same `LogFilterElement` construction logic.

 I’m fine treating that as follow-up rather than scope for this change. If you file a tracking issue for broader JSON-RPC API-level coverage, please link it here so we have a concrete follow-up item.

0xbigapple:
> > That said, I think it would be great to have a dedicated tracking issue to cover API-level tests for all JSON-RPC endpoints holistically, not just these two.
> 
> @0xbigapple Thanks, that makes sense to me overall.
> 
> I agree this does not need to block the current PR. For `eth_getFilterLogs`, the existing `eth_getLogs` positive-path test gives reasonable confidence because they share the same query path. My only nuance is that `eth_getFilterChanges` does not fully share the same API flow, since it goes through filter creation / queueing / pop semantics, even though it does reuse the same `LogFilterElement` construction logic.
> 
> I’m fine treating that as follow-up rather than scope for this change. If you file a tracking issue for broader JSON-RPC API-level coverage, please link it here so we have a concrete follow-up item.

Thanks for the clarification on `eth_getFilterChanges` — you're right that the filter creation / queueing / pop flow is distinct from the direct query path, so sharing `LogFilterElement` construction isn't full coverage.

Agreed this is better tackled as a dedicated effort rather than squeezed into this PR. I don't have the bandwidth to open the tracking issue right away, but I'll circle back and file it when I do, and link it here.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
0xbigapple: Ok, I have fixed it in the last commit.

waynercheung:
> Agreed this is better tackled as a dedicated effort rather than squeezed into this PR. I don't have the bandwidth to open the tracking issue right away, but I'll circle back and file it when I do, and link it here.

Sounds good. Agree this can be handled as follow-up and does not need to block this PR. Please link the tracking issue  here once it is filed.

wink-link (approved):

waynercheung (approved):
LGTM

bladehan1 (approved):

317787106 (approved):

PR #6673, opened 2026-04-10
build(p2p): internalize libp2p as local module

barbatos2011:
## What

Internalize `io.github.tronprotocol:libp2p:2.2.7` source code as a new `p2p/` gradle submodule, replacing the external Maven dependency. No functional changes to P2P behavior, protocol, or consensus.

## Why

Upstream `tronprotocol/libp2p` activity has been **intermittent**: the repo saw sparse maintenance from 2023-04 to 2024-10 (0–3 commits/month, several months with no commits), with renewed activity in 2025-03 (v2.2.6) and 2025-08 to 2025-11 (v2.2.7). Cross-repo maintenance creates friction:

- **Version coordination overhead**: Modifying P2P behavior requires publishing libp2p first, then updating java-tron's dependency version
- **Debugging difficulty**: IDE cannot jump into external artifact source; requires downloading source JARs
- **CI dependency**: Build depends on external Maven/JitPack artifact availability
- **Industry alignment**: geth (go-ethereum) maintains its `p2p/` package within the monorepo; this follows the same pattern

Internalization makes future P2P changes a single PR instead of a cross-repo dance, and does not preclude back-porting individual commits to upstream libp2p when appropriate.

### Commit 1: `chore(p2p)`: add libp2p v2.2.7 source as p2p module

- Copy 90 main Java files (including embedded `org.web3j` crypto utils) + 2 proto files + 26 test files from `tronprotocol/libp2p` tag `v2.2.7`
- Create `p2p/build.gradle` with protobuf plugin configuration
- Add `p2p/lombok.config` to temporarily override root logger field name (`log` vs `logger`)
- Move example code (`StartApp`) to `src/test/java` (not packaged in JAR, but available for reference)
- Register module in `settings.gradle`, update `gradle/verification-metadata.xml`

### Commit 2: `style(p2p)`: replace log field with logger

- Replace all `@Slf4j`-generated `log` references with `logger` (35 files) to match root `lombok.config` (`lombok.log.fieldName=logger`)
- Remove temporary `p2p/lombok.config` override

### Commit 3: `build(common)`: switch from external libp2p to local p2p module

- `common/build.gradle`: replace 17 lines (Maven dependency + 11 exclude rules) with single `api project(':p2p')`
- Remove `io.github.tronprotocol:libp2p:2.2.7` entry from `gradle/verification-metadata.xml`

### Commit 4: `style(p2p)`: fix checkstyle violations and enable checkstyle

- Apply `google-java-format` for bulk formatting (indentation, imports, whitespace)
- Manual fixes: lambda indentation, line length wrapping, empty catch blocks, star imports
- Enable checkstyle plugin with proto-generated code excluded (`exclude '**/protos/**'`)
- ~1374 of 1454 violations were from proto-generated code; after exclusion only 80 remained in handwritten code

### Commit 5: `fix(p2p)`: address review findings

- Replace `BasicThreadFactory.builder()` (requires commons-lang3 3.12+) with `new BasicThreadFactory.Builder()` (available since 3.0) to prevent implicit global version upgrade from 3.4 to 3.18.0
- Remove unused `grpc-core` dependency; retain `grpc-netty` as Netty transport provider (p2p uses 30+ Netty imports for TCP/UDP but zero gRPC imports)
- Remove gRPC protoc plugin from protobuf config (p2p protos define only messages, no services)
- Add `p2p/.gitignore` to exclude proto-generated code for all developers

### Commit 6: `fix(p2p)`: replace Math with StrictMath and fix flaky network tests

- Replace `Math.min/max` with `StrictMath.min/max` in 5 files to satisfy CI `check-math` rule (integer operations, results identical)
- Rewrite `NetUtilTest.testGetIP` with mocked `URLConnection` — original test called 3 external IP services which failed in CI (libp2p's own CI never ran tests with `-xtest`)
- New mock tests cover: valid IP response, connection failure, invalid IP, empty response
- Remove `testExternalIp` (covered by mock tests), fix `testGetLanIP` to not depend on external connectivity

### Commit 7: `test(p2p)`: improve test coverage from 35% to 76%

Add comprehensive unit tests to meet the CI coverage gate (changed files > 60%, overall delta < -0.1%).

- 27 new test files across `connection/`, `dns/`, `discover/`, `utils/web3j`, `stats`, `exception`, plus `P2pService` and `P2pConfig`
- Exclude proto-generated code from jacoco coverage (consistent with the checkstyle exclusion)

### Commit 8: `chore(p2p)`: move tests to framework, introduce example sourceSet

Addresses review feedback from @xxo1shine: align with java-tron's project-wide convention of consolidating module unit tests under `framework/src/test/`.

- Move 53 unit test files from `p2p/src/test/java/` to `framework/src/test/java/` (+4 from `org.web3j` package); 100% git renames, package names preserved
- Add `testImplementation` for AWS Route53 + Aliyun DNS SDKs in `framework/build.gradle` (required by 3 DNS-publish tests moved here; these SDKs are `implementation`-scope in `:p2p` so not transitively visible). Minimal excludes only.
- Fix pre-existing flaky `DnsManagerTest` NPE by adding `@Before initP2pConfig()` (test result depended on whether prior tests had left `Parameter.p2pConfig` non-null)
- Drop p2p's `jacocoTestReport` block (no local tests now; coverage is captured by `:framework:jacocoTestReport`)
- Move reference examples (`StartApp`, `DnsExample1/2`, `ImportUsing`, `README.md`) to a dedicated `p2p/src/example/` sourceSet — compiles for API validation but **not packaged in `p2p.jar`** and **not executed as tests**

### Commit 9: `fix(p2p)`: drop duplicate srcDir calls in example sourceSet

Removes the explicit `srcDir 'src/example/{java,resources}'` calls inside the example sourceSet — Gradle already defaults custom-sourceSet srcDirs to those paths, so the explicit calls duplicate them and broke `:p2p:processExampleResources` with "Entry README.md is a duplicate" on CI.

## Key Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Module name | `p2p` | Matches package `org.tron.p2p`, geth `p2p/` convention, project short-name style (`common`, `crypto`, etc.) |
| Proto files | Stay in p2p module | P2P protocol is separate from chain protocol in `protocol/`; avoids coupling |
| Unit test location | `framework/src/test/` | Project convention — `actuator`, `chainbase`, `consensus`, `common` etc. all keep their unit tests under framework |
| Example code | Dedicated `p2p/src/example/` sourceSet | Compiled for API validation, not packaged in jar, not run as tests; preserves passive library usability |
| commons-lang3 | 3.4 (global consistent) | Rewrite `builder()` → `new Builder()` (3.0 API) instead of upgrading globally |
| grpc-netty retained | Yes | Provides Netty transitively; same as original libp2p; version auto-aligns with protocol module |
| grpc-core + grpc plugin removed | Yes | p2p protos have no `service` definitions, zero `io.grpc` imports in source |
| Checkstyle proto exclude | `**/protos/**` | Proto-generated code has ~1374 violations; consistent with protocol module approach |
| `Math` → `StrictMath` | All 5 call sites | CI `check-math` rule; integer min/max results are identical |
| `testGetIP` mock rewrite | URLConnection mock | Original depended on external services; libp2p CI never ran tests |

## Scope

- Does NOT change any P2P protocol behavior or message formats
- Does NOT change consensus rules or transaction validation logic
- Does NOT modify protobuf definitions
- Does NOT require hard fork
- All files importing `org.tron.p2p` are unchanged (package name preserved)
- Existing nodes upgrading need only `git pull` and rebuild
- p2p module remains passively usable as a standalone library: self-contained code, default `p2p.jar` artifact, examples kept as compilable reference

## Known Issues (pre-existing in libp2p)

The following issues exist in the original libp2p v2.2.7 codebase and are NOT introduced by this PR. They will be addressed in follow-up PRs:

**Critical**:
- Node ID generation uses weak PRNG (`java.util.Random` instead of `SecureRandom`) — eclipse attack vector
- Handshake lacks cryptographic authentication (self-reported node ID, no signature verification)
- Neighbor injection has no count limit per response

**High**:
- TOCTOU race in `ChannelManager.handMessage` (double onConnect)
- DNS sync uses predictable `java.util.Random`
- `Algorithm.compressPubKey` drops leading zeros from public key X-coordinate

**Medium** (from Cubic AI review):
- `Node.equals` compares byte[] via `new String(id)` instead of `Arrays.equals`
- `Channel.hashCode()` NPE when `inetSocketAddress` is null
- `InterruptedException` swallowed in DiscoverServer and PeerClient
- Several minor NPE risks and unclosed resources

## Test

- p2p module compile + example sourceSet compile: PASS
- Full build: BUILD SUCCESSFUL (~14m, **2780 tests**, 0 real failures; 1 pre-existing flaky `testPbft` passed on retry)
- `checkstyleMain` + `checkstyleTest`: PASS for all modules including p2p
- `check-math`: PASS
- Node integration test: FullNode started with mainnet config, P2P handshake protocol verified by message exchange (HelloMessage / TOO_MANY_PEERS / channel close)
- Flaky net tests (PbftMsgHandler, TronStatsManager): confirmed pre-existing, pass on retry


xxo1shine:
@barbatos2011 What is the primary motivation for integrating libp2p into java-tron at this stage?
Also, considering that libp2p is undergoing protocol upgrades, would it be more appropriate to wait until the protocol stabilizes before proceeding with the integration?

xxo1shine:
@barbatos2011 Should the unit tests related to the libp2p module be moved to the framework module?

barbatos2011:
@xxo1shine Thanks for the questions. Let me address both:

### On the motivation for integrating now

The main driver is **reducing cross-repo maintenance friction**, not adding features. Three concrete pain points with the current setup:

1. **Version coordination overhead** — any P2P change today requires: modify `tronprotocol/libp2p` → publish artifact → bump `java-tron`'s dependency version → CI roundtrip. Post-internalization, it's a single PR, which actually **accelerates** future changes.

2. **Debugging difficulty** — IDE navigation into external artifact source requires downloading source JARs; stack traces point to decompiled code during incidents.

3. **Build-time external dependency** — CI depends on Maven/JitPack artifact availability for every build.

This follows the **same pattern as go-ethereum**, which keeps its `p2p/` package in the monorepo for the same reasons.

### On waiting for protocol stability

Fair concern, but the data doesn't support "active upgrade" framing:

- From 2023-04 to 2024-10 the upstream repo saw only sporadic maintenance (0-3 commits/month, many months with none). Development resumed properly in 2025-03 (v2.2.6) and 2025-08~11 (v2.2.7). So "waiting for stability" doesn't match the pattern — the repo isn't mid-upgrade, it's in a pattern of long quiet periods with release bursts.
- There is no published roadmap for incoming protocol changes.
- Internalization is explicitly **scoped to zero functional change** — no P2P behavior, protocol format, or consensus logic is modified. The PR is purely a build/source-layout change, so the risk surface of integration itself is minimal.

If libp2p protocol upgrades do come later, they can be applied as normal PRs to the internalized module — in fact that will be **easier** than the current cross-repo flow. Waiting would mean indefinite deferral without a clear signal for what "stable" means here.

Happy to discuss further if you see specific upgrade work that would be disrupted by this change.

xxo1shine (approved):
ok

barbatos2011:
Done in 7858e437.

**What was moved**

- 53 unit test files from `p2p/src/test/java/` → `framework/src/test/java/` (+4 from `org.web3j` package).
- All are 100% git renames; package names preserved.

**What stays in `:p2p`**

Only reference/example code from upstream libp2p v2.2.7, moved to a dedicated Gradle sourceSet:

```
p2p/src/example/java/org/tron/p2p/example/
  StartApp.java, DnsExample1.java, DnsExample2.java, ImportUsing.java
p2p/src/example/resources/
  README.md
```

The `example` sourceSet compiles against `main`'s output (so API changes surface as compile errors) but is **not** packaged in `p2p.jar` and **not** executed as tests.

**Other changes bundled in**

- `framework/build.gradle`: added `testImplementation` for route53 + aliyun DNS SDKs (required by 3 DNS-publish tests moved here; these SDKs are `implementation`-scope in `:p2p` so not transitively visible).
- `DnsManagerTest`: added `@Before initP2pConfig()` to fix a pre-existing flaky NPE — the test's result depended on whether prior tests had left `Parameter.p2pConfig` non-null.
- `p2p/build.gradle`: removed the now-empty `jacocoTestReport` block (no local tests; coverage is captured by `:framework:jacocoTestReport`).

**Verified**

- `:p2p:compileJava` / `:p2p:compileExampleJava` / `:p2p:jar` ✓
- `:framework:compileTestJava` ✓
- Full test suite `./gradlew test`: 2780 tests, 0 real failures (1 flaky `testPbft` pre-existing, passed on retry).
- `:framework:buildFullNodeJar`: p2p + web3j classes intact (163).
- Manual node start: P2P handshake protocol works correctly (verified HelloMessage exchanges with mainnet peers).

PR #6683, opened 2026-04-16
fix(crypto): deprecate events field in shielded TRC20 scan APIs

Federico2014:
**What does this PR do?**

Deprecates the `events` field in shielded TRC20 scan API requests and removes server-side log-type filtering.

1. Marked the `events` field as `[deprecated = true]` in `IvkDecryptTRC20Parameters` and `OvkDecryptTRC20Parameters` request messages in `api.proto`
2. Added runtime rejection: gRPC returns `INVALID_ARGUMENT`, HTTP returns 400 when the `events` field contains at least one non-empty string. An `events` list that is absent or contains only empty strings is treated as not set and passes through normally.
3. Removed `topicList` parameter from internal method signatures in `Wallet`, `RpcApiService`, and HTTP servlet handlers (`ScanShieldedTRC20NotesByIvkServlet`, `ScanShieldedTRC20NotesByOvkServlet`)
4. All log types (Mint/Transfer/Burn) are now always inspected during a scan
5. Added gRPC and HTTP test coverage for both the rejection path and the empty-events pass-through path across Full, Solidity, and PBFT stubs

The `events` field allowed callers to filter shielded TRC20 event classes by passing user-controlled log topic strings that were hashed and matched server-side. This was an unnecessary attack surface — callers should always scan all log types and filter client-side. Removing the field simplifies the API contract and eliminates potential information leakage through selective scanning.

⚠️ Breaking Changes:
- `scanShieldedTRC20NotesByIvk/Ovk` — the `events` field is rejected when it contains at least one non-empty value (`INVALID_ARGUMENT` on gRPC, HTTP 400). An absent field or a field containing only empty strings is accepted and treated as if the field were not set. Clients still sending non-empty event values must remove them. Clients that relied on the field to limit scan scope will now receive results for all log types.
- Other malformed inputs (e.g. invalid block index format in GET) continue to return the existing error response and are not affected by this change.

- [x] Unit Tests — `RpcApiServicesTest`, `WalletMockTest`, `ShieldedTRC20BuilderTest`
- [x] Build passes (`./gradlew clean build`)

**Follow up**

No consensus logic or on-chain state transitions affected.

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -49,7 +49,6 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Range;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
-import com.google.protobuf.ProtocolStringList;
 import java.math.BigInteger;
 import java.security.SignatureException;
 import java.util.ArrayList;
@@ -163,7 +162,6 @@ import org.tron.core.capsule.ProposalCapsule;
 import org.tron.core.capsule.TransactionCapsule;
 import org.tron.core.capsule.TransactionInfoCapsule;
 import org.tron.core.capsule.TransactionResultCapsule;
-import org.tron.core.capsule.TransactionRetCapsule;
 import org.tron.core.capsule.VotesCapsule;
 import org.tron.core.capsule.WitnessCapsule;
 import org.tron.core.capsule.utils.MarketUtils;
@@ -780,7 +778,7 @@ public class Wallet {
     if (limit > WITNESS_COUNT_LIMIT_MAX) {
       limit = WITNESS_COUNT_LIMIT_MAX;
     }
-    
+
     /*
       In the maintenance period, the VoteStores will be cleared.
       To avoid the race condition of VoteStores deleted but Witness vote counts not updated,
@@ -1502,8 +1500,8 @@ public class Wallet {
     builder.addChainParameter(Protocol.ChainParameters.ChainParameter.newBuilder()
         .setKey("getAllowTvmSelfdestructRestriction")
         .setValue(dbManager.getDynamicPropertiesStore().getAllowTvmSelfdestructRestriction())
-        .build());                      
-    
+        .build());
+
     builder.addChainParameter(Protocol.ChainParameters.ChainParameter.newBuilder()
         .setKey("getProposalExpireTime")
         .setValue(dbManager.getDynamicPropertiesStore().getProposalExpireTime())
@@ -3852,8 +3850,8 @@ public class Wallet {
     return builder.build(false);
   }
 
-  private int getShieldedTRC20LogType(TransactionInfo.Log log, byte[] contractAddress,
-      ProtocolStringList topicsList) throws ZksnarkException {
+  private int getShieldedTRC20LogType(TransactionInfo.Log log, byte[] contractAddress)
+      throws ZksnarkException {
     byte[] logAddress = log.getAddress().toByteArray();
     byte[] addressWithoutPrefix = new byte[20];
     if (ArrayUtils.isEmpty(contractAddress) || contractAddress.length != 21) {
@@ -3866,33 +3864,14 @@ public class Wallet {
       for (ByteString bs : logTopicsList) {
         topicsBytes = ByteUtil.merge(topicsBytes, bs.toByteArray());
       }
-      if (Objects.isNull(topicsList) || topicsList.isEmpty()) {
-        if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_MINT)) {
-          return 1;
-        } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_TRANSFER)) {
-          return 2;
-        } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_BURN_LEAF)) {
-          return 3;
-        } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_BURN_TOKEN)) {
-          return 4;
-        }
-      } else {
-        for (String topic : topicsList) {
-          byte[] topicHash = Hash.sha3(ByteArray.fromString(topic));
-          if (Arrays.equals(topicsBytes, topicHash)) {
-            if (topic.toLowerCase().contains("mint")) {
-              return 1;
-            } else if (topic.toLowerCase().contains("transfer")) {
-              return 2;
-            } else if (topic.toLowerCase().contains("burn")) {
-              if (topic.toLowerCase().contains("leaf")) {
-                return 3;
-              } else if (topic.toLowerCase().contains("token")) {
-                return 4;
-              }
-            }
-          }
-        }
+      if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_MINT)) {
+        return 1;
+      } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_TRANSFER)) {
+        return 2;
+      } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_BURN_LEAF)) {
+        return 3;
+      } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_BURN_TOKEN)) {
+        return 4;
       }
     }
     return 0;
@@ -3942,8 +3921,7 @@ public class Wallet {
   }
 
   private DecryptNotesTRC20 queryTRC20NoteByIvk(long startNum, long endNum,
-      byte[] shieldedTRC20ContractAddress, byte[] ivk, byte[] ak, byte[] nk,
-      ProtocolStringList topicsList)
+      byte[] shieldedTRC20ContractAddress, byte[] ivk, byte[] ak, byte[] nk)
       throws BadItemException, ZksnarkException, ContractExeException {
     if (!(startNum >= 0 && endNum > startNum && endNum - startNum <= 1000)) {
       throw new BadItemException(
@@ -3964,7 +3942,7 @@ public class Wallet {
             Optional<DecryptNotesTRC20.NoteTx> noteTx;
             int index = 0;
             for (TransactionInfo.Log log : logList) {
-              int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress, topicsList);
+              int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress);
               if (logType > 0) {
                 noteBuilder = DecryptNotesTRC20.NoteTx.newBuilder();
                 noteBuilder.setTxid(ByteString.copyFrom(txId));
@@ -4048,12 +4026,11 @@ public class Wallet {
 
   public DecryptNotesTRC20 scanShieldedTRC20NotesByIvk(
       long startNum, long endNum, byte[] shieldedTRC20ContractAddress,
-      byte[] ivk, byte[] ak, byte[] nk, ProtocolStringList topicsList)
+      byte[] ivk, byte[] ak, byte[] nk)
       throws BadItemException, ZksnarkException, ContractExeException {
     checkAllowShieldedTransactionApi();
 
-    return queryTRC20NoteByIvk(startNum, endNum,
-        shieldedTRC20ContractAddress, ivk, ak, nk, topicsList);
+    return queryTRC20NoteByIvk(startNum, endNum, shieldedTRC20ContractAddress, ivk, ak, nk);
   }
 
   private Optional<DecryptNotesTRC20.NoteTx> getNoteTxFromLogListByOvk(
@@ -4127,7 +4104,7 @@ public class Wallet {
   }
 
   public DecryptNotesTRC20 scanShieldedTRC20NotesByOvk(long startNum, long endNum,
-      byte[] ovk, byte[] shieldedTRC20ContractAddress, ProtocolStringList topicsList)
+      byte[] ovk, byte[] shieldedTRC20ContractAddress)
       throws ZksnarkException, BadItemException {
     checkAllowShieldedTransactionApi();
 
@@ -4149,7 +4126,7 @@ public class Wallet {
             Optional<DecryptNotesTRC20.NoteTx> noteTx;
             int index = 0;
             for (TransactionInfo.Log log : logList) {
-              int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress, topicsList);
+              int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress);
               if (logType > 0) {
                 noteBuilder = DecryptNotesTRC20.NoteTx.newBuilder();
                 noteBuilder.setTxid(ByteString.copyFrom(txid));
--- a/framework/src/main/java/org/tron/core/services/RpcApiService.java
+++ b/framework/src/main/java/org/tron/core/services/RpcApiService.java
@@ -95,6 +95,7 @@ import org.tron.core.exception.StoreException;
 import org.tron.core.exception.VMIllegalException;
 import org.tron.core.exception.ZksnarkException;
 import org.tron.core.metrics.MetricsApiService;
+import org.tron.core.services.http.Util;
 import org.tron.core.utils.TransactionUtil;
 import org.tron.core.zen.address.DiversifierT;
 import org.tron.core.zen.address.IncomingViewingKey;
@@ -290,6 +291,18 @@ public class RpcApiService extends RpcService {
     } else {
       return Status.INTERNAL.withDescription("unknown").asRuntimeException();
     }
+  }
+
+  private static boolean rejectIfEventsPresent(
+      StreamObserver<?> responseObserver, ProtocolStringList events) {
+    if (events.stream().anyMatch(s -> !s.isEmpty())) {
+      logger.info(Util.EVENTS_DEPRECATED_MSG);
+      responseObserver.onError(Status.INVALID_ARGUMENT
+          .withDescription(Util.EVENTS_DEPRECATED_MSG)
+          .asRuntimeException());
+      return true;
+    }
+    return false;
   }
 
   /**
@@ -722,18 +735,19 @@ public class RpcApiService extends RpcService {
     @Override
     public void scanShieldedTRC20NotesByIvk(IvkDecryptTRC20Parameters request,
         StreamObserver<DecryptNotesTRC20> responseObserver) {
+      if (rejectIfEventsPresent(responseObserver, request.getEventsList())) {
+        return;
+      }
       long startNum = request.getStartBlockIndex();
       long endNum = request.getEndBlockIndex();
       byte[] contractAddress = request.getShieldedTRC20ContractAddress().toByteArray();
       byte[] ivk = request.getIvk().toByteArray();
       byte[] ak = request.getAk().toByteArray();
       byte[] nk = request.getNk().toByteArray();
-      ProtocolStringList topicsList = request.getEventsList();
 
       try {
         responseObserver.onNext(
-            wallet.scanShieldedTRC20NotesByIvk(startNum, endNum, contractAddress, ivk, ak, nk,
-                topicsList));
+            wallet.scanShieldedTRC20NotesByIvk(startNum, endNum, contractAddress, ivk, ak, nk));
 
       } catch (Exception e) {
         responseObserver.onError(getRunTimeException(e));
@@ -744,15 +758,16 @@ public class RpcApiService extends RpcService {
     @Override
     public void scanShieldedTRC20NotesByOvk(OvkDecryptTRC20Parameters request,
         StreamObserver<DecryptNotesTRC20> responseObserver) {
+      if (rejectIfEventsPresent(responseObserver, request.getEventsList())) {
+        return;
+      }
       long startNum = request.getStartBlockIndex();
       long endNum = request.getEndBlockIndex();
       byte[] contractAddress = request.getShieldedTRC20ContractAddress().toByteArray();
       byte[] ovk = request.getOvk().toByteArray();
-      ProtocolStringList topicList = request.getEventsList();
       try {
         responseObserver
-            .onNext(wallet
-                .scanShieldedTRC20NotesByOvk(startNum, endNum, ovk, contractAddress, topicList));
+            .onNext(wallet.scanShieldedTRC20NotesByOvk(startNum, endNum, ovk, contractAddress));
       } catch (Exception e) {
         responseObserver.onError(getRunTimeException(e));
       }
@@ -2412,6 +2427,9 @@ public class RpcApiService extends RpcService {
     public void scanShieldedTRC20NotesByIvk(
         IvkDecryptTRC20Parameters request,
         StreamObserver<org.tron.api.GrpcAPI.DecryptNotesTRC20> responseObserver) {
+      if (rejectIfEventsPresent(responseObserver, request.getEventsList())) {
+        return;
+      }
       long startNum = request.getStartBlockIndex();
       long endNum = request.getEndBlockIndex();
       try {
@@ -2419,8 +2437,7 @@ public class RpcApiService extends RpcService {
             request.getShieldedTRC20ContractAddress().toByteArray(),
             request.getIvk().toByteArray(),
             request.getAk().toByteArray(),
-            request.getNk().toByteArray(),
-            request.getEventsList());
+            request.getNk().toByteArray());
         responseObserver.onNext(decryptNotes);
       } catch (BadItemException | ZksnarkException e) {
         responseObserver.onError(getRunTimeException(e));
@@ -2438,13 +2455,15 @@ public class RpcApiService extends RpcService {
     public void scanShieldedTRC20NotesByOvk(
         OvkDecryptTRC20Parameters request,
         StreamObserver<org.tron.api.GrpcAPI.DecryptNotesTRC20> responseObserver) {
+      if (rejectIfEventsPresent(responseObserver, request.getEventsList())) {
+        return;
+      }
       long startNum = request.getStartBlockIndex();
       long endNum = request.getEndBlockIndex();
       try {
         DecryptNotesTRC20 decryptNotes = wallet.scanShieldedTRC20NotesByOvk(startNum, endNum,
             request.getOvk().toByteArray(),
-            request.getShieldedTRC20ContractAddress().toByteArray(),
-            request.getEventsList());
+            request.getShieldedTRC20ContractAddress().toByteArray());
         responseObserver.onNext(decryptNotes);
       } catch (Exception e) {
         responseObserver.onError(getRunTimeException(e));
--- a/framework/src/main/java/org/tron/core/services/RpcApiService.java
+++ b/framework/src/main/java/org/tron/core/services/RpcApiService.java
@@ -295,7 +295,7 @@ public class RpcApiService extends RpcService {
 
   private static boolean rejectIfEventsPresent(
       StreamObserver<?> responseObserver, ProtocolStringList events) {
-    if (events.stream().anyMatch(s -> !s.isEmpty())) {
+    if (Util.hasMeaningfulEvents(events)) {
       logger.info(Util.EVENTS_DEPRECATED_MSG);
       responseObserver.onError(Status.INVALID_ARGUMENT
           .withDescription(Util.EVENTS_DEPRECATED_MSG)
--- a/framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
@@ -2,7 +2,6 @@ package org.tron.core.services.http;
 
 import com.alibaba.fastjson.JSONArray;
 import com.alibaba.fastjson.JSONObject;
-import java.util.stream.Collectors;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
@@ -41,15 +40,20 @@ public class ScanShieldedTRC20NotesByIvkServlet extends RateLimiterServlet {
       IvkDecryptTRC20Parameters.Builder ivkDecryptTRC20Parameters = IvkDecryptTRC20Parameters
           .newBuilder();
       JsonFormat.merge(params.getParams(), ivkDecryptTRC20Parameters, params.isVisible());
-
+      try {
+        Util.rejectIfEventsPresent(ivkDecryptTRC20Parameters.getEventsList());
+      } catch (IllegalArgumentException e) {
+        response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+        Util.processError(e, response);
+        return;
+      }
       GrpcAPI.DecryptNotesTRC20 notes = wallet
           .scanShieldedTRC20NotesByIvk(ivkDecryptTRC20Parameters.getStartBlockIndex(),
               ivkDecryptTRC20Parameters.getEndBlockIndex(),
               ivkDecryptTRC20Parameters.getShieldedTRC20ContractAddress().toByteArray(),
               ivkDecryptTRC20Parameters.getIvk().toByteArray(),
               ivkDecryptTRC20Parameters.getAk().toByteArray(),
-              ivkDecryptTRC20Parameters.getNk().toByteArray(),
-              ivkDecryptTRC20Parameters.getEventsList());
+              ivkDecryptTRC20Parameters.getNk().toByteArray());
       response.getWriter().println(convertOutput(notes, params.isVisible()));
     } catch (Exception e) {
       Util.processError(e, response);
@@ -57,6 +61,13 @@ public class ScanShieldedTRC20NotesByIvkServlet extends RateLimiterServlet {
   }
 
   protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    try {
+      Util.rejectIfEventsPresent(request.getParameterValues("events"));
+    } catch (IllegalArgumentException e) {
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      Util.processError(e, response);
+      return;
+    }
     try {
       boolean visible = Util.getVisible(request);
       long startNum = Long.parseLong(request.getParameter("start_block_index"));
@@ -74,7 +85,7 @@ public class ScanShieldedTRC20NotesByIvkServlet extends RateLimiterServlet {
       GrpcAPI.DecryptNotesTRC20 notes = wallet
           .scanShieldedTRC20NotesByIvk(startNum, endNum,
               ByteArray.fromHexString(contractAddress), ByteArray.fromHexString(ivk),
-              ByteArray.fromHexString(ak), ByteArray.fromHexString(nk), null);
+              ByteArray.fromHexString(ak), ByteArray.fromHexString(nk));
       response.getWriter().println(convertOutput(notes, visible));
     } catch (Exception e) {
       Util.processError(e, response);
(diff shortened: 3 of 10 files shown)


framework/src/main/java/org/tron/core/services/RpcApiService.java
```diff
@@ -722,18 +721,22 @@ public void isSpend(NoteParameters request, StreamObserver<SpendResult> response
     @Override
     public void scanShieldedTRC20NotesByIvk(IvkDecryptTRC20Parameters request,
         StreamObserver<DecryptNotesTRC20> responseObserver) {
```
alan-eth, line 735: Really clean approach to the deprecation — rejecting early before any business logic runs is exactly the right pattern for deprecated fields!

Minor observation: the same `events` field check appears 8 times across this file and the two servlets (4 in RpcApiService + 2 in each servlet). Would it be worth extracting a tiny static helper like `rejectIfEventsPresent(ProtocolStringList events)` to keep the rejection message and logic in one place? If the error message ever needs updating, a single change point would be nice. Totally fine as-is though — it's a straightforward guard clause.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
```diff
@@ -41,23 +40,33 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
-              ivkDecryptTRC20Parameters.getEventsList());
+              ivkDecryptTRC20Parameters.getNk().toByteArray());
       response.getWriter().println(convertOutput(notes, params.isVisible()));
+    } catch (IllegalArgumentException e) {
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      Util.processError(e, response);
     } catch (Exception e) {
       Util.processError(e, response);
     }
   }
```
alan-eth, line 80: Love that both doGet and doPost are covered — easy to forget the GET handler in servlet deprecation work!

One subtle difference I noticed: in `doPost`, the check is `!ivkDecryptTRC20Parameters.getEventsList().isEmpty()` (protobuf list non-empty), while in `doGet`, the check is `request.getParameter("events") != null`. The doGet version will also reject `?events=` (empty string), which is slightly stricter than the protobuf-based check. This is probably the right call for a deprecation, but wanted to flag in case the distinction matters for client compatibility.

framework/src/main/java/org/tron/core/services/RpcApiService.java
Federico2014: Good suggestion! Fixed in the latest commit. Added a private static `rejectIfEventsPresent(StreamObserver<?>, ProtocolStringList)` helper in `RpcApiService` to consolidate all 4 gRPC-level checks. For the HTTP layer, added two overloaded static helpers in `ScanShieldedTRC20NotesByIvkServlet` — one for `ProtocolStringList` (doPost) and one for `String` (doGet) — with the error message extracted into a `EVENTS_DEPRECATED_MSG` constant. `ScanShieldedTRC20NotesByOvkServlet` reuses both helpers directly.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
Federico2014: Good catch! Fixed in the latest commit. The `doGet` check now uses `eventsParam != null && !eventsParam.isEmpty()` to align with the `doPost` behavior — `?events=` (empty string) is no longer rejected, consistent with how an empty protobuf list is treated.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
```diff
@@ -35,29 +38,45 @@ public static String convertOutput(GrpcAPI.DecryptNotesTRC20 notes, boolean visi
...
           .scanShieldedTRC20NotesByIvk(ivkDecryptTRC20Parameters.getStartBlockIndex(),
               ivkDecryptTRC20Parameters.getEndBlockIndex(),
               ivkDecryptTRC20Parameters.getShieldedTRC20ContractAddress().toByteArray(),
               ivkDecryptTRC20Parameters.getIvk().toByteArray(),
               ivkDecryptTRC20Parameters.getAk().toByteArray(),
-              ivkDecryptTRC20Parameters.getNk().toByteArray(),
-              ivkDecryptTRC20Parameters.getEventsList());
+              ivkDecryptTRC20Parameters.getNk().toByteArray());
       response.getWriter().println(convertOutput(notes, params.isVisible()));
+    } catch (IllegalArgumentException e) {
```
waynercheung, line 73: [SHOULD] This catch scope is broader than the deprecation path itself. It will also turn unrelated IllegalArgumentException subclasses into HTTP 400 responses, for example NumberFormatException from `Long.parseLong(...)` in doGet and other IllegalArgumentException cases from `JsonFormat.merge(...)` in doPost. That expands the behavior change beyond "reject deprecated events". Please short-circuit the deprecated-field path directly, or use a dedicated exception type so only the events rejection is mapped to 400.

Same issue also applies to the corresponding catch blocks in `ScanShieldedTRC20NotesByOvkServlet`.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
```diff
@@ -35,29 +38,45 @@ public static String convertOutput(GrpcAPI.DecryptNotesTRC20 notes, boolean visi
...
+      response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+      Util.processError(e, response);
     } catch (Exception e) {
       Util.processError(e, response);
     }
   }
 
   protected void doGet(HttpServletRequest request, HttpServletResponse response) {
     try {
+      rejectIfEventsPresent(request.getParameter("events"));
```
waynercheung: [SHOULD] This only checks the single value returned by `request.getParameter("events")`. If the query string contains repeated parameters, for example `events=&events=mint`, container-specific behavior may make this check inconsistent and let the request pass. Since the external contract here is "events is rejected", please consider using `getParameterValues("events")` and rejecting when any supplied value is non-empty. If you go that route, the shared GET-side helper will probably need to move from a String overload to a `String[]-based` check as well.

Same issue also applies to the `doGet` path in `ScanShieldedTRC20NotesByOvkServlet`.

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
```diff
@@ -556,6 +558,54 @@ public void testScanShieldedTRC20NotesByOvk() {
...
 
+  @Test
+  public void testScanShieldedTRC20NotesByIvkRejectsDeprecatedEvents() {
+    IvkDecryptTRC20Parameters message = IvkDecryptTRC20Parameters.newBuilder()
+        .setStartBlockIndex(1)
+        .setEndBlockIndex(10)
+        .addEvents("mint")
+        .build();
+
+    StatusRuntimeException fullException = assertThrows(StatusRuntimeException.class,
```
waynercheung, line 570: [SHOULD] These new tests only assert the error description. Since the PR explicitly promises `INVALID_ARGUMENT` for gRPC, please also assert the status code, for example: 
```java
Assert.assertEquals(Status.Code.INVALID_ARGUMENT, fullException.getStatus().getCode());
```

Please apply the same assertion to the corresponding `Full/Solidity/PBFT` cases in both the `IVK` and `OVK` rejection tests.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -3964,7 +3944,7 @@ private DecryptNotesTRC20 queryTRC20NoteByIvk(long startNum, long endNum,
             Optional<DecryptNotesTRC20.NoteTx> noteTx;
             int index = 0;
```
waynercheung, line 3943: [DISCUSS] In the `IVK` scan path, index is incremented for every `logType > 0`, but `getNoteTxFromLogListByIvk(...)` only emits `noteTx` for `logType < 4`. That means a tx like `[BURN_TOKEN, decryptable MINT]` can return `NoteTx.index = 1` even though it is the first returned note. This behavior predates this PR, but always scanning all log types makes it more visible, and the Ovk path does not have the same behavior because it handles type 4. Should we document index as a log sequence within the tx, or track a follow-up to increment only when a noteTx is actually emitted?

waynercheung:
[SHOULD] The new rejection coverage is currently gRPC-only, but HTTP 400 is also part of the external contract introduced by this PR. Please add servlet-level regression tests for `scanShieldedTRC20NotesByIvk/Ovk` GET and POST to verify that requests with events return HTTP 400 and include the deprecation message. A small smoke test for the `Solidity/PBFT` HTTP wrappers would also help confirm the status code is preserved through the futureGet(...) wrapper.

framework/src/main/java/org/tron/core/Wallet.java
waynercheung: [NIT] `ProtocolStringList` and `TransactionRetCapsule` both appear to be unused now after this refactor. Please drop the stale imports while touching this file.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
Federico2014: Thanks for flagging this. On reflection, I think the broader catch scope may actually be the right behavior here — before this PR, `NumberFormatException` from invalid query parameters fell through to `catch (Exception)` which returned HTTP 200 with an error body (since `Util.processError` doesn't set a status code). Returning 400 for malformed client input seems more semantically correct. Would you agree, or do you still feel we should narrow the scope to the deprecation path only?

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
Federico2014: Good catch! Fixed in the latest commit — switched to `request.getParameterValues("events")` in both `ScanShieldedTRC20NotesByIvkServlet` and `ScanShieldedTRC20NotesByOvkServlet`, and updated the shared helper signature from `String` to `String[]` so it iterates all supplied values and rejects if any is non-empty.

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
Federico2014: Fixed. Added `Assert.assertEquals(Status.Code.INVALID_ARGUMENT, ...)` before each description assertion in both `testScanShieldedTRC20NotesByIvkRejectsDeprecatedEvents()` and `testScanShieldedTRC20NotesByOvkRejectsDeprecatedEvents()`, covering all three stubs (Full, Solidity, PBFT).

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Good observation, but `logType == 4` (BURN_TOKEN) will never be reached in the IVK scan path. The IVK (incoming viewing key) is used to decrypt notes received by a shielded address — only MINT (1), TRANSFER (2), and BURN_LEAF (3) produce new shielded notes visible to the recipient. BURN_TOKEN emits a log only when burning back to a transparent address, which carries no incoming shielded note, so this log type is never emitted in transactions relevant to IVK scanning. The `logType > 0` guard is therefore effectively equivalent to `logType > 0 && logType < 4` in this code path, and no index gap can occur.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Fixed. Removed both unused imports.

Federico2014:
@waynercheung  Fixed. Added `ScanShieldedTRC20NotesServletTest` covering all four cases — IVK POST, IVK GET, OVK POST, OVK GET — each asserting HTTP 400 and the deprecation message in the response body.

For the Solidity/PBFT wrappers: `futureGet` is synchronous and simply sets a DB cursor before delegating to `super.doPost/doGet`, so the `rejectIfEventsPresent` check and the `SC_BAD_REQUEST` status are set before any cursor operation is reached. The behavior is identical to the main servlet, with no additional code path to verify.

3for:
There are currently four predefined topic constants:

* `SHIELDED_TRC20_LOG_TOPICS_MINT`
* `SHIELDED_TRC20_LOG_TOPICS_TRANSFER`
* `SHIELDED_TRC20_LOG_TOPICS_BURN_LEAF`
* `SHIELDED_TRC20_LOG_TOPICS_BURN_TOKEN`

With the current change, when `events` is not provided, the behavior remains the same (matching all four standard topics). However, previously supported **subset filtering** (e.g., `events=["mint"]`) is no longer accepted and now results in `400 / INVALID_ARGUMENT`.

It seems this change also disables **custom topic scanning**, which is understandable since the previous `sha3 + substring match` approach was not very reliable. 

For clarity, the behavior change can be summarized as:

| Client Behavior              | Old Server             | New Server              | Impact          |
|-----------------------------|------------------------|--------------------------|-----------------|
| No events param             | Scan all 4 types       | Scan all 4 types         | No change       |
| events=["mint"] (subset filtering)             | Return only mint       | 400 / INVALID_ARGUMENT   | Breaking change |
| Custom topic (non-standard event)                | sha3 + substring match (unreliable)   | 400 / INVALID_ARGUMENT   | Breaking change |

### Suggestion

Would it be possible to **only disable custom/non-standard event scanning**, while still allowing filtering within the predefined topic set?

In other words:

* ✅ Keep support for standard subsets such as `events=["mint"]`, `["transfer"]`, or combinations
* ❌ Disallow arbitrary/custom topics outside the predefined constants

This would preserve backward compatibility for existing clients while still addressing the reliability concerns of custom topic matching.

Federico2014:
@3for Thanks for the detailed breakdown. To clarify: the old code never implemented subset filtering either. When `events` was empty, it always scanned all 4 predefined types (exact match against the 4 log topic constants). When `events` was non-empty, it did a `sha3(topic_string)` comparison — which only matched if the caller passed the exact full ABI event signature string, and even then, the classification still relied on a fragile substring check. There was no "pass `mint`, get only mint results" path.

The PR removes this non-functional code path and makes the behavior explicit: always scan all predefined types, reject any non-empty `events` field. This is the intended design.

framework/src/test/java/org/tron/core/services/http/ScanShieldedTRC20NotesServletTest.java
```diff
@@ -0,0 +1,77 @@
...
+  public void testOvkGetRejectsDeprecatedEvents() throws Exception {
+    MockHttpServletRequest request = createRequest(HttpGet.METHOD_NAME);
+    request.addParameter("events", "burn");
+    MockHttpServletResponse response = new MockHttpServletResponse();
+    scanShieldedTRC20NotesByOvkServlet.doGet(request, response);
+    Assert.assertEquals(HttpServletResponse.SC_BAD_REQUEST, response.getStatus());
+    Assert.assertTrue(response.getContentAsString().contains(EVENTS_DEPRECATED_MSG));
+  }
+
+}
```
bladehan1, line 115: [DISCUSS] Consider adding a positive-path test for requests without events field

All four new test cases cover the rejection path only (events field present → HTTP 400). A positive-path test (no events field → guard not triggered) would catch accidental inversion of the guard logic. However, constructing such a test in this mock context may require a fully initialized Wallet to handle the actual scan, which could be non-trivial.

Suggestion: If a lightweight positive-path assertion is feasible (e.g., verify the response status is not 400 even when the wallet returns an empty result), it would strengthen the test suite. If the setup cost is too high, consider a unit test directly on `rejectIfEventsPresent` instead.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
```diff
@@ -35,29 +38,49 @@ public static String convertOutput(GrpcAPI.DecryptNotesTRC20 notes, boolean visi
     }
   }
 
+  static void rejectIfEventsPresent(ProtocolStringList events) {
```
bladehan1: [SHOULD] POST and GET empty-string events handling is inconsistent — please unify

POST `{"events":[""]}` is rejected (ProtocolStringList.isEmpty() returns false for a list containing one empty string), while GET `?events=` is not rejected (String[] overload filters out empty strings via `!v.isEmpty()`). Both inputs represent semantically equivalent "no meaningful events" intent but produce different HTTP responses.

Suggestion: Align the ProtocolStringList overload to also ignore empty strings so that the two code paths behave identically for the same semantic input.

```java
// Change from:
if (!events.isEmpty()) { ... }
// To:
if (events.stream().anyMatch(s -> !s.isEmpty())) { ... }
```

framework/src/main/java/org/tron/core/services/RpcApiService.java
```diff
@@ -292,6 +292,17 @@ private StatusRuntimeException getRunTimeException(Exception e) {
     }
   }
 
+  private static boolean rejectIfEventsPresent(
+      StreamObserver<?> responseObserver, ProtocolStringList events) {
+    if (!events.isEmpty()) {
+      responseObserver.onError(Status.INVALID_ARGUMENT
```
bladehan1, line 300: [NIT] No structured log when deprecated events field is rejected

`rejectIfEventsPresent` silently returns an error to the caller with no server-side log entry. Node operators cannot identify which clients are still sending the deprecated field or track usage to determine when it is safe to remove the rejection code entirely.

Suggestion: Add a `logger.info` line inside the non-empty branch before calling `responseObserver.onError`. Note: java-tron convention uses `logger` (from `@Slf4j`), not `log`.

```java
logger.info("Received deprecated 'events' field in shielded TRC20 scan request");
```

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByOvkServlet.java
```diff
@@ -24,23 +23,28 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
       OvkDecryptTRC20Parameters.Builder ovkDecryptTRC20Parameters = OvkDecryptTRC20Parameters
           .newBuilder();
       JsonFormat.merge(params.getParams(), ovkDecryptTRC20Parameters, params.isVisible());
+      ScanShieldedTRC20NotesByIvkServlet
```
bladehan1: [DISCUSS] OvkServlet depends on IvkServlet's static method for input validation

`ScanShieldedTRC20NotesByOvkServlet` delegates its `events` field validation to `ScanShieldedTRC20NotesByIvkServlet.rejectIfEventsPresent(...)`. This creates a class-level dependency between two otherwise independent servlets. If `IvkServlet` is renamed, refactored, or moved to a different package, `OvkServlet` will break.

Suggestion: Extract `rejectIfEventsPresent` to an existing utility class (e.g., `Util`) so both servlets depend on a neutral shared helper rather than on each other. Acceptable to defer to a follow-up PR.

framework/src/test/java/org/tron/core/services/http/ScanShieldedTRC20NotesServletTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.mock.web.MockHttpServletResponse;
+import org.tron.common.BaseTest;
+import org.tron.common.TestConstants;
+import org.tron.core.config.args.Args;
+
+public class ScanShieldedTRC20NotesServletTest extends BaseTest {
+
```
bladehan1, line 21: [NIT] EVENTS_DEPRECATED_MSG duplicated in test class instead of referencing the production constant

The test class defines its own `private static final String EVENTS_DEPRECATED_MSG` that mirrors the constant in `ScanShieldedTRC20NotesByIvkServlet`. If the error message is updated in production, the test will still compile but will fail at runtime when the assertion no longer matches.

Suggestion: Reference `ScanShieldedTRC20NotesByIvkServlet.EVENTS_DEPRECATED_MSG` directly from the test class.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
Federico2014: Fixed in 87f9d573. Changed both the HTTP (`rejectIfEventsPresent(ProtocolStringList)`) and gRPC versions from `!events.isEmpty()` to `events.stream().anyMatch(s -> !s.isEmpty())`, so a list containing only empty strings is now treated the same as an absent field. Added regression tests `testIvkPostEmptyEventsPassesGuard` and `testOvkPostEmptyEventsPassesGuard` to cover this case.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
waynercheung: [SHOULD] Fair point that returning HTTP 400 for malformed client input is more REST-correct than the previous 200 + error-body behavior. I'm OK with keeping the broader catch if that's the intended direction.

One ask: since this now extends the behavior change beyond deprecated `events`, please make that explicit in the PR description / Breaking Changes. More precisely, these two endpoints now return HTTP 400 for malformed inputs that surface as `IllegalArgumentException` (for example `Long.parseLong(...)` in GET or `JsonFormat.merge(...)` cases in POST), not only when `events` is populated.

That makes the contract change clear for downstream SDKs. A follow-up to align similar servlets would also help consistency across the HTTP surface.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
waynercheung: [SHOULD] Thanks, this correctly aligns the empty-string semantics across GET / POST / gRPC.

One remaining gap worth closing: the gRPC code path in `RpcApiService.rejectIfEventsPresent(...)` changed in 87f9d57, but `RpcApiServicesTest` still only covers rejection with non-empty events. Please add symmetric IVK/OVK regression cases using `.addEvents("")` and assert the call does not fail with `INVALID_ARGUMENT`, so the new gRPC-side contract is pinned down the same way the HTTP POST path is now covered.

Also, please sync the PR description / Breaking Changes with the implemented behavior after 87f9d57. The current wording ("events request field is now rejected outright") is stricter than the code now enforces; the actual contract appears to be that non-empty `events` values are rejected, while an empty-string-only `events` list is treated as absent.

framework/src/main/java/org/tron/core/services/RpcApiService.java
Federico2014: Done — added `logger.warn(EVENTS_DEPRECATED_MSG)` in both HTTP overloads and the gRPC version. Used `warn` instead of `info` since it indicates a caller still using a deprecated API.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByOvkServlet.java
Federico2014: Done — moved both overloads and the `EVENTS_DEPRECATED_MSG` constant into `Util`. The HTTP servlets and gRPC service now both reference `Util.rejectIfEventsPresent` / `Util.EVENTS_DEPRECATED_MSG`.

framework/src/test/java/org/tron/core/services/http/ScanShieldedTRC20NotesServletTest.java
Federico2014: Done — `EVENTS_DEPRECATED_MSG` is now a `public static final` on `Util`, and the test class imports it via `import static org.tron.core.services.http.Util.EVENTS_DEPRECATED_MSG`.

framework/src/test/java/org/tron/core/services/http/ScanShieldedTRC20NotesServletTest.java
Federico2014: Done — added four positive-path tests: `testIvkPostWithoutEventsPassesGuard`, `testOvkPostWithoutEventsPassesGuard` (no events field at all), and `testIvkPostEmptyEventsPassesGuard`, `testOvkPostEmptyEventsPassesGuard` (`events=[""]` — empty string, should pass). For gRPC, `testScanShieldedTRC20NotesByIvk` / `testScanShieldedTRC20NotesByOvk` in `RpcApiServicesTest` already cover the no-events path.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
Federico2014: Good point. To preserve backward compatibility, the `IllegalArgumentException` catch has been scoped to only the events validation: in POST, it's wrapped in a nested try-catch around `rejectIfEventsPresent` with an early return; in GET, the events check runs before the main try block entirely. All other exceptions (e.g. `NumberFormatException` from `Long.parseLong`, parse errors from `JsonFormat.merge`) continue to fall through to the generic `Exception` catch and return the original 200 + error body behavior.

framework/src/main/java/org/tron/core/services/http/ScanShieldedTRC20NotesByIvkServlet.java
Federico2014: Done on both fronts:

1. Added `testScanShieldedTRC20NotesByIvkEmptyEventsPassesGuard` and `testScanShieldedTRC20NotesByOvkEmptyEventsPassesGuard` in `RpcApiServicesTest`, covering Full/Solidity/PBFT stubs with `.addEvents("")` — asserts the call succeeds rather than throwing `INVALID_ARGUMENT`.

2. Updated the PR description / Breaking Changes to reflect the actual contract: non-empty `events` values are rejected; an absent field or an empty-string-only list is treated as not set.

framework/src/main/java/org/tron/core/services/RpcApiService.java
0xbigapple: This rejection is client-induced: the server handles it correctly and returns `INVALID_ARGUMENT` / `HTTP 400`, with no action required from the operator. WARN can be flooded by a retrying client and trigger false alerts.
Suggest downgrading to `INFO`, consistent with PR #6693, which recently downgraded the same class of client-induced exceptions in Wallet.broadcastTransaction from `WARN` to `INFO`.
If usage tracking matters, a counter under `common/metrics` would be a better fit than a WARN log.

Federico2014: Good point — downgraded all three `logger.warn` calls (one in `RpcApiService.rejectIfEventsPresent` and two in `Util.rejectIfEventsPresent`) to `logger.info`, consistent with the pattern in PR #6693.

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -80,6 +84,24 @@ public class Util {
   public static final String FUNCTION_PARAMETER = "parameter";
   public static final String CALL_DATA = "data";
 
+  public static void rejectIfEventsPresent(ProtocolStringList events) {
```
bladehan1, line 91: [NIT] Empty-check logic still duplicated between Util.rejectIfEventsPresent(ProtocolStringList) and RpcApiService.rejectIfEventsPresent

Now that the constant and helpers live in `Util`, the gRPC-side `RpcApiService.rejectIfEventsPresent` and `Util.rejectIfEventsPresent(ProtocolStringList)` apply the same condition (`events.stream().anyMatch(s -> !s.isEmpty())`) and emit the same log; only the error-propagation strategy differs (gRPC `responseObserver.onError` vs HTTP `IllegalArgumentException`). Future tweaks to the rejection criterion would need synchronized edits in two places.

Suggestion: Extract the predicate into a single `Util` helper (e.g. `boolean hasMeaningfulEvents(ProtocolStringList events)`) and have both error-style wrappers reuse it. Non-blocking, follow-up acceptable.

framework/src/test/java/org/tron/core/services/http/ScanShieldedTRC20NotesServletTest.java
```diff
@@ -0,0 +1,115 @@
...
+    MockHttpServletResponse response = new MockHttpServletResponse();
+    scanShieldedTRC20NotesByOvkServlet.doGet(request, response);
+    Assert.assertEquals(HttpServletResponse.SC_BAD_REQUEST, response.getStatus());
+    Assert.assertTrue(response.getContentAsString().contains(EVENTS_DEPRECATED_MSG));
+  }
+
+  @Test
+  public void testIvkPostEmptyEventsPassesGuard() throws Exception {
+    MockHttpServletRequest request = createRequest(HttpPost.METHOD_NAME);
+    request.setContentType("application/json");
```
bladehan1, line 78: [NIT] Positive-path tests assert only `!= SC_BAD_REQUEST`, which is too weak

The four new positive-path tests (`test*Without/EmptyEventsPassesGuard`) assert `Assert.assertNotEquals(SC_BAD_REQUEST, response.getStatus())`. In the BaseTest mock context the wallet may not return a real success — the response could legitimately be 500 or 200-with-error-body and still satisfy `!= 400`. If the rejection guard is ever inverted but downstream code happens to short-circuit before reaching a 400, these tests would still pass.

Suggestion: Either (a) assert that the response body does not contain `EVENTS_DEPRECATED_MSG`, which directly proves the guard did not fire, or (b) add a Mockito spy on `Util.rejectIfEventsPresent` to verify it returned without throwing.

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
```diff
@@ -556,6 +559,84 @@ public void testScanShieldedTRC20NotesByOvk() {
...
+        .contains("'events' field is deprecated and no longer supported"));
+  }
+
+  @Test
+  public void testScanShieldedTRC20NotesByIvkEmptyEventsPassesGuard() {
+    IvkDecryptTRC20Parameters message = IvkDecryptTRC20Parameters.newBuilder()
+        .setStartBlockIndex(1)
+        .setEndBlockIndex(10)
+        .addEvents("")
+        .build();
```
bladehan1, line 622: [NIT] *EmptyEventsPassesGuard tests use `assertNotNull`, which is near-tautological

`testScanShieldedTRC20NotesByIvk/OvkEmptyEventsPassesGuard` assert `assertNotNull(blockingStub*.scan...(message))`. A protobuf message returned by a gRPC call is essentially never null; the test effectively passes as long as the stub does not throw `INVALID_ARGUMENT`. The current behaviour does work (the guard actually throws on rejection, the stub propagates it as `StatusRuntimeException`, and the test fails) — but the assertion does not document that contract.

Suggestion: Wrap the call in `try { stub.scan(...); } catch (StatusRuntimeException e) { Assert.fail("empty events should pass the guard, got: " + e.getStatus()); }`, or simply remove the wrapper and rely on JUnit failing on uncaught exception with a comment explaining the contract.

bladehan1 (commented):
Followup on the previous review — all 5 received comments are addressed; thanks for the thorough iteration. A few small follow-ups inline.

bladehan1:
[SHOULD] `testGetShieldedTRC20LogType2` in `WalletMockTest.java` does not assert the return value

After removing the `topicsList` parameter, `Wallet.getShieldedTRC20LogType` performs a pure constant comparison against `SHIELDED_TRC20_LOG_TOPICS_MINT/TRANSFER/BURN_LEAF/BURN_TOKEN`. The test currently only asserts that no exception is thrown — it does not verify that any of the four expected log types actually return the correct int value (1, 2, 3, 4). If the constants are wrong or the byte-comparison logic has a bug, the test will still pass.

**Suggestion:** Add `assertEquals` assertions after invoking the method with each of the four topic constants, verifying the returned value is 1, 2, 3, and 4 respectively, plus assert that an unknown topic returns 0. Could be tackled in a follow-up PR if you prefer to keep this one focused on the deprecation rollout.

framework/src/main/java/org/tron/core/services/http/Util.java
Federico2014: Good call — extracted the predicate into `Util.hasMeaningfulEvents(ProtocolStringList)` and updated both call sites (`Util.rejectIfEventsPresent(ProtocolStringList)` on the HTTP path and `RpcApiService.rejectIfEventsPresent` on the gRPC path) to delegate to it. Future tweaks to the rejection criterion now require a single edit. Error-propagation strategy stays differentiated by wrapper (HTTP throws `IllegalArgumentException`, gRPC routes through `responseObserver.onError` with `INVALID_ARGUMENT`).

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
Federico2014: Agreed — `assertNotNull` on a protobuf return value doesn't document the "guard must not fire" contract. Updated `testScanShieldedTRC20NotesBy{Ivk,Ovk}EmptyEventsPassesGuard` to wrap the three stub calls in `try { ... } catch (StatusRuntimeException e) { Assert.fail("empty events should pass the guard, got: " + e.getStatus()); }`. Now the test will fail loudly on any `INVALID_ARGUMENT` (or other status), and the intent is explicit.

framework/src/test/java/org/tron/core/services/http/ScanShieldedTRC20NotesServletTest.java
Federico2014: Fair point — `!= SC_BAD_REQUEST` is satisfied even by a 500 or a 200-with-error-body, so the guard could be inverted without these tests catching it. Went with option (a): added `Assert.assertFalse(response.getContentAsString().contains(EVENTS_DEPRECATED_MSG))` to all four positive-path tests (`testIvkPostEmptyEventsPassesGuard`, `testOvkPostEmptyEventsPassesGuard`, `testIvkPostWithoutEventsPassesGuard`, `testOvkPostWithoutEventsPassesGuard`). This directly proves the rejection guard didn't fire, independent of the eventual response code from the mocked wallet path. Avoided the Mockito-spy alternative to keep the test style consistent with the rest of the servlet test file.

Federico2014:
@bladehan1  Agreed — the existing tests only validated the input-validation path and the unknown-topic path; the four happy branches (return 1/2/3/4) had no coverage, which would have masked a bug in either the constants or the byte-comparison. Addressed in this PR rather than deferred since the simplification of `getShieldedTRC20LogType` is itself one of the deliverables.

Changes:
- Added `testGetShieldedTRC20LogTypeReturnsCorrectInt`: hashes each of the four event signatures (`MintNewLeaf`, `TransferNewLeaf`, `BurnNewLeaf`, `TokenBurn`) with `Hash.sha3(...)` and asserts the method returns 1, 2, 3, 4 respectively.
- Tightened `testGetShieldedTRC20LogType1` and `testGetShieldedTRC20LogType2` to capture the return value and `assertEquals(0, ...)`, making the unknown-topic contract explicit instead of just "didn't throw".

Build + targeted run pass: all 4 `testGetShieldedTRC20LogType*` tests green.

bladehan1 (approved):
LGTM

0xbigapple (approved):
LGTM

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -80,6 +84,28 @@ public class Util {
   public static final String FUNCTION_PARAMETER = "parameter";
   public static final String CALL_DATA = "data";
 
+  public static boolean hasMeaningfulEvents(ProtocolStringList events) {
+    return events.stream().anyMatch(s -> !s.isEmpty());
+  }
+
+  public static void rejectIfEventsPresent(ProtocolStringList events) {
+    if (hasMeaningfulEvents(events)) {
+      logger.info(EVENTS_DEPRECATED_MSG);
```
waynercheung, line 93: [SHOULD] This rejection path sits on a public/high-frequency API surface, so logging every deprecated-events hit at INFO may create unnecessary log noise if stale clients retry. Per the usual "avoid INFO/WARN on hot paths" guideline,`debug` seems safer here, or we can drop the log entirely and rely on the 400 / INVALID_ARGUMENT response itself.

 Util.java:102, and RpcApiService.java:299 have the same problem.

waynercheung (approved):
LGTM

PR #6687, opened 2026-04-16
fix(chainbase): guard divide-by-zero

Little-Peony:
## What does this PR do?

Fix audit warnings in energy-related code.

### Changes

- **`TransactionTrace`**: guard `newArea / newSize` with a `newSize == 0` check in `resetAccountUsage`. A zero window size means no valid time window exists, so usage should be 0 rather than dividing by zero. Also fixes a minor comment typo (`pre consumed` → `preconsumed`).

## Test plan

- [x] Existing tests pass — no behaviour change for the normal (non-zero `newSize`) path


chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
```diff
@@ -295,13 +295,14 @@ private void resetAccountUsage(AccountCapsule accountCap,
...
-    // Drop the pre consumed frozen energy
+    // Drop the preconsumed frozen energy
     long newArea = currentUsage * currentSize
         - (mergedUsage * mergedSize - usage * size);
     // If area merging happened during suicide, use the current window size
     long newSize = mergedSize == currentSize ? size : currentSize;
-    // Calc new usage by fixed x-axes
-    long newUsage = max(0, newArea / newSize, dynamicPropertiesStore.disableJavaLangMath());
+    // A zero window size means no valid time window exists and thus zero usage
+    long newUsage = newSize == 0 ? 0
```
alan-eth, line 304: The zero-division guard is well-reasoned — `newSize` can indeed be 0 when `mergedSize == currentSize` and `size` from the receipt is 0, even though `getWindowSize()` itself never returns 0. Great job tracing the actual root cause rather than just adding a blanket check! 👏

Since this is a subtle edge case that's easy to regress on, would it be worth adding a unit test in `TransactionTraceTest` that exercises the `newSize == 0` path? This would also serve as documentation for future readers about when this condition can occur.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
```diff
@@ -295,13 +295,14 @@ private void resetAccountUsage(AccountCapsule accountCap,
     }
     long currentSize = accountCap.getWindowSize(ENERGY);
     long currentUsage = accountCap.getEnergyUsage();
-    // Drop the pre consumed frozen energy
+    // Drop the preconsumed frozen energy
```
waynercheung, line 298: [NIT] `preconsumed` reads a bit awkwardly here and may trigger spell-check warnings. Something like `previously consumed` (for example, `Drop the previously consumed frozen energy`) would be clearer.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
waynercheung: [SHOULD] Please add a regression test for the `newSize == 0` path in `TransactionTraceTest`. This is a subtle edge case, and a test would both lock in the intended behavior (`newUsage = 0` with window reset) and document when this condition is expected to occur.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
```diff
@@ -295,13 +295,14 @@ private void resetAccountUsage(AccountCapsule accountCap,
...
     long newArea = currentUsage * currentSize
```
waynercheung, line 299: [QUESTION] `newArea` still relies on raw `long` multiplications/subtractions here. I realize this is pre-existing code rather than introduced by this PR, but since this method is already being touched for numeric safety, do we want a follow-up to harden these arithmetic operations with the existing exact-helper pattern as well?

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
waynercheung, line 304: [NIT] `git diff --check` reports trailing whitespace on this line (same issue at line 322). Please clean up both before merge.
Or we can use:
```
long newUsage =
        newSize == 0 ? 0 : max(0, newArea / newSize, dynamicPropertiesStore.disableJavaLangMath());
```

waynercheung:
[DISCUSS] If `newSize == 0` is truly reachable during block execution, then this change does more than prevent a local exception: it changes the behavior from an uncaught `ArithmeticException` during finalization to continuing with `newUsage = 0`. In that case, the PR should explicitly document the consensus/upgrade impact (for example, whether coordinated node upgrade or any activation mechanism is needed).

At the moment, though, the reachability of `newSize == 0` is not obvious from the current call path, because the related window sizes are populated via `getWindowSize()` in `VMActuator`, which normally does not return `0`. Please clarify the exact trigger path, or describe this as a defensive guard if it is only intended as protection against an unexpected zero window.

chainbase/src/main/java/org/tron/core/db2/core/Chainbase.java
```diff
@@ -35,7 +35,7 @@ public enum Cursor {
   //true:fullnode, false:soliditynode
   private ThreadLocal<Cursor> cursor = new ThreadLocal<>();
   private ThreadLocal<Long> offset = new ThreadLocal<>();
-  private Snapshot head;
+  private volatile Snapshot head;
```
halibobo1205, line 38: [SHOULD] Could you clarify the motivation for this change?

Under the revokable DB’s weak-consistency model, reading a slightly stale head reference doesn’t violate any invariant — it’s equivalent to the query having been issued a few milliseconds earlier.

To confirm: is there an actually observed issue (logs, a reproducible scenario), or is this a preventive change prompted by a static analyzer / audit?
If there’s no concrete trigger, I’d suggest dropping it. volatile doesn’t cover the multi-read consistency inside head(), nor the lifecycle race where a reader still holds the old reference after reset() closes the old snapshot — adding just this one line risks misleading future maintainers into thinking the concurrency concern is already fixed.

Little-Peony: You're right. volatile only addresses field visibility but doesn't prevent the lifecycle race where a reader holds a stale reference after reset() closes the snapshot, nor the multi-read inconsistency inside head(). The proper fix would require a ReadWriteLock across all read and write paths — that's a broader refactor and out of scope here. Removed the volatile.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
Little-Peony: Fixed. Changed to `previously consumed`.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
Little-Peony: Fixed. Removed trailing whitespace and reformatted to a single line as suggested.

Little-Peony: Done. Added two regression tests in `TransactionTraceTest`:
- `testResetAccountUsageZeroWindowSize`: covers the `resetAccountUsage` path (V2 disabled) with `mergedSize == currentSize` and `size == 0`, which produces `newSize == 0`.
- `testResetAccountUsageV2ZeroWindowSize`: same scenario via the `resetAccountUsageV2` path (V2 enabled).

Both verify that `energyUsage` is reset to 0 without an `ArithmeticException`.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
Sunny6889: Good observation. While there's a theoretical overflow risk in `newArea =  currentUsage * currentSize - (mergedUsage * mergedSize - usage * size)`, in practice this won't occur in java-tron's business context — both  `currentUsage (energy usage)` and `currentSize (window size, capped at WINDOW_SIZE_MS / BLOCK_PRODUCED_INTERVAL = 28800)` are tightly bounded by the protocol's energy limits, so the product stays well within long range. But I agree it's worth noting as a pre-existing pattern; happy to track it in a 
follow-up if you'd like.

Sunny6889:
> [DISCUSS] If `newSize == 0` is truly reachable during block execution, then this change does more than prevent a local exception: it changes the behavior from an uncaught `ArithmeticException` during finalization to continuing with `newUsage = 0`. In that case, the PR should explicitly document the consensus/upgrade impact (for example, whether coordinated node upgrade or any activation mechanism is needed).
> 
> At the moment, though, the reachability of `newSize == 0` is not obvious from the current call path, because the related window sizes are populated via `getWindowSize()` in `VMActuator`, which normally does not return `0`. Please clarify the exact trigger path, or describe this as a defensive guard if it is only intended as protection against an unexpected zero window.


The `size` parameter passed to `resetAccountUsage` is recorded from `account.getWindowSize(ENERGY)` in `VMActuator` (lines 583 and 751). `AccountCapsule.getWindowSize(ENERGY)` has an explicit fallback: when the stored proto field is `0`, it returns `WINDOW_SIZE_MS / BLOCK_PRODUCED_INTERVAL = 28800` — so it **never returns 0**. Likewise, `currentSize` is read from `accountCap.getWindowSize(ENERGY)` at the point `resetAccountUsage` is called, subject to the same fallback. Since neither `size` nor `currentSize` can be 0, `newSize` (`= mergedSize == currentSize ? size : currentSize`) cannot be 0 in any normal execution path.

This is therefore a **defensive guard** — the existing comment *"A zero window size means no valid time window exists and thus zero usage"* already expressed the intended behavior, but the zero check was missing. There is no reachable code path that produces `newSize == 0`, so no consensus impact and no coordinated upgrade is required.

chainbase/src/main/java/org/tron/core/db/TransactionTrace.java
```diff
@@ -295,13 +295,14 @@ private void resetAccountUsage(AccountCapsule accountCap,
...
     long currentUsage = accountCap.getEnergyUsage();
-    // Drop the pre consumed frozen energy
+    // Drop the previously consumed frozen energy
     long newArea = currentUsage * currentSize
         - (mergedUsage * mergedSize - usage * size);
     // If area merging happened during suicide, use the current window size
     long newSize = mergedSize == currentSize ? size : currentSize;
-    // Calc new usage by fixed x-axes
-    long newUsage = max(0, newArea / newSize, dynamicPropertiesStore.disableJavaLangMath());
+    // A zero window size means no valid time window exists and thus zero usage
```
bladehan1, line 303: [NIT] New comment removes the algorithm context from 'fixed x-axes'

The original comment `// Calc new usage by fixed x-axes` explained the area-based energy accounting model (usage × window = area; new_usage = area / new_window). The replacement comment only explains the zero-guard case, so readers of the non-zero path lose the rationale for `newArea / newSize`.

The same applies to the equivalent line in `resetAccountUsageV2` (line ~320).

Suggestion: Merge both meanings, e.g.

```java
// Calculate new usage from area/window model; zero window means no valid time context and yields zero usage
```

Little-Peony:
Closed, as it is only theoretical guard fix.

PR #6690, opened 2026-04-16
feat(db,lite): add --exclude-historical-balance for lite snapshot

### LINKED ISSUE #6597: [Feature] Exclude historical balance DBs from lite snapshot
## **Background**

Lite nodes retain all state data, while non-state data is limited to the most recent 65,536 blocks of block and transaction data.

`toolkit db lite` uses an `archiveDbs` list to determine which databases are excluded when splitting a lite node snapshot:

```java
// plugins/.../DbLite.java:61-66
private static final List<String> archiveDbs = Arrays.asList(
    BLOCK_DB_NAME,                    // "block"
    BLOCK_INDEX_DB_NAME,              // "block-index"
    TRANS_DB_NAME,                    // "trans"
    TRANSACTION_RET_DB_NAME,          // "transactionRetStore"
    TRANSACTION_HISTORY_DB_NAME);     // "transactionHistoryStore"
```

Currently, `account-trace` and `balance-trace` are not included in the list above, so they are copied in full into the lite node snapshot. However, both databases are space-intensive and effectively unusable in a lite node context: they are enabled via the CLI flag `--history-balance-lookup` or the config option `storage.balance.history.lookup`, written during block and transaction execution, serves the historical balance query API (`getAccountBalance`).

Since lite nodes retain only the most recent 65,536 blocks, historical block data before the snapshot point (e.g., `block`, `trans`) is already excluded, and the historical balance query API is inherently unavailable on lite nodes — retaining these two databases provides no practical value.

| Database | Contents | Measured Size |
| ----- | ----- | ----- |
| `balance-trace` | Historical balance change records at the block and transaction level | ≈ 690 GB |
| `account-trace` | Historical account balances indexed by address \+ block number | ≈ 180 GB |
| Total | Based on the mainnet full node snapshot measured on 2026-03-11 | ≈ 870 GB |

---

## **Problem Statement**

When `toolkit db lite` splits a Lite node snapshot, it does not include `account-trace` and `balance-trace` in the `archiveDbs` exclusion list. As a result, approximately 870 GB of effectively unused data is copied in full into the lite node snapshot, increasing storage costs and data transfer overhead while contributing nothing to lite node functionality.

---

## **Rationale**

**Why should this feature exist?**

* **Storage savings**: Each split operation avoids copying and transferring approximately 870 GB of unused data.  
* **No functional impact**: The two excluded databases are not involved in any online state computation on lite nodes; they serve only the historical balance query API, which is already unavailable on lite nodes. Furthermore, both databases are optional features that most nodes do not enable by default.

**What are the use cases?**

* Node operators using `toolkit db lite` to generate Lite Node snapshots, reducing disk usage and network transfer costs.  
* Snapshot distribution scenarios where a significantly smaller snapshot size lowers the barrier to initial sync.  
* Storage-constrained environments where eliminating unused data preserves valuable disk space.

**Who would benefit from this feature?**

Node operators, lite node deployers, and snapshot distribution service providers.

---

### **Specification**

**1\. Split (`lite`)**

Add `account-trace` and `balance-trace` to the `archiveDbs` list in `DbLite.java` so they are automatically excluded when splitting a lite node snapshot, and classified under the archive (historical) portion:

```java
private static final List<String> archiveDbs = Arrays.asList(
    BLOCK_DB_NAME,
    BLOCK_INDEX_DB_NAME,
    TRANS_DB_NAME,
    TRANSACTION_RET_DB_NAME,
    TRANSACTION_HISTORY_DB_NAME,
    ACCOUNT_TRACE_DB_NAME,    // new
    BALANCE_TRACE_DB_NAME);   // new
```

**2\. Merge (`merge`)**

The `merge` operation combines a lite node snapshot with historical data to restore a full node. Since `account-trace` and `balance-trace` are now classified on the archive side, the merge logic must be updated and validated accordingly:

**Merge strategy:**

| Source | Action |
| ----- | ----- |
| `account-trace` / `balance-trace` from archive | Retain entries with height \< `lite_height` |
| `account-trace` / `balance-trace` from lite snapshot | Append only entries with height \> `archive_height` |
| Final result | Two contiguous segments, no overlap, no gap |

* **API Changes**: None.  
* **Configuration Changes**: None.  
* **Protocol Changes**: None.

---

### **Test Scenarios**

1. After running `toolkit db lite`, verify that the lite snapshot directory does not contain `account-trace` or `balance-trace`, and that the archive directory does contain them (if the source node had the feature enabled).  
2. Compare snapshot sizes before and after the change to confirm the reduction matches expectations (≈ 870 GB).  
3. Start a lite node from the new snapshot and verify that block sync, state queries, and other core functions work correctly with no errors.  
4. **Merge validation** (source node with `history-balance-lookup` enabled): after merging, verify that the full node contains complete `account-trace` and `balance-trace` data and that the historical balance query API functions correctly.

### **Performance Considerations**

The change affects only file-handling logic during the snapshot split and merge phases; it has no impact on node runtime performance.

---

## **Scope of Impact**

* [ ] Core protocol  
* [ ] API/RPC  
* [x] Database  
* [ ] Network layer  
* [ ] Smart contracts  
* [ ] Documentation  
* [x] Other: toolkit `db lite.`

### **Breaking Changes**

None.

### **Backward Compatibility**

No modifications are made to existing full node database directories; only the behavior of `toolkit db lite` splitting is affected. Previously generated lite node snapshots are not impacted.

---

## **Implementation**

Append `ACCOUNT_TRACE_DB_NAME` and `BALANCE_TRACE_DB_NAME` to the `archiveDbs` constant in `DbLite.java` — a minimal code change. Add incremental merge logic to the merge flow for these two databases, scoped by block height range.

* [x] Yes, I can implement this feature

### **Estimated Complexity**

* [x] Medium (moderate changes)

---

## **Alternatives Considered**

None.

---

### **Related Issues/PRs**

None.

### **References**

* [`getAccountBalance`](https://developers.tron.network/reference/getaccountbalance) API documentation

halibobo1205:
### What changed
Add a boolean flag `--exclude-historical-balance` (default `false`) to`java -jar Toolkit.jar db lite -o split -t snapshot`. When enabled, the lite snapshot omits the `balance-trace` and `account-trace` stores.

### Default behavior: unchanged
With the default, both `split` and`merge` behave **exactly as before this PR**. The majority of operators
(running with `historyBalanceLookup=off`, the project default) need not think about this flag and will see no difference.

### Opt-in behavior: smaller lite snapshots
Passing `--exclude-historical-balance` to `split -t snapshot` produces a lite snapshot that does not carry `balance-trace` or `account-trace`. For source full nodes that ran with `historyBalanceLookup=true` (where the two stores can grow to ~900 GB), which is what makes lite-snapshot slicing operationally feasible.

`split -t history` and `merge` **ignore this flag** — the merge pipeline itself is not modified by this PR.

### Explicit warning at split time
When the flag is enabled, both the runtime output and the `--help` text spell out the contract:

- The flag has functional impact only when the source full node runs with `historyBalanceLookup=true`. Default configured operators are unaffected either way.
- If `historyBalanceLookup` was enabled, the loss is **permanent**: a lite node booted from such a snapshot cannot answer historical balance lookups (`getBlockBalance` / `getAccountBalance`), and running `merge` afterwards will **not** restore the feature.
- Operators who need historical balance lookup on the resulting lite node must **not** enable this flag.

### Design choices
- Scope the flag strictly to `split -t snapshot`. The toolkit stays a thin physical-split tool; the merge pipeline is left intact, avoiding any need to reason about history-pack trace presence, bak replay, or cross-toolkit-version compatibility.
- Keep the default at the legacy behavior. Existing workflows are not silently disrupted; operators who want the size saving opt in explicitly.
- The loss only matters for `historyBalanceLookup=true` users. Under the default configuration, the trace stores are empty Spring-initialized directories, so excluding them changes nothing of substance.

### Tests
- Existing default-path tests remain unchanged (zero regression):  `DbLiteRocksDbTest`, `DbLiteRocksDbV2Test`.
- New `DbLiteExcludeTraceRocksDbTest` covers the opt-in path, asserting that the snapshot produced with `--exclude-historical-balance`  contains neither `balance-trace` nor `account-trace`.

close #6597

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ Wallet
 
 /framework/propPath
 .cache
+
+#AI
+.claude
--- a/plugins/src/main/java/common/org/tron/plugins/DbLite.java
+++ b/plugins/src/main/java/common/org/tron/plugins/DbLite.java
@@ -120,11 +120,12 @@ public class DbLite implements Callable<Integer> {
           + "This flag only has a functional impact when the source full node ran with "
           + "`historyBalanceLookup=true` (off by default; most operators are unaffected). "
           + "WARNING: when historyBalanceLookup was enabled, this loss is permanent: a lite "
-          + "node booted from such a snapshot cannot answer historical balance lookups "
-          + "(getBlockBalance / getAccountBalance), and running merge afterwards will NOT "
-          + "restore the feature. If you need to keep historyBalanceLookup working on the "
-          + "resulting lite node, do NOT enable this flag. `split -t history` and `merge` "
-          + "ignore this flag.",
+          + "node booted from such a snapshot cannot safely serve historical balance lookups "
+          + "(getBlockBalance may fail, and getAccountBalance may return balance=0 when "
+          + "account-trace data is missing). Running merge afterwards will NOT restore the "
+          + "feature. If you need to keep historyBalanceLookup working on the resulting "
+          + "lite node, do NOT enable this flag. `split -t history` and `merge` ignore "
+          + "this flag.",
       order = 5)
   private boolean excludeHistoricalBalance;
 
@@ -143,8 +144,8 @@ public class DbLite implements Callable<Integer> {
     try {
       switch (this.operate) {
         case split:
-          warnIfExcludingHistoricalBalance();
           if (Type.snapshot == this.type) {
+            warnIfExcludingHistoricalBalance();
             generateSnapshot(fnDataPath, datasetPath);
           } else if (Type.history == type) {
             generateHistory(fnDataPath, datasetPath);
@@ -307,12 +308,14 @@ public class DbLite implements Callable<Integer> {
         + "will be excluded from the lite snapshot. This only matters when the source full "
         + "node ran with historyBalanceLookup=true (off by default; most operators are "
         + "unaffected). When that switch was enabled, this loss is permanent: lite nodes "
-        + "booted from this snapshot cannot answer historical balance lookups "
-        + "(getBlockBalance / getAccountBalance), and running merge afterwards will NOT "
-        + "restore the feature. If you need to keep historyBalanceLookup working on the "
-        + "resulting lite node, do NOT use this flag.";
+        + "booted from this snapshot cannot safely serve historical balance lookups "
+        + "(getBlockBalance may fail, and getAccountBalance may return balance=0 when "
+        + "account-trace data is missing). Running merge afterwards will NOT restore the "
+        + "feature. If you need to keep historyBalanceLookup working on the resulting "
+        + "lite node, do NOT use this flag.";
     logger.warn(msg);
-    spec.commandLine().getErr().println(msg);
+    spec.commandLine().getErr().println(spec.commandLine().getColorScheme()
+        .errorText(msg));
   }
 
   private List<String> getSnapshotDbs(String sourceDir) {
@@ -784,6 +787,5 @@ public class DbLite implements Callable<Integer> {
     }
   }
 }
-
 
 
--- a/plugins/src/main/java/common/org/tron/plugins/DbLite.java
+++ b/plugins/src/main/java/common/org/tron/plugins/DbLite.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
 import java.util.stream.LongStream;
+import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
 import me.tongfei.progressbar.ProgressBar;
 import org.rocksdb.RocksDBException;
@@ -57,6 +58,8 @@ public class DbLite implements Callable<Integer> {
   private static final String TRANSACTION_HISTORY_DB_NAME = "transactionHistoryStore";
   private static final String PROPERTIES_DB_NAME = "properties";
   private static final String TRANS_CACHE_DB_NAME = "trans-cache";
+  private static final String BALANCE_TRACE_DB_NAME = "balance-trace";
+  private static final String ACCOUNT_TRACE_DB_NAME = "account-trace";
 
   private static final List<String> archiveDbs = Arrays.asList(
       BLOCK_DB_NAME,
@@ -64,6 +67,10 @@ public class DbLite implements Callable<Integer> {
       TRANS_DB_NAME,
       TRANSACTION_RET_DB_NAME,
       TRANSACTION_HISTORY_DB_NAME);
+
+  private static final List<String> traceDbs = Arrays.asList(
+      BALANCE_TRACE_DB_NAME,
+      ACCOUNT_TRACE_DB_NAME);
 
   enum Operate { split, merge }
 
@@ -105,8 +112,25 @@ public class DbLite implements Callable<Integer> {
   private String datasetPath;
 
   @CommandLine.Option(
-      names = {"--help", "-h"},
+      names = {"--exclude-historical-balance"},
+      defaultValue = "false",
+      description = "only used with `operate=split -t snapshot`: when true, balance-trace "
+          + "and account-trace are excluded from the lite snapshot. "
+          + "Default: ${DEFAULT-VALUE} (legacy behavior; trace stores stay in the snapshot). "
+          + "This flag only has a functional impact when the source full node ran with "
+          + "`historyBalanceLookup=true` (off by default; most operators are unaffected). "
+          + "WARNING: when historyBalanceLookup was enabled, this loss is permanent: a lite "
+          + "node booted from such a snapshot cannot answer historical balance lookups "
+          + "(getBlockBalance / getAccountBalance), and running merge afterwards will NOT "
+          + "restore the feature. If you need to keep historyBalanceLookup working on the "
+          + "resulting lite node, do NOT enable this flag. `split -t history` and `merge` "
+          + "ignore this flag.",
       order = 5)
+  private boolean excludeHistoricalBalance;
+
+  @CommandLine.Option(
+      names = {"--help", "-h"},
+      order = 6)
   private boolean help;
 
 
@@ -119,6 +143,7 @@ public class DbLite implements Callable<Integer> {
     try {
       switch (this.operate) {
         case split:
+          warnIfExcludingHistoricalBalance();
           if (Type.snapshot == this.type) {
             generateSnapshot(fnDataPath, datasetPath);
           } else if (Type.history == type) {
@@ -253,12 +278,50 @@ public class DbLite implements Callable<Integer> {
     spec.commandLine().getOut().format("Merge history finished, take %d s.", during).println();
   }
 
+  /**
+   * Compute the directories to exclude from the lite snapshot.
+   * <p>
+   * Default ({@code --exclude-historical-balance=false}): the legacy archive set
+   * (5 dbs); {@link #BALANCE_TRACE_DB_NAME} / {@link #ACCOUNT_TRACE_DB_NAME}
+   * stay with the snapshot as state-style stores.
+   * <p>
+   * Opt-in ({@code --exclude-historical-balance=true}): the trace stores are
+   * additionally excluded, producing a smaller lite snapshot at the cost of
+   * dropping historical balance lookup support on the resulting lite node.
+   * Only {@code split -t snapshot} consults this. {@code split -t history}
+   * and {@code merge} always use the legacy archive set.
+   */
+  private List<String> snapshotExclusion() {
+    if (!excludeHistoricalBalance) {
+      return archiveDbs;
+    }
+    return Stream.concat(archiveDbs.stream(), traceDbs.stream())
+        .collect(Collectors.toList());
+  }
+
+  private void warnIfExcludingHistoricalBalance() {
+    if (!excludeHistoricalBalance) {
+      return;
+    }
+    String msg = "WARNING: --exclude-historical-balance is enabled. balance-trace / account-trace "
+        + "will be excluded from the lite snapshot. This only matters when the source full "
+        + "node ran with historyBalanceLookup=true (off by default; most operators are "
+        + "unaffected). When that switch was enabled, this loss is permanent: lite nodes "
+        + "booted from this snapshot cannot answer historical balance lookups "
+        + "(getBlockBalance / getAccountBalance), and running merge afterwards will NOT "
+        + "restore the feature. If you need to keep historyBalanceLookup working on the "
+        + "resulting lite node, do NOT use this flag.";
+    logger.warn(msg);
+    spec.commandLine().getErr().println(msg);
+  }
+
   private List<String> getSnapshotDbs(String sourceDir) {
     List<String> snapshotDbs = Lists.newArrayList();
     File basePath = new File(sourceDir);
+    List<String> excluded = snapshotExclusion();
     Arrays.stream(Objects.requireNonNull(basePath.listFiles()))
             .filter(File::isDirectory)
-            .filter(dir -> !archiveDbs.contains(dir.getName()))
+            .filter(dir -> !excluded.contains(dir.getName()))
             .forEach(dir -> snapshotDbs.add(dir.getName()));
     return snapshotDbs;
   }
--- a/plugins/src/main/java/common/org/tron/plugins/DbLite.java
+++ b/plugins/src/main/java/common/org/tron/plugins/DbLite.java
@@ -6,6 +6,7 @@ import com.google.common.collect.Maps;
 import com.google.common.primitives.Bytes;
 import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
+import com.google.protobuf.ByteString;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -30,6 +31,7 @@ import org.tron.plugins.utils.db.DBInterface;
 import org.tron.plugins.utils.db.DBIterator;
 import org.tron.plugins.utils.db.DbTool;
 import org.tron.protos.Protocol;
+import org.tron.protos.contract.BalanceContract;
 import picocli.CommandLine;
 
 @Slf4j(topic = "lite")
@@ -57,13 +59,17 @@ public class DbLite implements Callable<Integer> {
   private static final String TRANSACTION_HISTORY_DB_NAME = "transactionHistoryStore";
   private static final String PROPERTIES_DB_NAME = "properties";
   private static final String TRANS_CACHE_DB_NAME = "trans-cache";
+  private static final String BALANCE_TRACE_DB_NAME = "balance-trace";
+  private static final String ACCOUNT_TRACE_DB_NAME = "account-trace";
 
   private static final List<String> archiveDbs = Arrays.asList(
       BLOCK_DB_NAME,
       BLOCK_INDEX_DB_NAME,
       TRANS_DB_NAME,
       TRANSACTION_RET_DB_NAME,
-      TRANSACTION_HISTORY_DB_NAME);
+      TRANSACTION_HISTORY_DB_NAME,
+      BALANCE_TRACE_DB_NAME,
+      ACCOUNT_TRACE_DB_NAME);
 
   enum Operate { split, merge }
 
@@ -522,24 +528,42 @@ public class DbLite implements Callable<Integer> {
     DBInterface blockDb = DbTool.getDB(liteDir, BLOCK_DB_NAME);
     DBInterface transDb = DbTool.getDB(liteDir, TRANS_DB_NAME);
     DBInterface tranRetDb = DbTool.getDB(liteDir, TRANSACTION_RET_DB_NAME);
-
+    DBInterface balanceTraceDb = DbTool.getDB(liteDir, BALANCE_TRACE_DB_NAME);
+    DBInterface accountTraceDb = DbTool.getDB(liteDir, ACCOUNT_TRACE_DB_NAME);
 
     ProgressBar.wrap(LongStream.rangeClosed(start, end)
         .boxed()
-        .sorted((a, b) -> Long.compare(b, a)), "trimHistory").forEach(n -> {
+        .sorted((a, b) -> Long.compare(b, a)), "trimHistory")
+        .map(ByteArray::fromLong).forEach(n -> {
           try {
-            byte[] blockIdHash = blockIndexDb.get(ByteArray.fromLong(n));
+            byte[] blockIdHash = blockIndexDb.get(n);
             Protocol.Block block = Protocol.Block.parseFrom(blockDb.get(blockIdHash));
             // delete transactions
             for (Protocol.Transaction e : block.getTransactionsList()) {
               transDb.delete(DBUtils.getTransactionId(e).getBytes());
             }
             // delete transaction result
-            tranRetDb.delete(ByteArray.fromLong(n));
+            tranRetDb.delete(n);
             // delete block
             blockDb.delete(blockIdHash);
             // delete block index
-            blockIndexDb.delete(ByteArray.fromLong(n));
+            blockIndexDb.delete(n);
+            byte[] balanceTrace = balanceTraceDb.get(n);
+            if (balanceTrace != null) {
+              // delete account trace
+              long blockNum = ByteArray.toLong(n);
+              BalanceContract.BlockBalanceTrace.parseFrom(balanceTrace)
+                  .getTransactionBalanceTraceList().stream()
+                  .flatMap(tx -> tx.getOperationList().stream())
+                  .map(BalanceContract.TransactionBalanceTrace.Operation::getAddress)
+                  .distinct()
+                  .map(ByteString::toByteArray)
+                  .map(address -> Bytes.concat(address, Longs.toByteArray(
+                      blockNum ^ Long.MAX_VALUE)))
+                  .forEach(accountTraceDb::delete);
+              // delete balance trace
+              balanceTraceDb.delete(n);
+            }
           } catch (IOException e) {
             throw new RuntimeException(e);
           }
@@ -570,7 +594,8 @@ public class DbLite implements Callable<Integer> {
         DBInterface bakDb = DbTool.getDB(bakDir.toString(), dbName);
         DBInterface destDb = DbTool.getDB(liteDir, dbName);
         try (DBIterator iterator = bakDb.iterator()) {
-          if (TRANS_DB_NAME.equals(dbName) || TRANSACTION_HISTORY_DB_NAME.equals(dbName)) {
+          if (TRANS_DB_NAME.equals(dbName) || TRANSACTION_HISTORY_DB_NAME.equals(dbName)
+              || ACCOUNT_TRACE_DB_NAME.equals(dbName)) {
             iterator.seekToFirst();
           } else {
             iterator.seek(head);
--- a/plugins/src/test/java/org/tron/plugins/DbLiteTest.java
+++ b/plugins/src/test/java/org/tron/plugins/DbLiteTest.java
@@ -71,7 +71,7 @@ public class DbLiteTest {
     context.close();
   }
 
-  public void init(String dbType) throws IOException {
+  public void init(String dbType, boolean historyBalanceLookup) throws IOException {
     dbPath = folder.newFolder().toString();
     Args.setParam(new String[] {
         "-d", dbPath, "-w", "--p2p-disable", "true", "--storage-db-engine", dbType},
@@ -80,6 +80,7 @@ public class DbLiteTest {
     Args.getInstance().setAllowAccountStateRoot(1);
     Args.getInstance().setRpcPort(PublicMethod.chooseRandomPort());
     Args.getInstance().setRpcEnable(true);
+    Args.getInstance().setHistoryBalanceLookup(historyBalanceLookup);
     databaseDir = Args.getInstance().getStorage().getDbDirectory();
     // init dbBackupConfig to avoid NPE
     Args.getInstance().dbBackupConfig = DbBackupConfig.getInstance();
@@ -99,7 +100,8 @@ public class DbLiteTest {
       throws InterruptedException, IOException {
     logger.info("dbType {}, checkpointVersion {}, excludeHistoricalBalance {}",
         dbType, checkpointVersion, excludeHistoricalBalance);
-    init(dbType);
+    boolean historyBalanceLookup = excludeHistoricalBalance;
+    init(dbType, historyBalanceLookup);
     final String[] argsForSnapshot = excludeHistoricalBalance
         ? new String[] {"-o", "split", "-t", "snapshot", "--fn-data-path",
             dbPath + File.separator + databaseDir, "--dataset-path",
--- a/plugins/src/test/java/org/tron/plugins/DbLiteTest.java
+++ b/plugins/src/test/java/org/tron/plugins/DbLiteTest.java
@@ -1,11 +1,14 @@
 package org.tron.plugins;
 
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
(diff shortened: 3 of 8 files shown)


plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -522,24 +528,42 @@ private void trimExtraHistory(String liteDir, BlockNumInfo blockNumInfo)
     DBInterface blockDb = DbTool.getDB(liteDir, BLOCK_DB_NAME);
     DBInterface transDb = DbTool.getDB(liteDir, TRANS_DB_NAME);
     DBInterface tranRetDb = DbTool.getDB(liteDir, TRANSACTION_RET_DB_NAME);
-
+    DBInterface balanceTraceDb = DbTool.getDB(liteDir, BALANCE_TRACE_DB_NAME);
```
waynercheung: [SHOULD] `DbTool.getDB(liteDir, BALANCE_TRACE_DB_NAME)` opens with create-if-missing. On a default-configured node (`historyBalanceLookup=false`), `balance-trace` / `account-trace` may not exist at all; this change will create empty DB directories during merge, which is not quite "no impact for default-configured nodes". Please guard these opens with an existence check and skip the whole trace-trimming path when the DB is absent. The same issue also applies to the later bak-side opens in `mergeBak2Database()`.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -522,24 +528,42 @@ private void trimExtraHistory(String liteDir, BlockNumInfo blockNumInfo)
...
-            tranRetDb.delete(ByteArray.fromLong(n));
+            tranRetDb.delete(n);
             // delete block
             blockDb.delete(blockIdHash);
             // delete block index
-            blockIndexDb.delete(ByteArray.fromLong(n));
+            blockIndexDb.delete(n);
+            byte[] balanceTrace = balanceTraceDb.get(n);
+            if (balanceTrace != null) {
+              // delete account trace
```
waynercheung: [SHOULD] This reconstructs the `account-trace` composite key (`address || Longs.toByteArray(blockNum ^ Long.MAX_VALUE)`) inline. The same encoding also exists in `AccountTraceStore#recordBalanceWithBlock()` and in `AccountTraceStoreTest`, so there is no single source of truth.

Drift here is silent: if the key format ever changes (XOR mask, byte order, address layout/prefix, etc.), this trim path will stop deleting the intended entries and leave orphaned rows in `account-trace`. Since `AccountTraceStore#getPrevBalance()` later reads directly from this key space, that drift can affect historical balance lookups rather than just leaving internal DB dirt.

Please extract a shared helper for this key encoding. Since `plugins` does not depend on `chainbase`, the helper may need to live in a lower-level common package. At minimum, add a contract test that writes via `AccountTraceStore` and verifies `DbLite` reconstructs the same key, so future drift breaks CI instead of failing silently.

plugins/src/test/java/org/tron/plugins/DbLiteTest.java
```diff
@@ -89,7 +89,13 @@ public void clear() {
 
   public void testTools(String dbType, int checkpointVersion)
```
waynercheung, line 94: [SHOULD] The 2-arg overload now defaults `advanceSnapshot=false`, which reduces existing coverage for the legacy test matrix and also leaves the new `ACCOUNT_TRACE_DB_NAME` path in `mergeBak2Database()` unexercised for checkpoint v2.
Please either preserve the previous behavior for this overload or add explicit coverage, at minimum a `(ROCKSDB, 2, true, true)` case, so the new merge branch is actually tested.

plugins/src/test/java/org/tron/plugins/DbLiteTest.java
```diff
@@ -117,8 +124,7 @@ public void testTools(String dbType, int checkpointVersion)
     cli.execute(argsForSnapshot);
```
waynercheung, line 130: [SHOULD] The test currently proves startup/merge still works end-to-end, but it does not directly verify the packaging contract introduced by this PR. Please add assertions in `testTools` for both split outputs:

- after `cli.execute(argsForSnapshot)`, assert `snapshot/` does not contain `balance-trace` or `account-trace`
- after `cli.execute(argsForHistory)`, assert:
    - when `historyBalanceLookup=true`, `history/` contains both DBs with at least one entry, not just empty RocksDB directories"
    - when `historyBalanceLookup=false`, `history/` contains neither

When `historyBalanceLookup=true`, it would also help to sanity-check that the source DB produced these two stores before split, so the snapshot assertions do not pass vacuously. This is the regression guard for #6597; otherwise a future change to `archiveDbs` could silently pass CI.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -570,7 +594,8 @@ private void mergeBak2Database(String liteDir, BlockNumInfo blockNumInfo) throws
         DBInterface bakDb = DbTool.getDB(bakDir.toString(), dbName);
         DBInterface destDb = DbTool.getDB(liteDir, dbName);
         try (DBIterator iterator = bakDb.iterator()) {
-          if (TRANS_DB_NAME.equals(dbName) || TRANSACTION_HISTORY_DB_NAME.equals(dbName)) {
+          if (TRANS_DB_NAME.equals(dbName) || TRANSACTION_HISTORY_DB_NAME.equals(dbName)
```
waynercheung: [DISCUSS] `ACCOUNT_TRACE_DB_NAME` now falls back to `seekToFirst()`, so merge replays the entire bak DB instead of appending only entries with height > `historyMax` as described in #6597. Is the invariant here that the overlapping segment is guaranteed to be identical on both sides, so full replay is safe? If yes, please document that assumption; otherwise I’d prefer filtering by the encoded block-number suffix to keep the merge strictly height-scoped.

waynercheung:
[NIT] A few minor cleanup items:
- Please move the unrelated `.gitignore` hunk out of this PR so the review stays scoped.
- The `// delete account trace` comment in `DbLite.java` reads as if it belongs to `long blockNum = ...`; please move/reword it so it describes the stream below more clearly.
- The nested ternary in `DbLiteTest.java` is hard to parse; please extract it into a local variable before calling `generateSomeTransactions(...)`.

.gitignore
```diff
@@ -57,3 +57,6 @@ Wallet
 
 /framework/propPath
 .cache
+
+#AI
+.claude
```
xxo1shine: [NIT] This `.claude` entry is unrelated to the feature in this PR. Consider removing it here or moving it to a separate commit.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -570,7 +594,8 @@ private void mergeBak2Database(String liteDir, BlockNumInfo blockNumInfo) throws
...
+              || ACCOUNT_TRACE_DB_NAME.equals(dbName)) {
             iterator.seekToFirst();
           } else {
             iterator.seek(head);
```
xxo1shine, line 642: [QUESTION] `ACCOUNT_TRACE_DB_NAME` is added to the `seekToFirst()` branch because its keys are `address + blockNum` and cannot be positioned by block number alone. But `BALANCE_TRACE_DB_NAME` is not included and falls through to `iterator.seek(head)`. What is the byte format of `head` here? If it is a block ID hash rather than a serialized block number, it would be incompatible with the block-number-keyed layout of balance-trace and the seek would land at the wrong position. Please confirm `head` and the balance-trace key format are consistent.

plugins/src/test/java/org/tron/plugins/rocksdb/DbLiteWithHistoryRocksDbTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.io.IOException;
+import org.junit.Test;
+import org.tron.plugins.DbLiteTest;
+
+public class DbLiteWithHistoryRocksDbTest extends DbLiteTest {
+
+  @Test
+  public void testToolsWithTrimHistory() throws InterruptedException, IOException {
+    testTools("ROCKSDB", 1, true, true);
```
xxo1shine: [NIT] Both new test classes (`DbLiteWithHistoryRocksDbTest` and `DbLiteWithHistoryRocksDbV2Test`) use the method name `testToolsWithTrimHistory`, but they cover checkpoint v1 and v2 respectively. The names are indistinguishable in test reports. Consider suffixing them, e.g. `testToolsWithTrimHistoryV1` / `testToolsWithTrimHistoryV2`.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
halibobo1205: `balance-trace` / `account-trace`  is initiated by spring, not controlled by `historyBalanceLookup=false`, see `AccountTraceStore` and `BalanceTraceStore`.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
halibobo1205: ` if the key format ever changes`: never happened, if it does, it will be a breaking change.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
halibobo1205: `head = ByteArray.fromLong(start)` is an 8-byte block number, and `BalanceTraceStore#putBlockBalanceTrace` writes keys as `ByteArray.fromLong(blockId.getNum())` (BalanceTraceStore.java:129). So the key format matches and `seek(head)` lands at the right position for `balance-trace`. `account-trace` needs `seekToFirst()` only because its key is `address || (blockNum ^ Long.MAX_VALUE)` and can't be positioned by block number alone.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -522,24 +528,42 @@ private void trimExtraHistory(String liteDir, BlockNumInfo blockNumInfo)
...
+            tranRetDb.delete(n);
             // delete block
             blockDb.delete(blockIdHash);
             // delete block index
-            blockIndexDb.delete(ByteArray.fromLong(n));
+            blockIndexDb.delete(n);
+            byte[] balanceTrace = balanceTraceDb.get(n);
+            if (balanceTrace != null) {
+              long blockNum = ByteArray.toLong(n);
+              // delete account trace: one row per (address, blockNum) touched in this block
```
waynercheung: [SHOULD] This still reconstructs the `account-trace` composite key inline. The same encoding also exists in `AccountTraceStore` and `AccountTraceStoreTest`, so there is still no single source of truth.

Drift here is silent: deleting a non-existent key is a no-op, so if the key format ever changes, trim will stop deleting the intended rows and leave orphaned `account-trace` entries behind. Since `AccountTraceStore#getPrevBalance()` later reads from the same key space, this can leak into historical balance lookups rather than just leaving internal DB dirt.

Please extract a shared helper in a lower-level module accessible to both sides, or at minimum add a contract test that keeps the store-side write path and the toolkit-side reconstruction in sync.

plugins/src/test/java/org/tron/plugins/DbLiteTest.java
```diff
@@ -115,14 +125,22 @@ public void testTools(String dbType, int checkpointVersion)
...
-    generateSomeTransactions(checkpointVersion == 1 ? 6 : 18);
+    int postSnapshotTxSeconds = advanceSnapshot && checkpointVersion == 1 ? 6 : 18;
+    generateSomeTransactions(postSnapshotTxSeconds);
     // stop the node
     shutdown();
     // generate history
     cli.execute(argsForHistory);
+    // history carries balance-trace / account-trace regardless of lookup setting
+    Path historyDir = Paths.get(dbPath, "history");
+    assertTrue(historyDir.resolve("balance-trace").toFile().exists());
```
waynercheung: [SHOULD] These assertions only check `File.exists()`, which is weaker than the packaging contract this PR is trying to protect. In this repo `balance-trace` / `account-trace` directories can already exist as empty stores, so `exists()` cannot distinguish "history was correctly populated" from "an empty DB dir is present".

When `historyBalanceLookup=true`, please strengthen this with an actual-content check, e.g. by opening the DB via `DbTool` and asserting `iterator().hasNext()`. The current `exists()` assertions are still useful as a lower-bound check; they just are not sufficient as the positive proof for the history-enabled case.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
halibobo1205: ` if the key format ever changes`: never happened, if it does, it will be a breaking change.

xxo1shine (approved):
approved!

plugins/src/main/java/common/org/tron/plugins/DbLite.java
waynercheung: Could we at least add a one-line comment on both sides pointing at each other? Something like:

    // AccountTraceStore.java, above recordBalanceWithBlock:
    // NOTE: key format is also reconstructed in plugins DbLite#trimExtraHistory.

    // DbLite.java, above the composite key line:
    // NOTE: must match AccountTraceStore#recordBalanceWithBlock key format.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
waynercheung, line 449: [MUST] Functional regression for snapshot-based lite nodes with historyBalanceLookup=true.

Adding balance-trace and account-trace to archiveDbs removes both stores from every snapshot, but fillSnapshotBlockAndTransDb() only rebuilds block, block-index, and trans. As a result, a lite node started directly from the snapshot has retained blocks in block DB, but no matching balance/account trace state for that retained range:

Wallet#getBlockBalance() throws ItemNotFoundException
Wallet#getAccountBalance() can silently return 0, because AccountTraceStore#getPrevBalance() falls back to Pair.of(number, 0L) on an empty lookup
Pre-PR, snapshots generated from a source node with historyBalanceLookup=true carried these stores, so this is a regression for pre-merge snapshot operation, not just a known limitation.

Please preserve the trace state required for retained-block balance queries, or explicitly disable these APIs on snapshot-only lite nodes until history has been merged, and add coverage for that behavior.

Note that copying only the recent-window account-trace rows is not sufficient: getPrevBalance() needs the latest entry at or before the queried block, which may precede the retained window for accounts that did not change inside it.

halibobo1205:
## problem summary

### A · Pre-existing lite-mechanism defects (unrelated to this PR)

**A1. `transactionRetStore` is not aligned with the block window**  
A lite node retains the most recent 65,535 blocks, but `transactionRetStore` does not hold a complete set of receipts for that same window — there are gaps.

**A2. `section-bloom` and block body are asymmetric**  
`section-bloom` is kept in full, while block bodies are trimmed to the last 65,535. Queries can match the bloom filter but find no corresponding block/transaction.

---

### B · Semantic nature of historical balance lookup

**B1. State-tree semantics, not log semantics**  
`balance-trace` / `account-trace` must be accumulated continuously from genesis to correctly answer "what was the balance at any historical height." **Any gap is a correctness bug.**

**B2. Only valid on a full node that had the feature enabled from genesis**  
A lite node has no legitimate "partially enabled" state — even if the data is present, a lite node cannot answer the query independently.

**B3. The switch can be flipped at runtime**  
Because `historyBalanceLookup` can be toggled on/off at will, the same data directory may contain interleaved enabled/disabled segments. **The original full node's store may itself already be corrupt by this definition.**

**B4. No persisted "enabled window" metadata on the node side**  
Nothing like `firstEnabledBlockNum` is written to disk, so neither the toolkit nor the runtime can determine which block range the data legitimately covers.

---

### C · Current state of the code (important correction)

**C1. The two trace DBs are currently treated as state stores**  
Before this PR, `balance-trace` / `account-trace` are classified alongside `account` and get packaged into the lite snapshot — they are **not** archive DBs.

**C2. These two stores are close to 900 GB in size**  
On a full node with lookup enabled, the snapshot drags along that 900 GB, which makes **the resulting lite snapshot operationally unusable** (distribution, provisioning, and startup costs all become unacceptable).

**C3. This is the actual motivation behind PR**  
Moving these DBs from snapshot into `archiveDbs` is what makes "slice a usable lite snapshot out of a lookup-enabled full node" **actually feasible**. The space benefit is not "≈0" — that only holds for the small case where lookup is off on both ends. In the real-world lookup-enabled scenario, the benefit is on the order of **−900 GB**.

**C4. Illegal merge combinations still exist**  
Of the four (lite off/on × history off/on) combinations, only (on, on) from the same source is semantically valid; the other three can still silently happen. But given B3 + B4, any toolkit-level check can only catch the obvious misconfigurations, not the semantic ones — and a half-check offers false assurance.

---

### D · Re-evaluation of this PR's value

**D1. What the PR does**  
Moves the two trace DBs from the snapshot set into `archiveDbs` (so `split -t history` packages them, and `merge` restores them).

**D2. Impact on the default user (lookup off on both ends — the majority)**  
The two DBs are Spring-initialized empty directories and never occupied meaningful space in the snapshot to begin with. The PR brings **no net gain here, but also no cost** — just two empty dirs travel via history instead.

**D3. Impact on lookup-enabled full nodes (niche but real)**  
**Removes 900 GB from the snapshot**, turning lite-snapshot slicing from "infeasible" into "feasible." This is the core value of the PR.

**D4. The functional loss is not something this PR introduces**  
A lite node cannot serve historical balance lookup — that is a consequence of B2, independent of this PR. Whether or not we merge this PR, the lite node cannot answer the query. The PR neither causes nor worsens this; it only makes lite-snapshotting a lookup-enabled node physically viable.

**D5. The PR adds no warnings or validation**  
`split` does not notify the operator; `merge` does not check for illegal combinations.

---
### E · Guiding principle

**The toolkit is just a tool. The correctness of the node's own data is the operator's responsibility, and that responsibility should not be pushed onto the tool.**

- `DbLite` physically moves files between directories; it has no authoritative view of whether the source data is semantically valid (see B3, B4).
- Any validation logic the toolkit layers on top can only reject obvious misconfigurations. It cannot — and should not be expected to — detect semantic corruption caused by runtime switch flipping, interleaved enabled windows, cross-source mixing, etc.
- Adding heavy checks inside the toolkit therefore trades real engineering cost for **a false sense of safety**: users who skip node-side hygiene will still produce broken data; users who do node-side hygiene do not need the toolkit check.
- The right place to enforce continuity guarantees is the node itself (persisted enabled-window metadata, immutable-once-enabled semantics, etc. — see B4 as a follow-up). Until that exists, the toolkit should stay thin, do exactly one physical job well, and document the semantic contract clearly rather than pretend to police it.

**Corollary for this PR**: keep the scope to the physical split; do not introduce validation or compatibility scaffolding that we cannot honor consistently. **On the above reasoning, I consider this PR to have already absorbed enough compatibility work** — it delivers the real, concrete need ("let a lookup-enabled node produce a usable lite snapshot") with the minimal code change needed to physically separate the two stores, without pulling semantic responsibilities into the toolkit that do not belong there.

 cc @waynercheung @xxo1shine @Sunny6889

plugins/src/main/java/common/org/tron/plugins/DbLite.java
halibobo1205: See the latest comment.

waynercheung:
Reconsidered after @halibobo1205's A-E analysis. The operational motivation is valid, and "operator responsibility" is a reasonable product stance here, provided the limitation is documented clearly.

Withdrawing my earlier [MUST] on the snapshot-range query regression.
Given that the silent-0 behavior in `Wallet.getAccountBalance` is pre-existing (for `historyBalanceLookup=false` nodes), this PR enlarges rather than introduces that failure surface, so I no longer consider that point blocking for this PR.

Code changes LGTM. Ready to approve once one doc item lands; the other two are lightweight follow-ups.

[MUST] Update the PR description to state explicitly that lite snapshots no longer contain `balance-trace` / `account-trace` for retained blocks, and that `getBlockBalance` / `getAccountBalance` must not be relied on until `merge` has completed. If applicable, please also mention this in release notes / upgrade notes. Operator responsibility only works when operators are told about the changed contract up front.

[SHOULD] Please file a follow-up issue - this one is independent of the key-format question below, and is not covered by the "never happened / breaking change" argument. `AccountTraceStore#getPrevBalance` returning Pair.of(n, 0L) on empty lookup means Wallet.getAccountBalance surfaces balance=0 with no error signal. Because 0 is a legitimate balance value, clients cannot distinguish "account truly has 0" from "index isn't populated". This PR enlarges the set of configurations that trigger this. Just a tracking issue - no fix needed in this PR.

[SHOULD] Given that the on-disk key format is stable in practice, I'll drop the full helper/contract-test ask. But please add reciprocal cross-reference comments so the DbLite <-> AccountTraceStore coupling is discoverable:

 `chainbase/.../AccountTraceStore.java`, above recordBalanceWithBlock:
```
      // NOTE: the composite key layout (address || Longs.toByteArray(
      // number ^ Long.MAX_VALUE)) is also reconstructed in plugins
      // DbLite#trimExtraHistory. Keep them in sync.
```

`plugins/.../DbLite.java`, above the composite key build:
```
      // Must match AccountTraceStore#recordBalanceWithBlock key layout.
```

halibobo1205:
@waynercheung done.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
waynercheung, line 620: [MUST] This early return can permanently discard snapshot-side trace data.

At this point step 2 has already moved the lite node's archive DBs into `.bak_*`, and step 3 has copied whatever trace DBs the history dataset happens to contain. If the history dataset does not provide complete `balance-trace` / `account-trace` coverage for this merge case -- for example:

- history pack produced by a toolkit version before this PR, where the trace stores were not archived at all
- history pack whose trace stores exist on disk but do not provide usable post-genesis coverage for the range this merge assumes
- manually prepared / pruned pack

-- then `start > end` skips backup replay entirely here, and step 6 deletes the only remaining copy in `.bak_*`. Recent `getBlockBalance` / `getAccountBalance` data that existed on the snapshot side before merge silently disappears, with no error signal.

Please either:
- fail fast before any destructive step when the history side does not provide the required trace coverage, or
- restore the snapshot-side trace DBs from `.bak_*` even in the `start > end` case before deleting the backup.

See the comment below: both failure modes come from treating incomplete history-side trace coverage as authoritative.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -570,7 +595,13 @@ private void mergeBak2Database(String liteDir, BlockNumInfo blockNumInfo) throws
...
+          // These DBs cannot be positioned by block number:
+          //   TRANS_DB_NAME / TRANSACTION_HISTORY_DB_NAME are keyed by txId;
+          //   ACCOUNT_TRACE_DB_NAME is keyed by address || (blockNum ^ Long.MAX_VALUE).
+          // Full replay is safe: the overlapping segment [0, historyMaxNum] is
+          // identical on both sides, so re-putting it is idempotent.
+          if (TRANS_DB_NAME.equals(dbName) || TRANSACTION_HISTORY_DB_NAME.equals(dbName)
+              || ACCOUNT_TRACE_DB_NAME.equals(dbName)) {
             iterator.seekToFirst();
           } else {
             iterator.seek(head);
```
waynercheung, line 642: [MUST] `balance-trace` cannot safely use `seek(head)` unless the history dataset is guaranteed to provide complete `balance-trace` coverage up to `historyMaxNum`.

The branch above replays `ACCOUNT_TRACE_DB_NAME` from `seekToFirst()`, while `balance-trace` lands here on `seek(head)`. When the history side under-populates `balance-trace`, the overlap `<= historyMaxNum` is restored from bak for `account-trace` but skipped for `balance-trace`. After such a merge, the two stores can diverge: `getAccountBalance` may still return a value from replayed `account-trace`, while `getBlockBalanceTrace` returns `ItemNotFoundException` for overlapping blocks.

Thus the current logic is only correct under the invariant: "history-side `balance-trace` coverage is complete up to `historyMaxNum`". Either:
- enforce that invariant before any destructive step, or
- replay `balance-trace` from backup in the same conservative way as `account-trace` when the history side is missing or incomplete.

Please add tests for both paths:
- `historyMaxNum >= snapshotMaxNum` with incomplete history-side trace coverage
- `historyMaxNum < snapshotMaxNum` with incomplete history-side trace coverage

The current suite only exercises symmetric history packs and will not catch either regression.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -253,12 +278,50 @@ public void completeHistoryData(String historyDir, String liteDir) {
...
+   * additionally excluded, producing a smaller lite snapshot at the cost of
+   * dropping historical balance lookup support on the resulting lite node.
+   * Only {@code split -t snapshot} consults this. {@code split -t history}
+   * and {@code merge} always use the legacy archive set.
+   */
+  private List<String> snapshotExclusion() {
+    if (!excludeHistoricalBalance) {
+      return archiveDbs;
+    }
+    return Stream.concat(archiveDbs.stream(), traceDbs.stream())
```
waynercheung, line 299: [MUST] When this exclusion is enabled on a source node that had `historyBalanceLookup=true`, the resulting lite snapshot can still serve historical-balance RPCs, but not safely.

For blocks that are still retained in the snapshot `block-index`, `account-trace` is recreated as an empty store on boot, `AccountTraceStore#getPrevBalance()` maps an empty lookup to `Pair.of(number, 0L)`, and `Wallet.getAccountBalance()` returns that as a successful `balance=0`. For the same block, `getBlockBalance()` throws because `balance-trace` is absent.

So this flag does not merely "disable" historical balance lookup; it can silently return wrong account balances. That is a correctness issue, and it does not match the current help / warning / README wording of "cannot answer".

Please make this mode fail closed before merge, for example by persisting a marker in the snapshot and rejecting `getAccountBalance` / `getBlockBalance` on lite nodes created with this flag, or by retaining the trace coverage needed for snapshot-retained blocks. At minimum, the operator-facing text must describe the real asymmetry explicitly.

plugins/src/test/java/org/tron/plugins/DbLiteTest.java
```diff
@@ -89,10 +92,19 @@ public void clear() {
 
   public void testTools(String dbType, int checkpointVersion)
       throws InterruptedException, IOException {
-    logger.info("dbType {}, checkpointVersion {}", dbType, checkpointVersion);
+    testTools(dbType, checkpointVersion, false);
+  }
+
+  public void testTools(String dbType, int checkpointVersion, boolean excludeHistoricalBalance)
```
waynercheung, line 99: [SHOULD] This opt-in test path never enables `historyBalanceLookup=true` on the source node, so it only covers the default configuration where this flag has no functional effect.

`config-localtest.conf` inherits `storage.balance.history.lookup=false`, which means the source node never produces meaningful `balance-trace` / `account-trace` data. The assertions here prove "the directories were excluded", but they do not cover the only configuration this flag is meant for.

Please add a case with `historyBalanceLookup=true` and assert the post-snapshot behavior of `getAccountBalance` / `getBlockBalance` for a snapshot-retained block.

plugins/src/main/java/common/org/tron/plugins/DbLite.java
```diff
@@ -119,6 +143,7 @@ public Integer call() {
     try {
       switch (this.operate) {
         case split:
+          warnIfExcludingHistoricalBalance();
```
waynercheung: [NIT] `warnIfExcludingHistoricalBalance()` runs for every `split`, so `split -t history --exclude-historical-balance` will still print a snapshot-specific permanent-loss warning even though history split ignores this flag.

waynercheung (approved):
LGTM

xxo1shine (approved):

PR #6693, opened 2026-04-17
feat(api): optimize and harden HTTP/JSON-RPC API layer

### LINKED ISSUE #6606: [Feature] Block Interface JSON Serialization Optimization
## Problem Statement

　　The current block-related HTTP interfaces have redundant conversion paths in the JSON serialization process: first fully serialize BlockList / Block into a JSON string, then deserialize it into an object, and finally overwrite some fields. This process introduces the following issues: repeated serialization overhead (multiple conversions of protobuf → JSON → object), causing unnecessary CPU consumption; increased memory and GC pressure due to the creation of a large number of intermediate objects, which amplifies GC overhead under high-frequency calls; increased interface latency, as block queries are high-frequency APIs and the serialization path becomes a performance bottleneck.

## Proposed Solution

　　Optimize the JSON construction approach of block interfaces by removing the full serialization process and switching to on-demand JSON construction.

### Specification

　　Core optimization: avoid full conversion of protobuf → JSON → object, and instead construct JSON manually (field-by-field population).
　　Method-level changes: printBlockList no longer fully serializes BlockList, but directly constructs a JSONObject and populates the block array; printBlockToJSON no longer fully serializes Block, but only serializes block_header and manually populates blockID and transactions (when non-empty).

**API Changes** (if applicable)
　　The following interfaces are affected, but the returned fields remain unchanged: GetBlockByLimitNext, GetBlockByLatestNum, GetBlock, GetNowBlock, GetBlockByNum, GetBlockByIdS; returned fields remain unchanged: blockID, block_header, transactions (when present).

## Scope of Impact
　　API
　　Tests

**Breaking Changes**
　　None

**Backward Compatibility**
　　Fully compatible, response structure remains unchanged, no modification required for callers.

## Implementation

　　Changes: Modify Util.java to construct JSON on demand, remove the full serialization path, while maintaining consistency of output fields; enhance tests: add structural assertions in UtilMockTest.testPrintBlockList.

　　Yes

**Estimated Complexity**
　　Low, changes are localized with no cross-module impact, low risk.

## Testing Strategy

**Test Scenarios**
　　Single block interface validation: response contains blockID, block_header, transactions (when present); block list interface: correct length of block array and consistent structure for each element; visible mode regression (visible=true / false): consistent address and field formats; transaction scenarios: blocks with no transactions and blocks with transactions.

　　Compare interfaces before and after optimization (getblockbylatestnum, getblockbylimitnext), focusing on metrics: average latency, P95 / P99 latency, CPU usage, GC frequency, and object allocation; expected results: reduced serialization time, decreased memory allocation, and lower GC pressure.

## Alternatives Considered (Optional)

　　Keep full serialization + caching: limited benefits and introduces cache consistency complexity; replace JSON library: large scope of changes and high risk; trim response at upper layer: cannot solve the underlying repeated serialization problem.

　　This optimization is a hot path performance optimization: targeting high-frequency block query interfaces, reducing redundancy in the serialization pipeline to improve API throughput and response stability; applicable scenarios include batch queries by block explorers, block replay by wallets/gateways, and high-frequency block detail queries.

**Related Issues/PRs**
　　(To be added)

bladehan1:
**What does this PR do?**

1. Optimize block JSON serialization in `printBlockList` / `printBlockToJSON`: remove the full protobuf → JSON → object round-trip; construct the JSON field-by-field instead
2. Replace `Class.forName` with a strict whitelist in `RateLimiterServlet`; unknown adapter names now fail-fast via `TronError(RATE_LIMITER_INIT)` from `@PostConstruct`, so a misconfigured node aborts startup instead of silently running with the default limiter
3. Add a length guard for oversized hex block-number input at the JSON-RPC input layer — new `JsonRpcApiUtil.parseBlockNumber(blockNumOrTag)` helper enforces a 100 hex-char max and maps parse failures to `JsonRpcInvalidParamsException`; used at five call sites in `TronJsonRpcImpl` / `Wallet`
4. Downgrade broadcast transaction failures caused by user-side input (sig error, contract validation, insufficient balance, etc.) from WARN to INFO in `Wallet.broadcastTransaction`
5. Downgrade `NonUniqueObjectException` from ERROR to DEBUG in `RpcApiService.getAssetIssueByName` — the `name` field is no longer unique since the business-logic change, so collisions are an expected condition
6. Remove unused `Wallet.getMerkleTreeOfBlock` and `Wallet.deployContract`

1. Block query is a high-frequency API; the redundant intermediate object creation increases CPU usage, GC pressure, and interface latency
2. `Class.forName` allowed arbitrary class loading via a tampered config file. A failed adapter instantiation should abort startup (fail-fast) to surface the misconfiguration rather than silently degrade security policy
3. Oversized hex strings cause excessive CPU consumption during `BigInteger` parsing, enabling DoS. The length validation belongs in the API input layer, not in the generic low-level byte utility — `ByteArray.hexToBigInteger` is left unchanged
4. These are expected client-side rejections; WARN-level alerting creates operational noise and obscures real issues
5. Same rationale — expected condition should not log at ERROR
6. Dead code increases maintenance burden

**Follow-ups (not in this PR):**
- Handle malformed POST JSON bodies with an explicit HTTP 400 at the servlet `doPost` entry. An earlier attempt added this as a catch in `Util.checkGetParam`, but the utility layer conflated "malformed body" with "missing key" — rolled back per review, to be revisited at the correct layer in a separate PR.

- Unit Tests (`RateLimiterServletTest`, `UtilMockTest`, `JsonRpcApiUtilTest`)
- Manual Testing

Closes #6606


framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -541,9 +543,13 @@ private static String checkGetParam(HttpServletRequest request, String key) thro
...
-        JSONObject jsonObject = JSON.parseObject(value);
-        if (jsonObject != null) {
-          return jsonObject.getString(key);
+        try {
+          JSONObject jsonObject = JSON.parseObject(value);
+          if (jsonObject != null) {
+            return jsonObject.getString(key);
+          }
+        } catch (JSONException e) {
+          logger.debug("Invalid JSON input: {}", e.getMessage());
```
317787106, line 552: It should throw an Exception here, so http user can get this error message.

framework/src/main/java/org/tron/core/services/RpcApiService.java
```diff
@@ -418,7 +418,7 @@ public void getAssetIssueByName(BytesMessage request,
           responseObserver.onNext(wallet.getAssetIssueByName(assetName));
         } catch (NonUniqueObjectException e) {
           responseObserver.onNext(null);
-          logger.error("Solidity NonUniqueObjectException: {}", e.getMessage());
+          logger.debug("Solidity NonUniqueObjectException: {}", e.getMessage());
```
317787106, line 421: It should not swallow `NonUniqueObjectException` here, such as using `responseObserver.onError(xxx)`;

framework/src/test/java/org/tron/core/services/http/UtilMockTest.java
```diff
@@ -44,14 +46,67 @@ public void testPrintTransactionFee() {
   public void testPrintBlockList() {
     BlockCapsule blockCapsule1 = new BlockCapsule(1, Sha256Hash.ZERO_HASH,
         System.currentTimeMillis(), Sha256Hash.ZERO_HASH.getByteString());
-    BlockCapsule blockCapsule2 = new BlockCapsule(1, Sha256Hash.ZERO_HASH,
+    BlockCapsule blockCapsule2 = new BlockCapsule(2, Sha256Hash.ZERO_HASH,
         System.currentTimeMillis(), Sha256Hash.ZERO_HASH.getByteString());
     GrpcAPI.BlockList list = GrpcAPI.BlockList.newBuilder()
```
317787106, line 51: Maybe a test case should be added to compare the change of `printBlockToJSON`, verify they are same as before.

common/src/main/java/org/tron/common/utils/ByteArray.java
```diff
@@ -143,7 +143,22 @@ public static String toJsonHex(String x) {
...
+   * representations, with headroom. The exact value is not critical for
+   * performance — BigInteger parsing at this scale is negligible.
+   */
+  private static final int MAX_HEX_BIG_INTEGER_LEN = 100;
+
   public static BigInteger hexToBigInteger(String input) {
+    if (input == null) {
+      throw new IllegalArgumentException("hex input is null");
+    }
+    if (input.length() > MAX_HEX_BIG_INTEGER_LEN) {
```
Sunny6889, line 158: The length check should not be placed in hexToBigInteger. This function is a low-level conversion utility and should stay free of business logic. The maximum length is an API-level constraint that belongs in the JSON-RPC input validation layer, not in a shared utility that has no knowledge of its callers' requirements.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
```diff
@@ -40,43 +57,36 @@ public abstract class RateLimiterServlet extends HttpServlet {
...
+  }
+
+  static IRateLimiter buildAdapter(String cName, String params, String name) throws Exception {
+    Class<? extends IRateLimiter> c = ALLOWED_ADAPTERS.get(cName);
+    if (c != null) {
       try {
-        IRateLimiter rateLimiter = new DefaultBaseQqsAdapter(QpsStrategy.DEFAULT_QPS_PARAM);
-        container.add(KEY_PREFIX_HTTP, name, rateLimiter);
+        return c.getConstructor(String.class).newInstance(params);
       } catch (Exception e) {
```
Sunny6889, line 81: Should this place throw error as the original logic?

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
```diff
@@ -40,43 +57,36 @@ public abstract class RateLimiterServlet extends HttpServlet {
...
-        this.throwTronError("DefaultBaseQqsAdapter", QpsStrategy.DEFAULT_QPS_PARAM, name, e);
+        logger.warn("Failed to create rate limiter '{}' for servlet '{}', "
+            + "falling back to default. Cause: {}", cName, name, e.getMessage());
       }
+    } else {
```
Sunny6889, line 85: Suggest put this in front of the above long logic:
```
if (c == null) {
   logger.warn("Unknown rate limiter adapter '{}' for servlet '{}', "
          + "falling back to default.", cName, name);
}
```

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -541,9 +543,13 @@ private static String checkGetParam(HttpServletRequest request, String key) thro
...
 
-        JSONObject jsonObject = JSON.parseObject(value);
-        if (jsonObject != null) {
-          return jsonObject.getString(key);
+        try {
+          JSONObject jsonObject = JSON.parseObject(value);
+          if (jsonObject != null) {
+            return jsonObject.getString(key);
+          }
+        } catch (JSONException e) {
```
waynercheung, line 551: [MUST] We should not swallow `JSONException` and return `null` here. With the current behavior, malformed JSON and a missing key become indistinguishable to callers, and downstream servlets such as `GetBrokerageServlet` / `GetRewardServlet` may continue with a null address and return a normal-looking `0` response. Please keep the debug log if useful, but rethrow a controlled exception (for example `InvalidParameterException` or another 4xx-style exception) so the HTTP client gets an explicit error.

framework/src/test/java/org/tron/core/services/http/UtilMockTest.java
```diff
@@ -44,14 +46,67 @@ public void testPrintTransactionFee() {
...
         .addBlock(blockCapsule1.getInstance())
         .addBlock(blockCapsule2.getInstance())
         .build();
     String out = Util.printBlockList(list, true);
```
waynercheung, line 55: [SHOULD] Please add a regression test that compares the old and new `printBlockToJSON` behavior on the same input, instead of only asserting the new shape. At minimum, it would be good to cover an empty-transactions block, a block with transactions, and `visible=true/false`, so we can prove this optimization did not change response semantics.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
waynercheung: [DISCUSS] The fallback policy here needs to be explicit. For an unknown adapter name, I would lean toward fail-fast instead of fail-open, otherwise a bad config can silently start the node with the default limiter and hide an operational/security misconfiguration. Falling back only for a known adapter with bad params/constructor failure seems safer. The current implementation matches the PR description, but I think the intended policy should be confirmed before merge.

common/src/main/java/org/tron/common/utils/ByteArray.java
waynercheung: [DISCUSS] This length guard feels more like an API / JSON-RPC input constraint than the responsibility of a shared low-level conversion utility. If we keep it in `hexToBigInteger`, please update the comment to describe it as a security/DoS guard and consider aligning `jsonHexToInt` / `jsonHexToLong` as well. Otherwise, I would prefer moving the max-length validation to the caller layer.

framework/src/main/java/org/tron/core/services/RpcApiService.java
waynercheung: [DISCUSS] I agree with the concern about swallowing `NonUniqueObjectException`, but that behavior is not newly introduced by this PR; this change only downgrades the log level. Switching to `onError(...)` would change the existing gRPC response semantics, so I would prefer to handle that in a separate fix PR. At minimum, this additional log-level change should be mentioned in the PR description.

framework/src/main/java/org/tron/core/Wallet.java
waynercheung, line 250: [NIT] This import looks stale after removing `getMerkleTreeOfBlock`. Please delete it to keep the file clean.

waynercheung:
[NIT]
- Remove the stale import in `Wallet.java`.
- Remove the `System.out.println` calls in `ByteArrayTest`.
- Update the `MAX_HEX_BIG_INTEGER_LEN` comment so it reflects the security/DoS motivation rather than "performance headroom".

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
```diff
@@ -40,43 +57,36 @@ public abstract class RateLimiterServlet extends HttpServlet {
...
+      container.add(KEY_PREFIX_HTTP, name, buildAdapter(cName, params, name));
+    } catch (Exception e) {
+      this.throwTronError(cName, params, name, e);
     }
-    if (!success) {
-      // if the specific rate limiter strategy of servlet is not defined or fail to add,
-      // then add a default Strategy.
+  }
+
+  static IRateLimiter buildAdapter(String cName, String params, String name) throws Exception {
```
0xbigapple, line 76: SHOULD: throws Exception is no longer needed — every checked exception from newInstance is caught internally, and the fallback path has no checked throws.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
```diff
@@ -40,43 +57,36 @@ public abstract class RateLimiterServlet extends HttpServlet {
...
+    Class<? extends IRateLimiter> c = ALLOWED_ADAPTERS.get(cName);
+    if (c != null) {
       try {
-        IRateLimiter rateLimiter = new DefaultBaseQqsAdapter(QpsStrategy.DEFAULT_QPS_PARAM);
-        container.add(KEY_PREFIX_HTTP, name, rateLimiter);
+        return c.getConstructor(String.class).newInstance(params);
       } catch (Exception e) {
-        this.throwTronError("DefaultBaseQqsAdapter", QpsStrategy.DEFAULT_QPS_PARAM, name, e);
+        logger.warn("Failed to create rate limiter '{}' for servlet '{}', "
```
0xbigapple, line 82: NIT: "falling back to default" doesn't say which default. Including the fallback `DEFAULT_ADAPTER_NAME` + `QpsStrategy.DEFAULT_QPS_PARAM` saves a code lookup when triaging.

framework/src/main/java/org/tron/core/services/RpcApiService.java
bladehan1: The business logic has changed; the `name` field is no longer unique. See the `Wallet/getAssetIssueByName` interface. Similar to exceptions, these are debug logs.

framework/src/main/java/org/tron/core/services/http/Util.java
bladehan1: Kept the pre-PR behavior — the JSONException catch I added was rolled back after review, so malformed JSON no longer collapses into null. Tracking a follow-up to convert malformed POST bodies into an explicit 400 at the servlet `doPost` entry rather than inside this field-lookup utility.

common/src/main/java/org/tron/common/utils/ByteArray.java
bladehan1: Agreed. `hexToBigInteger` is back to its original signature — no length/null guards. Added a new `JsonRpcApiUtil.parseBlockNumber(blockNumOrTag)` helper that enforces the 100-hex-char max-length and maps parse failures to `JsonRpcInvalidParamsException`. All five JSON-RPC call sites now go through it; tests moved to `JsonRpcApiUtilTest`.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
bladehan1: Yes — restored fail-fast. Unknown adapter now throws `TronError(RATE_LIMITER_INIT)` from `@PostConstruct`, aborting Spring startup like the pre-whitelist `Class.forName` path did. The temporary fallback-to-default was masking misconfigurations.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
bladehan1: Went with fail-fast per @waynercheung's DISCUSS. There's no fallback branch any more, so the warn log (and its placement) is moot — unknown adapter immediately raises `TronError` carrying cName/params/servlet in the message. Spring startup aborts rather than quietly booting with the default limiter.

framework/src/main/java/org/tron/core/services/http/Util.java
bladehan1: Rolled back — `checkGetParam` propagates `JSONException` again, so malformed body and missing key are no longer indistinguishable. Separately tracking a follow-up PR that returns HTTP 400 for malformed POST bodies at the servlet `doPost` entry (or the shared body-reader). That's the right layer — fixing it inside a field-lookup utility had the semantic-collapse issue you called out, and also re-parsed the body once per field.

framework/src/main/java/org/tron/core/Wallet.java
bladehan1: Removed.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
bladehan1: Fixed — `addRateContainer` no longer declares `throws Exception`. Checked exceptions from `buildAdapter` are wrapped into `TronError` (unchecked) inside the method.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
bladehan1: Moot after switching to fail-fast — the fallback warn log is gone. Unknown adapter now raises `TronError` with cName/params/servlet in the message, so triage reads the cause from the startup stack trace rather than a separate log line.

bladehan1:
> [NIT]
> 
> * Remove the stale import in `Wallet.java`.
> * Remove the `System.out.println` calls in `ByteArrayTest`.
> * Update the `MAX_HEX_BIG_INTEGER_LEN` comment so it reflects the security/DoS motivation rather than "performance headroom".

Done

framework/src/test/java/org/tron/core/services/http/UtilMockTest.java
bladehan1: Update: the test suite has been rounded out to verify old-vs-new parity more tightly.

Final `printBlockToJSON` coverage in `UtilMockTest`:
- `testPrintBlockToJSONEmptyTransactions` / `testPrintBlockToJSONWithTransactions` — structural shape; the non-empty case now loops `visible=true/false` in one test instead of two duplicates.
- `testPrintBlockToJSONVisibleFlagAffectsAddressEncoding` — pins that the `visible` flag still reaches the underlying `JsonFormat` call (`witness_address` renders as Base58 only under `visible=true`).
- `testPrintBlockToJSONTransactionsKeyMatchesLegacyImpl` — uses the previous `JsonFormat.printToString(block, true)` output as ground truth and asserts key-presence parity on both empty and non-empty blocks, so any behavioral divergence from the pre-optimization path fails here.
- `testPrintBlockToJSONCoversAllProtoTopLevelFields` — reflects over `Block.getDescriptor()` so a newly added top-level proto field that the hand-rolled assembler forgets to handle fails in CI instead of being silently dropped.

framework/src/test/java/org/tron/core/services/http/UtilMockTest.java
bladehan1: Coverage expanded to address this. In `UtilMockTest`:
- `testPrintBlockToJSONEmptyTransactions` covers the no-tx case; `testPrintBlockToJSONWithTransactions` now loops `visible=true/false` to exercise both flags in one test.
- `testPrintBlockToJSONVisibleFlagAffectsAddressEncoding` verifies the `visible` flag is still threaded through so address fields re-encode (Base58 `T...` when true, raw bytes when false).
- `testPrintBlockToJSONTransactionsKeyMatchesLegacyImpl` is the regression test you asked for: on the same block input it builds the pre-optimization output via `JSONObject.parseObject(JsonFormat.printToString(block, true))` and asserts the new impl matches on `transactions`-key presence for both empty and non-empty blocks. If protobuf JsonFormat behavior ever changes, the first half of the assertion flags it; if only the new impl drifts, the second half does.
- `testPrintBlockToJSONCoversAllProtoTopLevelFields` guards against proto drift: if someone adds a new top-level `Block` field later, the hand-rolled assembler would silently drop it — this reflects over `Block.getDescriptor()` to fail fast.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
```diff
@@ -31,56 +33,66 @@
...
 public abstract class RateLimiterServlet extends HttpServlet {
   private static final String KEY_PREFIX_HTTP = "http_";
-  private static final String ADAPTER_PREFIX = "org.tron.core.services.ratelimiter.adapter.";
+
+  static final Map<String, Class<? extends IRateLimiter>> ALLOWED_ADAPTERS;
+  static final String DEFAULT_ADAPTER_NAME = DefaultBaseQqsAdapter.class.getSimpleName();
+
+  static {
+    Map<String, Class<? extends IRateLimiter>> m = new HashMap<>();
+    for (Class<? extends IRateLimiter> c : new Class[]{
```
waynercheung: [NIT] `new Class[]{...}` is a raw type and produces an "unchecked conversion" warning from javac.
Suggest using `Arrays.asList` with an explicit generic target so no `@SuppressWarnings` is needed:
```java
      List<Class<? extends IRateLimiter>> adapters = Arrays.asList(
          GlobalPreemptibleAdapter.class,
          QpsRateLimiterAdapter.class,
          IPQPSRateLimiterAdapter.class,
          DefaultBaseQqsAdapter.class);
      for (Class<? extends IRateLimiter> c : adapters) {
        m.put(c.getSimpleName(), c);
      }
```

`Arrays.asList` is `@SafeVarargs`; with the explicit `List<Class<? extends IRateLimiter>>` target,  javac infers `T = Class<? extends IRateLimiter>` cleanly - zero warnings, no `@SuppressWarnings`.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -515,6 +516,32 @@ public static long parseEnergyFee(long timestamp, String energyPriceHistory) {
...
+   */
+  private static final int MAX_BLOCK_NUM_HEX_LEN = 100;
+
+  private static final String BLOCK_NUM_ERROR = "invalid block number";
+
+  /**
+   * Parse a JSON-RPC block number (hex "0x..." or decimal) into a BigInteger,
+   * enforcing the {@link #MAX_BLOCK_NUM_HEX_LEN} length limit.
+   */
+  public static BigInteger parseBlockNumber(String blockNumOrTag)
```
waynercheung: [MUST] `parseBlockNumber()` currently validates only syntax/length, but it still allows negative and out-of-range values through. The two consuming call sites ([Wallet.java:746] and [TronJsonRpcImpl.java:960]) then narrow with `longValue()`, which truncates silently on overflow.

Concrete examples:
  - `"0x10000000000000000"` -> `2^64` -> `longValue()` = `0L` -> block 0 (genesis) can be returned
    instead of an `invalid block number` error.
  - `"0xffffffffffffffff"` -> `longValue()` = `-1L`
  - `"-1"` (and even `"0x-1"` with the current parser) are also accepted.

I’m aware the `.longValue()` behavior predates this PR, but since this PR introduces `parseBlockNumber()` as the shared JSON-RPC block-number validator, this helper is now the right place to fix the range check once instead of duplicating it at each consumer.

Suggested fix: make `parseBlockNumber()` return a checked `long` directly:
  1. parse to `BigInteger`
  2. reject `value.signum() < 0`
  3. use `value.longValueExact()`
  4. map failures to `JsonRpcInvalidParamsException("invalid block number")`

Please also add regression tests for `0x10000000000000000`, `0xffffffffffffffff`,
  `-1`, and `0x-1`.

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
bladehan1: Applied. Replaced the raw `new Class[]{...}` with `Arrays.asList(...)` targeting `List<Class<? extends IRateLimiter>>` so the element type is inferred explicitly — no unchecked conversion, no `@SuppressWarnings`. Whitelist contents unchanged. Existing `RateLimiterServletTest.testWhitelistContents` continues to cover all four entries.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
bladehan1: Applied all four steps. `parseBlockNumber` now returns `long` directly, using `signum()` to reject negatives (`-1`, `0x-1`) and `longValueExact()` to reject overflow; both paths map to `JsonRpcInvalidParamsException("invalid block number")`.

Callers `Wallet#getByJsonBlockId` and `TronJsonRpcImpl#ethGetBlockTransactionCountByNumber` drop the obsolete `.longValue()`; the other three call sites already discarded the return value and keep working unchanged.

Regression tests added for all four cases you called out: `-1`, `0x-1`, `0xffffffffffffffff` (uint64), and `0x10000000000000000` (2^64). Also added `0x7fffffffffffffff` / `0x8000000000000000` to pin the signed-long boundary. All nine `JsonRpcApiUtilTest` tests pass.

waynercheung (approved):
LGTM

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
```diff
@@ -31,56 +35,66 @@
...
+    try {
+      return c.getConstructor(String.class).newInstance(params);
+    } catch (Exception e) {
+      throw throwTronError(cName, params, name, e);
     }
   }
 
-  private void throwTronError(String strategy, String params, String servlet,  Exception e) {
-    throw new TronError("failure to add the rate limiter strategy. servlet = " + servlet
+  private static TronError throwTronError(String strategy, String params, String servlet,
```
SeriousCoding789: [NIT] rename `throwTronError` to `buildTronError`, as above `throw throwTronError(...)` a bit confused.

Sunny6889 (approved):

framework/src/main/java/org/tron/core/services/http/RateLimiterServlet.java
bladehan1: done

bladehan1:
Resolved conflict with the develop branch.

Sunny6889 (approved):

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -518,6 +519,56 @@ public static long parseEnergyFee(long timestamp, String energyPriceHistory) {
...
+    }
+  }
+
+  /**
+   * Parse a block tag or hex block number. Tags resolve via parseBlockTag;
+   * numeric inputs go through ByteArray.jsonHexToLong (strict 0x prefix).
+   * Pathologically long inputs are rejected up front to avoid worst-case
+   * parsing cost (DoS guard mirroring single-arg parseBlockNumber).
+   */
+  public static long parseBlockNumber(String blockNumOrTag, Wallet wallet)
```
0xbigapple, line 597: [SHOULD] **`parseBlockNumber` quietly relaxes the logs path**

The javadoc still says "numeric inputs go through `ByteArray.jsonHexToLong` (strict 0x prefix)", but the body now delegates to the single-arg `parseBlockNumber(String)`, which uses `ByteArray.hexToBigInteger` — `0x...` -> hex, otherwise decimal. No 0x prefix required.

Before this PR, the logs path (`LogFilterWrapper`) was strict and the other endpoints were lenient — already inconsistent. This PR consolidates them, but picks lenient, so the logs path silently loses its 0x-prefix requirement. The PR title says "harden" but this path gets weaker.

The Ethereum spec defines `QUANTITY` as `^0x(0|[1-9a-f][0-9a-f]*)$` — bare decimal is not valid. A spec-compliant client sending `"10"` expects `0x10 = 16`; we'd return block 10 instead.

Suggestion: at a minimum, keep the logs path as strict as it was pre-PR (require `0x`); ideally, align `parseBlockNumber` with the ETH spec regex.

bladehan1: Good catch — fixed. The two-arg `parseBlockNumber(String, Wallet)` is now reverted to upstream's exact body (`isBlockTag` → `parseBlockTag`, otherwise `ByteArray.jsonHexToLong`), so the log path requires `0x` again.

Went with the "at minimum" option rather than ETH-regex strictness:
- Rejecting uppercase (`0xABC`) and leading zeros (`0x01`) would break currently valid input that `Long.parseLong(s, 16)` accepts — backward-incompatible.
- Bare-decimal silent misinterpretation (your `"10"` vs. expected `0x10 = 16` example) only affects the single-arg `parseBlockNumber(String)` used by other endpoints. Tightening that risks breaking non-spec Tron clients sending bare decimals; left for a separate discussion.

Test expectations updated to match: `"Incorrect hex syntax"` for `"abc"`; `For input string: "xabc"` for `"0xxabc"`.

0xbigapple:
LGTM

0xbigapple (approved):

PR #6694, opened 2026-04-18
feat(crypto): shielded transaction API security enhancement

### LINKED ISSUE #6616: [Feature] Shielded transaction API security enhancement
The current default configuration of java-tron node has security and code quality issues that need to be addressed.

### Security Risks

1. **Key Exposure Risk**: `allowShieldedTransactionApi = true` is enabled by default, exposing sensitive cryptographic key generation APIs:
   - `GetSpendingKey` - Random spending key (sk)
   - `GetExpandedSpendingKey` - Expanded key components (ask, nsk, ovk)
   - `GetNewShieldedAddress` - Complete shielded address with all secrets

2. **Remote Attack Surface**: Keys transmitted over network can be intercepted; misconfigured firewalls may expose keys to the internet

3. **Bug Bounty False Positives**: These endpoints frequently trigger security reports requiring manual investigation

4. **No Legitimate Remote Use Case**: Key generation should be performed locally

### Code Quality Issue

Shielded transaction-related code in `ShieldedTRC20ParametersBuilder.java` and `ZenTransactionBuilder.java` uses `+=` and `-=` arithmetic operators. For overflow-safety and explicit behavior, these should use `StrictMathWrapper.addExact`/`subtractExact` which throw `ArithmeticException` on overflow.

### Deprecated File

The `sprout-verifying.key` file is a leftover from the Sprout circuit implementation, which is no longer used (Sapling only).

### Important Note (v4.8.1+)

Starting from v4.8.1, the `allowShieldedTransactionApi` configuration **only affects API endpoint availability**. Nodes can still **validate and process shielded transactions** even when the API is disabled.

#### Configuration Changes

**Actual default value location**: The default is set in `Args.java` when the config key is absent (line ~729), not in `CommonParameter.java`.

Change the fallback default value of `allowShieldedTransactionApi` from `true` to `false`:

```java
// Args.java - applyConfigParams method
PARAMETER.allowShieldedTransactionApi = false;  // was: true (fallback when config key absent)
```

```conf
# config.conf
# WARNING: Some shielded transaction APIs require sending private keys as parameters.
# Calling these APIs on untrusted or remote nodes may leak your private keys.
# It is recommended to invoke them locally for development and testing.
# allowShieldedTransactionApi = false
```

#### Overflow-Safety Code Improvements

Replace arithmetic operators with `StrictMathWrapper` for explicit overflow detection:

| File | Changes |
|------|---------|
| `ShieldedTRC20ParametersBuilder.java` | 5 occurrences: 3 `addExact` + 2 `subtractExact` |
| `ZenTransactionBuilder.java` | 6 occurrences: 4 `addExact` + 2 `subtractExact` |

**Note**: Using `addExact`/`subtractExact` will throw `ArithmeticException` on overflow, providing explicit failure mode for value boundary violations.

#### Cleanup Deprecated File

Delete `sprout-verifying.key` - deprecated Sprout circuit key file. (Need to verify packaging/release scripts before deletion.)

### API Changes

**Affected APIs (disabled by default when `allowShieldedTransactionApi = false`):**

The `allowShieldedTransactionApi` flag gates a **broader set of shielded wallet/helper APIs** than only the secret-material endpoints. This includes all shielded-related wallet APIs accessible via HTTP/gRPC.

**HTTP API examples:**
- `/wallet/getspendingkey`
- `/wallet/getexpandedspendingkey`
- `/wallet/getnewshieldedaddress`
- `/wallet/getakfromask`
- `/wallet/getnkfromnsk`
- `/wallet/getincomingviewingkey`
- `/wallet/getpaymentaddress`
- ... (all shielded wallet helper APIs)

**gRPC API examples:**
- `GetSpendingKey`
- `GetExpandedSpendingKey`
- `GetNewShieldedAddress`
- `GetAkFromAsk`
- `GetNkFromNsk`
- `GetIncomingViewingKey`
- `GetPaymentAddress`
- ... (all shielded wallet helper APIs)

**Implementation note**: Exposed HTTP/gRPC entries are rejected at runtime rather than conditionally unregistered.

### Configuration Changes

Users can restore previous behavior by explicitly configuring:

```conf
node {
  allowShieldedTransactionApi = true
}
```

### Protocol Changes

None - Shielded transaction validation logic is not affected (API control separated since v4.8.1+)

## Scope of Impact

- [ ] Core protocol
- [x] API/RPC
- [ ] Database
- [ ] Network layer
- [ ] Smart contracts
- [x] Documentation
- [x] Other: Security configuration, Code quality (overflow-safety)

## Breaking Changes

**Yes** - Nodes relying on the default `true` value will have shielded APIs disabled after upgrade.

## Backward Compatibility

Previous behavior can be restored through explicit configuration:

```conf
node {
  allowShieldedTransactionApi = true
}
```

## Implementation

- [x] Yes, implementation is complete

**Estimated Complexity**
- [x] Low code complexity, **non-trivial compatibility/documentation impact**

## Regression Test Plan

The following test coverage will be added:

- [ ] Default disabled behavior (APIs return error when not explicitly enabled)
- [ ] Explicit opt-in via `node.allowShieldedTransactionApi = true`
- [ ] Shielded transaction validation/processing still works when API is disabled
- [ ] `ArithmeticException` is properly handled in overflow scenarios

Federico2014:
**What does this PR do?**

Enhances the security of shielded transaction APIs by adding overflow protection, tightening default configuration, and cleaning up unused artifacts.

1. Added integer overflow protection in `ZenTransactionBuilder` and `ShieldedTRC20ParametersBuilder`: replaced raw `+=`/`-=` on `valueBalance` with `StrictMathWrapper.addExact()`/`subtractExact()`, which throw `ArithmeticException` on overflow. Overflow validation is performed before mutating builder state (`spends`/`receives` lists), ensuring the builder remains consistent if an overflow is detected.
2. Wrapped `createShieldedTransaction`, `createShieldedTransactionWithoutSpendAuthSig`, `createShieldedContractParameters`, and `createShieldedContractParametersWithoutAsk` in `try/catch ArithmeticException` to convert overflow to `ZksnarkException("shielded amount overflow")`
3. Changed the default value of `allowShieldedTransactionApi` from `true` to `false` to reduce exposure by default, since some shielded transaction APIs require passing private keys as parameters
4. Updated `config.conf` with an explicit warning advising users to only invoke these APIs locally
5. Removed `sprout-verifying.key` from the repository root — the file was not referenced by any code
6. Updated test fixtures to explicitly enable `allowShieldedTransactionApi` and properly save/restore the flag

- `valueBalance` was accumulated via raw `+=`/`-=` without overflow checks, allowing crafted inputs to silently wrap around `long` range and produce invalid balances
- Shielded transaction APIs accept private keys as parameters — enabling them by default on public-facing nodes risks private key leakage
- `sprout-verifying.key` was an unreferenced artifact that added unnecessary repository bloat

⚠️ Breaking Changes:
- `allowShieldedTransactionApi` default changed from `true` to `false` — nodes that previously relied on the default will find shielded transaction APIs disabled after upgrade. Fix: explicitly set `allowShieldedTransactionApi = true` in `config.conf`.

- [x] Unit Tests — `ShieldedTransferActuatorTest`, `ArgsTest`, `ShieldedTRC20BuilderTest`, `CreateSpendAuthSigServletTest`, `MerkleContainerTest`

**Follow up**

No consensus logic or on-chain state transitions affected.

**Extra details**
Close https://github.com/tronprotocol/java-tron/issues/6616


3for:
> 3. Changed the default value of `allowShieldedTransactionApi` from `true` to `false` to reduce exposure by default, since some shielded transaction APIs require passing private keys as parameters

The default value of `allowShieldedTransactionApi` has been changed from `true` to `false`. This introduces a breaking change for nodes that relied on the previous default behavior.

After the upgrade, approximately 26 Wallet/gRPC/HTTP endpoints are disabled. However, many of these endpoints do **not** involve private key handling, for example:

* `getMerkleTreeVoucherInfo`
* `getShieldTransactionHash`
* `getTriggerInputForShieldedTRC20Contract`
* `getDiversifier`

Additionally, several **read-only / scanning APIs** are implicitly disabled, including:

* `getPaymentAddress(ivk)`
* `scanNoteByIvk(ivk, …)`
* `scanNoteByOvk(ovk, …)`
* `scanShieldedTRC20NotesByIvk(ivk, …)`
* `scanShieldedTRC20NotesByOvk(ovk, …)`

Disabling these endpoints has significant side effects:

1. Wallets that previously relied on **public RPC nodes** to synchronize shielded balances (via IVK/OVK scanning) will stop functioning on nodes where this flag is `false`.
2. If most public nodes switch to `false` by default, users will be forced to rely on a small number of nodes that keep it enabled, introducing **centralization risks** — effectively leading to a “few-node monopoly” over shielded services.

The [TRON developer documentation](https://developers.tron.network/reference/createshieldedcontractparameters) already recommends constructing shielded transaction parameters locally. The ecosystem tool [wallet-cli](https://github.com/tronprotocol/wallet-cli) also provides **local shielded key derivation** that does not depend on node APIs and is therefore unaffected by this flag.

However, when `allowShieldedTransactionApi` is disabled:

* `wallet-cli` currently **cannot** provide a fully local workflow for shielded transaction construction and signing.
* This leaves a gap where users must either run a full node or rely on centralized RPC providers.

## Suggestions

1. **Introduce granular control instead of a global switch**
   Rather than disabling all shielded APIs, consider:

   * Disabling only **high-risk endpoints** (e.g., those accepting private keys)
   * Keeping **safe, read-only endpoints** (e.g., scanning/query APIs) enabled
   * Adding a clear startup log indicating which categories of APIs are disabled

2. **Add explicit warnings for risky endpoints**
   If a node enables endpoints that accept private keys:

   * Return a visible warning in API responses
   * Clearly inform users about the associated security risks

3. **Provide a lightweight local alternative**
   Compared to requiring users to run a full node, a better approach would be:

   * Enhance ecosystem tools such as `toolkit` or `wallet-cli`
   * Provide **fully local shielded transaction construction and signing**
   * Enable users to handle all private key–related operations offline

Overall, the current change improves security posture, but the **all-or-nothing behavior introduces usability and decentralization trade-offs** that could be mitigated with finer-grained controls and better local tooling support.

mufasacubana8-pixel:
class request {
justo=this.name
}

Federico2014:
@3for Thanks for the thorough analysis. A few points in response:

**On the breaking change**: The `clearParam: false` change means nodes that have never explicitly configured this flag will see it disabled after upgrade. This is intentional — shielded transactions are a niche feature that is not actively promoted on TRON, and a conservative default better reflects the expected usage pattern for most node operators.

**On the API categorization**: Your observation that some APIs under this flag (e.g., `getMerkleTreeVoucherInfo`, `getDiversifier`, IVK/OVK scanning) don't involve spending keys is technically correct. However, even viewing keys (IVK/OVK) passed to a remote node carry privacy risk — the node learns which notes belong to the user. Given the low usage of shielded features overall, we think a single flag is an appropriate trade-off for now.

**On centralization risk**: We acknowledge the concern, but in practice shielded transaction activity on TRON is extremely limited. On-chain data shows very low usage for shielded TRC20 contracts (e.g., [contract 1](https://tronscan.org/#/contract/TQEuSEVRk1GtfExm5q9T8a1w84GvgQJ13V/transfers), [contract 2](https://tronscan.org/#/contract/TWukE3P7LDPgaHX9GcphGZfsikoqzuxbey/transfers)), suggesting the real-world impact on users is minimal. Operators who specifically support shielded use cases can explicitly set `allowShieldedTransactionApi = true`.

We're open to revisiting granular controls if shielded transaction usage grows significantly, but at this stage the added complexity isn't warranted.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -3658,77 +3658,80 @@ public ShieldedTRC20Parameters createShieldedContractParameters(
                 scaledToAmount, shieldedReceives.get(0).getNote().getValue(),
             dbManager.getDynamicPropertiesStore().disableJavaLangMath()));
       } catch (ArithmeticException e) {
-        throw new ZksnarkException("Unbalanced burn!");
+        throw new ZksnarkException("shielded amount overflow");
```
alan-eth: Really nice catch on unifying the overflow handling across all four `createShielded*` entry points — the before-mutate check in the builders makes the state-consistency story much cleaner! 🎯

Minor concern: this particular catch used to throw `ZksnarkException("Unbalanced burn!")`, which specifically signaled a burn-amount imbalance at `totalToAmount = scaledToAmount + receiveValue`. After the change, both this inner catch and the new outer catch throw the identical `"shielded amount overflow"` message, so on-call responders lose the hint about *which* arithmetic step tripped. Would it be worth keeping the inner message (e.g. `"shielded burn amount overflow"`) so the two call sites remain distinguishable in logs?

Federico2014: Good point — reverted the inner catch back to `"Unbalanced burn!"` to preserve the diagnostic signal. The outer `ArithmeticException` catches on `builder.build()` keep `"shielded amount overflow"` since those cover general arithmetic overflow across the build path.

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
```diff
@@ -150,6 +150,7 @@ public class RpcApiServicesTest {
   public static void init() throws IOException {
     Args.setParam(new String[] {"-d", temporaryFolder.newFolder().toString()},
         TestConstants.TEST_CONF);
+    getInstance().allowShieldedTransactionApi = true;
```
3for, line 152: The `allowShieldedTransactionApi` is set to `true` but not restored afterward.

Since this class uses `@BeforeClass` without a corresponding `@AfterClass`, it may leave global state mutated. Subsequent tests in the same JVM that rely on the default value could be affected. Other test classes seem to follow a save/restore pattern, but this one appears to be missing it.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -2320,58 +2320,58 @@ public TransactionCapsule createShieldedTransaction(PrivateParameters request)
...
-    shieldedOutput(shieldedReceives, builder, ovk);
+      // output
+      shieldedOutput(shieldedReceives, builder, ovk);
 
-    TransactionCapsule transactionCapsule = null;
-    try {
-      transactionCapsule = builder.build();
+      return builder.build();
+    } catch (ArithmeticException e) {
+      throw new ZksnarkException("shielded amount overflow");
```
3for: In `Wallet.java` (lines 2370, 2474, 3732, 3858), multiple places catch `ArithmeticException` and rethrow `new ZksnarkException("shielded amount overflow")`, which drops the original cause. This makes debugging harder since the stack trace no longer points to where the overflow occurred.

It might be better to preserve the cause, for example:

```java
new ZksnarkException("shielded amount overflow", e);
```

framework/src/test/java/org/tron/core/services/RpcApiServicesTest.java
Federico2014: Thanks for the review. `RpcApiServicesTest` does have an `@AfterClass` (`destroy()`), which calls `Args.clearParam()` → `CommonParameter.reset()`. `reset()` replaces the entire singleton with `PARAMETER = new CommonParameter()`, which resets `allowShieldedTransactionApi` to the new default (`false`) unconditionally — so the state is not leaked to subsequent tests.

The explicit save/restore pattern in `CreateSpendAuthSigServletTest` / `MerkleContainerTest` is needed there specifically because those classes don't call `clearParam()` in teardown.

That said, the field is already annotated with `// clearParam: true`, which documents that `clearParam()` covers it. Happy to add an inline comment to the `@BeforeClass` pointing to `destroy()` if that makes the intent clearer.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -2320,58 +2320,58 @@ public TransactionCapsule createShieldedTransaction(PrivateParameters request)
 
     checkCmValid(shieldedSpends, shieldedReceives);
 
-    // add
-    if (!ArrayUtils.isEmpty(transparentFromAddress)) {
-      builder.setTransparentInput(transparentFromAddress, fromAmount);
-    }
+    try {
```
3for, line 2325: Functions such as `createShieldedTransaction` and similar methods expand the scope of the `try` block. Previously, `throw new ZksnarkException(PAYMENT_ADDRESS_FORMAT_WRONG);` would be thrown directly, but now it is caught by `catch (ZksnarkException e)` and rewrapped as `throw new ZksnarkException(e.toString())`.

As a result, the error message changes from:

```
"Payment address format is wrong"
```

to:

```
"org.tron.core.exception.ZksnarkException: Payment address format is wrong"
```

This may impact clients that rely on exact error message matching. If possible, it would be better to keep the error message consistent with the previous behavior.

Federico2014: Good catch. After checking the codebase, no tests or known clients match on the exact message string — assertions only check that a `ZksnarkException` is thrown, not its content. Given that this is an input-validation path (malformed payment address) on a niche, now-disabled-by-default API, the practical impact is negligible.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Good point. Fixed: added a `ZksnarkException(String, Throwable)` constructor and updated all four overflow catch sites to pass the original cause:

```java
throw new ZksnarkException("shielded amount overflow", e);
```

3for (approved):
LGTM. However, I think we should be cautious about disabling all shielded transaction API by default until viable alternative tools are mature and available.

framework/src/main/java/org/tron/core/Wallet.java
waynercheung, line 3659: [DISCUSS] This burn-path sum still uses deprecated `Maths.addExact(..., disableJavaLangMath())`, while the rest of this PR standardizes overflow-sensitive arithmetic onto `StrictMathWrapper`. If preserving the runtime switch here is intentional, please add a short comment explaining why this path cannot be aligned; otherwise it would be cleaner to standardize these two remaining adds as well.

waynercheung, line 3795: The same problem with `Wallet.java:3659`

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -2287,58 +2322,58 @@ public TransactionCapsule createShieldedTransaction(PrivateParameters request)
...
 
-    TransactionCapsule transactionCapsule = null;
-    try {
-      transactionCapsule = builder.build();
+      return builder.build();
+    } catch (ArithmeticException e) {
+      throw new ZksnarkException("shielded amount overflow", e);
     } catch (ZksnarkException e) {
       logger.error("createShieldedTransaction except, error is " + e.toString());
       throw new ZksnarkException(e.toString());
```
waynercheung: [SHOULD] This catch still rewraps a `ZksnarkException` as `new ZksnarkException(e.toString())`, which drops the original cause and rewrites the message into `"org.tron.core.exception.ZksnarkException: ..."`. The same pattern exists in `createShieldedTransactionWithoutSpendAuthSig`. Since this PR already introduced `ZksnarkException(String, Throwable)`, it would be better to preserve the original exception here (`throw e;`) so the RPC/HTTP layer keeps the real root cause.

framework/src/main/java/org/tron/core/Wallet.java
waynercheung: [SHOULD] This catch still drops the underlying `ArithmeticException`. The same pattern exists in `createShieldedContractParametersWithoutAsk`. Since this PR already added `ZksnarkException(String, Throwable)` and uses it for the outer `"shielded amount overflow"` catches, please preserve the cause here as well with `new ZksnarkException("Unbalanced burn!", e)` for consistent diagnostics.

Federico2014: Good catch on the inconsistency. Keeping these two callsites on the deprecated `Maths.addExact(..., disableJavaLangMath())` is intentional for this PR's scope — the goal here is the builder-internal `valueBalance` arithmetic, and these two `Wallet`  callsites are not on the consensus path. The runtime fork flag is also a no-op here in practice: for `long/int addExact`, `java.lang.Math` and `StrictMath` are spec-equivalent,  so the `disableJavaLangMath()` switch never produced divergent behavior on this path.                                                                                                                                         I'd rather not expand this PR's diff into a broader `Maths → StrictMathWrapper` migration.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Good catch — the rewrap drops the original cause and the `e.toString()` message produces a doubly-prefixed `org.tron.core.exception.ZksnarkException: ...` string downstream. This is also inconsistent with the chained-cause pattern this PR just introduced for `ArithmeticException` (and the new `ZksnarkException(String, Throwable)` constructor added for it), so it's in scope.

Fixed in both `createShieldedTransaction` and `createShieldedTransactionWithoutSpendAuthSig` by rethrowing the original exception (`throw e;`) instead of rewrapping.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Agreed — and consistent with the chained-cause pattern this PR already uses for the outer `"shielded amount overflow"` catches. Updated both call sites in `createShieldedContractParameters` and `createShieldedContractParametersWithoutAsk` to `throw new ZksnarkException("Unbalanced burn!", e);` so the underlying `ArithmeticException` reaches diagnostics.

common/src/test/java/org/tron/core/config/args/NodeConfigTest.java
```diff
@@ -286,7 +286,7 @@ public void testLegacyAliasTakesPriorityOverModernKey() {
   @Test
   public void testShieldedApiDefaultsToTrueWhenNeitherKeySet() {
```
3for: The function name contradicts the new default behavior and should be renamed to `testShieldedApiDefaultsToFalseWhenNeitherKeySet`.

common/src/test/java/org/tron/core/config/args/NodeConfigTest.java
3for: The comment here should be updated to `allowShieldedTransactionApi = false` to facilitate future maintenance.

waynercheung (approved):
LGTM

PR #6696, opened 2026-04-20
fix(trie): make TrieImpl.insert() idempotent for duplicate key-value

### LINKED ISSUE #6608: bug: TrieImpl.insert() is not idempotent for duplicate key-value pairs
## Summary

`TrieImpl.insert()` violates the Merkle Trie invariant: **the same dataset should always produce the same root hash, regardless of insertion order.** When a duplicate key-value pair is inserted (same key, same value), the node is unconditionally marked as `dirty`, which invalidates the internal hash cache and causes the final root hash to become insertion-order-dependent.

## Root Cause

In `TrieImpl.java`, the `kvNodeSetValueOrNode()` method (L873-878) unconditionally sets `dirty = true` without checking whether the new value equals the existing value:

```java
public Node kvNodeSetValueOrNode(Object valueOrNode) {
    parse();
    assert getType() != NodeType.BranchNode;
    children[1] = valueOrNode;    // overwrites even if same value
    dirty = true;                  // always marks dirty
    return this;
}
```

This is triggered from `insert()` (L188-189) when `commonPrefix.equals(k)`:

```java
} else if (commonPrefix.equals(k)) {
    return n.kvNodeSetValueOrNode(nodeOrValue);  // no same-value check
}
```

## Reproduction

The existing `TrieTest.testOrder()` (currently `@Ignore`) demonstrates the issue:

1. Insert keys 1-99, then re-insert key 10 with the same value
2. Shuffle the insertion order and rebuild
3. Root hashes differ ~3% of the time depending on shuffle order

## Impact

**Current: None.** The `allowAccountStateRoot` chain parameter (proposal 25) has not been enabled on mainnet. `AccountStateCallBack` checks `allowAccountStateRoot()` before using `TrieImpl`, so this code path is never executed.

**Future: Consensus-level risk.** If proposal 25 is approved, `AccountStateCallBack` will use `TrieImpl` to compute per-block account state root hashes stored in block headers. Different nodes could compute different root hashes for the same block if duplicate `trie.put(address, state)` calls occur, leading to `BadBlockException` and potential chain forks.

## Suggested Fix

Add a same-value check in `kvNodeSetValueOrNode()` to short-circuit when the value hasn't changed:

```java
public Node kvNodeSetValueOrNode(Object valueOrNode) {
    parse();
    assert getType() != NodeType.BranchNode;
    if (valueOrNode instanceof byte[] && children[1] instanceof byte[]
        && Arrays.equals((byte[]) valueOrNode, (byte[]) children[1])) {
        return this;  // no change, skip dirty marking
    }
    children[1] = valueOrNode;
    dirty = true;
    return this;
}
```

Then remove the `@Ignore` on `TrieTest.testOrder()` to verify the fix.

## Related

- PR #6586 — `@Ignore` annotation and root cause analysis in `TrieTest.java`
- `AccountStateCallBack.java` — production consumer of `TrieImpl`

halibobo1205:
## Summary

Fix a **correctness bug** in `TrieImpl.insert()` that breaks Merkle Trie idempotency when the **same key** is put more than once (regardless of whether the value changes), and add a related **optimization** that avoids unnecessary recomputation on the same code path:

1. **Correctness (root-hash corruption).** Reorder the condition checks in `insert()` so that `commonPrefix.equals(k)` is evaluated before `commonPrefix.isEmpty()`. The two conditions overlap only when both `k` and `currentNodeKey` are empty, which the recursion reaches exclusively on a **duplicate put of an already-inserted key** on a fully-split KV leaf. In that state, the old order incorrectly triggered the "no common prefix" branch, replacing `KVNode("", v_old)` with `BranchNode{terminal: v_new}`, thereby corrupting the trie structure and, in turn, the root hash. This happens whether or not the new value equals the old one.

2. **Optimization (avoid unnecessary recomputation).** Short-circuit `kvNodeSetValueOrNode()` when the new `byte[]` value equals the existing one (by reference or byte content) to skip redundant dirty marking and downstream hash recomputation on the KV-leaf path. This is **purely a performance optimization — it is not part of the correctness fix**, and it does not by itself address the root-hash corruption; it only reduces cache churn. Branch-terminal writes and `Node`-valued writes are intentionally not short-circuited (parent nodes still need dirty propagation when a child is mutated in place).

Closes #6608.

## What this fix really does

Restore the **Merkle Trie invariant**: the root hash must be a pure function of the `{key → value}` set, independent of insertion order. Before this fix, `TrieImpl` violated that definition — it only produced consistent roots in practice because callers fed it data in a deterministic order. After this fix, `TrieImpl` is a true Merkle Trie by definition.

## When the latent defect is exercised

The defect lies in `TrieImpl.insert()` itself, not in any caller. Whether it has any *observable* effect depends on two independent conditions:

**1. Is the buggy code path reachable?**

- ❌ Every key put at most once per trie → path not reachable, no effect.
- ✅ Same key put ≥ 2 times (same or different value) → **necessary but not sufficient**. The buggy branch is reached only when the recursion lands on a node whose stored key and the incoming key both reduce to empty — i.e. the trie has evolved into a fully-split KV leaf at that position. This structural condition is rare in random workloads but not negligible, and is exactly what the regression sequences in `testOrder` reproduce.

  In practice, the only known production write path exercising `TrieImpl.put()` for account-state-root generation is `AccountStateCallBack`. Its `trieEntryList` is an undeduplicated `ArrayList`: if multiple transactions in a block touch the same account, `trie.put(addr, …)` is invoked repeatedly with the same key, so the precondition for triggering the path is satisfied during normal operation.

**2. Does it produce a divergent root?**

- ❌ All callers feed the trie in the **same** order → all of them compute the same (non-canonical) root. No divergence, no observable problem.
- ✅ Two callers feed the trie in **different** orders for the same `{key → value}` set → they compute different roots.

  On-chain block processing is deterministic: every node executes the same transactions in the same order, so condition 2 is **never met in production**. It is only met in synthetic scenarios — e.g. the unit test, which deliberately `shuffles` the input to expose the defect.

**Net effect on a running chain:** the bug is exercised, but its symptom is masked by deterministic execution order. All nodes converge on the same root. This is why it has gone unnoticed and why it is not a consensus issue — but it also means `TrieImpl` is not a true Merkle Trie: its root depends on insertion order rather than on the `{key → value}` set alone.

## Consensus impact

**None on mainnet and Nile.** `AccountStateCallBack` is gated behind proposal 25 (`ALLOW_ACCOUNT_STATE_ROOT`), which is **not** activated on either network — the computed state root does not enter block headers and does not participate in consensus.

Even if proposal 25 were activated, **all honest nodes would still agree on the root with or without this fix**, because transaction execution order within a block is deterministic and identical across nodes. The `shuffle` used in the unit test is a synthetic construct to surface the latent defect; no such reordering occurs in production. In other words, on-chain `TrieImpl` happens to produce consistent roots today *because the caller feeds it a deterministic sequence*, not because the implementation satisfies the Merkle invariant.

So the fix is about **correctness of the data structure itself**, not about preventing a fork.

## ⚠️ Compatibility note for downstream fork chains

Any fork chain that has already activated proposal 25 in production has committed `accountStateRoot` values to block headers that were computed with the **buggy** `TrieImpl`. For those chains, this fix is **not** a transparent in-place upgrade:

- **Historical-block replay** would recompute different roots than those stored in existing headers → validation failures.
- **New-block consensus** would split between pre- and post-upgrade nodes.

Such forks must coordinate the upgrade via a hard-fork activation height (or gate the new behavior behind a dedicated proposal). Mainnet and Nile are **not** affected since proposal 25 has never been activated there; they can take this fix as an ordinary release.

## Release scope

- **Mainnet / Nile:** safe to include in any upcoming release. No consensus implications.
- **Recommended (defense-in-depth):** include this fix before any future release that could enable proposal 25, so that the data structure is correct by definition before the state root is ever committed to headers.

See #6608 for the full reproduction and affected code paths.

### DIFF
--- a/framework/src/main/java/org/tron/core/trie/TrieImpl.java
+++ b/framework/src/main/java/org/tron/core/trie/TrieImpl.java
@@ -179,6 +179,10 @@ public class TrieImpl implements Trie<byte[]> {
     } else {
       TrieKey currentNodeKey = n.kvNodeGetKey();
       TrieKey commonPrefix = k.getCommonPrefix(currentNodeKey);
+      // NOTE: equals(k) MUST precede isEmpty(). They overlap only when both k and
+      // currentNodeKey are empty (duplicate put on a fully-split KV leaf); in that
+      // case the correct behavior is an in-place value update, not conversion to
+      // a BranchNode. Swapping the order corrupts the trie structure. See #6608.
       if (commonPrefix.equals(k)) {
         return n.kvNodeSetValueOrNode(nodeOrValue);
       } else if (commonPrefix.isEmpty()) {
--- a/framework/src/main/java/org/tron/core/trie/TrieImpl.java
+++ b/framework/src/main/java/org/tron/core/trie/TrieImpl.java
@@ -179,14 +179,14 @@ public class TrieImpl implements Trie<byte[]> {
     } else {
       TrieKey currentNodeKey = n.kvNodeGetKey();
       TrieKey commonPrefix = k.getCommonPrefix(currentNodeKey);
-      if (commonPrefix.isEmpty()) {
+      if (commonPrefix.equals(k)) {
+        return n.kvNodeSetValueOrNode(nodeOrValue);
+      } else if (commonPrefix.isEmpty()) {
         Node newBranchNode = new Node();
         insert(newBranchNode, currentNodeKey, n.kvNodeGetValueOrNode());
         insert(newBranchNode, k, nodeOrValue);
         n.dispose();
         return newBranchNode;
-      } else if (commonPrefix.equals(k)) {
-        return n.kvNodeSetValueOrNode(nodeOrValue);
       } else if (commonPrefix.equals(currentNodeKey)) {
         insert(n.kvNodeGetChildNode(), k.shift(commonPrefix.getLength()), nodeOrValue);
         return n.invalidate();
@@ -873,6 +873,11 @@ public class TrieImpl implements Trie<byte[]> {
     public Node kvNodeSetValueOrNode(Object valueOrNode) {
       parse();
       assert getType() != NodeType.BranchNode;
+      if (valueOrNode instanceof byte[] && children[1] instanceof byte[]
+          && (children[1] == valueOrNode
+          || Arrays.equals((byte[]) children[1], (byte[]) valueOrNode))) {
+        return this;
+      }
       children[1] = valueOrNode;
       dirty = true;
       return this;
--- a/framework/src/test/java/org/tron/core/tire/TrieTest.java
+++ b/framework/src/test/java/org/tron/core/tire/TrieTest.java
@@ -41,6 +41,7 @@ public class TrieTest {
   private static String doge = "doge";
   private static String test = "test";
   private static String dude = "dude";
+  private static final long SHUFFLE_SEED = 0xC0FFEEL;
 
   @Test
   public void test() {
@@ -64,8 +65,6 @@ public class TrieTest {
     boolean result = trie
         .verifyProof(trieCopy.getRootHash(), new byte[]{1, 1}, (LinkedHashMap<byte[], Node>) map);
     Assert.assertTrue(result);
-    System.out.println(trieCopy.prove(RLP.encodeInt(5)));
-    System.out.println(trieCopy.prove(RLP.encodeInt(6)));
     assertTrue(RLP.encodeInt(5), trieCopy);
     assertTrue(RLP.encodeInt(5), RLP.encodeInt(6), trieCopy);
     assertTrue(RLP.encodeInt(6), trieCopy);
@@ -122,16 +121,16 @@ public class TrieTest {
 
   /*
    * Verifies that TrieImpl root hash is insertion-order-independent even when
-   * a key is inserted twice with the same value (idempotent put).
+   * the same key is put more than once (idempotent put).
    *
-   * Covers both known-failing sequences (regression) and a random shuffle.
-   * Previously flaky due to two bugs in TrieImpl.insert():
-   * 1. commonPrefix.isEmpty() was checked before commonPrefix.equals(k), causing
-   *    KVNode("",v) to be incorrectly replaced with BranchNode{terminal:v} on
-   *    duplicate put of a fully-split key — corrupting the root hash.
-   * 2. kvNodeSetValueOrNode() unconditionally marked the node dirty even when
-   *    the value was unchanged, causing unnecessary hash recomputation.
-   * Both are now fixed.
+   * Covers both known-failing sequences (regression) and a seeded random
+   * shuffle. Previously flaky due to a correctness bug in TrieImpl.insert():
+   * commonPrefix.isEmpty() was checked before commonPrefix.equals(k), causing
+   * KVNode("", v_old) to be incorrectly replaced with BranchNode{terminal:v_new}
+   * on a duplicate put of a fully-split key — this is the actual root-hash
+   * corruption. A separate, non-correctness optimization in
+   * kvNodeSetValueOrNode() additionally short-circuits same-value writes to
+   * avoid unnecessary dirty marking / hash recomputation.
    */
   @Test
   public void testOrder() {
@@ -145,7 +144,12 @@ public class TrieTest {
     trie.put(RLP.encodeInt(10), String.valueOf(10).getBytes());
     value.add(10);
     byte[] rootHash1 = trie.getRootHash();
-    Collections.shuffle(value, new Random(0xC0FFEEL));
+    TrieImpl baseline = new TrieImpl();
+    for (int i = 1; i < n; i++) {
+      baseline.put(RLP.encodeInt(i), String.valueOf(i).getBytes());
+    }
+    Assert.assertArrayEquals(baseline.getRootHash(), rootHash1);
+    Collections.shuffle(value, new Random(SHUFFLE_SEED));
     assertTrieRootHash(rootHash1, value);
     String[] sequences = {
         "95,10,66,10,67,2,98,31,85,89,81,96,19,68,44,49,43,40,62,87,4,38,17,18,8,"
@@ -206,7 +210,7 @@ public class TrieTest {
     return result;
   }
 
-  private void assertTrieRootHash(byte[] rootHash1, List<Integer> value) {
+  private static void assertTrieRootHash(byte[] rootHash1, List<Integer> value) {
     TrieImpl trie2 = new TrieImpl();
     for (int i : value) {
       trie2.put(RLP.encodeInt(i), String.valueOf(i).getBytes());
@@ -249,7 +253,7 @@ public class TrieTest {
       trie.put(RLP.encodeInt(i), String.valueOf(i).getBytes());
     }
     byte[] rootHash1 = trie.getRootHash();
-    Collections.shuffle(value, new java.util.Random(42));
+    Collections.shuffle(value, new Random(42));
     TrieImpl trie2 = new TrieImpl();
     for (int i : value) {
       trie2.put(RLP.encodeInt(i), String.valueOf(i).getBytes());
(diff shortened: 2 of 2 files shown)


framework/src/main/java/org/tron/core/trie/TrieImpl.java
```diff
@@ -179,14 +179,14 @@ private Node insert(Node n, TrieKey k, Object nodeOrValue) {
     } else {
       TrieKey currentNodeKey = n.kvNodeGetKey();
       TrieKey commonPrefix = k.getCommonPrefix(currentNodeKey);
-      if (commonPrefix.isEmpty()) {
+      if (commonPrefix.equals(k)) {
```
waynercheung, line 182: [SHOULD] Please add a short note here explaining that the order is semantically significant, otherwise this is easy to "clean up" back into the broken form later.

`commonPrefix.equals(k)` must be checked before `commonPrefix.isEmpty()`. The two conditions overlap only when both `k` and `currentNodeKey` are empty and terminal (duplicate put on a fully split KV leaf), and in that case the correct behavior is an in-place value update, not conversion to a `BranchNode`.

Suggested wording:

  // NOTE: equals(k) MUST precede isEmpty(). They overlap only when both k and
  // currentNodeKey are empty+terminal (duplicate put on a fully split KV leaf);
  // the equals(k) branch is the correct in-place update. See #6608.

framework/src/main/java/org/tron/core/trie/TrieImpl.java
```diff
@@ -873,6 +873,11 @@ public Object kvNodeGetValueOrNode() {
     public Node kvNodeSetValueOrNode(Object valueOrNode) {
       parse();
       assert getType() != NodeType.BranchNode;
+      if (valueOrNode instanceof byte[] && children[1] instanceof byte[]
```
waynercheung, line 876: [SHOULD] The short-circuit added here currently covers only `KVNodeValue(byte[])`. If the PR wants to claim "avoid unnecessary dirty marking and downstream hash recomputation", that statement is still broader than the implementation.

In particular, duplicate writes to a branch terminal slot still go through `branchNodeSetValue()` and mark the node dirty unconditionally. I would suggest either:
  - adding the symmetric `byte[]` equality fast path there as well, or
  - narrowing the PR description so it is clear this optimization currently applies only to KV-leaf `byte[]` updates.

I would avoid a generic same-reference short-circuit for `Node` values here, because parent nodes still need dirty propagation when a child node is mutated in place.

framework/src/test/java/org/tron/core/tire/TrieTest.java
```diff
@@ -121,29 +121,18 @@ public void test2() {
...
-   *    Merkle Trie invariant.
-   *
-   * Production impact: low. AccountStateCallBack uses TrieImpl to build per-block
-   * account state tries. Duplicate put(key, sameValue) is unlikely in production
-   * because account state changes between transactions (balance, nonce, etc.).
-   *
-   * Proper fix: TrieImpl.insert() should short-circuit when the existing value
-   * equals the new value, avoiding unnecessary invalidate(). See TrieImpl:188-192.
+   * Covers both known-failing sequences (regression) and a random shuffle.
+   * Previously flaky due to two bugs in TrieImpl.insert():
```
waynercheung, line 128: [SHOULD] This comment slightly overstates the role of the `kvNodeSetValueOrNode()` short-circuit. As written, "Previously flaky due to two bugs ... Both are now fixed" reads as if the dirty short-circuit was also a direct root-hash corruption fix.

My understanding is:
- the `commonPrefix.equals(k)` / `commonPrefix.isEmpty()` ordering bug is the actual trie-structure / root-hash correctness issue;
- the `kvNodeSetValueOrNode()` short-circuit avoids unnecessary recomputation / cache churn, but is not by itself the root-hash corruption mechanism.

I would reword this comment to make that distinction explicit.

framework/src/test/java/org/tron/core/tire/TrieTest.java
```diff
@@ -156,7 +145,68 @@ public void testOrder() {
...
+  private static List<Integer> parseSeq(String csv) {
+    String[] parts = csv.split(",");
+    List<Integer> result = new ArrayList<>(parts.length);
+    for (String p : parts) {
+      result.add(Integer.parseInt(p));
+    }
+    return result;
+  }
+
+  private void assertTrieRootHash(byte[] rootHash1, List<Integer> value) {
```
waynercheung, line 209: [NIT] A few small test-style items to clean up together here:

- `assertTrieRootHash` does not use instance state and can be `private static`, consistent with `parseSeq`;
- at line 252, `new java.util.Random(42)` can be simplified to `new Random(42)` since `Random` is already imported;
- at line 148, consider extracting the shuffle seed into a named constant (for example `private static final long SHUFFLE_SEED = 0xC0FFEEL`) to make failures easier to reproduce.

framework/src/test/java/org/tron/core/tire/TrieTest.java
```diff
@@ -156,7 +145,68 @@ public void testOrder() {
     trie.put(RLP.encodeInt(10), String.valueOf(10).getBytes());
     value.add(10);
     byte[] rootHash1 = trie.getRootHash();
```
waynercheung, line 147: [QUESTION] Would it make sense to add one stronger invariant here: the baseline built with the duplicate `put(10, sameValue)` should also have the same root hash as a baseline built without the duplicate insertion at all?

For example:
```java
  TrieImpl base = new TrieImpl();
  for (int i = 1; i < 100; i++) {
    base.put(RLP.encodeInt(i), String.valueOf(i).getBytes());
  }
  Assert.assertArrayEquals(base.getRootHash(), rootHash1);
```
Not required, but it would lock in the "duplicate put has zero effect on the root" invariant more directly.

waynercheung:
[MUST] The consensus-impact analysis is already documented in #6608, but per the repo’s consensus-change requirements, please mirror a short summary into the PR description itself and explicitly document:

1. the target release / upgrade scope for this fix;
2. that this fix must be included in any release intended to run before / when Proposal 25 (`ALLOW_ACCOUNT_STATE_ROOT`) may activate.

A short summary referencing #6608 is sufficient; no need to restate the full analysis.

Optional cleanup: since this PR already touches `TrieTest.java`, it would also be nice to remove the two pre-existing `System.out.println` lines at `TrieTest.java:67-68` in this PR if convenient. If you prefer to keep this PR tightly scoped, a small follow-up cleanup PR is also fine.

halibobo1205:
@waynercheung 

1. The PR description is updated.
2.  TrieTest.java:67-68 is cleaned up.

lvs0075:
I have a question regarding the description: the current code does not affect the consensus. The modification is merely intended to ensure that the results of the random write are consistent.

halibobo1205:
@lvs0075 

 You're right, and thanks for pointing this out — I'll reword the description.

 To be precise: this fix does **not** affect consensus on mainnet or Nile. On-chain write order is deterministic, so all honest nodes compute the same root regardless of this bug. The `shuffle` in the unit test is a synthetic scenario to expose the defect; it does not correspond to anything that happens in production execution.

 What this PR actually does is restore the Merkle trie invariant itself: the root must be a pure function of the {key → value} set, independent of insertion order. Without this fix, `TrieImpl` is not a true Merkle trie by definition — it just happens to produce consistent roots in practice because the caller feeds it a deterministic order.

 One caveat worth flagging in the description (not a consensus issue on mainnet/Nile, but a compatibility note for downstream): any fork chain that has already activated proposal 25 in production has committed `accountStateRoot` values computed with the buggy TrieImpl. Those forks cannot hot-upgrade to this fix — they'd need a coordinated activation height, otherwise historical-block validation would fail, and new blocks would split between pre- and post-upgrade nodes. Mainnet / Nile are unaffected since proposal 25 is not activated.

 I'll update the PR description to remove the misleading "must be merged before proposal 25 or nodes will fork" framing and replace it with the invariant-restoration framing plus the downstream warning.

waynercheung (approved):
LGTM

317787106:
A solid fix. If Proposal 25 is planned to be activated, the TrieImpl code should undergo further review and testing.

317787106 (approved):

lvs0075 (approved):
good job

PR #6698, opened 2026-04-23
fix(consensus,framework,actuator): use Locale.ROOT for case-insensitive

halibobo1205:
## Summary

Fix a **locale-dependent code hygiene issue**: no-arg `String.toLowerCase()` / `toUpperCase()` calls use the JVM default locale, and under a minority of locales (tr/az) the folding of `'I'` diverges from every other locale. Switching uniformly to `Locale.ROOT` decouples the behavior from the deployment environment, backed by a compile-time guard and a one-time data normalization.

This PR does three things:

1. **Correctness.** Replace every no-arg `toLowerCase()` / `toUpperCase()` call with its `Locale.ROOT` counterpart so that string folding is independent of the JVM's startup locale.
2. **Defense-in-depth.** Enable ErrorProne's `StringCaseLocaleUsage` checker at `ERROR` level, and add a custom `StringCaseLocaleUsageMethodRef` checker to cover the `String::toLowerCase` method-reference form that upstream misses. Any future no-arg call fails compilation.
3. **Data hygiene.** `MigrateTurkishKeyHelper` runs once at `Manager.init()` to normalize any non-ROOT legacy keys that may exist in `AccountIdIndexStore`.

## What this fix really does

Make the case-insensitive index behavior a **pure function of the input bytes**, no longer implicitly dependent on the JVM's startup locale. Before this fix, `AccountIdIndexStore` was, in principle, "case-insensitive relative to whichever locale this JVM happens to use" rather than truly locale-independent. After the fix it is the latter.

This is a **data-structure-level correctness convergence** — remove an implicit environment dependency and align the implementation with the intended semantics.

## Scope

**Only one site reaches persistent storage**: `AccountIdIndexStore.getLowerCaseAccountId` — the lowercased result becomes a DB key. All other touched call sites (DB engine selection, disabled-API list, log topic classification, hex display, OS detection, etc.) are in-process runtime comparisons — they neither write to disk nor cross nodes.

**The input domain itself is tightly constrained**: input to `setAccountId` is gated by `validReadableBytes` to printable ASCII (0x21–0x7E). Within this domain, the locale divergence in lowercase only manifests on a single character; every other character (letters, digits, symbols) folds identically under any locale.

 ### Auditability of the migration scope

 The "only one site reaches persistent storage" claim above can be independently verified by inspection. Every other touched call site falls into one of:

 - **(a) Runtime engine selection** — `dbEngine.toUpperCase()` in `TronDatabase`, `TronStoreWithRevoking`, `TxCacheDB`. Locale only affects which DB engine the running process selects; it is never written to disk.
 - **(b) In-memory config/membership checks** — `disabledApiList`, `HttpMethod` string comparisons in `Util` / access filters, `Account.accountType` switch. Comparison results are consumed in-process and never persisted.
 - **(c) Display/formatting** — `Hex.toHexString().toUpperCase()` in `DataWord`, help-text formatting in `Args`. Output to logs/console, never read back.
 - **(d) Host-environment probing** — `os.name` / `os.arch` / `java.vendor` reads in `Arch`, `WalletUtils.getDefaultKeyDirectory`, `KeystoreFactory`. Local environment branching never crosses nodes.

 None of these reach disk or cross-node boundaries, so the migration scope (`AccountIdIndexStore` only) is exact, not approximate.

## Why `MigrateTurkishKeyHelper` is bundled in

A code-only fix does not fully cover "nodes that were ever started under a non-default locale" — such nodes may carry legacy-format keys in their DB. The cleanest way to keep the code fix and the data state in sync is a one-shot normalization at startup.

**Design notes:**

- Follows the `MoveAbiHelper` pattern, gated by a `TURKISH_KEY_MIGRATION_DONE` flag in `DynamicPropertiesStore` so it runs at most once.
- Full-table scan (`AccountIdIndexStore` is a sparse index with very low cardinality — only 14 entries on mainnet today; scan cost is negligible).
- For each non-ROOT-form key: write a new entry under its ROOT-form key, then delete the legacy one.
- Conflict policy: if the ROOT-form key already exists, keep the ROOT entry and drop the legacy value — simple, auditable, and equivalent to the state of a node that was never affected.

For nodes that were never affected, the scan yields zero candidates and the helper is a no-op.

## Why the simplified normalization strategy

In principle one could enumerate every possible legacy variant of each ROOT-canonical key and merge them, but that enters a combinatorial space (per-key variant count can reach 2^k, where k is the occurrence count of the divergent character). Even with full enumeration, **picking the correct value across conflicts requires knowing the historical insertion order** — information the DB layer does not retain.

The conservative strategy chosen here:
- Simple and auditable
- Post-migration local semantics equal those of a node that only ever ran under ROOT
- In the extremely rare conflict case, one legacy value may be dropped — an acceptable cost under case-insensitive index semantics

## Impact

- **Mainnet / Nile.** Safe to include in any normal release. The code changes produce no byte-level behavioral difference in the expected locale environment; the migration scans a trivially small table and is a no-op for nodes that were never affected.
- **Node operators.** You'll see a one-time migration line in the startup log; it won't run again on subsequent restarts.
- **Long term.** The ErrorProne rule ensures no future patch can reintroduce no-arg lowercase/uppercase.

## Note for downstream fork chains

A downstream fork chain needs to treat this PR as a **non-transparent upgrade** only if **both** of the following hold:

1. All of its consensus nodes have historically run under the same minority locale that diverges in lowercase (e.g. all on tr/az).
2. Its history includes a committed `SetAccountId` transaction whose accountId contains a character that triggers the locale divergence (e.g. `I`).

On such a chain, the `AccountIdIndexStore` historical state was produced under non-ROOT folding; after this PR, `has()` lookups for the same accountId may yield different verdicts, which requires a coordinated hard-fork activation height and operator-defined historical data migration strategy.

Mainnet and Nile are **not** in this category: condition (1) has never held historically.

## Release scope

- Ship as a normal release — no extra coordination required.
- The ErrorProne rule takes effect with this PR; any no-arg lowercase/uppercase in future patches will be blocked at compile time.


framework/src/test/java/org/tron/core/db/AccountIdIndexStoreTest.java
```diff
@@ -128,4 +134,59 @@ public void testCaseInsensitive() {
...
+    };
+
+    byte[][] addresses = new byte[accountIds.length][];
+    byte[][] turkishKeys = new byte[accountIds.length][];
+
+    for (int i = 0; i < accountIds.length; i++) {
+      addresses[i] = randomBytes(21);
+      String turkishLower = accountIds[i].toLowerCase(TURKISH);
+      turkishKeys[i] = turkishLower.getBytes(StandardCharsets.UTF_8);
+      accountIdIndexStore.put(turkishKeys[i], new BytesCapsule(addresses[i]));
```
alan-eth, line 153: The test design is really crisp — 14 realistic accountIds covering both divergent (`I`-containing) and non-divergent names, symmetric pre-/post-migration assertions, and using `Locale.forLanguageTag("tr")` to faithfully reproduce legacy behavior is lovely. 🎯

One branch that doesn't get exercised today: the `if (ArrayUtils.isEmpty(existing))` guard in `MigrateTurkishKeyHelper.doWork()`. Because this setup only seeds Turkish-form keys and never seeds a pre-existing ROOT-form key for the same accountId, the conflict-recovery path (where a ROOT entry already exists and the legacy Turkish value is silently dropped) is always skipped. That branch encodes the conservative conflict policy you describe in the PR body — would you consider adding a small case that:

- puts a ROOT-form key for `"ISSRWallet"` with one address,
- then puts the Turkish-form key `"ıssrwallet"` with a different address,
- runs `doWork()`,
- asserts the ROOT value wins and the Turkish-form key is gone?

It would also make the intentional policy visible to future readers who might otherwise wonder whether a "keep legacy" policy would be safer.

framework/src/main/java/org/tron/core/db/api/MigrateTurkishKeyHelper.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Scan AccountIdIndexStore for keys containing Turkish dotless-ı (U+0131),
+   * replace them with ROOT-equivalent keys (ı → i), and delete the old keys.
+   *
+   * <p>Uses raw {@link IRevokingDB} access to bypass the fallback lookup
+   * logic in {@link AccountIdIndexStore#has(byte[])} — we need exact-match
+   * checks to determine whether the ROOT key already exists in the DB.
+   */
```
lxcmyf, line 42: [NIT] The justification here is inaccurate. `AccountIdIndexStore#has(byte[])` has no fallback lookup path — it just calls `getLowerCaseAccountId` (an idempotent ROOT-lowercase) and does a single `revokingDB.getUnchecked`. There's nothing to "bypass". Using raw `IRevokingDB` is still reasonable (iteration avoids the `BytesCapsule` wrap), but the comment should say that instead, so the next reader isn't left looking for a fallback branch that doesn't exist. Also worth noting: other startup migrations (`MoveAbiHelper`, `AssetUpdateHelper`) go through the high-level store API — consider aligning for consistency, or call out explicitly why this one is different.

framework/src/main/java/org/tron/core/db/api/MigrateTurkishKeyHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // Phase 2: migrate each Turkish key to its ROOT equivalent
+    for (Map.Entry<byte[], byte[]> entry : entriesToMigrate) {
+      String keyStr = new String(entry.getKey(), StandardCharsets.UTF_8);
+      byte[] rootKey = keyStr.replace(DOTLESS_I, 'i')
+          .getBytes(StandardCharsets.UTF_8);
+      // Only write if ROOT key doesn't already exist
+      byte[] existing = revokingDB.getUnchecked(rootKey);
+      if (ArrayUtils.isEmpty(existing)) {
```
lxcmyf: [SHOULD] `ArrayUtils.isEmpty(byte[])` returns `true` for both `null` **and** zero-length arrays. If a ROOT-form entry ever holds a zero-length value, this branch would treat it as "ROOT key doesn't exist" and overwrite it with the Turkish-form value — which directly contradicts the conflict policy documented in the PR description ("keep the ROOT entry and drop the legacy value"). In practice `AccountIdIndexStore` stores 21-byte addresses so the case is unlikely, but the contract of this migration should match what the PR body claims. Suggest `if (existing == null)` so the semantics are exactly "key is absent".

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -3880,14 +3881,14 @@ private int getShieldedTRC20LogType(TransactionInfo.Log log, byte[] contractAddr
         for (String topic : topicsList) {
           byte[] topicHash = Hash.sha3(ByteArray.fromString(topic));
           if (Arrays.equals(topicsBytes, topicHash)) {
-            if (topic.toLowerCase().contains("mint")) {
+            if (topic.toLowerCase(Locale.ROOT).contains("mint")) {
```
lxcmyf: [NIT] The same `topic` string is lowercased up to 5 times in this branch chain (`mint` → `transfer` → `burn` → `leaf` → `token`). This is on the event-subscribe topic-classification path that runs for every emitted log, so extracting it once is both clearer and cheaper:
```java
String lower = topic.toLowerCase(Locale.ROOT);
if (lower.contains("mint")) return 1;
else if (lower.contains("transfer")) return 2;
else if (lower.contains("burn")) {
  if (lower.contains("leaf")) return 3;
  else if (lower.contains("token")) return 4;
}
```

framework/src/main/java/org/tron/core/services/http/Util.java
```diff
@@ -525,10 +526,10 @@ public static byte[] getAddress(HttpServletRequest request) throws Exception {
   private static String checkGetParam(HttpServletRequest request, String key) throws Exception {
     String method = request.getMethod();
 
-    if (HttpMethod.GET.toString().toUpperCase().equalsIgnoreCase(method)) {
+    if (HttpMethod.GET.toString().toUpperCase(Locale.ROOT).equalsIgnoreCase(method)) {
```
lxcmyf, line 573: [NIT] These two lines contain pre-existing redundancy that's worth cleaning up while we're here:

- Line 529: `HttpMethod.GET.toString()` is always the literal `"GET"` (uppercase); `.toUpperCase(Locale.ROOT)` is a no-op; and `.equalsIgnoreCase(method)` already folds case. The chain collapses to `"GET".equalsIgnoreCase(method)`.
- Line 532: same deal — `HttpMethod.POST.toString().toUpperCase(Locale.ROOT).equals(method)` reduces to `"POST".equals(method)` (if case-sensitivity is intended) or `"POST".equalsIgnoreCase(method)` (if symmetric with GET).

The locale fix is correct in isolation, but the `toUpperCase` call here isn't really doing anything — better to drop it than keep it guarded by `Locale.ROOT`.

lxcmyf (commented):
Review findings：

framework/src/main/java/org/tron/core/db/api/MigrateTurkishKeyHelper.java
```diff
@@ -0,0 +1,92 @@
...
+        revokingDB.put(rootKey, entry.getValue());
+      }
+      keysToDelete.add(entry.getKey());
+    }
+
+    // Phase 3: delete old Turkish keys
+    for (byte[] key : keysToDelete) {
```
xxo1shine: [SHOULD]The `put` and `delete` operations can be performed directly in the first loop; there's no need for two loops.

errorprone/src/main/java/errorprone/StringCaseLocaleUsageMethodRef.java
```diff
@@ -0,0 +1,56 @@
...
+    implements BugChecker.MemberReferenceTreeMatcher {
+
+  @Override
+  public Description matchMemberReference(MemberReferenceTree tree, VisitorState state) {
+    String name = tree.getName().toString();
+    if (!"toLowerCase".equals(name) && !"toUpperCase".equals(name)) {
+      return Description.NO_MATCH;
+    }
+    // Verify the qualifier type is java.lang.String
+    Type qualifierType = ((com.sun.tools.javac.tree.JCTree) tree.getQualifierExpression())
```
xxo1shine: @halibobo1205,

[SHOULD] Avoid casting directly to `com.sun.tools.javac.tree.JCTree` to read the internal `.type` field. ErrorProne already wraps this in `ASTHelpers.getType()`:

```java
Type qualifierType = ASTHelpers.getType(tree.getQualifierExpression());
```

Benefits:
- Removes a hard dependency on the JDK-internal `com.sun.tools.javac.tree.JCTree` (requires `--add-exports` on JDK 9+ and is more likely to break across JDK versions).
- Stays consistent with the `ASTHelpers.getSymbol(tree)` style already used a few lines below.
- Easier to maintain across future ErrorProne upgrades.

The remaining `com.sun.tools.javac.code.Symbol` / `Type` imports still belong here (that's the standard BugChecker SDK), but dropping the explicit `JCTree` cast is a clear win.

errorprone/src/main/java/errorprone/StringCaseLocaleUsageMethodRef.java
```diff
@@ -0,0 +1,56 @@
...
+        .type;
+    if (qualifierType == null) {
+      return Description.NO_MATCH;
+    }
+    if (!state.getTypes().isSameType(
```
xxo1shine: @halibobo1205,

[NIT] The idiomatic ErrorProne way to compare types is via `ASTHelpers`:

```java
if (!ASTHelpers.isSameType(qualifierType, state.getSymtab().stringType, state)) {
  return Description.NO_MATCH;
}
```

Or, using the standard supplier for an even shorter form:

```java
import static com.google.errorprone.suppliers.Suppliers.STRING_TYPE;
...
if (!ASTHelpers.isSameType(qualifierType, STRING_TYPE.get(state), state)) { ... }
```

Functionally identical, but matches the convention used by ErrorProne's own built-in checkers — easier for future maintainers to recognize.

framework/src/main/java/org/tron/core/Wallet.java
halibobo1205: Thanks — we're intentionally keeping this PR's scope limited to "replace no-arg `toLowerCase()` / `toUpperCase()` with the `Locale.ROOT` variant", without refactoring surrounding logic even when the underlying call is redundant (the same argument applies to the repeated `topic.toLowerCase` in `Wallet.getShieldedTRC20LogType`). Replacing a no-arg call with `Locale.ROOT` is byte-for-byte neutral; removing the call entirely or collapsing it into `equalsIgnoreCase` is a semantic refactor that deserves its own PR. Happy to follow up with a dedicated cleanup.

framework/src/main/java/org/tron/core/services/http/Util.java
halibobo1205: this PR's scope limited to "replace no-arg toLowerCase() / toUpperCase() with the Locale.ROOT variant“

framework/src/main/java/org/tron/core/db/api/MigrateTurkishKeyHelper.java
halibobo1205: - `AccountIdIndexStore`'s only public `put` path stores `accountCapsule.getAddress().toByteArray()` — always a 21-byte TRON address, never empty or null.
 - More importantly, the store's own `has()` and `get()` methods both use `ArrayUtils.isEmpty(value)` as the absence check. The migration uses the same predicate deliberately, to stay consistent with how the rest of the store interprets a value.
 - Switching to `existing == null` would create an internal inconsistency: if a zero-length value ever existed, `store.has(k)` would report "absent" while the helper would report "present, skip" — same DB state, different verdicts.
 So the current `ArrayUtils.isEmpty` check is intentional and matches the store's conventions.

framework/src/test/java/org/tron/core/db/AccountIdIndexStoreTest.java
halibobo1205: Thanks — I considered this carefully and want to keep the test surface as-is.

 The conflict branch only fires when the same accountId is committed to the DB under both Turkish-form and ROOT-form. For that state to exist on any deployed node, the node would have had to either ：
(a) switch locale mid-operation while remaining in consensus, 
(b) restore a tr-locale snapshot onto a ROOT-locale node.
Both of which would have produced contradictory `SetAccountIdActuator.validate()` verdicts versus the rest of the network and forked the chain. Mainnet and Nile have not observed such a fork, so no deployed DB contains the state this branch handles.

The branch itself stays in the code as a defensive check (snapshot-stitching tools, debugging operations, etc., should still not corrupt ROOT entries), but exercising it in a unit test would document a behavior that production data can never actually reach. I'd rather keep the test suite focused on the paths that real DBs traverse.

yanghang8612:
The three-layer fix is well-shaped: `Locale.ROOT` substitution is byte-for-byte neutral within the input domain established by the PR description (the only ASCII codepoint that diverges across locales is `I`, only on tr/az), the ErrorProne `StringCaseLocaleUsage` rule plus the custom `StringCaseLocaleUsageMethodRef` checker close the regression door at compile time (including the method-reference form upstream misses), and `MigrateTurkishKeyHelper` — gated by `TURKISH_KEY_MIGRATION_DONE`, modelled on `MoveAbiHelper`, with a documented conflict policy — is the right shape for the only persisted site.

One angle worth recording in the PR body for future readers: the "only one site reaches persistent storage" claim is independently verifiable by inspection. Every other touched call site is one of (a) runtime engine selection (`dbEngine.toUpperCase()` in `TronDatabase` / `TronStoreWithRevoking` / `TxCacheDB`), (b) in-memory config or membership checks (`disabledApiList`, `HttpMethod` comparisons, `accountType` switch in `Account`), (c) display/formatting (`Hex.toHexString().toUpperCase()` in `DataWord`, help-text formatting in `Args`), or (d) host-environment probing (`os.name` / `os.arch` / `java.vendor` in `Arch`, `WalletUtils`, `KeystoreFactory`). None of these reach disk or cross nodes, so the migration scope is genuinely exact rather than approximate — worth a one-liner in the description so the audit is auditable later without re-running the sweep.

LGTM once the remaining NITs resolve.

halibobo1205:
@yanghang8612  Thanks for the detailed audit and the clean (a)–(d) categorization — updated the PR description accordingly.

lvs0075 (approved):

xxo1shine (approved):

yanghang8612 (approved):
LGTM

317787106 (approved):

PR #6700, opened 2026-04-23
feat(log): improve  log for gRPC, DB, and shutdown

### LINKED ISSUE #6583: Improve logging: SLF4J bridge, less startup noise, fix shutdown log loss
**Background**

The current java-tron logging system has three concrete problems affecting observability, startup clarity, and graceful shutdown:

* **gRPC logs bypass Logback and go directly to stderr.** grpc-java uses `java.util.logging` (JUL) internally. While the project already includes `jcl-over-slf4j`, the JUL→SLF4J bridge is absent. As a result, gRPC diagnostic messages — connection resets, TLS handshake failures, etc. — never reach `tron.log`, forcing operators to monitor both stderr and log files simultaneously during RPC troubleshooting (see grpc-java\#1577).  
* **DB stats logging generates excessive INFO noise at startup.** `DbStat.statProperty()` logs at INFO level, flooding startup output and burying critical initialization messages.  
* **Shutdown logs are incomplete on abnormal exits.** After receiving SIGTERM, `TronLogShutdownHook` waits up to 180 seconds before closing. If the node hangs during this window, critical shutdown-phase logs may never be flushed.

*Core Changes*

1. **JUL→SLF4J bridge:** Add `jul-to-slf4j` as a dependency and call `SLF4JBridgeHandler.install()` at the earliest possible point during node startup, after resetting the default JUL handlers. This routes all gRPC JUL output into SLF4J/Logback for centralized collection.

java

```java
// Remove default JUL handlers to prevent duplicate output
java.util.logging.LogManager.getLogManager().reset();
SLF4JBridgeHandler.install();
```

groovy

```
// build.gradle
implementation 'org.slf4j:jul-to-slf4j:1.7.36'
```

2. **DB stats log level downgrade:** Change `log.info(...)` calls in `DbStat.statProperty()` and similar high-frequency periodic stat methods to `log.debug(...)`, reducing startup noise without losing diagnostic value.  
3. **Shutdown log completeness:** Introduce a `volatile boolean shuttingDown` flag in `TronLogShutdownHook`, shorten unnecessary wait intervals, and ensure critical log buffers are flushed as a first-priority step in the JVM shutdown hook.

*Configuration Changes*

* Add an explicit logger level entry for `io.grpc` in `logback.xml` so operators can tune gRPC log verbosity independently.  
* DB stats loggers default to DEBUG; operators can promote them on demand.

*No API / Protocol Changes*

This change has no impact on external APIs or the network protocol.

**Testing Strategy**

1. Verify that gRPC diagnostic logs (connection resets, TLS failures, etc.) appear in `tron.log` and are no longer emitted to stderr.  
2. Compare INFO-level log entry counts during node startup before and after — confirm a significant reduction in DB stats noise.  
3. Simulate an abnormal shutdown (send SIGTERM, then stall the process) and verify that critical shutdown-phase entries are complete and timely.

* `SLF4JBridgeHandler` introduces marginal overhead for JUL log forwarding, but gRPC diagnostic logs are infrequent during normal operation — the impact is negligible.  
* Downgrading DB stats logs reduces log I/O volume with no runtime performance cost.

**Scope of Impact**

* **Other:** logging initialization, DB module stats log level, node shutdown lifecycle

**Breaking Changes:** None　　

**Backward Compatibility:** None

**Alternatives Considered** None

**References**

* grpc/grpc-java\#1577 
* [SLF4J JUL Bridge](https://www.slf4j.org/legacy.html#jul-to-slf4j)

halibobo1205:
**Summary**

Closes #6583.

Six operator-facing logging improvements:

1. **gRPC logs route through Logback into a dedicated `grpc.log`.** Install `SLF4JBridgeHandler` (jul-to-slf4j) at the earliest startup point so grpc-java's JUL output is routed through Logback instead of bypassing it to stderr; `io.grpc` is configured at INFO and written to `./logs/grpc/grpc.log`, so `NettyServerTransport`'s `"Transport failed"` JUL-INFO diagnostics (TLS handshake failures, unexpected resets, and other non-quiet exceptions) are captured in managed log files.
2. **DB stats no longer flood startup at INFO.** `DbStat.statProperty()` moved to DEBUG, so critical init messages are no longer buried.
3. **Shutdown logs are complete.** `TronLogShutdownHook` refactored: constants named, **max wait raised from 60 s to 180 s** to give the node's executor pools realistic headroom to drain.
4. **Slow signature verifications are visible.** `TransactionCapsule.logSlowSigVerify(...)` emits INFO for any single verify > 50 ms.
5. **LevelDB open stalls are no longer silent.** `LevelDbDataSourceImpl` wraps the JNI-blocking `factory.open(...)` in a watchdog: WARN after 60 s with a remediation hint pointing at `Toolkit.jar db archive`.
6. **Invalid `--log-config` fails fast.** `LogService.load(...)` throws `TronError(ErrCode.LOG_LOAD)` on an unreadable `--log-config` path.

**Impact**

- **grpc.log:** gRPC TLS/handshake/connection-reset diagnostics now land in `./logs/grpc/grpc.log` instead of stderr.
- **tron.log:** startup no longer dumps DB stats at INFO; slow signature verification and stuck LevelDB open diagnostics remain operator-visible in the main managed logs.
- **CI output:** test root logger moved from `DEBUG` to `INFO` to reduce CI log volume.

**Release scope**

Logging-only change: no API, no RPC, no network, no DB format change.


halibobo1205:
The coverage delta is dominated by `BlockEventGet: 87.65% → 54.73% (+80 missed lines)`. Root cause is `EventPluginLoader` singleton pollution from `FilterQueryTest` — the same bug being fixed in #6688 (commit [`[ab0db5f0](https://github.com/tronprotocol/java-tron/pull/6688/commits/ab0db5f0ea23e7f5e6c3f1a66dbe5beb5735b80a)`](https://github.com/tronprotocol/java-tron/pull/6688/commits/ab0db5f0ea23e7f5e6c3f1a66dbe5beb5735b80a)). None of the regressing classes are in this PR's diff. Will re-check after #6688 lands.

build.gradle
```diff
@@ -81,8 +81,9 @@ subprojects {
     }
 
     dependencies {
-        implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
-        implementation group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.25'
+        implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.36'
```
waynercheung, line 84: [SHOULD] This PR upgrades `slf4j-api` and `jcl-over-slf4j` from 1.7.25 to 1.7.36 while adding `jul-to-slf4j`. Please mention this in the PR description per the dependency-upgrade checklist: why the SLF4J artifacts are aligned on 1.7.36, why this remains compatible with `logback-classic` 1.2.13 / SLF4J 1.7.x, and the result of `:framework:dependencyInsight --configuration runtimeClasspath --dependency org.slf4j:slf4j-api` showing the runtime classpath resolves consistently.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
```diff
@@ -656,12 +660,27 @@ public boolean validatePubSignature(AccountStore accountStore,
...
 
+  /**
+   * INFO-logs when a single signature verification exceeds
+   * {@link #SLOW_SIG_VERIFY_MS}. Package-private so it can be exercised from
+   * tests without forcing a real slow crypto path.
+   */
+  void logSlowSigVerify(long startNs) {
+    long costMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs);
+    if (costMs > SLOW_SIG_VERIFY_MS) {
+      logger.info("slow verify: txId={}, sigCount={}, cost={} ms",
```
waynercheung: [SHOULD] This INFO log is on a per-transaction verification path. Please keep the operator-visible INFO signal, but bound its volume and include the missing `permissionId` from #6583. A GC/CPU stall or malformed-transaction burst can otherwise emit many slow-verify lines at once, and the 100-slot async appender queue will not protect `tron.log` from sustained bursts. Suggested payload: `txId`, `sigCount`, `permissionId`, and `costMs`; suggested control: time-based ate limiting or aggregation of slow-verify counts.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
```diff
@@ -656,12 +660,27 @@ public boolean validatePubSignature(AccountStore accountStore,
...
   }
 
+  /**
+   * INFO-logs when a single signature verification exceeds
+   * {@link #SLOW_SIG_VERIFY_MS}. Package-private so it can be exercised from
+   * tests without forcing a real slow crypto path.
+   */
+  void logSlowSigVerify(long startNs) {
+    long costMs = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNs);
+    if (costMs > SLOW_SIG_VERIFY_MS) {
```
waynercheung, line 678: [QUESTION] #6583 proposed 10 ms, while this PR uses 50 ms based on the stated 2-6 ms baseline. The 50 ms threshold looks reasonable, but please call out the intentional deviation in the PR description so downstream reviewers do not have to reconcile the issue and implementation manually.

chainbase/src/main/java/org/tron/common/storage/leveldb/LevelDbDataSourceImpl.java
```diff
@@ -121,6 +136,15 @@ private void openDatabase(Options dbOptions) throws IOException {
     if (!Files.isSymbolicLink(dbPath.getParent())) {
       Files.createDirectories(dbPath.getParent());
     }
+    final long openStartNs = System.nanoTime();
+    final AtomicReference<String> manifestInfo = new AtomicReference<>();
+    ScheduledExecutorService watchdog = ExecutorServiceManager
```
waynercheung, line 134: [SHOULD] `openDatabase()` allocates a fresh `ScheduledExecutorService` (and its daemon thread) per store and tears it down in `finally`. java-tron opens dozens of stores serially during `Spring context.refresh()` and almost all complete well under the 60 s threshold, so on the happy path this is pure thread create/teardown churn for an event that rarely fires.

Per-DB attribution is **not** lost by sharing the scheduler — the WARN message body already carries `dataBaseName` (`"DB {} open still in progress after {}s…"`), and because stores open serially during startup, at most one watchdog is alive at any moment; `jstack` on a stuck startup will pin the offender via the main thread's `factory.open(...)` frame, not the watchdog's thread name. So the per-DB executor pattern doesn't buy diagnostic value beyond what the log line already provides.

**Suggest one shared static daemon scheduler:**

```java
private static final ScheduledExecutorService OPEN_WATCHDOG =
    ExecutorServiceManager.newSingleThreadScheduledExecutor("db-open-watchdog", true);
```

Capture `dataBaseName` in the lambda; cancel the task in `finally`. If you want the per-task name to show up in `jstack`(cheap, optional), set `Thread.currentThread().setName("db-open-watchdog[" + dataBaseName + "]")` at the top of the task and restore it in a `finally` inside the lambda. Also call `setRemoveOnCancelPolicy(true)` on the underlying `ScheduledThreadPoolExecutor` so cancelled delayed tasks don't accumulate in the queue (or add that knob to `ExecutorServiceManager`).

common/src/main/java/org/tron/common/log/LogService.java
```diff
@@ -12,18 +14,42 @@ public class LogService {
   public static void load(String path) {
     LoggerContext lc = (LoggerContext) LoggerFactory.getILoggerFactory();
     try {
-      File file = new File(path);
-      if (!file.exists() || !file.isFile() || !file.canRead()) {
-        return;
+      // we must fail fast rather than silently fall back and mislead them into
```
waynercheung: [NIT] Rolling style items into one comment per project convention:

- **LogService.java:17–18** — comment fragment "we must fail fast rather than silently fall back and mislead them..." has no clear antecedent for "them". Suggest:

```java
   // Fail fast rather than silently falling back to the classpath default —
   // that legacy behavior misled operators into thinking their custom
   // --log-config was active.
```

- **framework/src/test/java/org/tron/common/storage/leveldb/LevelDbDataSourceImplTest.java:151** 
local variable `final String WATCHDOG_DB_NAME = "slow-open-db";` should be lowerCamelCase per Google Java Style §5.2.2 (`SCREAMING_SNAKE_CASE` is reserved for static final constants). Rename to `watchdogDbName`.

- **chainbase/src/main/java/org/tron/common/storage/leveldb/LevelDbDataSourceImpl.java** - `git diff --check upstream/develop...HEAD` flags trailing whitespace on every added line in this file. Heads-up: the cause is pre-existing — the file is 503-line CRLF on develop, and the PR's new lines correctly match the existing line endings. **Please do not convert CRLF->LF in this PR** (it would touch every line and pollute `git blame`); a separate cleanup PR is the right place.

build.gradle
halibobo1205: Pre-PR, the runtime classpath had a version mismatch within the SLF4J family:
  - `org.slf4j:slf4j-api` resolved to **1.7.36** (transitively pulled to the highest version on the graph)
  - `org.slf4j:jcl-over-slf4j` was pinned to **1.7.25** in build.gradle and  nothing on the graph bumped it

SLF4J explicitly requires all artifacts in the family (api / bindings /bridges / over-jars) to share the same version — mismatched versions have undefined behavior (silent bridge failures, NoSuchMethodError on internal calls between api and bridge, etc.).

This PR brings every SLF4J artifact (`slf4j-api`, `jcl-over-slf4j`, and the newly-added `jul-to-slf4j`) onto **1.7.36**, matching the version `slf4j-api` that was already resolving to. Side benefit: SAST/lint tools no longer flag the explicit 1.7.25 declaration as a stale version, while the runtime sits on a different one.

Compatibility: 1.7.x is binary-compatible across the line by design, and `logback-classic` 1.2.13 ships against 1.7.32 in its parent POM but is tested across 1.7.x. Staying on 1.7.x avoids the 2.x ABI break that would also require bumping `logback-classic` to 1.4+.
dependencyInsight (post-PR, proves the family is now uniform):
```
$ ./gradlew :framework:dependencyInsight \
    --configuration runtimeClasspath --dependency org.slf4j:slf4j-api
$ ./gradlew :framework:dependencyInsight \
    --configuration runtimeClasspath --dependency org.slf4j:jcl-over-slf4j
<paste both outputs>
```

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
halibobo1205: **1. `permissionId` field — intentionally omitted**

 The `permissionId` requirement has been removed from issue #6583. In the specific context of "slow signature verification," this field carries no diagnostic value:

 - The real causes (multi-sig signature count / GC pauses / chain-level ECDSA→SM2 switch / CPU contention) are all independent of permission type
 - Owner / active / witness permissions all run through the same verify primitive — there is no per-permission cost difference
 - With `txId` already in the payload, an operator can pull the full permissionId and signature details offline via `getTransactionInfoById`. Including it in the log line would be redundant and actively misleading: any "group by permissionId" analysis will be dominated by the SR (witness) path simply because SR signing is constant, not because it is the cause

 Keeping the three fields that actually carry diagnostic signal: `txId`, `sigCount`, `costMs`.

 **2. Rate limiting — not needed**

 In real production, `verify > 50ms` is a very sparse event

 Additionally, logback's `AsyncAppender` defaults to `discardingThreshold = queueSize / 5 = 20`, so INFO is silently dropped under queue pressure — it never blocks the producer and never destabilizes `tron.log`. The natural backstop is already in place at the appender layer.

 No explicit rate-limit logic is introduced: a sparse event does not need throttling, and adding one would require maintaining an independent state machine (suppression window/counter reset / periodic summary output) that buys nothing here.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
halibobo1205: The issue has been updated to 50 ms.

chainbase/src/main/java/org/tron/common/storage/leveldb/LevelDbDataSourceImpl.java
halibobo1205: <p>Per-DB executor cost is <code>ScheduledThreadPoolExecutor</code> allocation + <code>scheduleAtFixedRate</code> + <code>cancel</code> + <code>shutdownNow</code>. On the happy path (DB opens within the 60 s <code>initialDelay</code>), the worker thread is never started thanks to lazy thread creation in <code>ScheduledThreadPoolExecutor</code>, so the actual per-DB cost collapses to microseconds — sub-millisecond. Across ~30 stores at startup, the total is at least an order of magnitude below LevelDB's own I/O time. Operationally invisible.</p>
<p>What it buys is exact lifecycle alignment with the monitored object and zero residual state after startup. A shared static would trade that "few milliseconds of imperceptible CPU" for "one daemon thread resident for the lifetime of a multi-week node process" — the trade-off runs the wrong direction.</p>
</blockquote>

common/src/main/java/org/tron/common/log/LogService.java
halibobo1205: fixed in  https://github.com/tronprotocol/java-tron/pull/6700/changes/df07f45b7127a37364d15e58503f80fe6a215029

waynercheung (approved):
LGTM

lxcmyf (approved):

317787106 (approved):

PR #6701, opened 2026-04-23
feat(framework,actuator,common): replace fastjson with jackson

### LINKED ISSUE #6607: [Feature] Replace fastjson with Jackson
## Problem Statement

Java-tron currently uses two JSON libraries side by side:

- **Jackson** (`com.fasterxml.jackson`) — primarily in the JSON-RPC、eventPlugin、keystore、TVM Trace、Transaction Cache、Peer Persist  layer
- **fastjson** (`com.alibaba:fastjson`) — primarily in the HTTP API layer

This dual-library situation creates three concrete problems:

1. **Maintenance burden** — Developers must learn and maintain two distinct serialization APIs (`JSONObject` / `JSONArray` vs. `ObjectMapper` / `JsonNode`), increasing cognitive overhead and onboarding cost.
2. **Security exposure** — fastjson has accumulated 20+ CVEs since 2017, including multiple critical RCE vulnerabilities (e.g., autoType bypass chains). The library's development focus has shifted to fastjson2, leaving 1.x in maintenance-only mode.
3. **Serialization inconsistency** — The HTTP layer and JSON-RPC layer diverge in field ordering, `null` handling, and type coercion, causing subtle integration issues for downstream API consumers.

| Dimension | fastjson | Jackson |
|---|---|---|
| Ecosystem reach | Alibaba-centric | Java ecosystem standard (Spring Boot, Quarkus, Dropwizard) |
| Security track record | 20+ CVEs incl. critical RCE | Mature vulnerability handling, well-audited |
| API richness | Basic parse / serialize | ObjectMapper, JsonNode tree model, streaming API, custom serializers, MixIn, modular extensions |
| Maintenance | Single-vendor driven | FasterXML org + broad community |

## Proposed Solution

Replace all `com.alibaba.fastjson` usages with the existing `com.fasterxml.jackson`.

### Specification

**Dependency Changes**

- **Remove**: `implementation 'com.alibaba:fastjson:1.2.83'` from `framework/build.gradle`
- **Upgrade**: `jackson-databind` from 2.18.3 → 2.18.6 (addresses CVE GHSA-72hv-8253-57qq)

**Code Changes**

- Replace `JSONObject` / `JSONArray` / `JSON.parseObject()` / `JSON.toJSONString()` with Jackson equivalents (`ObjectMapper`, `ObjectNode`, `ArrayNode`, `JsonNode`)
- Introduce `Util.MAPPER` (shared `ObjectMapper` singleton) and `Util.parseObject()` as centralized helpers in `Util.java.`
- Remove `ParserConfig.setSafeMode()` from `DefaultConfig.java` (Jackson is safe by default)
- Migrate all HTTP Servlet classes, `ContractEventParserJson`, `TronJsonRpcImpl`, and related test files

**API Changes**: None — all external JSON output formats remain identical.

**Configuration Changes**: None.

**Protocol Changes**: None.

## Scope of Impact

- [ ] Core protocol
- [x] API/RPC
- [ ] Database
- [ ] Network layer
- [ ] Smart contracts
- [ ] Documentation
- [ ] Other

**Breaking Changes**

None. External JSON response formats are preserved.

**Backward Compatibility**

Fully backward compatible. This is a pure internal refactoring — no external API, configuration, or protocol changes.

## Implementation

1. Centralize a shared `ObjectMapper` singleton in `Util.java` with safe defaults (`FAIL_ON_UNKNOWN_PROPERTIES = false`, etc.).
2. Start with `Util.java` as the pivot point — most HTTP Servlets delegate JSON operations through it, so converting this file cascades coverage broadly.
3. Migrate each Servlet file, replacing fastjson API calls with Jackson equivalents.
4. Update all corresponding test files.

Yes.

**Estimated Complexity**

Medium — Large volume of files (113+ changed).

## Testing Strategy

**Test Scenarios**

1. All existing HTTP Servlet unit tests pass without regression.
2. All JSON-RPC tests pass without regression.
3. JSON output semantic equivalence verified across all modified API endpoints (field names, value types, null handling, field ordering).

- Jackson's `ObjectMapper` is thread-safe and designed for singleton reuse — no per-request allocation overhead.
- No performance degradation expected; Jackson is the serialization engine already used by Spring and most Java frameworks at scale.
- Benchmark critical endpoints under concurrent load to confirm no throughput regression.

halibobo1205:
### Summary
Replace `com.alibaba:fastjson` with Jackson-backed drop-in wrappers (`org.tron.json.{JSON, JSONObject, JSONArray, JSONException}`). No external API changes — all HTTP and JSON-RPC responses remain identical.

### Motivation
- Fastjson 1.2.83 is EOL with 20+ CVEs including critical RCE
- Upgrade jackson-databind 2.18.3 → 2.18.6 (GHSA-72hv-8253-57qq)
- Unify JSON handling (previously split between Jackson and Fastjson)

#### (common):
- Add `org.tron.json` wrappers backed by a shared `ObjectMapper`
- Remove fastjson from `common/build.gradle`
- Expose `maxNestingDepth` / `maxTokenCount` on `CommonParameter` and `NodeConfig.HttpConfig`, with defaults `100` / `100_000` in `reference.conf`

#### (framework): HTTP & servlet changes
- Swap imports from `com.alibaba.fastjson` → `org.tron.json` across all HTTP servlets, JSON-RPC layer, and event/log parsers
#### Compatibility

| Area | Fastjson default behavior | Jackson behavior in this PR | Compatibility |
|---|---|---|---|
| Unquoted field names | Accepts `{a:1}` |  `ALLOW_UNQUOTED_FIELD_NAMES` | 🟢 Compatible |
| Single quotes | Accepts `{'a':'1'}` | `ALLOW_SINGLE_QUOTES` | 🟢 Compatible |
| Single trailing comma | Accepts `{"a":1,}` and `[1,2,]` | `ALLOW_TRAILING_COMMA` | 🟢 Compatible  |
| Repeated / arbitrary commas | Accepts cases like `{"a":1,,,,}` and `[1,,2]` | Rejects | 🔴 Not supported |
| `NaN` | Accepts and treats as `null` | Rejects | 🔴 Not supported |
| `Infinity` / `-Infinity` | Rejects | Rejects | 🟢 Compatible |
| Leading plus sign | Accepts `+123`, `+0.5` | `ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS` | 🟢 Compatible |
| Signed leading decimal | Accepts `+.5` |  `ALLOW_LEADING_DECIMAL_POINT_FOR_NUMBERS` | 🟢 Compatible |
| Unsigned leading decimal | Rejects `.5` |  `ALLOW_LEADING_DECIMAL_POINT_FOR_NUMBERS` | 🔴 More permissive than Fastjson. Jackson cannot accept `+.5` without also accepting `.5`. |
| Trailing decimal point | Accepts `5.` | `ALLOW_TRAILING_DECIMAL_POINT_FOR_NUMBERS` | 🟢 Compatible |
| Leading zeros | Accepts `007` as `7` | `ALLOW_LEADING_ZEROS_FOR_NUMBERS` | 🟢 Compatible |
| Unescaped control chars in strings | Accepts raw control chars such as newline/tab in strings | Accepts via `ALLOW_UNESCAPED_CONTROL_CHARS` | 🟢 Compatible |
| Non-standard backslash escapes | Rejects `"\q"` | Rejects  | 🟢 Compatible |
| Java-style comments | Accepts `// ...` and `/* ... */` | `ALLOW_JAVA_COMMENTS` | 🟢 Compatible |
| Uppercase `NULL` token | Accepts bare `NULL` as `null` | Rejects | 🔴 Not supported |
| Uppercase boolean tokens | Rejects `TRUE` / `FALSE` | Rejects | 🟢 Compatible |
| Floating point numeric type | Uses `BigDecimal` by default | `USE_BIG_DECIMAL_FOR_FLOATS` | 🟢 Compatible |
| Unknown POJO fields | Ignores unknown fields by default |`FAIL_ON_UNKNOWN_PROPERTIES = false` | 🟢 Compatible |
| Empty bean serialization | Serializes empty beans as `{}` |`FAIL_ON_EMPTY_BEANS = false` | 🟢 Compatible |
| Null-valued fields serialization | Omits null fields by default unless `WriteMapNullValue` is enabled | `JsonInclude.Include.NON_NULL` | 🟢 Compatible |
 > [!CAUTION]
> The table above covers only the currently known differences in Fastjson/Jackson behavior verified during this migration. There may still be other unknown edge-case differences. New code and clients are strongly encouraged to use strict standard JSON instead of relying on Fastjson-specific lenient parsing behavior.

### Configuration
Two new `node.http` keys (defaults shown):

```hocon
node {
  http {
    maxNestingDepth = 100      # Jackson default 1000
    maxTokenCount   = 100000   # Jackson default Long.MAX_VALUE
  }
}
```

Beyond either ceiling, parsing fails with `StreamConstraintsException` before any business logic runs. Defaults are sized to accommodate every legitimate java-tron HTTP payload comfortably.

### Build:
- Update Jackson to 2.18.6
- Remove fastjson

close #6607


common/src/main/java/org/tron/json/JSON.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private JSON() {
+  }
+
+  /**
+   * Returns {@code true} when {@code text} is null, blank, or a
+   * case-insensitive {@code "null"} literal — mirroring Fastjson's lenient
+   * treatment of these inputs as JSON {@code null}.
+   */
+  static boolean isNullLiteral(String text) {
```
lxcmyf, line 83: [MUST] Preserve unquoted NULL compatibility: `{a:NULL}` currently throws, but fastjson 1.2.83 accepted this and treated it as null. Since this PR claims drop-in behavior, this is a breaking compatibility regression for existing clients and should be fixed before merge.

actuator/src/main/java/org/tron/core/vm/trace/ProgramTrace.java
```diff
@@ -90,12 +90,12 @@ public void merge(ProgramTrace programTrace) {
     this.ops.addAll(programTrace.ops);
   }
 
-  public String asJsonString(boolean formatted) {
-    return serializeFieldsOnly(this, formatted);
+  public String asJsonString() {
```
lxcmyf: [SHOULD] Avoid breaking `ProgramTrace.asJsonString(boolean)` API: replacing it with a no-arg method introduces source/binary compatibility risk for external callers and removes compact-vs-pretty output control. Prefer keeping the original signature (or adding an overload) to preserve compatibility.

common/src/main/java/org/tron/json/JSON.java
```diff
@@ -0,0 +1,155 @@
...
+
+  public static final ObjectMapper MAPPER = JsonMapper.builder()
+      // Fastjson Feature.AllowUnQuotedFieldNames (default ON)
+      .enable(JsonReadFeature.ALLOW_UNQUOTED_FIELD_NAMES)
+      // Fastjson Feature.AllowSingleQuotes (default ON)
+      .enable(JsonReadFeature.ALLOW_SINGLE_QUOTES)
+      // Fastjson tolerates trailing commas (e.g. {"a":1,}) by default
+      .enable(JsonReadFeature.ALLOW_TRAILING_COMMA)
+      // Fastjson accepts NaN/Infinity as valid tokens
+      .enable(JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS)
```
waynercheung: [MUST] Please remove this flag from the shared request parser.

This is reachable on current wallet HTTP paths, not just a theoretical Fastjson-parity issue. `DeployContractServlet` and `TriggerSmartContractServlet` read `call_value`, `call_token_value`, `token_id`, and `fee_limit` through `Util.getJsonLongValue(...)`.

With this flag enabled, `Infinity` / `-Infinity` parse successfully, `getBigDecimal(...)` returns `null`, and those non-required fields become `0L` instead of rejecting the request. Before this PR, Fastjson rejected those payloads during parse.

framework/src/main/java/org/tron/core/services/http/Util.java
waynercheung, line 277: [SHOULD] This line controls the `"type"` field in every HTTP transaction JSON payload, but the new tests do not pin the exact serialized value here. 

`UtilMockTest#testPrintTransactionToJSON()` only asserts non-null, and most servlet tests only check "no Error" / key presence. Please add an exact-output assertion that `"type"` remains the expected string for at least one `TransferContract` and one `TriggerSmartContract` response.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
waynercheung, line 395: [QUESTION] `TransactionJson.transaction` is serialized by jsonrpc4j's own `ObjectMapper`, not `org.tron.json.JSON.MAPPER`. `@JsonValue` on `JSONObject.unwrap()` should bridge them, but I couldn't find a test that pins the on-wire bytes.

`BuildTransactionTest` stops at the rebuilt `ContractType`; `JsonrpcServiceTest` doesn't cover this serialization path.

Could we add one golden-output assertion for a representative `buildTransaction` response (transfer + trigger contract)? That's the most direct check on the "responses remain identical" claim for the JSON-RPC layer.

common/src/main/java/org/tron/common/utils/JsonUtil.java
```diff
@@ -1,14 +1,16 @@
 package org.tron.common.utils;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import org.springframework.util.StringUtils;
 
 public class JsonUtil {
 
+  private static final ObjectMapper om = new JsonMapper();
```
waynercheung, line 9: [SHOULD] This mapper backs persisted RocksDB JSON blobs (`NodePersistService`, `TxCacheDB`, and `Manager` recent-transaction persistence), so changes here affect restart-time compatibility, not just in-memory utility behavior. The current tests cover same-version round trips, but I don't see a compatibility-focused assertion against pre-existing stored JSON. A short comment or a focused regression test would make that storage-format role explicit.

framework/src/test/java/org/tron/core/services/http/GetDelegatedResourceAccountIndexV2ServletTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    verify(wallet).getDelegatedResourceAccountIndexV2(eq(expectedAddress));
+    String postContent = response.getContentAsString();
+    assertFalse(postContent.contains("\"Error\""));
+    assertFalse(postContent.trim().isEmpty());
+  }
+
+  @Test
+  public void testGet() throws Exception {
+    MockHttpServletRequest request =
```
waynercheung: [NIT] `git diff --check` still reports trailing whitespace here.

What about:
```java
MockHttpServletRequest request = getRequest("visible", "true", "value",
        "TBxSocpujP6UGKV5ydXNVTDQz7fAgdmoaB");
```

yanghang8612:
Direction is right — fastjson 1.2.83 has been a long-standing security overhang, and consolidating on Jackson + a thin wrapper is the obvious move. Waiting on the MUST items @lxcmyf and @waynercheung raised before LGTM.

One additional question worth pinning down in the PR description: after this change, are there any remaining `com.alibaba.fastjson` imports reachable anywhere in the tree — or does this fully retire the dependency? The diff summary doesn't show the `build.gradle` side, and a `grep -r 'com.alibaba.fastjson' --include='*.java'` result in the PR body would make the "no more fastjson" claim explicit and let dependency scanners (GHSA, Snyk) close the finding cleanly.

Also worth a sentence in the PR description on hot-path performance posture: fastjson 1.2.x has a historically-fast parse path, and Jackson has different characteristics under `LargeJsonPayload` + `/wallet/*` throughput. If there's even a rough benchmark for `DeployContractServlet` / `TriggerSmartContractServlet` JSON-decode latency pre/post, that closes the loop on "drop-in" covering both correctness *and* perf.

common/src/main/java/org/tron/json/JSON.java
halibobo1205: This scenario won't be triggered in java-tron in practice:

**1. Most servlets are guarded by `JsonFormat.merge`**

Transaction-related servlets (TransferServlet, DeployContractServlet, TriggerSmartContractServlet, etc.) all follow this flow:

```
HTTP body → JsonFormat.merge() → JSONObject.parseObject()
```

`JsonFormat` uses its own `Tokenizer` (not Jackson, not Fastjson) with strict lowercase-only null matching. I traced the full path through the code:

- **Known fields** — `handlePrimitive()` (`JsonFormat.java:648`): `"null".equals(tokenizer.currentToken())` — strict lowercase
- **Unknown fields** — `handleMissingField()` (`JsonFormat.java:609`): same `"null".equals(...)` check

Suppose a client sends `NULL` (uppercase), neither path matches. It then fails `lookingAtInteger()` (first char `N`), fails `lookingAtBoolean()` (only accepts `"true"`/`"false"`), and falls through to `consumeString()` which throws `ParseException` because `N` is not `"` or `'`. This is true for both known and unknown protobuf fields.

So if the input contains uppercase `NULL`, `JsonFormat.merge` throws `ParseException` before `parseObject` is ever reached. This behavior is the same before and after the migration — it has nothing to do with the JSON library.

**2. The few servlets that parse external input directly expect trivially simple formats**

Only 3 call sites in the entire codebase bypass `JsonFormat` and call `parseObject` directly on external input:
- `ValidateAddressServlet` — expects `{"address": "T..."}`
- `BroadcastHexServlet` — expects `{"transaction": "0a0b..."}`
- `Util.getParam()` — extracts a single string parameter

All field values in these interfaces are strings. There is no business scenario where `null` (in any case) would be a meaningful input.

**3. No standard JSON client produces uppercase `NULL`**

Java (Jackson/Gson), Python (json), JavaScript (JSON.stringify), Go (encoding/json) — all mainstream JSON libraries exclusively output lowercase `null`. java-tron's clients (wallet SDKs, TronWeb, DApps) all use these standard libraries. Fastjson's acceptance of uppercase `NULL` was its own non-standard lenient behavior, never part of java-tron's API contract.

Given the above, this is not a real compatibility regression and does not require special handling.

---

common/src/main/java/org/tron/json/JSON.java
halibobo1205: `ALLOW_NON_NUMERIC_NUMBERS` is intentionally enabled, with explicit test coverage in `testNonNumericNumbers`.

Fastjson's handling of Infinity/NaN is actually flawed:
- `NaN` → silently returns null (loses semantics)
- `Infinity` / `-Infinity` → throws an exception

Meanwhile, java-tron's protobuf parsing layer (`JsonFormat.Tokenizer.consumeDouble()`) already supports `Infinity`, `-Infinity`, and `NaN` via `DOUBLE_INFINITY`/`FLOAT_NAN` regex matching. Enabling this flag makes the Jackson layer consistent with the protobuf layer.

Regarding the security concern: critical fields like transaction amounts are defined as `int64` in protobuf and are not affected by this configuration. Infinity/NaN only applies to double/float fields (e.g., `avgLatency`, `cpuRate` in NodeInfo), which already accept these values through `JsonFormat`.

actuator/src/main/java/org/tron/core/vm/trace/ProgramTrace.java
halibobo1205: `asJsonString(boolean)` has zero callers within the project — the only call site is `toString()`, which always passed `true`. The three custom serializers (`DataWordSerializer`, `ByteArraySerializer`, `OpCodeSerializer`) in `Serializers.java` were also dead code — never registered on any mapper.

`ProgramTrace` is an internal class, not part of a public SDK or API. If we retain unused methods based on the possibility of external callers, then by the same logic, no existing method in tron should ever be removed, which isn't practical. Removing unused code reduces maintenance burden and keeps the codebase clean.

framework/src/main/java/org/tron/core/services/http/Util.java
halibobo1205: `contract.getType()` is a protobuf `ContractType` enum. Both Fastjson and Jackson serialize enums via `.name()` by default, producing identical strings.

`UtilTest` already covers exact-value round-trip validation for this:

- `testPackTransaction()` hardcodes `"type": "TransferContract"` in the JSON input, which is deserialized via `ContractType.valueOf("TransferContract")` (`Util.java:291`) — this is an exact string match, not a contains/non-null check. The test passes only if serialization and deserialization use the same format.
- `testPackTransactionWithInvalidType()` verifies that an incorrect type string (`"TransferContract11111"`) is properly rejected.

framework/src/test/java/org/tron/core/services/http/GetDelegatedResourceAccountIndexV2ServletTest.java
halibobo1205: Fixed, thanks.

halibobo1205:
@yanghang8612 
**Fastjson fully retired:**
Yes. The fastjson dependency has been removed from `build.gradle` ([diff](https://github.com/tronprotocol/java-tron/pull/6701/changes#diff-3dc11ca3a71905f437e1e170f963e9854da27943efcff23bf87bfe87abeda513L47)). If any `com.alibaba.fastjson` import remained, the build would fail at compile time — so a passing CI already proves zero remaining references.

**Performance:**
The JSON parsing hot path for `/wallet/*` endpoints goes through `JsonFormat.java` (a custom protobuf-to-JSON serializer/parser with its own `Tokenizer`), not through the Jackson `ObjectMapper`. Jackson is only used for lightweight operations: extracting a few fields (`visible`, `type`, `value`, etc.) from the top-level request JSON. The actual heavy lifting — protobuf message construction and field-by-field parsing — is handled by `JsonFormat.merge()`, which is unchanged in this PR. So there is no meaningful parse-latency delta on the hot path. Performance benchmarks will be added as a follow-up.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
halibobo1205: Add test in https://github.com/tronprotocol/java-tron/pull/6701/changes/5916672e52ddc0bb8def66b3754954e2f5f9fa9a.

common/src/main/java/org/tron/json/JSONObject.java
```diff
@@ -0,0 +1,338 @@
...
+    if (value == null) {
+      node.remove(key);
+    } else {
+      node.put(key, value);
+    }
+    return this;
+  }
+
+  public JSONObject put(String key, Integer value) {
+    if (value == null) {
```
xxo1shine, line 228: [SHOULD] The following logic is used in multiple places and can be encapsulated into a function.
```
    if (value == null) {
      node.remove(key);
    } else {
      node.put(key, value);
    }
    return this;
```

common/src/main/java/org/tron/json/TypeUtils.java
```diff
@@ -0,0 +1,310 @@
...
+
+    if (value instanceof Number) {
+      return ((Number) value).intValue() == 1;
+    }
+
+    if (value instanceof String) {
+      String strVal = (String) value;
+      if (strVal.isEmpty()
+          || "null".equals(strVal)
+          || "NULL".equals(strVal)) {
```
xxo1shine, line 49: [QUESTION]  What should be returned in the following situations: "Null",   "  NULL  "

common/src/main/java/org/tron/json/TypeUtils.java
```diff
@@ -0,0 +1,310 @@
...
+      }
+      if ("true".equalsIgnoreCase(strVal)
+          || "1".equals(strVal)) {
+        return Boolean.TRUE;
+      }
+      if ("false".equalsIgnoreCase(strVal)
+          || "0".equals(strVal)) {
+        return Boolean.FALSE;
+      }
+      if ("Y".equalsIgnoreCase(strVal)
```
xxo1shine, line 60: [QUESTION] What should be returned in the following situations: "yes", " yes ", "t", " t "

common/src/main/java/org/tron/json/TypeUtils.java
```diff
@@ -0,0 +1,310 @@
...
+
+  public static Boolean castToBoolean(Object value) {
+    if (value == null) {
+      return null;
+    }
+    if (value instanceof Boolean) {
+      return (Boolean) value;
+    }
+
+    if (value instanceof BigDecimal) {
```
xxo1shine, line 37: [QUESTION] Will there be any loss of precision? Some Fastjson implementations use `.equals(BigDecimal.ONE)`.

common/src/main/java/org/tron/json/TypeUtils.java
```diff
@@ -0,0 +1,310 @@
...
+      if ("false".equalsIgnoreCase(strVal)
+          || "0".equals(strVal)) {
+        return Boolean.FALSE;
+      }
+      if ("Y".equalsIgnoreCase(strVal)
+          || "T".equals(strVal)) {
+        return Boolean.TRUE;
+      }
+      if ("F".equalsIgnoreCase(strVal)
+          || "N".equals(strVal)) {
```
xxo1shine, line 65: [QUESTION] What should be returned in the following situations: "no", " no ", "n", " n "

common/src/main/java/org/tron/json/TypeUtils.java
```diff
@@ -0,0 +1,310 @@
...
+        return Boolean.FALSE;
+      }
+    }
+    throw new JSONException("can not cast to boolean, value : " + value);
```
xxo1shine, line 69: [QUESTION] Character type not handled (Fastjson supports this)
Some versions of Fastjson support:
'1' → true
'0' → false

common/src/main/java/org/tron/json/TypeUtils.java
halibobo1205: This class is from FastJson; the default behavior is not changed.

common/src/main/java/org/tron/json/JSONObject.java
halibobo1205: These `put` overloads delegate to different type-specific methods on Jackson's `ObjectNode` (`node.put(key, String)`, `node.put(key, int)`, `node.put(key, long)`, etc.), so they can't be unified into a single generic helper without introducing a lambda or routing through `valueToTree(Object)` — both of which would add indirection to what is currently a straightforward 4-line method. Keeping them explicit is clearer here.

waynercheung (approved):
LGTM

yanghang8612 (approved):
LGTM

xxo1shine (approved):

317787106 (approved):

PR #6703, opened 2026-04-24
feat(vm): add ABI semantic validation for /wallet/deploycontract

### LINKED ISSUE #6674: [Feature] Add ABI semantic validation for `/wallet/deploycontract`
# Summary
Add semantic ABI validation to the `/wallet/deploycontract` interface so malformed or non-standard ABI definitions are rejected before transaction construction. The validation rules should follow the Solidity ABI specification to improve tooling compatibility and reduce audit findings around "ABI injection".

# Problem
### Motivation
The `/wallet/deploycontract` interface currently accepts user-supplied ABI data with only basic JSON-to-protobuf parsing. This leaves room for malformed ABI definitions to pass through even when they do not conform to common ABI expectations.

### Current State
In the current implementation, the `abi` field in the deploy contract request is parsed into protobuf structures, but there is no complete semantic validation of ABI entry types or parameter types.

Examples of inputs that may not be properly rejected today include:
- invalid parameter types such as `uint`, malformed array types, or illegal `bytesN`
- duplicate `fallback` or `receive` entries
- `receive` entries with non-`payable` mutability
- tuple-related definitions that cannot be fully represented by the current schema

### Limitations or Risks
This creates several issues:
- malformed ABI metadata may be accepted and persisted
- downstream tooling may interpret the stored ABI inconsistently
- audit tools may report this as an ABI injection issue
- developer experience suffers because invalid ABI is not rejected early with clear errors

# Proposed Solution

### Proposed Design
Introduce a dedicated ABI validation step for the `/wallet/deploycontract` interface and reject malformed ABI definitions before building the transaction.

This proposal changes the behavior of that specific interface: requests with an invalid `abi` field should fail fast during request handling / transaction construction, instead of being accepted and converted into protobuf ABI data.

At minimum, the validation should:
- validate top-level ABI entry types
- validate parameter types in `inputs` and `outputs`
- reject duplicate `fallback` and `receive` entries
- require `receive.stateMutability == payable`
- reject tuple definitions that cannot be safely stored or parsed by the current ABI schema

The implementation should follow the Solidity ABI specification, while keeping the final logic aligned with java-tron's own schema and compatibility requirements.

### Key Changes
- Module(s): `/wallet/deploycontract` request handling
- Module(s): ABI parsing / validation utility
- API: stricter validation for the `abi` field in `/wallet/deploycontract` requests
- Test: add coverage for representative invalid ABI cases

# Impact
This change is expected to improve:
- Security: stronger input validation and less room for malformed ABI payloads
- Stability: fewer inconsistencies in stored ABI metadata
- Developer Experience: earlier and clearer error reporting for invalid ABI input
- Tooling Compatibility: behavior closer to common ABI expectations

# Compatibility
- Breaking Change: No
- Default Behavior Change: Invalid ABI payloads sent to the `/wallet/deploycontract` interface that were previously accepted may now be rejected during request handling / deploy transaction construction
- Migration Required: No for valid ABI definitions

# References (Optional)
- Prior art: Solidity ABI specification and Solidity documentation
- Related issue: audit findings describing the problem as "ABI injection"

# Additional Notes

yanghang8612:
Closes #6674.

Add a dedicated `AbiValidator` and invoke it inside the existing `CreateSmartContract` branch of `Wallet.createTransactionCapsule`, so both the HTTP `/wallet/deploycontract` path and the gRPC `deployContract` path fail fast on malformed ABI input with a field-path-anchored error.

**Scope:** validation runs only at the API entry point (transaction construction). Block replay and re-validation of already-persisted contracts are unaffected — no consensus-layer behaviour change.

Rules align with go-ethereum's `accounts/abi` parser:

- reject `uint` / `int` shorthand and out-of-range `uintN` / `intN` widths
- reject `bytesN` with N outside `[1, 32]`
- reject the entire `fixed` / `ufixed` family
- reject malformed array suffixes
- reject leading / trailing whitespace in any type string (so persisted bytes equal the validated bytes)
- reject duplicate `constructor` / `fallback` / `receive`
- require `receive` to be payable (legacy `payable` flag honoured)
- reject `fallback` / `receive` carrying inputs or outputs
- reject `UnknownEntryType`
- require a name for `function`; require a name for `event` only when not `anonymous`; require a name for `error`
- keep `tuple` / `tuple[]` / `tuple[N]` permissive, since the proto schema cannot represent `components`

Also reuse the validator in `PublicMethod.jsonStr2Abi` and `TvmTestUtils.jsonStr2Abi` so test/util ABI parsing and request-handling share the same rule set; both helpers now also map `"receive"` and `"error"` so receive-/error-entry test ABIs reach the proper validator branch instead of falling to `UNRECOGNIZED`.

## Tests

- `AbiValidatorTest` — covers each rejection branch (widths, bytesN bounds, fixed/ufixed, array suffixes, whitespace, duplicates, payable/inputs/outputs constraints, UnknownEntryType, missing-name).
- `DeployContractServletTest` — end-to-end HTTP `/wallet/deploycontract` rejection.
- `TvmTestUtilsTest`, `PublicMethodTest` — confirm `receive` / `error` entry mapping reaches the validator.

### DIFF
--- a/actuator/src/main/java/org/tron/core/utils/AbiValidator.java
+++ b/actuator/src/main/java/org/tron/core/utils/AbiValidator.java
@@ -1,0 +1,189 @@
+package org.tron.core.utils;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.tron.core.exception.ContractValidateException;
+import org.tron.protos.contract.SmartContractOuterClass.SmartContract.ABI;
+import org.tron.protos.contract.SmartContractOuterClass.SmartContract.ABI.Entry;
+import org.tron.protos.contract.SmartContractOuterClass.SmartContract.ABI.Entry.EntryType;
+import org.tron.protos.contract.SmartContractOuterClass.SmartContract.ABI.Entry.Param;
+import org.tron.protos.contract.SmartContractOuterClass.SmartContract.ABI.Entry.StateMutabilityType;
+
+public final class AbiValidator {
+
+  private static final Pattern INT_TYPE = Pattern.compile("^(u?int)(\\d*)$");
+  private static final Pattern BYTES_N_TYPE = Pattern.compile("^bytes(\\d+)$");
+  private static final Pattern ARRAY_SUFFIX = Pattern.compile("\\[(\\d*)]$");
+
+  private static final Set<String> BASE_TYPES = ImmutableSet.of(
+      "address", "bool", "string", "bytes", "function", "tuple", "trcToken");
+
+  private AbiValidator() {
+  }
+
+  public static void validate(ABI abi) throws ContractValidateException {
+    if (abi == null || abi.getEntrysCount() == 0) {
+      return;
+    }
+
+    int constructorCount = 0;
+    int fallbackCount = 0;
+    int receiveCount = 0;
+
+    for (int i = 0; i < abi.getEntrysCount(); i++) {
+      Entry entry = abi.getEntrys(i);
+      EntryType type = entry.getType();
+
+      if (type == EntryType.UnknownEntryType || type == EntryType.UNRECOGNIZED) {
+        throw new ContractValidateException(
+            String.format("abi entry #%d: unknown entry type", i));
+      }
+
+      switch (type) {
+        case Constructor:
+          constructorCount++;
+          break;
+        case Fallback:
+          fallbackCount++;
+          if (entry.getInputsCount() > 0 || entry.getOutputsCount() > 0) {
+            throw new ContractValidateException(String.format(
+                "abi entry #%d: fallback function must not have inputs or outputs", i));
+          }
+          break;
+        case Receive:
+          receiveCount++;
+          if (entry.getInputsCount() > 0 || entry.getOutputsCount() > 0) {
+            throw new ContractValidateException(String.format(
+                "abi entry #%d: receive function must not have inputs or outputs", i));
+          }
+          if (entry.getStateMutability() != StateMutabilityType.Payable && !entry.getPayable()) {
+            throw new ContractValidateException(String.format(
+                "abi entry #%d: receive function must be payable", i));
+          }
+          break;
+        case Function:
+          if (entry.getName().isEmpty()) {
+            throw new ContractValidateException(String.format(
+                "abi entry #%d: function must have a name", i));
+          }
+          break;
+        case Event:
+          if (entry.getName().isEmpty() && !entry.getAnonymous()) {
+            throw new ContractValidateException(String.format(
+                "abi entry #%d: non-anonymous event must have a name", i));
+          }
+          break;
+        case Error:
+          if (entry.getName().isEmpty()) {
+            throw new ContractValidateException(String.format(
+                "abi entry #%d: error must have a name", i));
+          }
+          break;
+        default:
+          break;
+      }
+
+      validateParams(i, "inputs", entry.getInputsList());
+      validateParams(i, "outputs", entry.getOutputsList());
+    }
+
+    if (constructorCount > 1) {
+      throw new ContractValidateException("abi: only one constructor is allowed");
+    }
+    if (fallbackCount > 1) {
+      throw new ContractValidateException("abi: only one fallback function is allowed");
+    }
+    if (receiveCount > 1) {
+      throw new ContractValidateException("abi: only one receive function is allowed");
+    }
+  }
+
+  private static void validateParams(int entryIdx, String side, List<Param> params)
+      throws ContractValidateException {
+    for (int j = 0; j < params.size(); j++) {
+      String type = params.get(j).getType();
+      String reason = checkType(type);
+      if (reason != null) {
+        throw new ContractValidateException(String.format(
+            "abi entry #%d %s[%d] type '%s': %s", entryIdx, side, j, type, reason));
+      }
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


actuator/src/main/java/org/tron/core/utils/AbiValidator.java
```diff
@@ -0,0 +1,189 @@
...
+      }
+    }
+  }
+
+  // Returns null when the type is acceptable, otherwise a short failure reason.
+  private static String checkType(String raw) {
+    if (raw == null || raw.isEmpty()) {
+      return "type must not be empty";
+    }
+    String t = raw.trim();
```
alan-eth: Really nice helper layout — `checkType` returning either `null` or a short reason makes the call site read very cleanly, and isolating the regex/base-type tables at the top is great for future tweaking. ✨

One subtle gap: `t = raw.trim()` lets a type like `" uint256 "` pass validation, but the validator never normalizes the proto — the on-chain `Param.type` keeps the original whitespace. Per the linked issue's "fewer inconsistencies in stored ABI metadata" goal, would it make sense to reject whitespace-padded types upfront, e.g.:

```java
if (!raw.equals(raw.trim())) {
  return "type must not contain leading/trailing whitespace";
}
```

That way the validator's "this passed" implies "what gets persisted is exactly what got checked", which matches the stated goal of strict-matching tooling compatibility.

framework/src/test/java/org/tron/common/runtime/TvmTestUtils.java
```diff
@@ -489,6 +490,8 @@ private static SmartContract.ABI.Entry.EntryType getEntryType(String type) {
         return SmartContract.ABI.Entry.EntryType.Event;
       case "fallback":
         return SmartContract.ABI.Entry.EntryType.Fallback;
+      case "error":
+        return SmartContract.ABI.Entry.EntryType.Error;
```
alan-eth, line 496: Great call adding `case "error"` here — without it any test ABI carrying an `error` entry would fall into `UNRECOGNIZED` and now hit the new "unknown entry type" rejection. Nice symmetry with `PublicMethod.getEntryType` which already had it. 🎯

Minor follow-up: both this helper and `PublicMethod.getEntryType` still omit `case "receive"`, even though the new validator has dedicated handling for `Receive` (payability + IO checks). Any test that wants to exercise a `receive` entry through these helpers will still get `UNRECOGNIZED` and bounce off the new rejection. Worth adding the one-line case in both files while you're here, so the validator's full surface is reachable from test utilities?

yanghang8612: Good catch. Adding `case "receive"` to both `TvmTestUtils.getEntryType` and `PublicMethod.getEntryType` so the validator's `Receive` handling (payability + IO checks) is actually reachable from test helpers — otherwise any test that exercises a receive entry bounces off "unknown entry type" before it hits the rules the validator is there to enforce. Folds into the same commit as the whitespace fix above.

actuator/src/main/java/org/tron/core/utils/AbiValidator.java
yanghang8612: Thanks @alan-eth — that lands. The whole point of this validator is that "passed validation" should match "what gets persisted"; accepting `" uint256 "` breaks the invariant silently and the proto write never normalizes it back. Adding the upfront whitespace reject is a cleaner contract than threading a `trim()` through the write path.

Going in the next commit.

actuator/src/main/java/org/tron/core/utils/AbiValidator.java
```diff
@@ -0,0 +1,192 @@
...
+  private static final Pattern BYTES_N_TYPE = Pattern.compile("^bytes(\\d+)$");
+  private static final Pattern ARRAY_SUFFIX = Pattern.compile("\\[(\\d*)]$");
+
+  private static final Set<String> BASE_TYPES = ImmutableSet.of(
+      "address", "bool", "string", "bytes", "function", "tuple", "trcToken");
+
+  private AbiValidator() {
+  }
+
+  public static void validate(ABI abi) throws ContractValidateException {
```
CodeNinjaEvan, line 27: Is validate(ABI abi) too strict? It may reject legitimate ABIs generated by older or non-standard compilers.

yanghang8612: Looked into this carefully — checked solc v0.4.0 source (Sep 2016, predates TRON mainnet by ~2 years):

- IntegerType::toString (Types.cpp:323-328) returns `prefix + bits` unconditionally, so ABIs always carry "uint256"/"int128" etc., never bare "uint"/"int".
- FixedBytesType::toString (Types.h:455) returns `"bytes" + N` unconditionally — the `byte` keyword is source-language only, never reaches ABI JSON.
- Fallback gets its own ABI entry (InterfaceHandler.cpp:78-86 in v0.4.0): method["type"] = "fallback". Never emitted as type=function with empty name.

Combined with TRON's solc fork being 0.4.x-based, every contract ever deployed on TRON mainnet has gone through canonical-emitting solc. The strict validator does not reject any solc-produced ABI, modern or historical.

Hand-written ABIs that use shorthand can canonicalize client-side; storing non-canonical types on-chain would actively work against #6674's "fewer inconsistencies in stored ABI metadata" goal.

Keeping the strict rules.

yanghang8612:
Closing without merging — after further discussion we've decided not to pursue this feature. Full rationale (marginal necessity given the <0.3% sampling result + compatibility risk against ABIs from historical solc versions) is in tronprotocol/java-tron#6674. Thanks everyone for the review.

PR #6704, opened 2026-04-24
feat(net): optimize disconnectRandom by tracking block receive time

### LINKED ISSUE #6504: Node fails to synchronize blocks after a peer is randomly disconnected
#### Software Versions
OS : Linux
Version : 4.8.0 

#### Expected behaviour
When the node has established multiple valid peer connections, it should be able to fetch block data from multiple available peers.
Even if a single peer is disconnected (including random disconnection), the node should automatically switch to other already connected peers to continue block synchronization, ensuring continuity and stability of the sync process.

#### Actual behaviour
During runtime, the following abnormal behavior was observed:
- The node successfully establishes approximately 30 P2P connections.
- However, during block synchronization, it appears that the node continuously requests blocks from only a single peer, while the other connections are not effectively utilized.
- When this peer is disconnected by the random elimination strategy, the following log is printed:
```
19:22:47.166 INFO  [peerWorker-5] [net](PeerConnection.java:184) Send peer /IP:50816 message type: P2P_DISCONNECT
reason: RANDOM_ELIMINATION
```

After this connection is terminated, the node does not automatically resume block requests from other already connected peers. The block height stops updating, and the node enters a state where it can no longer continue block synchronization.

Based on logs and observed behavior:
- There appears to be a strong single-peer dependency for block data syncing.
- When the random elimination strategy is triggered, the only actively used peer is disconnected, which raises the question of whether there is an issue with the random elimination strategy.

#### Frequency
This issue is not consistently reproducible and was not clearly observed in earlier runs.

xxo1shine:
**What does this PR do?**

Optimize disconnectRandom by tracking block receive time peer

- Add blockRcvTime/blockRcvTimeCmp fields to PeerConnection to track when a peer last delivered a valid block
- Set blockRcvTime in BlockMsgHandler after each block is received
- Fix lastInteractiveTime update in InventoryMsgHandler: only update for block inventories above current head block num, preventing attackers from forging activity via stale block hashes
- Add getRandomDisconnectionPeers() to ResilienceService: narrows the disconnect candidate pool to the oldest half by blockRcvTime, so peers that recently delivered blocks are protected from random eviction

Fixes https://github.com/tronprotocol/java-tron/issues/6504

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
@@ -6,6 +6,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.tron.common.utils.Sha256Hash;
+import org.tron.core.capsule.BlockCapsule.BlockId;
 import org.tron.core.config.args.Args;
 import org.tron.core.exception.P2pException;
 import org.tron.core.exception.P2pException.TypeEnum;
@@ -44,7 +45,10 @@ public class InventoryMsgHandler implements TronMsgHandler {
       peer.getAdvInvReceive().put(item, System.currentTimeMillis());
       advService.addInv(item);
       if (type.equals(InventoryType.BLOCK) && peer.getAdvInvSpread().getIfPresent(item) == null) {
-        peer.setLastInteractiveTime(System.currentTimeMillis());
+        long headNum = tronNetDelegate.getHeadBlockId().getNum();
+        if (new BlockId(id).getNum() > headNum) {
+          peer.setLastInteractiveTime(System.currentTimeMillis());
+        }
       }
     }
   }
--- a/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
+++ b/framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
@@ -88,6 +88,10 @@ public class PeerConnection {
   @Setter
   private volatile long lastInteractiveTime;
 
+  @Setter
+  @Getter
+  private volatile long blockRcvTime;
+
   @Getter
   @Setter
   private volatile TronState tronState = TronState.INIT;
--- a/framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
+++ b/framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
@@ -3,8 +3,10 @@ package org.tron.core.net.service.effective;
 import static org.tron.common.math.Maths.ceil;
 import static org.tron.common.math.Maths.max;
 
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -44,7 +46,7 @@ public class ResilienceService {
 
   @Autowired
   private ChainBaseManager chainBaseManager;
-
+ 
   public void init() {
     if (Args.getInstance().isOpenFullTcpDisconnect) {
       executor.scheduleWithFixedDelay(() -> {
@@ -86,6 +88,7 @@ public class ResilienceService {
         .collect(Collectors.toList());
 
     if (peers.size() >= minBroadcastPeerSize) {
+      peers = getRandomDisconnectionPeers(peers);
       long now = System.currentTimeMillis();
       Map<Object, Integer> weights = new HashMap<>();
       peers.forEach(peer -> {
@@ -120,6 +123,14 @@ public class ResilienceService {
     }
   }
 
+
+  private List<PeerConnection> getRandomDisconnectionPeers(List<PeerConnection> peers) {
+    Map<PeerConnection, Long> snapshot = new IdentityHashMap<>(peers.size());
+    peers.forEach(p -> snapshot.put(p, p.getBlockRcvTime()));
+    List<PeerConnection> sorted = new ArrayList<>(peers);
+    sorted.sort(Comparator.comparingLong(snapshot::get));
+    return sorted.subList(0, sorted.size() / 2);
+  }
 
   private void disconnectLan() {
     if (!isLanNode()) {
--- a/framework/src/main/java/org/tron/core/net/messagehandler/BlockMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/BlockMsgHandler.java
@@ -150,6 +150,7 @@ public class BlockMsgHandler implements TronMsgHandler {
 
     try {
       tronNetDelegate.processBlock(block, false);
+      peer.setBlockRcvTime(System.currentTimeMillis());
       witnessProductBlockService.validWitnessProductTwoBlock(block);
 
       Item item = new Item(blockId, InventoryType.BLOCK);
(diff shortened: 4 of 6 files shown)


framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
```diff
@@ -88,6 +88,14 @@ public class PeerConnection {
   @Setter
   private volatile long lastInteractiveTime;
 
+  @Setter
+  @Getter
+  private volatile long blockRcvTime;
+
+  @Setter
+  @Getter
+  private volatile long blockRcvTimeCmp;
```
warku123: [QUESTION]Why does `blockRcvTimeCmp` live in `PeerConnection`?

Its only use is inside `ResilienceService.getRandomDisconnectionPeers()` to snapshot `blockRcvTime` before sorting, so the comparator reads a stable value even if another thread updates `blockRcvTime` mid-sort. That's a valid concern, but it's an implementation detail of one method in one service — it doesn't describe any property of the peer itself.

A local snapshot map achieves the same safety without polluting the peer domain object:

```java
private List<PeerConnection> getRandomDisconnectionPeers(List<PeerConnection> peers) {
    Map<PeerConnection, Long> snapshot = new IdentityHashMap<>();
    peers.forEach(p -> snapshot.put(p, p.getBlockRcvTime()));
    List<PeerConnection> sorted = new ArrayList<>(peers);
    sorted.sort(Comparator.comparingLong(snapshot::get));
    return sorted.subList(0, sorted.size() / 2);
}
```

Is there a reason this needs to be a field rather than a method-local variable?

xxo1shine: The reason `blockRcvTimeCmp` was introduced as a field was to solve a concrete problem: `blockRcvTime` is updated concurrently by network threads during the sort, which causes the `Comparator` to observe inconsistent values across comparisons and triggers `IllegalArgumentException: Comparison method violates its general contract` — failing the sort. Making it a field on the peer was the way I picked at the time to give the comparator a stable snapshot value to read.

That said, your direction is right — a stable snapshot is purely an implementation detail of `getRandomDisconnectionPeers` and shouldn't leak onto `PeerConnection`. Switching to a method-local `IdentityHashMap<PeerConnection, Long>` populated once before the sort and read-only during the sort gives the same stability guarantee without polluting the domain model.

xxo1shine: I've optimized it according to your suggestions.

framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
```diff
@@ -40,7 +41,10 @@ public void processMessage(PeerConnection peer, TronMessage msg) {
       peer.getAdvInvReceive().put(item, System.currentTimeMillis());
       advService.addInv(item);
       if (type.equals(InventoryType.BLOCK) && peer.getAdvInvSpread().getIfPresent(item) == null) {
-        peer.setLastInteractiveTime(System.currentTimeMillis());
+        long headNum = tronNetDelegate.getHeadBlockId().getNum();
+        if (new BlockId(id).getNum() > headNum) {
```
halibobo1205, line 49: [QUESTION] Small wording suggestion. The PR description says this prevents "attackers from forging activity via stale block hashes" — the *stale-replay* part is correct and a real improvement, but I don't think arbitrary forgery is prevented:

`BlockId(Sha256Hash)` reinterprets the first 8 bytes of the hash as a big-endian long (BlockCapsule.java:347-352) without verifying the hash matches any real block. An adversary can craft a 32-byte hash whose high bytes encode `headNum + 1` and still pass the check.

The check still has clear value (this is the only path that refreshes `lastInteractiveTime` on INV — `P2pEventHandlerImpl.updateLastInteractiveTime` doesn't include `INVENTORY` in its switch, so without this, tightening any historical block hash refreshes the inactivity clock). Just suggest narrowing the description to "prevent stale block hashes from refreshing lastInteractiveTime" rather than the broader "forging activity" framing — keeps expectations accurate.

framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
```diff
@@ -40,7 +41,10 @@ public void processMessage(PeerConnection peer, TronMessage msg) {
...
+          peer.setLastInteractiveTime(System.currentTimeMillis());
+        }
```
317787106, line 51: [SHOULD] Good comparing. If new BlockId(id).getNum() < solidity Num, we should set the LastInteractiveTime as very old time.

framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
```diff
@@ -121,6 +124,14 @@ private void disconnectRandom() {
   }
 
 
+  private List<PeerConnection> getRandomDisconnectionPeers(List<PeerConnection> peers) {
+    Map<PeerConnection, Long> snapshot = new IdentityHashMap<>(peers.size());
+    peers.forEach(p -> snapshot.put(p, p.getBlockRcvTime()));
+    List<PeerConnection> sorted = new ArrayList<>(peers);
+    sorted.sort(Comparator.comparingLong(snapshot::get));
+    return sorted.subList(0, sorted.size() / 2);
```
317787106, line 132: [SHOULD] There are two problems:

1. Exclude latest peer is enough, not half of peers. 
2. Is there any necessaary to create new Map instead of sorting by peer's blockRcvTime ?

framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
xxo1shine: This modification doesn't completely solve the update problem, but rather increases the cost of malicious activity, preventing attackers from refreshing the time using expired block hashes—an attack with very low cost.

The attack you mentioned does exist, but attackers would need to construct malicious blocks, significantly increasing the cost of such an attack.

framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
xxo1shine: If the time isn't refreshed, the time will fall behind, so the added logic isn't very significant. It's also possible that the other party, due to network latency, is slightly behind and might be mistakenly affected.

framework/src/main/java/org/tron/core/net/service/effective/ResilienceService.java
xxo1shine: The main purpose is to increase randomness; things that are too certain are more vulnerable to attack. If not sorted by blockRcvTime, then what should be used for sorting?

framework/src/main/java/org/tron/core/net/messagehandler/InventoryMsgHandler.java
317787106: Let's improve it, suppose if new BlockId(id).getNum() < solidityNum - THRESHOLD, it takes into account the impact of network latency under extreme conditions. Since you’ve modified this part, it’s recommended to imporove the logic here.

xxo1shine: I think this is redundant logic. Besides increasing complexity, it doesn't provide any defensive capability. Previously, malicious nodes would do this because broadcasting blocks below the solidified block level could refresh the time. Now that refreshing is no longer possible, and there's no longer a list of blocks below the solidified block level being broadcast.

halibobo1205 (approved):

lvs0075 (approved):
The review has been completed in full, and no significant issues have been identified.

317787106 (approved):

PR #6709, opened 2026-04-27
fix(jsonrpc): correct TransactionResult.nonce per JSON-RPC spec

### LINKED ISSUE #6547: The `nonce` returned by `eth_getTransactionByHash` is non-compliant
It should be 0x0, not 0x0000000000000000.

Match pattern: ^0x(0|[1-9a-f][0-9a-f]*)$

waynercheung:
## What this PR does

Corrects the `nonce` field in `TransactionResult`, the response body shared by `eth_getTransactionByHash`, `eth_getTransactionByBlockHashAndIndex`, `eth_getTransactionByBlockNumberAndIndex`, and the full-tx mode of `eth_getBlockByHash` / `eth_getBlockByNumber`.

Closes #6547.

## Why

Per [`ethereum/execution-apis`](https://github.com/ethereum/execution-apis/tree/main/src/schemas):

- `TransactionInfo.nonce` is `uint` (QUANTITY), pattern `^0x(0|[1-9a-f][0-9a-f]*)$` - no leading zeros; zero is `0x0`.

java-tron's `TransactionResult` previously emitted `ByteArray.toJsonHex(new byte[8])` -> `"0x0000000000000000"` (16 hex chars), which violates the pattern.

## What changes

| Field | Path | Before | After |
|-------|------|--------|-------|
| `nonce` | both `TransactionResult` constructors | `0x0000000000000000` | `0x0` |

## Why `Block.nonce` is intentionally NOT changed

`Block.nonce` is `bytes8` per the spec (pattern `^0x[0-9a-f]{16}$`), so the existing `0x0000000000000000` is already compliant. Shortening it to `0x0` would break conformance with the bytes8 length requirement.

`BlockResult.nonce` is therefore left untouched and existing `JsonrpcServiceTest` assertions on it (`"0x0000000000000000"`) remain valid.

## Compatibility

This PR makes a minimal, surgical change: only the `nonce` literal value is modified. All other fields in `TransactionResult` retain their existing serialization, preserving byte-for-byte compatibility with current production responses.

The numeric semantics of `nonce` are unchanged (`Long.parseLong` of either value yields `0`); only the string representation tightens to the spec-compliant form.

## Tests

`TransactionResultTest` is updated to:

- Assert `nonce == "0x0"` in both test methods (covering both `TransactionResult` constructors).
- Validate `nonce` against the schema regex `^0x(0|[1-9a-f][0-9a-f]*)$` via a small `assertQuantity` helper, so any future regression that re-introduces a leading-zero form will fail the test.

## Pre-submit checklist

- [x] Single module (jsonrpc), single issue (#6547), 2 files changed
- [x] Code style follows Google Java Style
- [x] No debug code / temporary comments / stray TODOs
- [x] No unsafe numeric ops introduced
- [x] No log-level changes
- [x] No DB schema / consensus / config changes
- [x] No dependency upgrades
- [x] Comments explain WHY (the spec rationale), not WHAT

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
@@ -98,7 +98,7 @@ public class TransactionResult {
     TransactionCapsule capsule = new TransactionCapsule(tx);
     byte[] txId = capsule.getTransactionId().getBytes();
     hash = ByteArray.toJsonHex(txId);
-    nonce = ByteArray.toJsonHex(new byte[8]); // no value
+    nonce = "0x0"; // no value, QUANTITY type per Ethereum JSON-RPC spec
     blockHash = ByteArray.toJsonHex(blockCapsule.getBlockId().getBytes());
     blockNumber = ByteArray.toJsonHex(blockCapsule.getNum());
     transactionIndex = ByteArray.toJsonHex(index);
@@ -133,7 +133,7 @@ public class TransactionResult {
     TransactionCapsule capsule = new TransactionCapsule(tx);
     byte[] txId = capsule.getTransactionId().getBytes();
     hash = ByteArray.toJsonHex(txId);
-    nonce = ByteArray.toJsonHex(new byte[8]); // no value
+    nonce = "0x0"; // no value, QUANTITY type per Ethereum JSON-RPC spec
     blockHash = "0x";
     blockNumber = "0x";
     transactionIndex = "0x";
--- a/framework/src/test/java/org/tron/core/services/jsonrpc/TransactionResultTest.java
+++ b/framework/src/test/java/org/tron/core/services/jsonrpc/TransactionResultTest.java
@@ -23,8 +23,16 @@ public class TransactionResultTest extends BaseTest {
   private static final String OWNER_ADDRESS = "41548794500882809695a8a687866e76d4271a1abc";
   private static final String CONTRACT_ADDRESS = "A0B4750E2CD76E19DCA331BF5D089B71C3C2798548";
 
+  // QUANTITY pattern from ethereum/execution-apis base-types schema (uint).
+  private static final String QUANTITY_PATTERN = "^0x(0|[1-9a-f][0-9a-f]*)$";
+
   static {
     Args.setParam(new String[] {"-d", dbPath()}, TestConstants.TEST_CONF);
+  }
+
+  private static void assertQuantity(String value) {
+    Assert.assertNotNull(value);
+    Assert.assertTrue("not a valid QUANTITY: " + value, value.matches(QUANTITY_PATTERN));
   }
 
   @Test
@@ -49,6 +57,8 @@ public class TransactionResultTest extends BaseTest {
         transactionResult.getHash());
     Assert.assertEquals(transactionResult.getGasPrice(), "0x1");
     Assert.assertEquals(transactionResult.getGas(), "0x64");
+    Assert.assertEquals("0x0", transactionResult.getNonce());
+    assertQuantity(transactionResult.getNonce());
   }
 
   @Test
@@ -65,7 +75,8 @@ public class TransactionResultTest extends BaseTest {
     Assert.assertEquals("0x5691531881bc44adbc722060d85fdf29265823db8e884b0d104fcfbba253cf11",
         transactionResult.getHash());
     Assert.assertEquals(transactionResult.getGasPrice(), "0x");
-    Assert.assertEquals(transactionResult.getNonce(), "0x0000000000000000");
+    Assert.assertEquals("0x0", transactionResult.getNonce());
+    assertQuantity(transactionResult.getNonce());
   }
 
 }


framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
```diff
@@ -58,9 +58,10 @@ public class TransactionResult {
   private void parseSignature(Transaction tx) {
 
     if (tx.getSignatureCount() == 0) {
-      v = ByteArray.toJsonHex(new byte[1]);
-      r = ByteArray.toJsonHex(new byte[32]);
-      s = ByteArray.toJsonHex(new byte[32]);
+      // v, r, s are QUANTITY per Ethereum JSON-RPC spec; "0x0" matches uint pattern.
+      v = "0x0";
+      r = "0x0";
+      s = "0x0";
```
bladehan1: [DISCUSS] Signed-path r/s may still violate uint pattern when high byte is 0x00

For signed transactions, `r` and `s` come from `Rsv.fromSignature(...)` as fixed 32-byte arrays and are serialised in the branch right below this via `ByteArray.toJsonHex(byte[])`, which preserves leading zeros. When the high byte of `r` (or `s`) happens to be `0x00` — roughly a 1/256 probability per byte — the output becomes `"0x00ab..."`, which does NOT match `^0x(0|[1-9a-f][0-9a-f]*)$` and will be rejected by the same strict clients this PR aims to unblock.

The PR description scopes the fix to no-value paths and states production responses are byte-identical, which is true. But the headline statement "strict clients can now parse the response" is slightly stronger than what is actually delivered: for live transactions, every signature still has a non-trivial chance of producing a non-compliant `r`/`s`.

Suggestion: Add a "Known limitation" paragraph in the PR description and open a follow-up issue covering the signed path; longer-term, an API-layer `toQuantity(byte[])` helper that strips leading zeros (and returns `"0x0"` for empty/all-zero) would close the gap without leaking spec policy into `common/utils/ByteArray`.

framework/src/test/java/org/tron/core/services/jsonrpc/TransactionResultTest.java
```diff
@@ -64,8 +89,22 @@ public void testBuildTransactionResult() {
...
+    Assert.assertEquals("0x0", transactionResult.getNonce());
+    Assert.assertEquals("0x0", transactionResult.getGasPrice());
+    Assert.assertEquals("0x0", transactionResult.getV());
+    Assert.assertEquals("0x0", transactionResult.getR());
+    Assert.assertEquals("0x0", transactionResult.getS());
+    assertQuantity(transactionResult.getNonce());
+    assertQuantity(transactionResult.getGasPrice());
+    assertQuantity(transactionResult.getV());
+    assertQuantity(transactionResult.getR());
+    assertQuantity(transactionResult.getS());
```
bladehan1: [NIT] Tighten test sentinels — 2 small follow-ups on TransactionResultTest

Two small gaps in the new assertions, both in the fallback-path test (around lines 94-107):

1. **No end-to-end null serialisation assertion.** The test asserts `Assert.assertNull(transactionResult.getBlockHash())` etc. on the Java object, but never serialises the DTO. `TransactionResult` has no `@JsonInclude(JsonInclude.Include.NON_NULL)`, so by default these fields render as `"blockHash": null` in JSON, which matches spec nullability. If a future PR adds `@JsonInclude(NON_NULL)` (class-level or via a global `ObjectMapper` config), the fields silently disappear from the wire — and strict clients distinguish "null" from "missing". Adding one `ObjectMapper.writeValueAsString(...)` assertion that the produced JSON contains `"blockHash":null`, `"blockNumber":null`, `"transactionIndex":null` would lock the wire contract.

2. **`gas` and `value` not covered by `assertQuantity`.** The fallback path sets `gas = "0x0"` and `value = "0x0"` when the contract list is empty. The new sentinels cover `nonce`/`gasPrice`/`v`/`r`/`s` but skip `gas` and `value`, leaving a hole: a future change setting `gas = ByteArray.toJsonHex(new byte[8])` would not be caught. The other test (`testBuildTransactionResultWithBlock`) has the same gap for `value`.

Suggestion: Add `assertQuantity(getGas())` and `assertQuantity(getValue())` to both test methods, and add one JSON-serialisation assertion on the fallback path.

framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
```diff
@@ -58,9 +58,10 @@ public class TransactionResult {
   private void parseSignature(Transaction tx) {
 
     if (tx.getSignatureCount() == 0) {
-      v = ByteArray.toJsonHex(new byte[1]);
-      r = ByteArray.toJsonHex(new byte[32]);
-      s = ByteArray.toJsonHex(new byte[32]);
+      // v, r, s are QUANTITY per Ethereum JSON-RPC spec; "0x0" matches uint pattern.
+      v = "0x0";
```
bladehan1: [NIT] Consider a QUANTITY_ZERO constant for the literal repeated 5 times

The literal `"0x0"` now appears in five places in this class (parseSignature: v/r/s; main constructor: nonce; fallback constructor: nonce, gasPrice — gas was already `"0x0"`). They all carry the same semantic — "spec-compliant uint zero". When the next ETH-shaped field is added (e.g. EIP-1559 `maxFeePerGas` / `maxPriorityFeePerGas` placeholders), it is easy to copy the wrong literal — `"0x"` is the empty-bytes value for DATA fields, and the two must not be conflated.

Suggestion: Introduce `private static final String QUANTITY_ZERO = "0x0";` (or a tiny helper in `JsonRpcApiUtil`) and reuse it. Pure cleanup; no behaviour change.

framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
waynercheung: Documented as a "Known limitation" section in the PR description, with the geth-equivalent fix path and reasoning for why the rewrite is out of scope for this PR. Will track as a follow-up issue.

framework/src/test/java/org/tron/core/services/jsonrpc/TransactionResultTest.java
waynercheung: Both points accepted:

(1) Adding `assertQuantity(getValue())` to both test methods, plus `assertQuantity(getGas())` to `testBuildTransactionResult` (the WithBlock variant already covers gas).

(2) Adding an `ObjectMapper.writeValueAsString(...)` assertion that pins the `"blockHash":null` / `"blockNumber":null` / `"transactionIndex":null` wire-level form, so the contract is locked even if a future `@JsonInclude(NON_NULL)` change tries to omit those fields.

Will push the follow-up commit shortly.

framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
waynercheung: Good catch - accepted. Will introduce `private static final String QUANTITY_ZERO = "0x0";` and replace the 6 literals introduced in this PR. Will also opportunistically replace the pre-existing `"0x0"` literals (type, gas, value) so the whole class is consistent. Coming in a follow-up commit on this branch.

waynercheung:
After review discussion, the scope of this PR has been narrowed to only the `nonce` field - the change directly addressing #6547. All other fields previously modified in earlier commits (`v` / `r` / `s` in the unsigned-signature branch, `blockHash` / `blockNumber` / `transactionIndex` / `gasPrice` in the fallback constructor) have been reverted to their original behavior to maintain strict byte-level compatibility with current production responses.

The branch has been force-pushed to a single minimal commit. Earlier inline comments on the now-removed changes are no longer applicable; please re-review the latest state.

halibobo1205 (approved):
LGTM

framework/src/main/java/org/tron/core/services/jsonrpc/types/TransactionResult.java
```diff
@@ -98,7 +102,7 @@ public TransactionResult(BlockCapsule blockCapsule, int index, Protocol.Transact
     TransactionCapsule capsule = new TransactionCapsule(tx);
     byte[] txId = capsule.getTransactionId().getBytes();
     hash = ByteArray.toJsonHex(txId);
-    nonce = ByteArray.toJsonHex(new byte[8]); // no value
+    nonce = QUANTITY_ZERO; // no value
```
0xbigapple, line 105: [NIT] Drop or rewrite the stale `// no value` comment on `nonce`
The `// no value` comment on both `nonce = QUANTITY_ZERO` lines describes the old `new byte[8]` placeholder. After this PR the value is a real spec-compliant zero, not a placeholder, so the comment is misleading.
  Suggestion: drop it, or replace with `// TRON has no Ethereum nonce, always 0`.

yanghang8612 (approved):
LGTM

bladehan1 (approved):
LGTM

0xbigapple (approved):
LGTM

PR #6711, opened 2026-04-28
fix(jsonrpc): harden ABI parser bounds and revert reason decoding

0xbigapple:
**What does this PR do?**

This PR hardens ABI decoding for malformed dynamic data in both contract event parsing and JSON-RPC revert reason handling.

It:
- in `subBytes`: rejects negative or out-of-bounds offsets/lengths (no longer silently truncates oversized lengths via `min`)
- in `parseDataBytes`: rejects negative dynamic lengths instead of treating them as empty `string` / `bytes`
- adds `tryDecodeRevertReason` to centralize duplicated revert error handling used by `eth_call` and `eth_estimateGas`
- adds focused tests for parser bounds checks and revert reason decoding

Previously, malformed ABI payloads could:
- reach `subBytes()` with invalid offsets or oversized lengths
- treat negative dynamic lengths as empty `string` / `bytes`
- affect event handling and revert reason decoding in `eth_call` and `eth_estimateGas`

- Unit Tests
  - `:framework:test --tests org.tron.common.logsfilter.EventParserTest`
  - `:framework:test --tests org.tron.core.services.jsonrpc.TronJsonRpcRevertReasonTest`
  - `:framework:test --tests org.tron.core.jsonrpc.JsonRpcCallAndEstimateGasTest` — end-to-end coverage of `eth_call` and `eth_estimateGas` over revert/non-revert` paths
  
- Manual Testing
  - deployed a test contract to a local private chain
  - verified malformed revert payloads through `eth_call`
  - verified malformed revert payloads through `eth_estimateGas`
  - verified malformed event payloads through emitted logs

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
@@ -169,6 +169,8 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
   private static final String NO_BLOCK_HEADER_BY_HASH = "header for hash not found";
 
   private static final String ERROR_SELECTOR = "08c379a0"; // Function selector for Error(string)
+  private static final int REVERT_REASON_SELECTOR_LENGTH = 4;
+  private static final int MAX_REVERT_REASON_PAYLOAD_BYTES = 4096;
   /**
    * thread pool of query section bloom store
    */
@@ -470,6 +472,36 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
   }
 
   /**
+   * Decodes an Error(string) revert reason when possible.
+   * Returns ": reason" on success, otherwise "".
+   */
+  static String tryDecodeRevertReason(byte[] resData) {
+    if (resData == null || resData.length <= REVERT_REASON_SELECTOR_LENGTH) {
+      return "";
+    }
+    if (!Hex.toHexString(resData, 0, REVERT_REASON_SELECTOR_LENGTH).equals(ERROR_SELECTOR)) {
+      return "";
+    }
+
+    int revertPayloadLength = resData.length - REVERT_REASON_SELECTOR_LENGTH;
+    if (revertPayloadLength > MAX_REVERT_REASON_PAYLOAD_BYTES) {
+      logger.debug("skip parsing oversized revert reason payload: {} bytes", revertPayloadLength);
+      return "";
+    }
+
+    try {
+      String reason = ContractEventParser.parseDataBytes(
+          org.bouncycastle.util.Arrays.copyOfRange(resData, REVERT_REASON_SELECTOR_LENGTH,
+              resData.length),
+          "string", 0);
+      return ": " + reason;
+    } catch (Exception e) {
+      logger.debug("parse revert reason failed", e);
+      return "";
+    }
+  }
+
+  /**
    * @param data Hash of the method signature and encoded parameters. for example:
    * getMethodSign(methodName(uint256,uint256)) || data1 || data2
    */
@@ -512,14 +544,8 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
       }
       result = ByteArray.toJsonHex(listBytes);
     } else {
-      String errMsg = retBuilder.getMessage().toStringUtf8();
       byte[] resData = trxExtBuilder.getConstantResult(0).toByteArray();
-      if (resData.length > 4 && Hex.toHexString(resData).startsWith(ERROR_SELECTOR)) {
-        String msg = ContractEventParser
-            .parseDataBytes(org.bouncycastle.util.Arrays.copyOfRange(resData, 4, resData.length),
-                "string", 0);
-        errMsg += ": " + msg;
-      }
+      String errMsg = retBuilder.getMessage().toStringUtf8() + tryDecodeRevertReason(resData);
 
       if (resData.length > 0) {
         throw new JsonRpcInternalException(errMsg, ByteArray.toJsonHex(resData));
@@ -677,15 +703,8 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
     }
 
     if (trxExtBuilder.getTransaction().getRet(0).getRet().equals(code.FAILED)) {
-      String errMsg = retBuilder.getMessage().toStringUtf8();
-
       byte[] data = trxExtBuilder.getConstantResult(0).toByteArray();
-      if (data.length > 4 && Hex.toHexString(data).startsWith(ERROR_SELECTOR)) {
-        String msg = ContractEventParser
-            .parseDataBytes(org.bouncycastle.util.Arrays.copyOfRange(data, 4, data.length),
-                "string", 0);
-        errMsg += ": " + msg;
-      }
+      String errMsg = retBuilder.getMessage().toStringUtf8() + tryDecodeRevertReason(data);
 
       if (data.length > 0) {
         throw new JsonRpcInternalException(errMsg, ByteArray.toJsonHex(data));
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
@@ -491,7 +491,7 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
 
     try {
       String reason = ContractEventParser.parseDataBytes(
-          org.bouncycastle.util.Arrays.copyOfRange(resData, REVERT_REASON_SELECTOR_LENGTH,
+          Arrays.copyOfRange(resData, REVERT_REASON_SELECTOR_LENGTH,
               resData.length),
           "string", 0);
       return reason.isEmpty() ? "" : ": " + reason;
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
@@ -495,7 +495,7 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
               resData.length),
           "string", 0);
       return reason.isEmpty() ? "" : ": " + reason;
-    } catch (Exception e) {
+    } catch (RuntimeException e) {
       logger.debug("parse revert reason failed", e);
       return "";
     }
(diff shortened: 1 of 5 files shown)


framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -469,6 +471,36 @@ private void estimateEnergy(byte[] ownerAddressByte, byte[] contractAddressByte,
...
+      logger.debug("skip parsing oversized revert reason payload: {} bytes", revertPayloadLength);
+      return "";
+    }
+
+    try {
+      String reason = ContractEventParser.parseDataBytes(
+          org.bouncycastle.util.Arrays.copyOfRange(resData, REVERT_REASON_SELECTOR_LENGTH,
+              resData.length),
+          "string", 0);
+      return ": " + reason;
```
alan-eth: Really nice job pulling this helper out — the dedup between `eth_call` and `eth_estimateGas` is much cleaner now, and centralizing the size guard + selector check in one place is a great win. 🎯

One small edge case worth thinking about: when a contract reverts with a legally-encoded *empty* string (e.g. Solidity `require(cond, "")`, which yields `08c379a0 | offset=0x20 | length=0`), `parseDataBytes` returns `""`, and this helper returns `": "`. The caller then surfaces the JSON-RPC error message as `"REVERT opcode executed: "` with a trailing colon and space. Not a correctness bug, just a small cosmetic wart in the error string.

Maybe something like:
```java
return reason.isEmpty() ? "" : ": " + reason;
```
Up to you — feel free to drop if you'd rather keep the helper minimal.

0xbigapple: Good catch — applied. Also added `testTryDecodeRevertReasonWithEmptyString` to pin the behavior.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -469,6 +471,36 @@ private void estimateEnergy(byte[] ownerAddressByte, byte[] contractAddressByte,
...
+      return "";
+    }
+
+    try {
+      String reason = ContractEventParser.parseDataBytes(
+          org.bouncycastle.util.Arrays.copyOfRange(resData, REVERT_REASON_SELECTOR_LENGTH,
+              resData.length),
+          "string", 0);
+      return reason.isEmpty() ? "" : ": " + reason;
+    } catch (Exception e) {
```
bladehan1: [SHOULD] Narrow `catch (Exception)` to `RuntimeException` (or named types)

`tryDecodeRevertReason` currently swallows every `Exception` and returns `""`. The actual call into `parseDataBytes` only throws `UnsupportedOperationException`; `Hex.toHexString(resData, 0, 4)` is guarded by the earlier `length <= REVERT_REASON_SELECTOR_LENGTH` check, and `Arrays.copyOfRange(resData, 4, resData.length)` is in-range. Catching `Exception` makes the helper appear to never fail, so any future regression in `parseDataBytes` (e.g. a new unchecked exception, an NPE, or an assertion error) is hidden behind a debug log that is invisible at production INFO level.

Suggestion: tighten the catch to `RuntimeException` or, better, the specific exception(s) you actually expect (`UnsupportedOperationException`).

framework/src/test/java/org/tron/core/jsonrpc/JsonRpcCallAndEstimateGasTest.java
```diff
@@ -0,0 +1,299 @@
...
+
+  @Test
+  public void testGetCallAppendsRevertReason() throws Exception {
+    byte[] revertData = ByteArray.fromHexString(ERROR_REVERT_HEX);
+
+    mockRpc = newRpcWithMockedFailedCall(revertData, EstimatePath.CONSTANT_CALL);
+
+    try {
+      mockRpc.getCall(newCallArgs(), "latest");
+      Assert.fail("expected JsonRpcInternalException");
```
bladehan1: [NIT] Use `assertThrows` instead of try/fail/catch for expected-exception tests

All 7 negative-path tests in this class follow the `try { ... ; fail(...); } catch (JsonRpcInternalException e) { Assert.assertEquals(...); }` pattern. The codebase already standardises on JUnit 4.13's `assertThrows` for this case (see `WalletMockTest`, `BandwidthProcessorTest`). The current pattern silently passes if `fail(...)` is forgotten, and is more verbose at every call site.

Suggestion: switch to `JsonRpcInternalException e = assertThrows(JsonRpcInternalException.class, () -> mockRpc.getCall(...));` and then assert on `e.getMessage()`.

framework/src/test/java/org/tron/core/jsonrpc/JsonRpcCallAndEstimateGasTest.java
```diff
@@ -0,0 +1,299 @@
...
+    CONSTANT_CALL,
+    ESTIMATE_ENERGY
+  }
+
+  private final boolean originalEstimateEnergy = CommonParameter.getInstance().isEstimateEnergy();
+  private TronJsonRpcImpl mockRpc;
+
+  @After
+  public void tearDown() throws Exception {
+    if (mockRpc != null) {
```
bladehan1, line 47: [NIT] `Mockito.clearAllCaches()` in tearDown is unnecessary

JUnit 4 instantiates a fresh test class per `@Test`, and each test creates new mocks via `mock(...)`. `mockRpc.close()` already releases the executor. `Mockito.clearAllCaches()` resets process-wide mock-class caches and is intended for memory-pressure scenarios, not per-test isolation; here it adds no isolation but does add hidden global state effects.

Suggestion: drop the call.

framework/src/test/java/org/tron/common/logsfilter/EventParserTest.java
```diff
@@ -101,6 +102,81 @@ public synchronized void testEventParser() {
...
+  }
+
+  @Test
+  public void testParseDataBytesNonEmptyString() {
+    // "hello world" is 11 ASCII bytes (68656c6c6f20776f726c64), padded to 32 bytes.
+    byte[] data = ByteArray.fromHexString(
+        "0000000000000000000000000000000000000000000000000000000000000020"
+            + "000000000000000000000000000000000000000000000000000000000000000b"
+            + "68656c6c6f20776f726c64000000000000000000000000000000000000000000");
+    Assert.assertEquals("hello world", ContractEventParser.parseDataBytes(data, "string", 0));
```
bladehan1, line 177: [NIT] Add a multi-byte UTF-8 case to lock in the charset fix

The charset change `new String(realBytes)` → `new String(realBytes, StandardCharsets.UTF_8)` is the main user-observable fix in `parseDataBytes`, but the only string test (`testParseDataBytesNonEmptyString`) uses ASCII `"hello world"`, which round-trips identically under any default charset. CI on Linux defaults to UTF-8, so the test cannot distinguish before/after.

Suggestion: add a case that ABI-encodes a multi-byte UTF-8 string (e.g. `"中文"` = `e4b8ad e69687`, length 6) and asserts the decoded value equals the original. This guards the fix against accidental future regressions on non-UTF-8 default JVMs.

framework/src/test/java/org/tron/core/services/jsonrpc/TronJsonRpcRevertReasonTest.java
```diff
@@ -0,0 +1,97 @@
...
+    Assert.assertEquals("", TronJsonRpcImpl.tryDecodeRevertReason(resData));
+  }
+
+  @Test
+  public void testTryDecodeRevertReasonAtPayloadLimit() {
+    // selector(4) + payload(4096) exactly at the limit: must go through parse, not size-reject.
+    byte[] resData = new byte[4100];
+    resData[0] = 0x08;
+    resData[1] = (byte) 0xc3;
+    resData[2] = 0x79;
```
bladehan1, line 86: [NIT] Cover the legal empty revert string case

The helper handles many malformed inputs but does not test the legal empty-reason case: `08c379a0 | offset=0x20 | length=0`, which is exactly what Solidity emits for `require(cond, "")`. This is a valid wire-level encoding and worth pinning down behavior on (today the helper returns `": "` — see the existing reviewer note on `TronJsonRpcImpl.java:497`).

Suggestion: add a `testTryDecodeRevertReasonWithEmptyReason` case that constructs the empty-string payload and asserts the chosen behavior (whatever you settle on for the trailing `": "` cosmetic).

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
0xbigapple: Good point. Tightened to `RuntimeException`. Because `parseDataBytes` has no `throws` clause declaring it only escapes `UnsupportedOperationException` — that's just an implicit contract of the current implementation. Future changes may throw other exceptions and leave the fallback gone.

framework/src/test/java/org/tron/core/jsonrpc/JsonRpcCallAndEstimateGasTest.java
0xbigapple: Removed.

framework/src/test/java/org/tron/common/logsfilter/EventParserTest.java
0xbigapple: Thanks, this is a fair point. Added `testParseDataBytesMultiByteUtf8String`.

framework/src/test/java/org/tron/core/jsonrpc/JsonRpcCallAndEstimateGasTest.java
0xbigapple: Thanks, good catch. Converted all 7 negative-path tests to
`JsonRpcInternalException e = assertThrows(JsonRpcInternalException.class, () -> …);`
followed by `Assert.assertEquals(...)` on `e.getMessage()`.

framework/src/test/java/org/tron/core/services/jsonrpc/TronJsonRpcRevertReasonTest.java
0xbigapple: Thanks for the suggestion — this case has actually been handled in an earlier thread on this PR. https://github.com/tronprotocol/java-tron/pull/6711#discussion_r3152582572  @alan-eth flagged that `require(cond, "")` was producing a trailing `": "` in the JSON-RPC error message, and we agreed to short-circuit empty reasons to `""` (commit `df05b161a`). The corresponding test `testTryDecodeRevertReasonWithEmptyString` was added in the same commit.

bladehan1 (approved):

317787106 (approved):

PR #6712, opened 2026-04-28
feat(net): add P2P message deduplication and length validation

### LINKED ISSUE #6667: Add deduplication and length checks for p2p messages
# Summary

TRON nodes currently do not perform duplicate or length checks for certain lists during the reception of inventory requests, transaction lists, and blockchain synchronization messages. By adding duplicate detection and length constraints at the message reception stage, the node can manage processing more accurately and improve predictability of resource usage.

# Root Cause
- `FetchInvDataMessage` may contain duplicate entries, but the system does not deduplicate at the reception stage.
- `TransactionMessage` may include duplicate transactions, and the current parsing stage does not check for duplicates, relying instead on caching to skip processing.
- The block summary list in `SyncBlockChainMessage` has no length limit, and the receiving node does not constrain the list size.

# Reproduction

- Start a node and connect it to the P2P network.
- Construct messages containing duplicate inventory entries, duplicate transactions, or an excessively long block summary list.
- Send the messages to the node and observe the reception and processing behavior.
- The node continues processing without interception at the reception stage.

# Impact
- Message processing statistics are not entirely accurate: duplicates are still parsed on first reception.
- Node resource management is slightly affected: parsing and caching operations may incur extra overhead when handling many duplicates or long lists.
- Network data handling efficiency may fluctuate slightly, but overall node availability is not impacted.

# Suggested Fix
- Add validation logic at the message reception stage:
  - Perform duplicate entry checks for FetchInvDataMessage and TransactionMessage lists.
  - Enforce a length limit on the block summary list in SyncBlockChainMessage.
- If validation fails, the node can terminate processing early and log the event, improving processing efficiency.

xxo1shine:
**What does this PR do?**

Add P2P message deduplication and length validation:

- FetchInvDataMsgHandler: reject messages with duplicate hashes
- TransactionsMsgHandler: reject messages with duplicate transactions
- SyncBlockChainMsgHandler: reject blockIds list exceeding 30 entries
- Add MAX_SYNC_CHAIN_IDS = 30 constant to NetConstants
- Add unit tests covering duplicate rejection and boundary values

All violations throw `P2pException(BAD_MESSAGE)`, triggering peer disconnect via existing `P2pEventHandlerImpl` error path.

Fixes https://github.com/tronprotocol/java-tron/issues/6667

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/common/src/main/java/org/tron/core/config/Parameter.java
+++ b/common/src/main/java/org/tron/core/config/Parameter.java
@@ -102,6 +102,7 @@ public class Parameter {
     public static final int MSG_CACHE_DURATION_IN_BLOCKS = 5;
     public static final int MAX_BLOCK_FETCH_PER_PEER = 100;
     public static final int MAX_TRX_FETCH_PER_PEER = 1000;
+    public static final int MAX_SYNC_CHAIN_IDS = 30;
   }
 
   public class DatabaseConstants {
--- a/framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
@@ -3,6 +3,7 @@ package org.tron.core.net.messagehandler;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Lists;
+import java.util.HashSet;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
@@ -153,6 +154,12 @@ public class FetchInvDataMsgHandler implements TronMsgHandler {
 
   private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg,
                      boolean isAdv) throws P2pException {
+    List<Sha256Hash> hashList = fetchInvDataMsg.getHashList();
+    if (hashList.size() != new HashSet<>(hashList).size()) {
+      throw new P2pException(TypeEnum.BAD_MESSAGE,
+          "FetchInvData contains duplicate hashes, size: " + hashList.size());
+    }
+
     MessageTypes type = fetchInvDataMsg.getInvMessageType();
 
     if (type == MessageTypes.TRX) {
--- a/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
+++ b/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
@@ -398,9 +398,9 @@ public class VMActuator implements Actuator2 {
       byte[] ops = newSmartContract.getBytecode().toByteArray();
       rootInternalTx = new InternalTransaction(trx, trxType);
 
-      long maxCpuTimeOfOneTx = rootRepository.getDynamicPropertiesStore()
-          .getMaxCpuTimeOfOneTx() * VMConstant.ONE_THOUSAND;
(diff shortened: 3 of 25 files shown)


common/src/main/java/org/tron/core/config/Parameter.java
```diff
@@ -102,6 +102,7 @@ public class NetConstants {
     public static final int MSG_CACHE_DURATION_IN_BLOCKS = 5;
     public static final int MAX_BLOCK_FETCH_PER_PEER = 100;
     public static final int MAX_TRX_FETCH_PER_PEER = 1000;
+    public static final int MAX_SYNC_CHAIN_IDS = 30;
```
alan-eth, line 105: Nice to have a named constant here instead of a magic number! 👍

Quick question — how was `30` chosen? Observed from real traffic, or estimated from `getBlockChainSummary()`'s log-step algorithm?

lvs0075 (approved):

waynercheung (approved):
LGTM

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
```diff
@@ -153,6 +154,12 @@ public boolean isAdvInv(PeerConnection peer, FetchInvDataMessage msg) {
 
   private void check(PeerConnection peer, FetchInvDataMessage fetchInvDataMsg,
                      boolean isAdv) throws P2pException {
+    List<Sha256Hash> hashList = fetchInvDataMsg.getHashList();
+    if (hashList.size() != new HashSet<>(hashList).size()) {
+      throw new P2pException(TypeEnum.BAD_MESSAGE,
+          "FetchInvData contains duplicate hashes, size: " + hashList.size());
+    }
```
317787106, line 161: Good check. Validation has already been added for `FetchInvDataMessage`, `SyncBlockChainMessage`, and `TransactionsMessage`. Why not also add deduplication checks for `InventoryMessage` and `ChainInventoryMessage`?

common/src/main/java/org/tron/core/config/Parameter.java
xxo1shine: Based on the chain digest algorithm, using a binary search from the solidified block to the head block, 30 blocks can generate 2^30 block digests, which is theoretically large enough, and 30 blocks will not put pressure on the system.

framework/src/main/java/org/tron/core/net/messagehandler/FetchInvDataMsgHandler.java
xxo1shine: Inventory Messages have rate limits and perform deduplication during processing, and they don't consume CPU, so there's no need to perform a second check before processing. ChainInventoryMessages have stricter checks than deduplication, while also ensuring that there are no duplicate lists.

317787106: ChainInventoryMessages's block number's order must be increasing, we can ignore it. But although inventory messages don't need consume CPU, deduplication checks are still needed to keep consistency with other message handling logic and to defend against malicious peers.

xxo1shine: InventoryMessages does not constitute an attack, and the lack of duplicate checks has no impact on the system.

317787106 (approved):

PR #6714, opened 2026-04-28
feat(net): optimize transaction rate limiting

### LINKED ISSUE #6684: Optimize `isBusy` logic by incorporating mempool transactions and enabling configurable `MAX_TRX_SIZE`
# Summary

TRON nodes currently determine whether they are in a busy state based only on the size of the transaction processing thread pool queues, without considering the number of transactions already cached in the mempool.
By incorporating both the thread pool queues and the mempool transaction count into the evaluation, the node can more accurately reflect its current load. This enables better control over transaction intake and improves the stability of resource management.
In addition, it is recommended that `MAX_TRX_SIZE` be configurable, allowing flexible adjustment based on the node's deployment environment.

# Root Cause

The current `isBusy()` logic is as follows:
```
public boolean isBusy() {
   return queue.size() + smartContractQueue.size() > MAX_TRX_SIZE;
}
```
This logic only accounts for:
- The standard transaction processing queue (queue)
- The smart contract processing queue (smartContractQueue)

It does not include transactions already stored in the mempool.
In the actual processing flow, once a transaction is executed successfully, it enters the mempool and waits to be packed into a block. Therefore, the overall node load is reflected not only by processing queues but also by the size of cached transactions.
The following mempool-related queues are currently not included in the load calculation:

- pendingTransactions
- rePushTransactions
- pushTransactionQueue

As a result, the load assessment is incomplete.

# Reproduction

- Start a node and connect it to the P2P network
- Continuously send transactions to the node so they accumulate in the mempool
- Control the thread pool processing speed to keep queue and smartContractQueue at relatively low levels
- Observe that `isBusy()` still reports the node as not busy
- The node continues accepting new transactions while the total cached transaction volume keeps increasing

# Impact

- Incomplete load evaluation: mempool transactions are not considered
- Inaccurate resource estimation: the total transaction load on the node is underestimated
- Suboptimal intake control: the node may continue accepting new transactions even when the mempool is already large

# Suggested Fix

Include mempool-related transaction counts in the `isBusy()` calculation and make `MAX_TRX_SIZE` configurable.
Example:
```
public boolean isBusy() {
   return queue.size()
       + smartContractQueue.size()
       + manager.getPendingSize()
       + manager.getRePushTransactions().size()
       + manager.getPushTransactionQueue().size()
       > MAX_TRX_SIZE; // configurable
}
```
By introducing these additional metrics and making `MAX_TRX_SIZE` configurable, the node can achieve a more comprehensive understanding of its load. This enables more accurate transaction intake control and allows operators to tune thresholds based on their deployment environment.

xxo1shine:
**What does this PR do?**

Optimize transaction rate limiting with accurate cache size check.

1. Add `Manager.getCachedTransactionSize() = pushTransactionQueue + pendingTransactions + rePushTransactions` to expose the true cached transaction count across all three queues.

2. Fix `isTooManyPending()` to include pushTransactionQueue, which was previously omitted, causing the pending threshold to be underestimated.

3. Update `TransactionsMsgHandler.isBusy()` to factor in the Manager cache size via `TronNetDelegate.getCachedTransactionSize()`, so the node stops accepting TRX INV messages when the full pipeline is busy.

4. Make the busy threshold configurable via `node.maxTrxCacheSize (default: 50000)`, replacing the hardcoded `MAX_TRX_SIZE` constant.

Fixes https://github.com/tronprotocol/java-tron/issues/6684

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/common/src/main/java/org/tron/common/parameter/CommonParameter.java
+++ b/common/src/main/java/org/tron/common/parameter/CommonParameter.java
@@ -492,6 +492,9 @@ public class CommonParameter {
   public long pendingTransactionTimeout;
   @Getter
   @Setter
+  public int maxTrxCacheSize;
+  @Getter
+  @Setter
   public boolean nodeMetricsEnable = false;
   @Getter
   @Setter
--- a/common/src/main/resources/reference.conf
+++ b/common/src/main/resources/reference.conf
@@ -344,6 +344,8 @@ node {
   receiveTcpMinDataLength = 2048
   maxTransactionPendingSize = 2000
   pendingTransactionTimeout = 60000
+  # total cached trx across handler queues + pending + rePush
+  maxTrxCacheSize = 50000
 
   # Consensus agreement
   agreeNodeCount = 0
--- a/framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java
@@ -32,7 +32,6 @@ import org.tron.protos.Protocol.Transaction.Contract.ContractType;
 @Component
 public class TransactionsMsgHandler implements TronMsgHandler {
 
-  private static int MAX_TRX_SIZE = 50_000;
   private static int MAX_SMART_CONTRACT_SUBMIT_SIZE = 100;
   @Autowired
   private TronNetDelegate tronNetDelegate;
@@ -41,7 +40,8 @@ public class TransactionsMsgHandler implements TronMsgHandler {
   @Autowired
   private ChainBaseManager chainBaseManager;
 
-  private BlockingQueue<TrxEvent> smartContractQueue = new LinkedBlockingQueue(MAX_TRX_SIZE);
+  private BlockingQueue<TrxEvent> smartContractQueue = new LinkedBlockingQueue(
+      Args.getInstance().getMaxTrxCacheSize());
 
   private BlockingQueue<Runnable> queue = new LinkedBlockingQueue();
 
@@ -71,7 +71,8 @@ public class TransactionsMsgHandler implements TronMsgHandler {
   }
 
   public boolean isBusy() {
-    return queue.size() + smartContractQueue.size() > MAX_TRX_SIZE;
+    return queue.size() + smartContractQueue.size()
+        + tronNetDelegate.getCachedTransactionSize() > Args.getInstance().getMaxTrxCacheSize();
   }
 
   @Override
--- a/framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
+++ b/framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
@@ -48,8 +48,6 @@ public class TransactionsMsgHandlerTest extends BaseTest {
   public void testProcessMessage() {
     TransactionsMsgHandler transactionsMsgHandler = new TransactionsMsgHandler();
     try {
-      Assert.assertFalse(transactionsMsgHandler.isBusy());
-
       transactionsMsgHandler.init();
 
       PeerConnection peer = Mockito.mock(PeerConnection.class);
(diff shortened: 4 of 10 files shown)


common/src/main/resources/reference.conf
```diff
@@ -351,6 +351,7 @@ node {
   receiveTcpMinDataLength = 2048
   maxTransactionPendingSize = 2000
   pendingTransactionTimeout = 60000
+  maxTrxCacheSize = 50000           // total cached trx across handler queues + pending + rePush
```
halibobo1205: [MUST] use `#` not `//`

framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
```diff
@@ -132,6 +132,23 @@ public void testProcessMessage() {
     }
   }
 
+  @Test
+  public void testIsBusyWithCachedTransactions() throws Exception {
+    TransactionsMsgHandler handler = new TransactionsMsgHandler();
+
+    TronNetDelegate tronNetDelegateMock = Mockito.mock(TronNetDelegate.class);
+    Mockito.when(tronNetDelegateMock.getCachedTransactionSize()).thenReturn(50_001);
```
alan-eth: Good targeted test for the new busy logic — mocking `TronNetDelegate` to control the cached size is exactly the right approach 👍

Tiny nit: the magic number `50_001` couples the test to the default `maxTrxCacheSize`. If someone tweaks the default later, this test silently keeps passing only because `50_001 > 50_000` happens to hold. Using `Args.getInstance().getMaxTrxCacheSize() + 1` would be future-proof, and adding a boundary case (e.g. exact-threshold returns `false`) would catch off-by-one regressions. Totally optional polish.

framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java
```diff
@@ -63,7 +63,8 @@ public void close() {
   }
 
   public boolean isBusy() {
-    return queue.size() + smartContractQueue.size() > MAX_TRX_SIZE;
+    return queue.size() + smartContractQueue.size()
+        + tronNetDelegate.getCachedTransactionSize() > Args.getInstance().getMaxTrxCacheSize();
```
317787106, line 75: [NIT] May be specify what happens when maxrxCacheSize <=0 .

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -468,6 +468,9 @@ public class CommonParameter {
   public long pendingTransactionTimeout;
   @Getter
   @Setter
+  public int maxTrxCacheSize; // clearParam: 50000
```
317787106: [SHOULD] Remove this comment; otherwise, you’ll need to keep two places in the code consistent.

xxo1shine: ok

common/src/main/resources/reference.conf
xxo1shine: ok

framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java
xxo1shine: The minimum value has been set to 2000.

framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
xxo1shine: Agreed — good catch. `isBusy()` uses strict `>`, so `cachedSize == maxTrxCacheSize` should indeed return `false`, and a boundary case is a clean way to guard against off-by-one regressions. Plan:

- Replace `50_001` with `Args.getInstance().getMaxTrxCacheSize() + 1`
- Add an assertion that `isBusy()` returns `false` when `cachedSize` equals `getMaxTrxCacheSize()`

Will push this in the next commit. Thanks for the review!

317787106 (approved):

halibobo1205 (approved):

PR #6716, opened 2026-04-28
feat(net): validate merkle root before block broadcast

xxo1shine:
**What does this PR do?**

- Validate merkle root before block broadcast.                                                                                                           
- Reset isVerified on owner permission revocation.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
@@ -15,6 +15,8 @@
 
 package org.tron.core.capsule;
 
+import static org.tron.core.exception.BadBlockException.TypeEnum.CALC_MERKLE_ROOT_FAILED;
+
 import com.google.common.primitives.Longs;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.CodedInputStream;
@@ -37,6 +39,7 @@ import org.tron.common.utils.Sha256Hash;
 import org.tron.common.utils.Time;
 import org.tron.core.capsule.utils.MerkleTree;
 import org.tron.core.config.Parameter.ChainConstant;
+import org.tron.core.exception.BadBlockException;
 import org.tron.core.exception.BadItemException;
 import org.tron.core.exception.ValidateSignatureException;
 import org.tron.core.store.AccountStore;
@@ -49,6 +52,7 @@ import org.tron.protos.Protocol.Transaction;
 public class BlockCapsule implements ProtoCapsule<Block> {
 
   public boolean generatedByMyself = false;
+  private volatile boolean merkleValidated = false;
   @Getter
   @Setter
   private TransactionRetCapsule result;
@@ -223,6 +227,19 @@ public class BlockCapsule implements ProtoCapsule<Block> {
         .collect(Collectors.toCollection(ArrayList::new));
 
     return MerkleTree.getInstance().createTree(ids).getRoot().getHash();
+  }
+
+  public void validateMerkleRoot() throws BadBlockException {
+    if (merkleValidated) {
+      return;
+    }
+    Sha256Hash actual = calcMerkleRoot();
+    if (!actual.equals(getMerkleRoot())) {
+      throw new BadBlockException(CALC_MERKLE_ROOT_FAILED,
+          String.format("merkle root mismatch for block %d: expected %s, actual %s",
+              getNum(), getMerkleRoot(), actual));
+    }
+    merkleValidated = true;
   }
 
   public void setMerkleRoot() {
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1307,12 +1307,7 @@ public class Manager {
         try (PendingManager pm = new PendingManager(this)) {
 
           if (!block.generatedByMyself) {
-            if (!block.calcMerkleRoot().equals(block.getMerkleRoot())) {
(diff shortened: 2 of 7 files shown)


framework/src/main/java/org/tron/core/net/TronNetDelegate.java
```diff
@@ -347,6 +347,11 @@ public boolean validBlock(BlockCapsule block) throws P2pException {
       throw new P2pException(TypeEnum.BAD_BLOCK,
               "time:" + time + ",block time:" + block.getTimeStamp());
     }
+    try {
+      block.validateMerkleRoot();
+    } catch (BadBlockException e) {
+      throw new P2pException(TypeEnum.BLOCK_MERKLE_ERROR, e.getMessage());
+    }
```
alan-eth, line 354: Nice work moving the merkle check before broadcast — that's exactly the right place to catch tampered blocks before we forward them. 🎯

Minor consistency nit: when wrapping `BadBlockException` here, you're passing `e.getMessage()` as the message string, but the existing flow in `processBlock` (around L302) wraps it as `throw new P2pException(TypeEnum.BLOCK_MERKLE_ERROR, e)` (the exception itself becomes the cause). Would it make sense to do the same here so the original stack trace is preserved when this propagates up? It also keeps the two `BLOCK_MERKLE_ERROR` throw sites symmetric.

lvs0075 (approved):

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -2106,6 +2101,13 @@ public void rePush(TransactionCapsule tx) {
       return;
     }
 
+    String ownerAddress = ByteArray.toHexString(tx.getOwnerAddress());
+    synchronized (this) {
+      if (ownerAddressSet.contains(ownerAddress)) {
+        tx.setVerified(false);
+      }
+    }
```
317787106, line 2133: [INT] Give code comment why should setVerified as false here.

xxo1shine: This if statement is already quite concise and clear. Transactions involving permission changes should reset this flag. Re-verification is required to prevent transactions with incorrect signatures from entering the pending queue.

framework/src/main/java/org/tron/core/net/TronNetDelegate.java
xxo1shine: The BLOCK_MERKLE_ERROR exception is quite obvious, and the message information is sufficient. There's no need to re-throw the original exception, and the previous logic doesn't need to be changed; there's no impact whatsoever.

chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
```diff
@@ -37,6 +37,8 @@
 import org.tron.common.utils.Time;
 import org.tron.core.capsule.utils.MerkleTree;
 import org.tron.core.config.Parameter.ChainConstant;
+import org.tron.core.exception.BadBlockException;
+import static org.tron.core.exception.BadBlockException.TypeEnum.CALC_MERKLE_ROOT_FAILED;
```
waynercheung: [SHOULD] Static import is interleaved with regular imports. Move it above all `import org...` lines as a separate group, matching the convention used in `TronNetDelegate.java:4` and `ManagerTest.java:11`. Likely a Checkstyle violation.

framework/src/test/java/org/tron/core/db/ManagerTest.java
```diff
@@ -1292,6 +1292,35 @@ public void blockTrigger() {
     Assert.assertEquals(TronError.ErrCode.EVENT_SUBSCRIBE_ERROR, thrown.getErrCode());
   }
 
+  @Test
```
waynercheung, line 1423: [SHOULD] The PR title is "validate merkle root before block broadcast", but `TronNetDelegate.validBlock()` - the function that actually realises this - has no direct test. Consider adding a test that constructs a block with a tampered merkle root and asserts `validBlock` throws `P2pException` with `TypeEnum.BLOCK_MERKLE_ERROR`. The existing tests only cover `BlockCapsule.validateMerkleRoot()` and the `Manager.pushBlock` path.

chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
xxo1shine: Confirmed — static imports should be placed as a separate group above all regular imports. Will move `import static org.tron.core.exception.BadBlockException.TypeEnum.CALC_MERKLE_ROOT_FAILED;` to the top of the file and run checkstyle to verify. Thanks for catching this!

framework/src/test/java/org/tron/core/db/ManagerTest.java
xxo1shine: Agreed. `BlockCapsule.validateMerkleRoot()` and the `Manager.pushBlock` path are covered, but `TronNetDelegate.validBlock()` — the actual pre-broadcast entry point — has no direct test (it's mocked out in `BlockMsgHandlerTest`). Plan: add a unit test that constructs a `BlockCapsule` with a tampered merkle root, calls `validBlock` directly, and asserts `P2pException` is thrown with `TypeEnum.BLOCK_MERKLE_ERROR`. Will push this in the next commit. Thanks for the catch!

waynercheung (approved):

317787106 (approved):

PR #6717, opened 2026-04-28
feat(net): reduce sync memory via lazy parsing and throttling

### LINKED ISSUE #6685: [Feature]Reduce memory usage in blocks sync via delayed serialization and cache limits
# Summary
During the block synchronization process, TRON nodes may experience cache bloat because blocks are serialized too early in the caching stage. By delaying serialization until the actual processing phase and limiting the size of the sync cache, memory usage and cache scale can be better controlled, thereby improving system stability.

# Problem

### Motivation

During block synchronization, nodes may need to temporarily cache a large number of block data. A more efficient memory management mechanism is required to prevent excessive resource consumption and potential impacts on node stability.

### Current State

In the current implementation, blocks are serialized when they are added to the sync cache. When the number of cached blocks is large (e.g., close to 4000), the serialized data significantly increases memory usage.

### Limitations or Risks

- Early serialization leads to cache size inflation
- High concurrency sync scenarios may cause significant memory pressure
- Inefficient resource utilization in the block synchronization pipeline

# Proposed Solution

### Proposed Design

Optimize the block synchronization process as follows:

- Delay block serialization to the actual processing stage
- Limit the number of blocks in the sync cache to prevent unbounded growth
- Avoid unnecessary serialization overhead and control cache size effectively

### Key Changes
- Block sync module (adjust cache management logic)
- Block processing flow (change serialization timing)
- Sync cache mechanism (introduce or strengthen maximum cache size limits)

# Impact

- Memory usage: Reduces memory inflation caused by early serialization
- System stability: Lowers memory pressure in extreme sync scenarios
- Performance: Reduces unnecessary serialization overhead during caching
- Scalability: Improves node adaptability in large-scale synchronization scenarios

# Compatibility

- Breaking Change: No
- Default Behavior Change: Yes (serialization is delayed)
- Migration Required: No

This optimization does not affect external APIs or network protocols; it only impacts internal synchronization and caching logic.

# Additional Notes

- Implementation ideas: Yes
- Willing to implement: Yes

xxo1shine:
**What does this PR do?**

Cap memory usage during block sync by deferring block deserialization and bounding in-flight block requests.                                                         
   
  1. **New `UnparsedBlock` carrier** — lightweight `(BlockId, byte[])` pair (`framework/.../net/service/sync/UnparsedBlock.java`). Holds the parsed block id (cheap) plus the raw protobuf bytes; the block body is not deserialized until the worker thread is ready to process it. Implements `equals`/`hashCode` on `blockId` so it works as a map key.                                                                                                                                                  
                  
  2. **Defer deserialization in sync queues** — in `SyncService`:                                                                                                      
     - `blockWaitToProcess` and `blockJustReceived` change from `Map<BlockMessage, PeerConnection>` to `Map<UnparsedBlock, PeerConnection>`
     - `processBlock(peer, blockMessage)` builds the `UnparsedBlock` from `blockMessage.getData()` instead of inserting the fully-parsed message                       
     - `handleSyncBlock` parses the bytes back into a `BlockMessage` only when the block is about to be applied                                                        
                                                                                                                                                                       
  3. **Throttle in-flight block requests via `node.maxPendingBlockNum`** — new config controlling the total budget of `requested + justReceived + waitToProcess` blocks across all peers:                                                                                                                                                   
     - Default `500`, clamped to `[50, 2000]` in `NodeConfig.postProcess()`                                                                                            
     - Wired through `NodeConfig.maxPendingBlockNum` → `Args.applyNodeConfig` → `CommonParameter.maxPendingBlockNum`                                                   
     - Default mirrored in `reference.conf`; example documented in `config.conf`                                                                                       
                                                                                                                                                                       
  4. **`startFetchSyncBlock` becomes budget-aware** — computes `remainNum = maxPendingBlockNum - requested - justReceived - waitToProcess`; once the budget is         
  exhausted it stops requesting new heights, with one exception: blocks at or below `maxRequestedBlockNum` (the previously-requested ceiling) are still allowed        
  through, so a slow peer can be retried without stalling the whole sync.                                                                                              
                  
  5. **Tests** — `SyncServiceTest` updated to construct `UnparsedBlock` carriers in place of raw `BlockMessage` puts, plus boundary-condition coverage for the throttle
   path.

 1. **Sync-time heap was unbounded.** Every received block was parsed eagerly into a `BlockMessage` and held in two maps until processing finished. Each `BlockMessage` keeps the parsed `Block` proto with every transaction expanded into Java objects, which inflates the on-heap footprint several times over the wire size. During catch-up sync from many peers, this routinely caused multi-GB heap growth and GC pressure, with documented OOM incidents on smaller-RAM nodes.          
                  
  2. **`UnparsedBlock` collapses the worst case.** Raw bytes are 1–2 MB per block; the parsed in-memory representation is significantly larger. Keeping the bytes raw until processing defers (and amortizes) the parse cost to the worker that actually needs it, and frees the in-flight buffer to grow proportional to *bytes*, not *parsed-object-graph*.                                                                                                                                               
                  
  3. **Concurrent peer fetching had no global cap.** A peer aggressively shipping inventory could push the `blockJustReceived` map far past safe levels, because the  receiver only tracked per-peer `MAX_BLOCK_FETCH_PER_PEER`. Bounding `requested + justReceived + waitToProcess` makes the worst-case in-flight memory predictable: `maxPendingBlockNum × avg block size`.                                                                                                                               
                  
  4. **The `maxRequestedBlockNum` exemption avoids deadlock.** A pure hard cap could stall sync if every peer holding a needed block goes idle and the budget is full —
   the only way to make progress is to retry the height, which requires re-requesting a block already in the budget. Allowing retries within the existing ceiling keeps sync forward-progressing under transient peer flakiness.                                                                                                            
                  
  5. **Operators need the knob.** Default `500` is conservative for nodes with ≤ 8 GB heap; large/dedicated nodes benefit from raising it for higher sync throughput,  and constrained environments can lower it. Surfacing it as `node.maxPendingBlockNum` (rather than a constant) lets operators tune without a release.
        
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

Fixes https://github.com/tronprotocol/java-tron/issues/6685

### DIFF
--- a/framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
+++ b/framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
@@ -5,6 +5,7 @@ import static org.tron.core.config.Parameter.NetConstants.MAX_BLOCK_FETCH_PER_PE
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -44,9 +45,9 @@ public class SyncService {
   @Autowired
   private PbftDataSyncHandler pbftDataSyncHandler;
 
-  private Map<BlockMessage, PeerConnection> blockWaitToProcess = new ConcurrentHashMap<>();
+  private Map<UnparsedBlock, PeerConnection> blockWaitToProcess = new ConcurrentHashMap<>();
 
-  private Map<BlockMessage, PeerConnection> blockJustReceived = new ConcurrentHashMap<>();
+  private Map<UnparsedBlock, PeerConnection> blockJustReceived = new ConcurrentHashMap<>();
 
   private long blockCacheTimeout = Args.getInstance().getBlockCacheTimeout();
   private Cache<BlockId, PeerConnection> requestBlockIds = CacheBuilder.newBuilder()
@@ -68,6 +69,10 @@ public class SyncService {
   private volatile boolean fetchFlag = false;
 
   private final long syncFetchBatchNum = Args.getInstance().getSyncFetchBatchNum();
+
+  private final int maxPendingBlockSize = Args.getInstance().getMaxPendingBlockSize();
+
+  private volatile long maxRequestedBlockNum = 0;
 
   public void init() {
     ExecutorServiceManager.scheduleWithFixedDelay(fetchExecutor, () -> {
@@ -135,7 +140,9 @@ public class SyncService {
 
   public void processBlock(PeerConnection peer, BlockMessage blockMessage) {
     synchronized (blockJustReceived) {
-      blockJustReceived.put(blockMessage, peer);
+      UnparsedBlock unparsedBlock = new UnparsedBlock(
+          blockMessage.getBlockId(), blockMessage.getData());
+      blockJustReceived.put(unparsedBlock, peer);
     }
     handleFlag = true;
     if (peer.isSyncIdle()) {
@@ -227,8 +234,18 @@ public class SyncService {
   }
 
   private void startFetchSyncBlock() {
+    Collection<PeerConnection> activePeers = tronNetDelegate.getActivePeer();
+    int reqNum = activePeers.stream()
+        .mapToInt(p -> p.getSyncBlockRequested().size()).sum();
+    int remainNum;
+    synchronized (blockJustReceived) {
+      remainNum = maxPendingBlockSize - reqNum
+          - blockJustReceived.size() - blockWaitToProcess.size();
+    }
+
     HashMap<PeerConnection, List<BlockId>> send = new HashMap<>();
-    tronNetDelegate.getActivePeer().stream()
+    int[] fetchingBlockSize = {0};
+    activePeers.stream()
         .filter(peer -> peer.isNeedSyncFromPeer() && peer.isSyncIdle())
         .filter(peer -> peer.isFetchAble())
         .forEach(peer -> {
@@ -238,9 +255,16 @@ public class SyncService {
           for (BlockId blockId : peer.getSyncBlockToFetch()) {
             if (requestBlockIds.getIfPresent(blockId) == null
                 && !peer.getSyncBlockInProcess().contains(blockId)) {
+              if (fetchingBlockSize[0] >= remainNum && blockId.getNum() > maxRequestedBlockNum) {
+                break;
+              }
+              if (blockId.getNum() > maxRequestedBlockNum) {
+                maxRequestedBlockNum = blockId.getNum();
+              }
               requestBlockIds.put(blockId, peer);
               peer.getSyncBlockRequested().put(blockId, System.currentTimeMillis());
               send.get(peer).add(blockId);
+              fetchingBlockSize[0]++;
               if (send.get(peer).size() >= MAX_BLOCK_FETCH_PER_PEER) {
                 break;
               }
@@ -269,29 +293,37 @@ public class SyncService {
 
       isProcessed[0] = false;
 
-      blockWaitToProcess.forEach((msg, peerConnection) -> {
+      blockWaitToProcess.forEach((unparsedBlock, peerConnection) -> {
         synchronized (tronNetDelegate.getBlockLock()) {
+          BlockId blockId = unparsedBlock.getBlockId();
           if (peerConnection.isDisconnect()) {
-            blockWaitToProcess.remove(msg);
-            invalid(msg.getBlockId(), peerConnection);
+            blockWaitToProcess.remove(unparsedBlock);
+            invalid(blockId, peerConnection);
             return;
           }
-          if (msg.getBlockId().getNum() <= solidNum) {
-            blockWaitToProcess.remove(msg);
-            peerConnection.getSyncBlockInProcess().remove(msg.getBlockId());
+          if (blockId.getNum() <= solidNum) {
+            blockWaitToProcess.remove(unparsedBlock);
+            peerConnection.getSyncBlockInProcess().remove(blockId);
             return;
           }
           final boolean[] isFound = {false};
           tronNetDelegate.getActivePeer().stream()
-              .filter(peer -> msg.getBlockId().equals(peer.getSyncBlockToFetch().peek()))
+              .filter(peer -> blockId.equals(peer.getSyncBlockToFetch().peek()))
               .forEach(peer -> {
                 isFound[0] = true;
               });
           if (isFound[0]) {
-            blockWaitToProcess.remove(msg);
+            blockWaitToProcess.remove(unparsedBlock);
             isProcessed[0] = true;
-            processSyncBlock(msg.getBlockCapsule(), peerConnection);
-            peerConnection.getSyncBlockInProcess().remove(msg.getBlockId());
+            BlockCapsule block;
+            try {
+              block = new BlockCapsule(unparsedBlock.getData());
+            } catch (Exception e) {
+              logger.warn("Deserialize block {} failed", blockId.getString(), e);
+              return;
+            }
+            processSyncBlock(block, peerConnection);
+            peerConnection.getSyncBlockInProcess().remove(blockId);
           }
         }
       });
--- a/framework/src/test/java/org/tron/core/net/services/SyncServiceTest.java
+++ b/framework/src/test/java/org/tron/core/net/services/SyncServiceTest.java
@@ -23,6 +23,7 @@ import org.tron.core.net.peer.PeerConnection;
 import org.tron.core.net.peer.PeerManager;
 import org.tron.core.net.peer.TronState;
 import org.tron.core.net.service.sync.SyncService;
+import org.tron.core.net.service.sync.UnparsedBlock;
 import org.tron.p2p.connection.Channel;
 import org.tron.protos.Protocol;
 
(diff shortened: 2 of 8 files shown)


framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
```diff
@@ -69,6 +70,10 @@ public class SyncService {
 
   private final long syncFetchBatchNum = Args.getInstance().getSyncFetchBatchNum();
 
+  private final int maxPendingBlockNum = Args.getInstance().getMaxPendingBlockNum();
+
+  private static volatile long maxRequestedBlockNum = 0;
```
waynercheung: [MUST] This should be an instance field, not `static`.

Every other piece of sync state on this class (`blockWaitToProcess`, `blockJustReceived`, `maxPendingBlockNum`) is instance-scoped — only this one is static, with no apparent reason. Any scenario that recreates the Spring context (integration tests, hot reload, future multi-instance use) will leak state across instances.

Direct evidence: `SyncServiceTest` has to reset it via reflection (`SyncServiceTest.java:185-187`) to avoid pollution between test cases.

Suggested change:

    private volatile long maxRequestedBlockNum = 0;

The reflection reset in the test can then be removed as well. If only the single-threaded `fetchExecutor` writes to it after this change, `volatile` can also be dropped.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
```diff
@@ -269,29 +293,37 @@ private synchronized void handleSyncBlock() {
...
                 isFound[0] = true;
               });
           if (isFound[0]) {
-            blockWaitToProcess.remove(msg);
+            blockWaitToProcess.remove(unparsedBlock);
             isProcessed[0] = true;
-            processSyncBlock(msg.getBlockCapsule(), peerConnection);
-            peerConnection.getSyncBlockInProcess().remove(msg.getBlockId());
+            BlockCapsule block;
+            try {
```
waynercheung, line 319: [MUST] On failure this branch only logs and returns, which leaves three problems:

1. `unparsedBlock` stays in `blockWaitToProcess` - next tick re-parses, re-fails, and the entry permanently consumes one `maxPendingBlockNum` slot.
2. `peer.getSyncBlockInProcess()` is not cleared for this `blockId`.
3. `invalid(blockId, peerConnection)` is not called, so the peer is not penalised and the block is not rescheduled to another peer.

The trigger probability is very low (bytes were already parsed once in the `BlockMessage` ctor), but the cleanup should mirror the `isDisconnect()` branch directly above:

```java
} catch (Exception e) {
  logger.warn("Deserialize block {} failed", blockId.getString(), e);
  blockWaitToProcess.remove(unparsedBlock);
  peerConnection.getSyncBlockInProcess().remove(blockId);
  invalid(blockId, peerConnection);
  return;
}
```

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
waynercheung, line 282: [SHOULD] Missing wake-up safety net when the throttle saturates.

When the budget is full, `startFetchSyncBlock` returns having issued zero requests, and `fetchFlag` has already been consumed (set to false). Re-arming it depends on a later event from `processBlock` / `invalid`. If no new block arrives and no peer disconnects in that window, the next fetch has to wait for the cron tick *and* an external event.

Since `handleSyncBlock` is exactly the place that frees budget, the cheapest fix is to set `fetchFlag = true;` at the end of the `if (isFound[0]) { ... }` block (after `processSyncBlock`). Negligible cost, removes the corner case.

framework/src/test/java/org/tron/core/net/services/SyncServiceTest.java
```diff
@@ -169,6 +180,33 @@ public void testStartFetchSyncBlock() throws Exception {
...
+
+    // fill blockWaitToProcess to reach maxPendingBlockNum (default 500)
+    int maxPendingBlockNum = (int) ReflectUtils.getFieldObject(service, "maxPendingBlockNum");
+    for (int i = 0; i < maxPendingBlockNum; i++) {
+      BlockCapsule.BlockId fillId = new BlockCapsule.BlockId(Sha256Hash.ZERO_HASH, 10000 + i);
+      blockWaitToProcess.put(new UnparsedBlock(fillId, new byte[0]), peer);
+    }
+    method.invoke(service);
+    // highBlockId must NOT be requested: remainNum <= 0 and num > maxRequestedBlockNum
+    Assert.assertNull(peer.getSyncBlockRequested().get(highBlockId));
```
waynercheung, line 208: [SHOULD] Missing test for the `maxRequestedBlockNum` exemption path.

Current coverage only asserts the hard-cap case (`budget exhausted + num > maxRequestedBlockNum` -> block not requested). The PR's deadlock-avoidance design specifically allows `budget exhausted + num <= maxRequestedBlockNum` to go through as a retry, and that path is untested - a future refactor could break it silently.

Suggestion: add a symmetric case where `maxRequestedBlockNum` is set to e.g. 100, `blockWaitToProcess` is filled to `maxPendingBlockNum`, the target `blockId.num = 50`, and assert `peer.getSyncBlockRequested().get(blockId) != null`.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
317787106: [SHOULD] The names of these two variables are very poorly chosen and have low readability. Based on the code, `maxPendingBlockNum` refers to the length of the block list, while `maxRequestedBlockNum` refers to an absolute block number—these meanings are quite different, yet the names are almost identical. It is recommended to rename `maxPendingBlockNum` to `maxPendingBlockSize`.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
```diff
@@ -227,8 +234,18 @@ private BlockId getBlockIdByNum(long num) throws P2pException {
...
+        .mapToInt(p -> p.getSyncBlockRequested().size()).sum();
+    int remainNum;
+    synchronized (blockJustReceived) {
+      remainNum = maxPendingBlockNum - reqNum
+          - blockJustReceived.size() - blockWaitToProcess.size();
+    }
+
     HashMap<PeerConnection, List<BlockId>> send = new HashMap<>();
-    tronNetDelegate.getActivePeer().stream()
+    int[] cnt = {0};
```
317787106: [SHOULD] This name `cnt` is no meaning. Give the variable a name that clearly conveys its meaning. Such as `fetchingBlockSize`.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
```diff
@@ -269,29 +293,37 @@ private synchronized void handleSyncBlock() {
...
               });
           if (isFound[0]) {
-            blockWaitToProcess.remove(msg);
+            blockWaitToProcess.remove(unparsedBlock);
             isProcessed[0] = true;
-            processSyncBlock(msg.getBlockCapsule(), peerConnection);
-            peerConnection.getSyncBlockInProcess().remove(msg.getBlockId());
+            BlockCapsule block;
+            try {
+              block = new BlockCapsule(unparsedBlock.getData());
```
317787106, line 320: [MUST] The purpose of replacing Block with `UnparsedBlock` is to save memory. But have you evaluated how much memory is actually saved? Encoding the data in a binary (proto) form doesn’t reduce memory usage significantly, so the benefit is minimal, while introducing several downsides:

- It increases the cognitive burden for users.
- If the data cannot be constructed into a `Block`, the peer should be disconnected as early as possible. However, nothing is handled here, which is inconsistent with the previous logic.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
xxo1shine: I accept the suggested changes; removing the static field is also fine.

xxo1shine: The variable names have been modified as suggested.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
xxo1shine: The variable names have been modified as suggested.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
xxo1shine: The deserialization was performed once before, so it's impossible for it to fail here. This is just to handle exceptions and doesn't require any additional logic.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
xxo1shine: This will result in a significant improvement. Since deserialization was performed once before, it's impossible for it to fail here. This is simply to handle exceptions and doesn't require any additional logic.

framework/src/main/java/org/tron/core/net/service/sync/SyncService.java
xxo1shine: processBlock already sets fetchFlag=true whenever a new block arrives, so the wake-up frequency during an active sync is already sufficient. Adding another fetchFlag=true here would, once the sync is near catch-up and peers have nothing left in syncBlockToFetch, cause startFetchSyncBlock to repeatedly walk activePeers for nothing — i.e. wasted fetch cycles. Leaving it as-is for now.

framework/src/test/java/org/tron/core/net/services/SyncServiceTest.java
xxo1shine: Okay, the single test has been completed.

halibobo1205 (approved):

waynercheung (approved):

317787106 (approved):

PR #6718, opened 2026-04-28
fix(event): correct fork rollback handling in solidity event maps

### LINKED ISSUE #6678: [Bug]Event cache not cleared on reorg causing duplicate and inconsistent event delivery
# Software Versions

OS: Linux
JVM: Oracle Corporation 1.8.0_161 amd64
Version: 4.8.1

# Expected Behavior

After a transaction is executed and events are generated, external subscription systems should receive unique finalized event data that is consistent with the chain.

# Actual Behavior

In a private network deployment, when the event service is using version 0, the event cache of the old fork chain is not cleared when a chain reorg occurs.

As a result, event data from both the forked chain and the new main chain are written into the database simultaneously. External subscribers may therefore receive duplicate or inconsistent event data.

Image

# Frequency

This issue may occur whenever a chain reorganization happens and contract event solidification (finalization) is enabled.

xxo1shine:
**What does this PR do?**

 Fix two correctness bugs in event service rollback handling that cause solidity-side event/log triggers to either leak fork- removed entries or be lost during reorg.                                                                                
                                                                                                                                                                       
                                                                                                                                                                       
  When a block is rolled back during fork resolution, its contract triggers are                                                                                         marked `removed=true`. The solidity event pipeline depends on this flag plus an  in-memory map (`Args.getSolidityContractEventTriggerMap` /  `getSolidityContractLogTriggerMap`) keyed by block number. Two paths through this pipeline were broken:     
                                                                                                                                         
                                                                                                                                                                       
  1. **Removed triggers reached the solidity sink.** `ContractTriggerCapsule.processTrigger()`                                                                         
     only checked `isSolidityEventTriggerEnable / isSolidityLogTriggerEnable`, never the                                                                               
     `removed` flag. Triggers from forked-out blocks were emitted as valid solidity events.                                                                            
                                                                                                                                                                       
  2. **Async enqueue raced with reorg cache clearing.** `Manager.postContractTrigger`                                                                                  
     posted to a global `triggerCapsuleQueue` for async draining, but                                                                                                  
     `reOrgContractTrigger` clears the in-memory cache synchronously. Capsules already                                                                                 
     in the queue could reference state that was about to be invalidated, leading to
     either lost events or stale cache reads.                                                                                                                          
                  
  3. **Solidity caches were not cleared on block-process failure or shutdown.**                                                                                        
     `getSolidityContract{Log,Event}TriggerMap` entries are populated per block.
     If `processBlock` threw after population, or on node shutdown, those entries                                                                                      
     leaked. 

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

Fixes https://github.com/tronprotocol/java-tron/issues/6678

### DIFF
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1377,6 +1377,7 @@ public class Manager {
             } catch (Throwable throwable) {
               logger.error(throwable.getMessage(), throwable);
               khaosDb.removeBlk(block.getBlockId());
+              clearSolidityContractTriggerCache(block.getNum());
               throw throwable;
             }
             long newSolidNum = getDynamicPropertiesStore().getLatestSolidifiedBlockNum();
@@ -2378,6 +2379,16 @@ public class Manager {
             getDynamicPropertiesStore().getLatestBlockHeaderHash());
       }
     }
+    clearSolidityContractTriggerCache(getHeadBlockNum());
+  }
+
+  private void clearSolidityContractTriggerCache(long blockNum) {
+    if (eventPluginLoaded
+        && (EventPluginLoader.getInstance().isSolidityEventTriggerEnable()
+        || EventPluginLoader.getInstance().isSolidityLogTriggerEnable())) {
+      Args.getSolidityContractLogTriggerMap().remove(blockNum);
+      Args.getSolidityContractEventTriggerMap().remove(blockNum);
+    }
   }
 
   private void postContractTrigger(final TransactionTrace trace, boolean remove, String blockHash) {
@@ -2397,9 +2408,14 @@ public class Manager {
             .getLatestSolidifiedBlockNum());
         contractTriggerCapsule.setBlockHash(blockHash);
 
-        if (!triggerCapsuleQueue.offer(contractTriggerCapsule)) {
-          logger.info("Too many triggers, contract log trigger lost: {}.",
-              trigger.getTransactionId());
+        // Process synchronously to avoid race condition between async queue and
+        // reOrgContractTrigger cache clearing. Performance is not impacted because
+        // processTrigger() only enqueues events into the plugin's internal queue
+        // without blocking on actual I/O.
+        try {
+          contractTriggerCapsule.processTrigger();
+        } catch (Throwable throwable) {
+          logger.warn("Post contract trigger failed.", throwable);
         }
       }
     }
--- a/framework/src/test/java/org/tron/core/db/ManagerTest.java
+++ b/framework/src/test/java/org/tron/core/db/ManagerTest.java
@@ -3,7 +3,9 @@ package org.tron.core.db;
 import static org.junit.Assert.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
 import static org.tron.common.utils.Commons.adjustAssetBalanceV2;
 import static org.tron.common.utils.Commons.adjustTotalShieldedPoolValue;
 import static org.tron.common.utils.Commons.getExchangeStoreFinal;
@@ -16,12 +18,15 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.protobuf.Any;
 import com.google.protobuf.ByteString;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -32,6 +37,8 @@ import org.tron.api.GrpcAPI;
 import org.tron.common.BaseMethodTest;
 import org.tron.common.TestConstants;
 import org.tron.common.crypto.ECKey;
+import org.tron.common.logsfilter.EventPluginLoader;
+import org.tron.common.logsfilter.trigger.ContractLogTrigger;
 import org.tron.common.runtime.RuntimeImpl;
 import org.tron.common.utils.ByteArray;
 import org.tron.common.utils.Commons;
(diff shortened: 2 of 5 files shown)


lvs0075 (approved):

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -2378,6 +2379,16 @@ private void reOrgContractTrigger() {
             getDynamicPropertiesStore().getLatestBlockHeaderHash());
       }
     }
+    clearSolidityContractTriggerCache(getHeadBlockNum());
+  }
+
+  private void clearSolidityContractTriggerCache(long blockNum) {
+    if (eventPluginLoaded
+        && (EventPluginLoader.getInstance().isSolidityEventTriggerEnable()
+        || EventPluginLoader.getInstance().isSolidityLogTriggerEnable())) {
```
0xbigapple, line 2388: [SHOULD] `clearSolidityContractTriggerCache` guard is narrower than the write side; redundancy path leaks the map.
`ContractTriggerCapsule.java:162` ( the event-as-log redundancy path) also writes to `solidityContractLogTriggerMap` under this combination:

```
isContractLogTriggerEnable      = true
isContractLogTriggerRedundancy  = true
isSolidityLogTriggerRedundancy  = true
isSolidityLogTriggerEnable      = false
isSolidityEventTriggerEnable    = false
```

outer condition `(isContractLogEnable && isContractLogRedundancy) || ...` is true → enters the redundancy block; inner `isSolidityLogTriggerRedundancy=true` → writes to the solidity log map. But both clear-side guard predicates are false → clear is skipped.

Subtler still: the consumer `postSolidityTrigger` is also gated by `isSolidityLogTriggerEnable`, so under this configuration the consumer does not run either. The result: `solidityContractLogTriggerMap` accumulates one entry per event-bearing block, never consumed, never cleared — a memory leak. It does not produce Issue #6678's "duplicate downstream events" symptom (no downstream is wired up), but since the PR formalizes the clearing responsibility, the clear's semantics should be aligned with the write side.

**Suggestion (preferred)**: simply drop the guard. `ConcurrentHashMap.remove(key)` on a non-existent key is an O(1) bucket lookup returning `null` — the cost is negligible. The guard was only a defensive optimization; without it, write-side condition changes won't desync the clear:
```java
private void clearSolidityContractTriggerCache(long blockNum) {
  Args.getSolidityContractLogTriggerMap().remove(blockNum);
  Args.getSolidityContractEventTriggerMap().remove(blockNum);
}
```

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -1377,6 +1377,7 @@ public void pushBlock(final BlockCapsule block)
             } catch (Throwable throwable) {
               logger.error(throwable.getMessage(), throwable);
               khaosDb.removeBlk(block.getBlockId());
+              clearSolidityContractTriggerCache(block.getNum());
```
0xbigapple, line 1380: [SHOULD] `switchFork` has two `applyBlock` failure paths that don't clear the cache.

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -2397,9 +2408,14 @@ private void postContractTrigger(final TransactionTrace trace, boolean remove, S
...
-          logger.info("Too many triggers, contract log trigger lost: {}.",
-              trigger.getTransactionId());
+        // Process synchronously to avoid race condition between async queue and
+        // reOrgContractTrigger cache clearing. Performance is not impacted because
+        // processTrigger() only enqueues events into the plugin's internal queue
+        // without blocking on actual I/O.
+        try {
+          contractTriggerCapsule.processTrigger();
+        } catch (Throwable throwable) {
+          logger.warn("Post contract trigger failed. {}", throwable.getMessage());
```
0xbigapple: [SHOULD] `logger.warn` loses stack trace; trigger failures will be undebuggable.

**Suggestion**:
```java
} catch (Throwable throwable) {
  logger.warn("Post contract trigger failed.", throwable);
}
```

framework/src/test/java/org/tron/core/db/ManagerTest.java
```diff
@@ -1292,6 +1299,79 @@ public void blockTrigger() {
     Assert.assertEquals(TronError.ErrCode.EVENT_SUBSCRIBE_ERROR, thrown.getErrCode());
   }
 
+  @Test
+  public void testReOrgContractTriggerClearsMap() throws Exception {
```
0xbigapple, line 1303: [NIT] `testReOrgContractTriggerClearsMap` doesn't actually call `reOrgContractTrigger` — the name is misleading

framework/src/main/java/org/tron/core/db/Manager.java
xxo1shine: Repeated chain switching is a historical oversight and will not be addressed in this PR. It can be discussed separately in a later issue.

framework/src/main/java/org/tron/core/db/Manager.java
xxo1shine: Enabling only `isContractLogTriggerEnable = true` and `isContractLogTriggerRedundancy = true` will not write data; only enabling the solid event switch will write data.

framework/src/test/java/org/tron/core/db/ManagerTest.java
xxo1shine: @0xbigapple, you're right. The test name promises the reorg path but only exercises `clearSolidityContractTriggerCache`, which also overlaps with `testClearSolidityContractTriggerCache` below. I'll rewrite this case to actually go through `reOrgContractTrigger` (construct a forked block and assert the corresponding entry is removed from the cache) so the name and behavior line up.

framework/src/main/java/org/tron/core/db/Manager.java
xxo1shine: ok

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -2414,9 +2408,14 @@ private void postContractTrigger(final TransactionTrace trace, boolean remove, S
             .getLatestSolidifiedBlockNum());
         contractTriggerCapsule.setBlockHash(blockHash);
 
-        if (!triggerCapsuleQueue.offer(contractTriggerCapsule)) {
-          logger.info("Too many triggers, contract log trigger lost: {}.",
-              trigger.getTransactionId());
+        // Process synchronously to avoid race condition between async queue and
+        // reOrgContractTrigger cache clearing. Performance is not impacted because
+        // processTrigger() only enqueues events into the plugin's internal queue
```
yanghang8612, line 2413: This comment understates the behavior change. `processTrigger()` does not only enqueue into an internal queue; it can synchronously call `postContractEventTrigger` / `postContractLogTrigger`, which may invoke plugin listeners or publish to `NativeMessageQueue`. This moves normal contract trigger delivery into the `applyBlock` path. Could we keep only the solidity map update synchronous, or adjust the comment/implementation to reflect the synchronous plugin delivery?

xxo1shine: This is mainly to solve the data concurrency problem. The underlying plugin is actually asynchronous. The plugin puts the data into the queue and then returns, so there will be no blocking behavior.

yanghang8612: OK, got it.

yanghang8612 (approved):
LGTM

0xbigapple (approved):

Issue #6266, opened 2025-03-28
Using a full node for reading


Mihail-source-arch:
Have a nice day and a good mood. Please help me figure out the issue.
I have a rather complex smart contract that performs large calculations, including calls to other smart contracts. I call it via triggerConstantContract, and get the data I need. But in some cases, the calculation process takes more than 80 ms. and my transaction fails with the error "class org.tron.core.vm.program.Program$OutOfTimeException : CPU timeout for 'SWAP1' operation executing". Is it possible to somehow circumvent this limitation when calling ConstantContract, and bring the smart contract function to completion? Is it possible to use a full read-only node? Thanks in advance, I'm sorry if my question is stupid, I'm a beginner.

Mihail-source-arch:
> We apologize for the bad experience.
> 
> Currently you can't avoid the `80ms` execution limit by accessing the `/wallet/triggerconstantcontract` interface through the official node.
> 
> However, you can start the node in `debug` mode by adding the startup command line parameter `--debug`, so that accessing the `/wallet/triggerconstantcontract` interface on that node will skip the 80ms execution time limit.

Thank you very much. It worked for me. All the best to you.

Mihail-source-arch:
Unfortunately, block synchronization stopped after some time in the --debug mode. After starting in normal mode, the synchronization has started. Please tell me, for what reasons could this happen?

Mihail-source-arch:
Thank you very much for your reply. Everything is fine now. I will definitely study the material that you sent me a link to.

vivian1912:
Hi @Mihail-source-arch,  For your question, please refer to this [documentation](https://developers.tron.network/docs/faq#20-is-there-a-way-to-temporarily-bypass-the-transaction-execution-time-limit-for-testing-or-debugging-purposes).

As this issue has been inactive for a while, we’ll go ahead and close it.
If you have any further questions, feel free to reopen it anytime. Thanks for your contribution!

PR #6719, opened 2026-04-28
feat(vm): add node-level vm.constantCallTimeoutMs for constant calls

### LINKED ISSUE #6681: [Feature] Add a node-level configuration to control the TVM execution time limit for constant calls
# Background

In the TRON network, the `triggerconstantcontract` API (and the Ethereum-compatible `eth_call`) is widely used for constant calls — including state queries via `view`/`pure` functions and transaction simulation (dry-runs). However, both are constrained by the TVM execution time limit (mainnet default: 80ms). When executing complex contract logic or intensive state access, this can lead to `out_of_time` exceptions, reducing the reliability of development tools, data services, and analytics platforms.

Running the FullNode with the `--debug` flag can remove the TVM timeout restriction for constant calls, but it also removes the timeout limit for transactions during normal block processing. As discussed in #6266, this can cause the local execution result to diverge from the result carried in the block (the block may report `OUT_OF_TIME` while the local node does not), which in turn causes the node to stop syncing. `--debug` is therefore not suitable for production nodes.

This issue supersedes #6288. After extensive discussion in that thread, the community reached consensus that a node-level configuration is preferable to the originally proposed request-body parameter approach. This new issue reflects the revised direction.

Related issues: #6266, #6288

# Rationale

Node operators should be able to extend the TVM execution time window **for constant calls only**, without affecting block validation behavior. This would:

- Improve the stability of off-chain analysis tools and data services
- Prevent false negatives when querying or simulating complex contracts
- Reduce misleading error codes that confuse developers
- Avoid the sync-halt side effect of `--debug`

## What are the use-cases?

- Data services performing bulk state queries via contract calls
- Deep cross-contract call simulation
- Simulating complex logic (e.g., loops or large mappings) that exceed the default execution window

## Why a node-level config instead of a request parameter?

As discussed in #6288, we decided against adding a `timeout_ms` field to the request body, for the following reasons:

- **API standard compatibility.** Adding a non-standard field to `eth_call` would diverge from the Ethereum JSON-RPC specification and complicate integration for Ethereum tooling ([ref](https://github.com/tronprotocol/java-tron/issues/6288#issuecomment-2811949987)).
- **Separation of responsibility.** Execution policy belongs to the node operator, not the caller. Operators that need extended simulation windows opt in explicitly; public-facing nodes keep the default ([ref](https://github.com/tronprotocol/java-tron/issues/6288#issuecomment-2918519825)).
- **No consensus risk.** Constant calls are not broadcast on-chain, so a node-level knob does not risk SR consensus divergence ([ref](https://github.com/tronprotocol/java-tron/issues/6288#issuecomment-2918861325)).
- **Backward compatible.** No API surface change, no SDK updates required.

# Specification

Introduce a new node configuration option (name TBD, e.g. `vm.constantCallTimeoutMs`) in `config.conf`:

```hocon
vm {
  # Max TVM execution time (ms) for constant calls
  # (triggerconstantcontract / eth_call). Default: 80.
  # Must be within [80, MAX_ALLOWED].
  constantCallTimeoutMs = 80
}
```

Behavior:

- Applies **only** to constant calls (`triggerconstantcontract`, `eth_call`, and equivalent read-only execution paths).
- Does **not** affect TVM execution during block processing or transaction broadcasting.
- A strict upper bound (e.g. 500ms) must be enforced to bound DoS risk.
- The effective timeout used for a given call should be observable (e.g. logged / returned in diagnostics) so operators can reason about behavior.

# Safeguards

Regardless of the exact value chosen, extending the window increases per-call cost. The implementation should include:

- A hard upper bound on the configurable value.
- Rate limiting / concurrency cap for constant calls, especially when the configured timeout exceeds the default.
- Clear operator-facing documentation describing the trade-off and recommended values for public vs. private node deployments.

# Test Specification

- Unit tests covering: default behavior unchanged; custom timeout respected for constant calls; block-processing timeout unchanged regardless of the new config.
- Integration tests verifying that a node with an extended `constantCallTimeoutMs` still syncs correctly against the main network (no `OUT_OF_TIME` divergence like `--debug`).
- Stress tests under concurrent constant-call load at the configured upper bound.

# Scope Of Impact

- **Non-breaking.** No API signature change; existing clients and SDKs unaffected.
- TVM executor must branch on call context (block processing vs. constant call) when selecting the timeout.
- Operator documentation and release notes must describe the new option and its risks.

# Implementation

Tentative outline:

- Add the config field and wire it through `Args` / VM config.
- Plumb a per-invocation timeout into the constant-call execution path (TVM executor) without touching the block-processing path.
- Enforce the upper bound at config-load time; reject invalid values with a clear error.
- Add rate-limiting hooks around constant-call execution.

Yes.

yanghang8612:
**What does this PR do?**

Adds a node-level config `vm.constantCallTimeoutMs` (any positive integer that fits VM deadline conversion; omit the property entirely to keep current behaviour) that extends the TVM execution-time window for the read-only RPC paths only — `triggerconstantcontract`, `triggersmartcontract` dispatched to view/pure functions, `estimateenergy`, `eth_call`, `eth_estimateGas`, and any other RPC routed through `Wallet.callConstantContract`. The block-processing deadline is unchanged.

When set, the configured value is used **verbatim** as the per-call deadline (no clamp against the network's `maxCpuTimeOfOneTx`, no `cpuLimitInUsRatio` scaling). When unset, constant calls fall back to the existing path that uses the network's `maxCpuTimeOfOneTx`.

Closes #6681.

`triggerconstantcontract` / `eth_call` and the other constant-call RPCs are widely used for state queries and dry-runs but cap out at the same TVM deadline as block-processing (mainnet 80 ms). Operators currently have only `--debug` to extend this window, which **also** removes the block-processing timeout — unsafe for any node serving traffic, since it can desync from the network (#6266). This PR gives operators a scoped lever for the constant-call window without that side effect.

**Design**

- **Block-processing is structurally unaffected.** The new branch is gated by `isConstantCall` in `VMActuator.create()` / `call()`, so it is unreachable from block validation — guaranteed by structure, not by convention.
- **Verbatim semantics.** When the operator sets a value, it is used as the deadline as-is. No `max(configured, networkMs)` clamp, no `cpuLimitInUsRatio` scaling. Operators get exactly the budget they configured.
- **Opt-in detected by `Config#hasPath`.** `vm.constantCallTimeoutMs` is intentionally absent from `reference.conf`, and the `VmConfig` field is excluded from `ConfigBeanFactory` auto-binding (`@Setter(AccessLevel.NONE)`) so the strict schema check doesn't require the path. `vmSection.hasPath(...)` then becomes a precise signal of "operator wrote this in `config.conf`". Same pattern as `BlockConfig` rejecting the legacy `committee.proposalExpireTime` location.
- **Validation only on opt-in.** When the operator configures the value, it must be `> 0` and fit VM deadline conversion (`<= Long.MAX_VALUE / 1000` ms); explicit `0`, negative, and overflowing values are rejected at config-load with `IllegalArgumentException`. Nodes that don't configure the property keep the existing default behaviour — no validation, no behaviour change.
- **Sub-frames inherit the deadline.** Cross-contract simulation uses the existing `Program` plumbing (`getVmShouldEndInUs()` propagated to internal `ProgramInvoke` instances), so the override applies to deep call trees — the headline use case.

- Unit Tests
  - `VmConfigTest`: default `0L` when absent; positive values accepted (`1`, `50`, `500`, `5000`); explicit `= 0`, negative values, and overflowing values rejected.
  - `VMActuatorTest`: configured constant-call timeout is used verbatim; unset constant calls fall back to the network-derived deadline; non-constant/block-processing paths ignore the override.
  - Existing `VMActuator` and `Wallet` callers' tests remain green.

**Migration from `--debug`**

Operators currently running with `--debug` **purely** to extend the constant-call window should switch to `vm.constantCallTimeoutMs=<ms>` once this lands. `--debug` additionally extends the block-processing timeout, which is unsafe for any node serving traffic (see #6266). The config reference includes this migration note; release notes should carry the same pointer.

### DIFF
--- a/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
+++ b/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
@@ -398,13 +398,9 @@ public class VMActuator implements Actuator2 {
       byte[] ops = newSmartContract.getBytecode().toByteArray();
       rootInternalTx = new InternalTransaction(trx, trxType);
 
-      long maxCpuTimeOfOneTx = rootRepository.getDynamicPropertiesStore()
-          .getMaxCpuTimeOfOneTx() * VMConstant.ONE_THOUSAND;
-      long thisTxCPULimitInUs = (long) (maxCpuTimeOfOneTx * getCpuLimitInUsRatio());
-      long constantCallTimeoutMs = CommonParameter.getInstance().getConstantCallTimeoutMs();
-      if (isConstantCall && constantCallTimeoutMs > 0L) {
-        thisTxCPULimitInUs = constantCallTimeoutMs * VMConstant.ONE_THOUSAND;
-      }
+      long thisTxCPULimitInUs = calculateCpuLimitInUs(isConstantCall,
+          rootRepository.getDynamicPropertiesStore().getMaxCpuTimeOfOneTx(),
+          getCpuLimitInUsRatio(), CommonParameter.getInstance().getConstantCallTimeoutMs());
       long vmStartInUs = System.nanoTime() / VMConstant.ONE_THOUSAND;
       long vmShouldEndInUs = vmStartInUs + thisTxCPULimitInUs;
       ProgramInvoke programInvoke = ProgramInvokeFactory
@@ -516,13 +512,9 @@ public class VMActuator implements Actuator2 {
         energyLimit = getTotalEnergyLimit(creator, caller, contract, feeLimit, callValue);
       }
 
-      long maxCpuTimeOfOneTx = rootRepository.getDynamicPropertiesStore()
-          .getMaxCpuTimeOfOneTx() * VMConstant.ONE_THOUSAND;
-      long thisTxCPULimitInUs = (long) (maxCpuTimeOfOneTx * getCpuLimitInUsRatio());
-      long constantCallTimeoutMs = CommonParameter.getInstance().getConstantCallTimeoutMs();
-      if (isConstantCall && constantCallTimeoutMs > 0L) {
-        thisTxCPULimitInUs = constantCallTimeoutMs * VMConstant.ONE_THOUSAND;
-      }
+      long thisTxCPULimitInUs = calculateCpuLimitInUs(isConstantCall,
+          rootRepository.getDynamicPropertiesStore().getMaxCpuTimeOfOneTx(),
+          getCpuLimitInUsRatio(), CommonParameter.getInstance().getConstantCallTimeoutMs());
       long vmStartInUs = System.nanoTime() / VMConstant.ONE_THOUSAND;
       long vmShouldEndInUs = vmStartInUs + thisTxCPULimitInUs;
       ProgramInvoke programInvoke = ProgramInvokeFactory
@@ -697,6 +689,14 @@ public class VMActuator implements Actuator2 {
     }
 
     return cpuLimitRatio;
+  }
+
+  static long calculateCpuLimitInUs(boolean isConstantCall, long maxCpuTimeOfOneTxMs,
+      double cpuLimitInUsRatio, long constantCallTimeoutMs) {
+    if (isConstantCall && constantCallTimeoutMs > 0L) {
+      return constantCallTimeoutMs * VMConstant.ONE_THOUSAND;
+    }
+    return (long) (maxCpuTimeOfOneTxMs * VMConstant.ONE_THOUSAND * cpuLimitInUsRatio);
   }
 
   public long getTotalEnergyLimitWithFixRatio(AccountCapsule creator, AccountCapsule caller,
--- a/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
+++ b/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
@@ -401,6 +401,10 @@ public class VMActuator implements Actuator2 {
       long maxCpuTimeOfOneTx = rootRepository.getDynamicPropertiesStore()
           .getMaxCpuTimeOfOneTx() * VMConstant.ONE_THOUSAND;
       long thisTxCPULimitInUs = (long) (maxCpuTimeOfOneTx * getCpuLimitInUsRatio());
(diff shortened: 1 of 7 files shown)


yanghang8612:
Revised the implementation per the review thread on #6681 and force-pushed `feat/constant-call-timeout` (history collapsed from four commits to two: one `feat(vm)` + one `test(vm)`).

Major changes vs. the original four commits:

1. **Validation range relaxed, with overflow guard.** The `[80, 500]` ms bound is dropped. Any positive integer that fits VM deadline conversion is accepted; explicit `0`, negative values, and overflowing values are rejected.
2. **Concurrency cap removed.** `vm.constantCallMaxConcurrency`, the `Semaphore`, the `ThreadLocal` reentrancy guard, and the `WalletConstantCallThrottleTest` suite are all gone. Operators handle throttling above the node (reverse proxy, API gateway, per-IP limits) — hard-gating at the node would false-positive on deployments that already throttle externally.
3. **Configured value is used verbatim.** No more `max(configured, networkMs)` clamp and no `cpuLimitInUsRatio` scaling on the constant-call branch. The operator's configured ms is the per-call deadline, full stop.
4. **`VmTimeoutPolicy` helper removed.** With the simplified policy, the deadline computation is a 4-line in-place patch in `VMActuator.create()` / `call()`. The helper class and its 12-case test suite are gone.
5. **Opt-in detection via `hasPath`.** `vm.constantCallTimeoutMs` is intentionally absent from `reference.conf`, and the field is excluded from `ConfigBeanFactory` binding (`@Setter(AccessLevel.NONE)`). This pattern follows `BlockConfig`'s rejection of the legacy `committee.proposalExpireTime` path: when a key is absent from `reference.conf` and isn't a bound bean property, `Config#hasPath` precisely reflects "operator wrote this in `config.conf`".

Net diff stays small and now includes focused deadline-policy coverage plus overflow validation.

aiden3885 (approved):

CodeNinjaEvan (approved):
LGTM

PR #6722, opened 2026-04-28
fix(jsonrpc): accept "input" as alias for "data" in call args

### LINKED ISSUE #6517: Are there plans to support parameter passing via the `input` field for `eth_call`?
As stated in the title, when calling `eth_call` in TRON's JSON-RPC, changing the parameter name from `data` to `input` will result in a JSON parsing exception.

waynercheung:
### What does this PR do?

Accept `input` alongside `data` on `eth_call`, `eth_estimateGas`, and `buildTransaction` call arguments, matching the Ethereum execution-apis spec and resolving #6517.

- Add `input` (raw field) to `CallArguments` and `BuildArguments`.
- Resolve calldata via `resolveData()` - pure precedence resolution with `input` winning over `data`, mirroring go-ethereum's `TransactionArgs.data()`.
- On the build path, conflicts between `input` and `data` (both set, not equal) are detected at `BuildArguments.getContractType()` - mirroring geth's `data()` / `setDefaults` split. The error message uses go-ethereum's wording verbatim. The query path stays lenient (`input` wins silently).
- Switch five production read sites in `TronJsonRpcImpl` from `getData()` to `resolveData()`.
- Hex validation is mode-driven via `JsonRpcApiUtil.requireValidHex` (`HexMode.STRICT` / `HexMode.LENIENT`):
  - `input` is STRICT: follows the execution-apis BYTES schema (`^0x[0-9a-f]*$`, even length; `""` is accepted as empty bytes per geth's `hexutil.Bytes.UnmarshalText`).
  - `data` keeps LENIENT parsing via `ByteArray.fromHexString` (accepts bare hex and odd length) for backward compatibility with existing callers - notably `BuildTransactionTest.testCreateSmartContract`, which submits bare-hex bytecode.
- The resolver is named `resolveData()` - verb prefix, not `getXxx` - so Jackson's and FastJSON's JavaBean introspection skips it with no per-serializer annotations needed.
- The Lombok-generated positional constructors widen by one slot (`CallArguments` 7->8 args, `BuildArguments` 16->17 args). The named-JSON deserialisation path is unaffected.

### Why are these changes required?

The Ethereum execution-apis spec uses `input` as the canonical calldata field; `data` is the legacy alias. go-ethereum accepts both, with `input` winning when equal and erroring when not. Notably, go-ethereum's `ethclient` has emitted only `input` since [ethereum/go-ethereum#28078](https://github.com/ethereum/go-ethereum/pull/28078), so any tron RPC user upgrading their geth client side hits this.

Before this PR, java-tron only read `data`. A spec-compliant client carrying only `input` failed two ways:

1. Jackson rejected the unknown property (`FAIL_ON_UNKNOWN_PROPERTIES=true`), returning `-32603` instead of executing the call.
2. Even when the field was tolerated, contract-creation paths (no `to`) tripped `invalid json request` because calldata appeared absent.

Both symptoms are captured in #6517.

### This PR has been tested by:

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

`CallArgumentsTest` (23 tests) + `BuildArgumentsTest` (29 tests) = 52 total. Coverage:

- Single-field cases: `input` only, `data` only, neither.
- Both fields equal — precedence resolves cleanly.
- Both fields not equal: `CallArguments` returns `input` silently; `BuildArguments.getContractType()` throws with geth-equivalent wording.
- Empty-string `""` is presence (matches geth).
- `0x` (zero-length payload) is presence and never collapses to absent.
- Case-insensitive byte equality (`0xDEAD == 0xdead`).
- Strict `input` validation rejects bare hex, odd length, non-hex chars.
- Lenient `data` validation still accepts bare hex and odd length; `BuildTransactionTest.testCreateSmartContract` still passes.
- Loser-field validation: malformed `data` is rejected even when `input` is the precedence winner, and vice versa.
- Jackson + FastJSON serialisation: `resolveData` never appears as a wire property and serialisation never throws even with conflicting `input` / `data` (regression guard for the verb-prefix naming).
- #6517 reproductions: Jackson deserialisation of `{"input": "0xdeadbeef"}` succeeds; contract-creation via `input` (no `to`) resolves to `CreateSmartContract`.

### Compatibility notes

Requests sending only `data` with valid hex are byte-for-byte unchanged. Two error paths shift, both improvements:

- **Issue #6517 itself** - any request carrying `input`. Pre-PR Jackson rejects the unknown property; jsonrpc4j surfaces this as `-32700 "JSON parse error"` with `id: "null"`. Post-PR the request parses cleanly and either succeeds or returns a specific `-32602` (on malformed hex or build-path conflict).

- **Invalid hex in `data`** (e.g. `data: "0xzz"`) - pre-PR BouncyCastle's `DecoderException` falls through to jsonrpc4j's default resolver:
```
code -32001
message "exception decoding Hex string: invalid characters encountered in Hex string"
data "org.bouncycastle.util.encoders.DecoderException"
```

Post-PR `requireValidHex` catches and rethrows as `JsonRpcInvalidParamsException`, mapped via `@JsonRpcError` to standard `-32602`:
```
code -32602
message "invalid hex string for "data""
data "{}"
```

Standard JSON-RPC code, deterministic message, no internal class leak. Clients hardcoded against `-32700` (#6517 symptom) or `-32001` (the BC leak) will see different behaviour.

### Follow up

- Add a JsonRpcServer-level integration test exercising the full Jackson -> handler -> response path for both `input` and `data` payloads. Useful but not blocking - the unit-level path is fully covered.
- Tighten `ByteArray.fromHexString` to the execution-apis BYTES schema, then drop the asymmetric branch in `requireValidHex`. Behaviour-changing for non-RPC callers, so deferred to a separate change.

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
@@ -442,6 +442,50 @@ public class JsonRpcApiUtil {
     return StringUtils.isEmpty(quantity) || quantity.equals("0x0");
   }
 
+  /**
+   * Validation mode for {@link #requireValidHex}.
+   */
+  public enum HexMode {
+    /**
+     * Execution-apis BYTES schema: requires {@code 0x} prefix and
+     * even total length; {@code ""} is accepted as empty bytes per
+     * geth's {@code hexutil.Bytes.UnmarshalText}.
+     */
+    STRICT,
+    /**
+     * {@link ByteArray#fromHexString}'s lenient parsing: accepts bare
+     * hex and odd-length input. Kept for backward compatibility.
+     */
+    LENIENT
+  }
+
+  /**
+   * Throws if {@code value} is not parseable hex under the given
+   * {@code mode}. {@code null} is treated as absent and returns
+   * silently. {@code fieldName} is used only in error messages.
+   */
+  public static void requireValidHex(String fieldName, String value, HexMode mode)
+      throws JsonRpcInvalidParamsException {
+    if (value == null) {
+      return;
+    }
+    if (mode == HexMode.STRICT) {
+      if (value.isEmpty()) {
+        return;
+      }
+      if (!value.startsWith("0x") || value.length() % 2 != 0) {
+        throw new JsonRpcInvalidParamsException(
+            "invalid hex string for \"" + fieldName + "\"");
+      }
+    }
+    try {
+      ByteArray.fromHexString(value);
+    } catch (Exception e) {
+      throw new JsonRpcInvalidParamsException(
+          "invalid hex string for \"" + fieldName + "\"");
+    }
+  }
+
   public static long parseQuantityValue(String value) throws JsonRpcInvalidParamsException {
     long callValue = 0L;
 
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
@@ -4,8 +4,10 @@ import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.addressCompatibleToB
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.paramQuantityIsNull;
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.paramStringIsNull;
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.parseQuantityValue;
+import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.requireValidHex;
 
 import com.google.protobuf.ByteString;
+import java.util.Arrays;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -13,9 +15,11 @@ import lombok.Setter;
 import lombok.ToString;
 import org.apache.commons.lang3.StringUtils;
 import org.tron.api.GrpcAPI.BytesMessage;
+import org.tron.common.utils.ByteArray;
 import org.tron.core.Wallet;
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidParamsException;
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidRequestException;
+import org.tron.core.services.jsonrpc.JsonRpcApiUtil.HexMode;
 import org.tron.protos.Protocol.Transaction.Contract.ContractType;
 import org.tron.protos.contract.SmartContractOuterClass.SmartContract;
 
@@ -42,6 +46,9 @@ public class BuildArguments {
   @Getter
   @Setter
   private String data;
+  @Getter
+  @Setter
+  private String input;
   @Getter
   @Setter
   private String nonce = ""; //not used
@@ -83,16 +90,50 @@ public class BuildArguments {
     gasPrice = args.getGasPrice();
     value = args.getValue();
     data = args.getData();
+    input = args.getInput();
+  }
+
+  /**
+   * Returns {@code input} if non-null, else {@code data}. Pure
+   * precedence resolution, mirroring go-ethereum's
+   * {@code TransactionArgs.data()}.
+   *
+   * <p>Both fields are first validated by
+   * {@link org.tron.core.services.jsonrpc.JsonRpcApiUtil#requireValidHex}
+   * — strict for {@code input}, lenient for {@code data} (see that
+   * method for the rules).
+   *
+   * <p>Conflict between {@code input} and {@code data} is not checked
+   * here. Build-path callers must route through
+   * {@link #getContractType(Wallet)} for the geth-equivalent
+   * {@code setDefaults} enforcement.
+   *
+   * <p>Java callers using positional constructors should pass
+   * {@code null} (not {@code ""}) for unset {@code input}.
+   *
+   * <p>Verb-prefix name (not {@code getXxx}) keeps Jackson and
+   * FastJSON's JavaBean introspection from invoking it during
+   * serialisation; two regression tests per DTO pin this invariant.
+   */
+  public String resolveData() throws JsonRpcInvalidParamsException {
+    requireValidHex("input", input, HexMode.STRICT);
+    requireValidHex("data", data, HexMode.LENIENT);
+    return input != null ? input : data;
   }
 
   public ContractType getContractType(Wallet wallet) throws JsonRpcInvalidRequestException,
       JsonRpcInvalidParamsException {
+    // Fail fast on bad hex / conflict before the state lookup;
+    // calldataEquals relies on resolveData() having validated hex first.
+    String resolvedData = resolveData();
+    validateCallDataConflict();
+
     ContractType contractType;
 
     // to is null
     if (paramStringIsNull(to)) {
       // data is null
-      if (paramStringIsNull(data)) {
+      if (paramStringIsNull(resolvedData)) {
         throw new JsonRpcInvalidRequestException("invalid json request");
       }
 
@@ -134,6 +175,26 @@ public class BuildArguments {
 
   private boolean availableTransferAsset() {
     return tokenId > 0 && tokenValue > 0 && paramQuantityIsNull(value);
+  }
+
+  /**
+   * Throws when both fields decode to non-equal bytes. Wording matches
+   * geth's setDefaults so existing tooling can detect the error string.
+   */
+  private void validateCallDataConflict() throws JsonRpcInvalidParamsException {
+    if (input != null && data != null && !calldataEquals(input, data)) {
+      throw new JsonRpcInvalidParamsException(
+          "both \"data\" and \"input\" are set and not equal. "
+              + "Please use \"input\" to pass transaction call data");
+    }
+  }
+
+  /**
+   * Byte-level equality, so {@code "0xDEAD"} equals {@code "0xdead"}. Both
+   * args must have passed {@code requireValidHex} first.
+   */
+  private static boolean calldataEquals(String a, String b) {
+    return Arrays.equals(ByteArray.fromHexString(a), ByteArray.fromHexString(b));
   }
 
 }
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
@@ -28,6 +28,16 @@ import org.tron.protos.contract.SmartContractOuterClass.SmartContract;
 @ToString
 public class BuildArguments {
 
+  /**
+   * Conflict error message wording. Mirrors go-ethereum's
+   * {@code setDefaults} verbatim — external EVM tooling may
+   * pattern-match this string. Do not change the wording without
+   * coordinating with downstream consumers.
+   */
+  private static final String CONFLICT_ERR_MSG =
+      "both \"data\" and \"input\" are set and not equal. "
+          + "Please use \"input\" to pass transaction call data";
+
   @Getter
   @Setter
   private String from;
@@ -183,9 +193,7 @@ public class BuildArguments {
    */
   private void validateCallDataConflict() throws JsonRpcInvalidParamsException {
     if (input != null && data != null && !calldataEquals(input, data)) {
-      throw new JsonRpcInvalidParamsException(
-          "both \"data\" and \"input\" are set and not equal. "
-              + "Please use \"input\" to pass transaction call data");
+      throw new JsonRpcInvalidParamsException(CONFLICT_ERR_MSG);
     }
   }
 
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
@@ -3,6 +3,7 @@ package org.tron.core.services.jsonrpc.types;
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.addressCompatibleToByteArray;
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.paramStringIsNull;
 import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.parseQuantityValue;
+import static org.tron.core.services.jsonrpc.JsonRpcApiUtil.requireValidHex;
 
 import com.google.protobuf.ByteString;
 import lombok.AllArgsConstructor;
@@ -15,6 +16,7 @@ import org.tron.api.GrpcAPI.BytesMessage;
 import org.tron.core.Wallet;
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidParamsException;
 import org.tron.core.exception.jsonrpc.JsonRpcInvalidRequestException;
+import org.tron.core.services.jsonrpc.JsonRpcApiUtil.HexMode;
 import org.tron.protos.Protocol.Transaction.Contract.ContractType;
 import org.tron.protos.contract.SmartContractOuterClass.SmartContract;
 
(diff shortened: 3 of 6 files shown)


0xbigapple:
LGTM

0xbigapple (approved):

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -442,6 +442,50 @@ public static boolean paramQuantityIsNull(String quantity) {
...
+  public static void requireValidHex(String fieldName, String value, HexMode mode)
+      throws JsonRpcInvalidParamsException {
+    if (value == null) {
+      return;
+    }
+    if (mode == HexMode.STRICT) {
+      if (value.isEmpty()) {
+        return;
+      }
+      if (!value.startsWith("0x") || value.length() % 2 != 0) {
```
bladehan1, line 476: [SHOULD] STRICT mode rejects 0X (uppercase) prefix, diverging from execution-apis spec

`value.startsWith("0x")` is case-sensitive, so an `input` like `0Xdeadbeef` is rejected with `invalid hex string for "input"`. go-ethereum's `hexutil.has0xPrefix` accepts both `0x` and `0X` (`input[1] == 'x' || input[1] == 'X'`), so a spec-compliant client emitting `0X...` is bounced by java-tron but accepted by geth — the same regression class this PR set out to fix.

Note that `ByteArray.fromHexString` itself is case-sensitive on the prefix and `Hex.decode` rejects `X` as a digit, so even after relaxing this gate the STRICT path needs to normalize the prefix (e.g. `value = "0x" + value.substring(2)`) before delegating to `fromHexString`, otherwise the call still throws downstream.

Suggestion: Make the STRICT prefix check case-insensitive and normalize the prefix before passing to `ByteArray.fromHexString`; add a `resolveData_inputUppercase0X_accepted` test in both `CallArgumentsTest` and `BuildArgumentsTest`.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
```diff
@@ -136,4 +177,24 @@ private boolean availableTransferAsset() {
...
+              + "Please use \"input\" to pass transaction call data");
+    }
+  }
+
+  /**
+   * Byte-level equality, so {@code "0xDEAD"} equals {@code "0xdead"}. Both
+   * args must have passed {@code requireValidHex} first.
+   */
+  private static boolean calldataEquals(String a, String b) {
+    return Arrays.equals(ByteArray.fromHexString(a), ByteArray.fromHexString(b));
```
bladehan1, line 205: [SHOULD] calldataEquals depends on an undocumented invariant that fromHexString("") never throws

The inline comment at line 127 states "calldataEquals relies on resolveData() having validated hex first", but `requireValidHex("input", "", STRICT)` short-circuits on `value.isEmpty()` and never calls `ByteArray.fromHexString`. So when `input=""` and `data="0xdeadbeef"`, conflict detection enters `calldataEquals` and calls `fromHexString("")` for the first time. Today this happens to return `[]` (because `Hex.decode("")` returns an empty array), so the test passes — but the PR description's own follow-up ("Tighten ByteArray.fromHexString to the execution-apis BYTES schema") would make this throw, and the unwrapped `RuntimeException` would escape `validateCallDataConflict`, get caught by the outer RPC handler, and downgrade the error code from -32602 to -32603.

Suggestion: Either wrap `calldataEquals` in a try-catch that rethrows as `JsonRpcInvalidParamsException`, or run a LENIENT validation on `input` before invoking `validateCallDataConflict` so `fromHexString` is exercised once before the conflict path needs it.

framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
```diff
@@ -43,21 +45,41 @@ public class CallArguments {
...
 
+  /**
+   * Returns {@code input} if non-null, else {@code data}. Pure
+   * precedence resolution, mirroring go-ethereum's
+   * {@code TransactionArgs.data()}; no conflict check on the query
+   * path (matches geth's {@code ToMessage}). See
+   * {@link BuildArguments#resolveData()} for the rationale on
+   * naming, validation split, and serialiser interaction.
+   */
+  public String resolveData() throws JsonRpcInvalidParamsException {
```
bladehan1, line 61: [SHOULD] resolveData() relies solely on a fragile naming convention; add explicit serializer-ignore annotations

The verb-prefix idiom (`resolveData` instead of `getResolvedData`) is the only reason Jackson and FastJSON skip this method during bean introspection. The 4 regression tests pin this for the default `ObjectMapper` and default `JSON.toJSONString` paths, but they cannot cover: (a) a future contributor renaming it to `getXxx`, (b) a custom `ObjectMapper` opting into `MapperFeature.ALLOW_FIELDS_WITHOUT_GETTERS`, (c) a Jackson MixIn injected at the project level, or (d) a subclass adding `@JsonProperty` that retunes the bean model. If the invariant ever breaks, the serializer invokes `resolveData()`, which throws `JsonRpcInvalidParamsException` for any conflicting in-flight DTO and breaks the whole RPC response path.

The same applies to `BuildArguments.resolveData()` at line 118.

Suggestion: Annotate both `resolveData()` methods with `@com.fasterxml.jackson.annotation.JsonIgnore` and `@com.alibaba.fastjson.annotation.JSONField(serialize = false, deserialize = false)` as a belt-and-braces guard.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
```diff
@@ -136,4 +177,24 @@ private boolean availableTransferAsset() {
...
   }
 
+  /**
+   * Throws when both fields decode to non-equal bytes. Wording matches
+   * geth's setDefaults so existing tooling can detect the error string.
+   */
+  private void validateCallDataConflict() throws JsonRpcInvalidParamsException {
+    if (input != null && data != null && !calldataEquals(input, data)) {
+      throw new JsonRpcInvalidParamsException(
+          "both \"data\" and \"input\" are set and not equal. "
```
bladehan1: [NIT] Extract the conflict error string to a constant with a do-not-change javadoc

The message `"both \"data\" and \"input\" are set and not equal. Please use \"input\" to pass transaction call data"` is a cross-project contract — external tooling may match this exact string to detect the geth/setDefaults error. Inlined as a literal it's too easy for a future refactor to "improve" the wording (e.g. wrap in `String.format` to include both byte values) and silently break that contract.

Suggestion: Lift to a `private static final String` field (e.g. `CONFLICT_ERR_MSG`) and add a one-line javadoc "wording mirrors go-ethereum's setDefaults; do not change — external tooling may match this string".

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
```diff
@@ -83,16 +90,50 @@ public BuildArguments(CallArguments args) {
...
+    requireValidHex("input", input, HexMode.STRICT);
+    requireValidHex("data", data, HexMode.LENIENT);
+    return input != null ? input : data;
   }
 
   public ContractType getContractType(Wallet wallet) throws JsonRpcInvalidRequestException,
       JsonRpcInvalidParamsException {
+    // Fail fast on bad hex / conflict before the state lookup;
+    // calldataEquals relies on resolveData() having validated hex first.
+    String resolvedData = resolveData();
```
bladehan1, line 138: [NIT] Hex validation runs twice on the build path

`getContractType()` calls `resolveData()` (which runs `requireValidHex` on both fields and decodes the winning value), and the handler later calls `args.resolveData()` again before `ByteArray.fromHexString(...)` to set the bytecode (`TronJsonRpcImpl` lines 1132 and 1177). For a 24KB CreateSmartContract bytecode this is two redundant hex decodes per request — small but measurable.

Suggestion: Either return a tuple of (ContractType, resolvedCalldata) from `getContractType` so the handler can reuse the already-validated result, or document on `resolveData()` that callers should invoke it once per request.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
```diff
@@ -83,16 +90,50 @@ public BuildArguments(CallArguments args) {
...
+    validateCallDataConflict();
+
     ContractType contractType;
 
     // to is null
     if (paramStringIsNull(to)) {
```
bladehan1, line 144: [NIT] Align javadoc/PR description with paramStringIsNull collapsing "" to absent

The javadoc and PR description both state that `""` is presence (matching geth's `hexutil.Bytes.UnmarshalText`), but `paramStringIsNull("")` returns true and treats `""` as absent — for example, `BuildArguments.getContractType()` with `input=""` and `data=null` falls into the `paramStringIsNull(resolvedData)` branch and throws "invalid json request" rather than recognising presence. The observable behaviour is correct (geth also rejects empty bytecode for contract creation), but the internal semantics drift from the stated invariant, which makes future readers misjudge edge cases.

Suggestion: Update the javadoc to clarify that on the wire `""` is presence, but inside the DTO `paramStringIsNull` collapses both `""` and `0x` to absent, matching the "no bytecode" semantics geth enforces in `setDefaults`.

framework/src/test/java/org/tron/core/services/jsonrpc/BuildArgumentsTest.java
```diff
@@ -29,45 +30,299 @@ public class BuildArgumentsTest extends BaseTest {
   public void initBuildArgs() {
     buildArguments = new BuildArguments(
```
bladehan1, line 31: [NIT] Replace 17-arg positional constructor in test fixture with setter chain

`initBuildArgs()` and `testBuildArgument()` build instances via `@AllArgsConstructor`'s positional form; this PR widened both signatures (16→17 and 7→8) and the next field addition will force the same edit again. Other test classes in the module (`WalletCursorTest`, `BuildTransactionTest`, `JsonRpcTest`) already use `new BuildArguments()` plus setters and are immune to this churn. The 17-arg literal is also hard to read — a mix of address strings, decimals, longs, and a trailing boolean.

Suggestion: Convert these two fixtures to no-arg constructor + setter calls so they don't need touching the next time the DTO grows.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -442,6 +442,50 @@ public static boolean paramQuantityIsNull(String quantity) {
...
+     */
+    LENIENT
+  }
+
+  /**
+   * Throws if {@code value} is not parseable hex under the given
+   * {@code mode}. {@code null} is treated as absent and returns
+   * silently. {@code fieldName} is used only in error messages.
+   */
+  public static void requireValidHex(String fieldName, String value, HexMode mode)
```
bladehan1, line 467: [NIT] Add direct unit tests for requireValidHex / HexMode

`HexMode` and `requireValidHex` are designed as a reusable utility (the comment hints at future fields adopting it), but coverage today comes only via `CallArgumentsTest` / `BuildArgumentsTest`. When the next caller (e.g. an `accessList` field, or a future RPC method) adopts it, regressing the utility itself will manifest as failures in those callers, which is hard to debug.

Suggestion: Add a `JsonRpcApiUtilTest.requireValidHex_*` series covering null, `""` for both modes, `0x` only, valid hex, no-prefix, odd-length, non-hex digits, and `0X` (uppercase) — the last one fails before the case-sensitivity fix and pins it once landed.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
```diff
@@ -136,4 +177,24 @@ private boolean availableTransferAsset() {
     return tokenId > 0 && tokenValue > 0 && paramQuantityIsNull(value);
   }
 
+  /**
+   * Throws when both fields decode to non-equal bytes. Wording matches
+   * geth's setDefaults so existing tooling can detect the error string.
+   */
+  private void validateCallDataConflict() throws JsonRpcInvalidParamsException {
```
bladehan1, line 194: [NIT] Document the precondition that calldataEquals requires pre-validated hex

The contract "both args must have passed `requireValidHex` first" lives on `calldataEquals` (line 192-195) but the actual ordering happens up in `getContractType` (line 127-129). A reader visiting `validateCallDataConflict` in isolation has no warning that calling it on un-validated input is unsafe — and the next P1 finding (calldataEquals + `fromHexString("")`) is exactly an instance of that precondition not holding for `input=""`.

Suggestion: Restate the precondition in `validateCallDataConflict`'s javadoc ("caller must ensure both fields have passed `requireValidHex`"), so the contract is documented at every entry point, not only on the leaf method.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
waynercheung: Acknowledged. The micro-overhead is real but bounded - `requireValidHex` on already-validated hex is a single pass over the string. Returning a `(ContractType, calldata)` tuple from `getContractType` would widen the public API beyond this PR's scope, and memoizing on a mutable POJO has correctness concerns. Leaving as-is; if profiling on the build path ever flags this, a dedicated optimization PR.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
waynercheung: Each layer documents itself correctly: `HexMode.STRICT`'s javadoc covers wire-level acceptance ("`""` is accepted as empty bytes per geth's `hexutil.Bytes.UnmarshalText`"), and `paramStringIsNull` is a general-purpose helper whose collapse-to-absent behavior is its own contract used across the jsonrpc layer. `resolveData()` itself doesn't claim `""` is presence; it points to `requireValidHex` for validation rules. End-to-end behavior is pinned by `resolveData_inputEmpty_dataNonEmpty_inputWinsAsEmpty` and `getContractType_inputEmptyDataNonEmpty_throwsConflict`. Adding a paragraph that crosses levels would create a synchronization point future refactors can drift from silently. Leaving documentation at its current level.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
waynercheung: Done in 4b9c77911. Lifted to a `private static final CONFLICT_ERR_MSG` with a "do not change wording" javadoc.

framework/src/main/java/org/tron/core/services/jsonrpc/types/BuildArguments.java
waynercheung: `ByteArray.fromHexString("")` returns `[]` today via `Hex.decode("")` and is pinned end-to-end by `getContractType_inputEmptyDataNonEmpty_throwsConflict`. The follow-up tightening of `ByteArray.fromHexString` will likely preserve geth-aligned `""` -> empty-bytes (otherwise STRICT mode self-contradicts). Even if it doesn't, that test would fail immediately and the follow-up author can add try-catch alongside the change that introduces the risk. Defending now would prepay for a hypothetical future change rather than for present correctness.

framework/src/main/java/org/tron/core/services/jsonrpc/types/CallArguments.java
waynercheung: We deliberately removed these in v2 in favor of the verb-prefix naming as a single source of truth. The four regression tests (`jacksonSerialize_*`, `fastjsonSerialize_*` on both DTOs) pin the wire shape at default ObjectMapper / FastJSON settings and would fail immediately on a `getXxx` rename. Custom MixIn or `ALLOW_FIELDS_WITHOUT_GETTERS` scenarios are explicit opt-ins by whoever introduces them — that contributor takes ownership of re-validating the wire shape. Keeping the DTO surface clean and relying on tests + naming.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
waynercheung: Fair point as a test infrastructure step. For this PR's scope (#6517 fix), the utility is exercised end-to-end through `CallArgumentsTest` and `BuildArgumentsTest` across all STRICT/LENIENT branches — utility regressions would surface there. Building out a dedicated `JsonRpcApiUtilTest` is forward-investment for the next caller of `requireValidHex` and most naturally lives alongside the `ByteArray.fromHexString` tightening listed in Follow up.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
waynercheung: Pre-PR `ByteArray.fromHexString` is already case-sensitive on the prefix (`startsWith("0x")` matches lowercase only) and BC's `Hex.decode` rejects `X` as a hex character — so `0X...` was rejected pre-PR too, just surfaced as -32001 with the BC class name leaked. This PR doesn't change `0X` acceptance; it only changes the error code/message. Relaxing the prefix to accept `0X` would also need `ByteArray.fromHexString` updated, which is the shared utility tightening already on the Follow up list. Better to handle prefix policy uniformly there. Will not change in this PR.

framework/src/test/java/org/tron/core/services/jsonrpc/BuildArgumentsTest.java
waynercheung: The 17-arg literal is indeed harder to read. Same argument applies whenever any DTO field is added — converting fixtures to setters is incremental modernization (Builder pattern is the real endpoint), not a single-PR fix. For this PR's scope, the existing fixtures pass and the change is mechanical. If module-wide test fixture cleanup is wanted, a dedicated PR scoping just that change would be cleaner than bundling with the bug fix.

yanghang8612 (approved):
LGTM

bladehan1 (approved):
LGTM

PR #6724, opened 2026-04-28
refactor(db): remove function for periodic backup of RocksDB database

### LINKED ISSUE #6595: [Feature] Remove periodic database backup in favor of dual-node failover
## Problem Statement
The database backup feature was originally introduced to enable fast recovery during disk failover and played a certain role in the early stages of chain development. In the past, there were efforts to optimize this feature, such as completing the implementation of the backup database.

However, as the database size has grown rapidly, a series of negative effects have emerged. For example, long backup times can block block synchronization, causing the drawbacks to significantly outweigh the benefits.

### Why should it be removed ?
The database backup feature is configured as follows:
```
storage.backup = {
  enable = false  // indicate whether enable the backup plugin
  propPath = "prop.properties" // record which bak directory is valid
  bak1path = "bak1/database" // you must set two backup directories to prevent application halt unexpected (e.g. kill -9).
  bak2path = "bak2/database"
  frequency = 10000   // indicate backup db once every 10000 blocks processed.
}
```

When this feature is enabled (enable = true), during execution of pushBlock(final BlockCapsule block), if `block number % frequency == 0`, all databases that implement the RevokingDatabase interface are copied to an alternative directory.

This mechanism was designed to address data corruption caused by sporadic disk failures, power outages, or abrupt process termination (e.g. kill -9) in the early stages of deployment.

Current Major Issues：

- After optimization and extensive testing, it has been confirmed that kill -9 does not corrupt the database, which significantly reduces the necessity of periodic database backups.

- State-related databases are extremely large. As of 2026-01-28, the mainnet state database is close to 3 TB, and a single copy operation can take several hours. During the backup window, the node is unable to synchronize blocks, which poses a fatal risk to service stability.

### Alternative Solution

An alternative to database backup is to deploy dual FullNodes in a primary–backup configuration. The configuration is as follows:
```
node.backup {
  port = 10001
  # my priority, each member should use different priority
  priority = 8
  # peer's ip list, can't contain mine
  members = [
    # "ip"
  ]
}
```

If a node becomes unavailable due to database corruption or other issues, traffic can be switched to the backup node.

### Specification

**API Changes** 
None

**Configuration Changes** 
Remove item `storage.backup` 

**Protocol Changes** 
None

## Scope of Impact

**Breaking Changes**
Section of Database will be impacted.

**Backward Compatibility**
Not compatible with v4.8.1 or older.

## Implementation

Yes

Yes

**Estimated Complexity**
Medium

317787106:
**What does this PR do?**

Removes the periodic database backup feature (`storage.backup`) entirely. This includes the backup configuration class, the AspectJ aspect that intercepted `pushBlock` to trigger copies, the backup utility, and all associated test fixtures and config files.

Closes tronprotocol/java-tron#6595.

The `storage.backup` feature was designed to guard against data corruption from disk failures or abrupt process termination (e.g. `kill -9`). It has since become more harmful than helpful:

1. **The original threat no longer applies.** Extensive testing has confirmed that `kill -9` does not corrupt RocksDB, which was the primary motivation for the feature.
2. **Copy times block block synchronisation.** As of early 2026, the mainnet state database is close to 3 TB. A single backup copy can take several hours, during which the node cannot sync blocks — a fatal risk to service stability.
3. **A better alternative exists.** Operators can deploy a primary–backup pair of full nodes using `node.backup` (UDP keep-alive failover). If one node becomes unavailable, traffic can be switched to the standby without any file-copy disruption.

**Changes:**

- **Deleted** `DbBackupConfig`, `BackupDbUtil`, `BackupRocksDBAspect`, `NeedBeanCondition` (`common` / `framework`).
- **Removed** `CommonParameter.dbBackupConfig` field and `Args.initRocksDbBackupProperty()` initialiser.
- **Removed** `ConfigKey.STORAGE_BACKUP_*` constants and the conditional `backupRocksDBAspect` Spring bean from `DefaultConfig`.
- **Deleted** `BackupDbUtilTest` and `config-test-dbbackup.conf`; updated `RocksDbDataSourceImplTest` and `SupplementTest` to use `TEST_CONF`.
- **Configuration change:** the `storage.backup` block is no longer read and has no effect. Nodes carrying this section in their config file can leave it in place without error, but it will be silently ignored.

**Backward Compatibility:** Not compatible with v4.8.1 or older.

- Unit Tests

**Follow up**

Operators who currently rely on `storage.backup` should migrate to the `node.backup` primary–backup failover configuration.

**Extra details**

N/A


halibobo1205 (approved):
LGTM

xxo1shine (commented):
[SHOULD] `framework/build.gradle` still references `BackupDbUtilTest.class` in two spots, but this PR deletes that test class:

- line 119: `t.exclude '**/BackupDbUtilTest.class'` (Windows-exclude block)
- line 164: `include 'org/tron/core/db/backup/BackupDbUtilTest.class'` (in the `testWithRocksDb` task)

Gradle silently ignores include/exclude patterns that match nothing, so the build won't break — but they're dead config. Please remove them in this PR so the cleanup is complete.

Sunny6889 (approved):
## Operator notice (suggest adding to release notes)

Nodes upgrading from a version where `storage.backup.enable = true` will have leftover artifacts on disk:

- The `propPath` file configured by the user (default: `prop.properties`)
- The `bak1path` / `bak2path` directories (defaults: `bak1/database/` and `bak2/database/`)

**Please call this out explicitly in the release notes**: operators should manually delete these paths after upgrading to reclaim disk space. Without this guidance, the dead backup directories will silently keep consuming storage on every node that previously had the feature enabled.

317787106:
@Sunny6889 This feature stopped working a long time ago. It might be worth mentioning this in the release notes documentation.

317787106:
> [SHOULD] `framework/build.gradle` still references `BackupDbUtilTest.class` in two spots, but this PR deletes that test class:
> 
> * line 119: `t.exclude '**/BackupDbUtilTest.class'` (Windows-exclude block)
> * line 164: `include 'org/tron/core/db/backup/BackupDbUtilTest.class'` (in the `testWithRocksDb` task)
> 
> Gradle silently ignores include/exclude patterns that match nothing, so the build won't break — but they're dead config. Please remove them in this PR so the cleanup is complete.

Remove it in `framework/build.gradle`, thanks.

PR #6727, opened 2026-04-28
feat(net): support domain names in peer configuration

### LINKED ISSUE #6634: [Feature] Support domain names in node.backup.members and seed.node.ip.list
## Problem Statement

Currently, `node.backup.members` and `seed.node.ip.list` in the node configuration only accept raw IP addresses. This creates operational friction in dynamic or cloud-native deployments where:

- Infrastructure IP addresses change frequently (auto-scaling groups, cloud instances, Kubernetes pods)
- Operators prefer to manage a small set of stable DNS records rather than updating config files across many nodes whenever IPs rotate
- High-availability backup setups require the member list to stay accurate even as IPs change at runtime

Neither field documents or handles hostname inputs — any domain name silently fails to resolve, causing nodes to be silently dropped from the member/seed list.

**Who benefits:** Node operators running large-scale deployments, SR infrastructure teams, and any operator managing nodes on infrastructure with dynamic IP allocation.

## Proposed Solution

Extend the parsers for both `node.backup.members` and `seed.node.ip.list` to accept hostname entries in addition to raw IPv4/IPv6 addresses.

Domain resolution should reuse the DNS resolution pattern from the libp2p library (see [tronprotocol/libp2p#130](https://github.com/tronprotocol/libp2p/pull/130)) — specifically the dnsjava `SimpleResolver` with explicit per-query timeout — rather than re-implementing resolution logic. Note: libp2p's existing `lookUpTxt` queries TXT records for node discovery; for resolving hostnames to IPs (A/AAAA records) a parallel `lookUpA` function following the same pattern is needed.

The two fields have **different reliability requirements** and therefore adopt different resolution strategies:

| Field | Resolution timing | Startup failure on DNS error | Periodic re-resolution
| --- | --- | --- | --- | 
| `node.backup.members` | **Synchronous**, blocks startup until done | **Fatal if entry fails** | **Yes** (fixed interval constant in `BackupManager`) | 
| `seed.node.ip.list` | **Asynchronous**, does not block startup | Warning per entry, startup continues | No (p2p handles dynamic expansion) | 

### Specification

**Configuration Changes**

`node.backup.members` — extend to accept hostnames (no port, since the backup port is global). No new config options are added; the re-resolution interval is a fixed constant inside `BackupManager`:

```hocon
node.backup {
  port = 10001
  priority = 8
  keepAliveInterval = 3000

  // support both peer's ip or domain list, but only fill one of them
  members = [
    // "192.168.1.10",            <- IP, used as-is
    // "backup-node.example.com"  <- domain, resolved at startup and refreshed periodically
  ]
}
```

`seed.node.ip.list` — extend to accept `host:port` where `host` may be a domain name. No new config options are added; resolution is asynchronous and best-effort:

```hocon
seed.node = {
  // List of seed nodes. Entries may be IP:port or domain:port.
  ip.list = [
    "52.8.46.215:18888",           // IP:port — used as-is
    "seed1.tron.network:18888"     // domain:port — resolved asynchronously at startup
  ]
}
```

**Resolution Behavior**

*`node.backup.members` (high-reliability, synchronous):*

- Due to the high reliability requirements of primary/standby election, domain resolution is **synchronous and blocks node startup** until resolution finished.
- If the domain entry failed to resolve, the node refuses to start and logs a prominent fatal error with the exact hostname and remediation advice:
  ```
  [FATAL] Failed to start: cannot resolve backup member "backup-node.example.com".
  Please check DNS configuration or replace the entry with a reachable IP address.
  ```
- After startup, a background task re-resolves the domain entry on a periodic schedule (interval defined as a constant in `BackupManager`, not configurable).
  - On refresh, successfully resolved IP replace the previous one for that domain.
  - If the domain temporarily fails to resolve on refresh, the **last known IP is retained** and a warning is logged — the node does not stop.

*`seed.node.ip.list` (best-effort, asynchronous):*

- To avoid lengthening startup time, domain entries are resolved **asynchronously** in the background. Peer discovery begins immediately using raw IP entries; resolved IPs are added to the seed pool as results arrive.
- All domain lookups are submitted **in parallel** so resolution finishes as quickly as possible.
- Per-entry failures are logged as warnings; the entry is skipped and startup always continues:
  ```
  [WARN] Cannot resolve seed node "seed1.tron.network" — skipping. Check DNS or update ip.list.
  ```
- No periodic re-resolution after startup. The p2p protocol dynamically discovers and expands the peer list; the seed list is only used for initial bootstrap.

**API Changes**

None — this is a configuration parsing enhancement only.

**Protocol Changes**

None.

## Scope of Impact

- Network layer (peer discovery, backup node election)
- Configuration parsing (`Args.java`, `CommonParameter.java`)

**Breaking Changes**

None. All existing IP-only configurations remain valid.

**Backward Compatibility**

Fully backward compatible. Raw IP entries continue to work as-is; the parser treats an entry as a domain only when it is not a valid IP literal (e.g., via Guava's `InetAddresses.isInetAddress()`).

## Implementation

**Ideas regarding implementation:**

1. In `Args.java`, after reading `node.backup.members` and `seed.node.ip.list`, pass each entry through a helper that distinguishes IP literals from hostnames.
2. Invoke the DNS resolver interface exposed by the libp2p library to perform hostname-to-IP resolution.
3. For `node.backup.members`:
   - At startup, submit the domain entry to the resolver and wait synchronously; fail fast with a descriptive message if the resolution failed.
   - In `BackupManager`, add a `ScheduledExecutorService` task that re-resolves the domain entry at a fixed interval (e.g. 10 minutes, defined as a constant) and atomically swaps the live `member ip` set. On refresh failure, retain the last known IP and log a warning.
4. For `seed.node.ip.list`:
   - At startup, submit all domain entries to a thread pool asynchronously (fire-and-forget with callbacks); node startup continues immediately using raw IP entries.
   - Each callback logs `WARN` on failure or adds resolved IPs to the seed pool on success.
   - No scheduled re-resolution task needed.

Yes.

**Estimated Complexity**

Medium — the core parsing/resolution change is straightforward, but the periodic re-resolution in `BackupManager` and the synchronous-vs-asynchronous split require careful implementation and testing.

## Testing Strategy

**Test Scenarios**

*`node.backup.members`:*

- IP entry: behavior unchanged.
- Hostname entry: resolved IP used correctly by backup election.
- For unresolvable hostname at startup: node refuses to start with a clear fatal error message.
- Periodic refresh: updated IP reflected in member set without restart.
- Transient refresh failure: last known IP retained, warning logged, node continues.

*`seed.node.ip.list`:*

- IP-only entries: behavior unchanged.
- Hostname entries: resolved IPs used as initial seeds.
- Mixed IP + hostname entries: both work correctly.
- Some or all hostnames unresolvable: warnings logged per entry, node starts and continues with available IPs.
- Asynchronous resolution: startup is not delayed; domain lookups complete in the background.
- Parallel resolution: all domain lookups submitted concurrently, total resolution time bounded by the slowest single lookup.

- `node.backup.members` resolution is synchronous and blocks startup. 
- `seed.node.ip.list` resolution is asynchronous and does not add to startup time; raw IP entries are usable immediately.
- Per-lookup timeout: libp2p's `SimpleResolver` uses `setTimeout(Duration.ofMillis(1000))` per attempt with up to 5 retries (worst-case ~5 s per domain). The same configuration should be reused for consistency.
- Periodic re-resolution in `BackupManager` must run on a dedicated scheduled thread, not on the backup UDP event loop.

## Alternatives Considered

- **Re-implement DNS resolution in java-tron** — rejected to avoid duplicating logic already maintained in the libp2p library.
- **Periodic refresh for seed nodes** — rejected; seed nodes are numerous and the p2p protocol already handles dynamic peer discovery after bootstrap, so the added complexity is not justified.

## Additional Context

**Related Issues/PRs**

- [tronprotocol/libp2p](https://github.com/tronprotocol/libp2p) — exposes the DNS resolver interface this feature would consume

**References**

- Current config: `framework/src/main/resources/config.conf` — `node.backup`, `seed.node`
- Backup member usage: `framework/src/main/java/org/tron/common/backup/BackupManager.java`
- Seed node config parsing: `framework/src/main/java/org/tron/core/config/args/Args.java`

317787106:
## Summary

Allows `node.seed.ip.list`, `node.active`, `node.passive`, `node.fastForward`, and `node.backup.members` configuration entries to specify domain names in addition to IP literals. Domains are resolved to IPs at startup; backup member IPs are refreshed automatically every 60 seconds when DNS returns a different address. 
Closes https://github.com/tronprotocol/java-tron/issues/6634

## Problem

All network peer configuration fields previously required numeric IP addresses. Operators running nodes behind dynamic DNS or cloud load-balancers had to manually update config files whenever IPs changed. Backup members in particular need continuous accuracy because they participate in keep-alive health checks.

### New: `InetUtil` 
Central DNS resolution utility used by both startup parsing and runtime refresh:

- `resolveInetSocketAddressList(List<String>)` — converts a list of `ipOrDomain:port` strings to `InetSocketAddress` objects. IP literals are used as-is; domain entries are resolved via `LookUpTxt.lookUpIp` (IPv4 first, IPv6 fallback). When there are multiple domain entries, resolution is parallelised using a bounded thread pool (up to 10 threads). Unresolvable entries are silently skipped.
- `resolveInetAddress(String)` — resolves a bare host (no port) to `InetAddress`; IP literals take the fast path without a DNS lookup.

### `BackupManager` (`framework`)
- At `init()`, each configured backup member is resolved via `InetUtil.resolveInetAddress`. The resolved IP is stored in `members`; entries with a domain host are additionally tracked in a `domainIpCache`.
- A new `dnsExecutorService` scheduler runs `refreshMemberIps()` every 60 seconds when `domainIpCache` is non-empty. If DNS returns a new IP, the old IP is removed from `members` and the new IP is added; `localIp` is never added.
- `dnsExecutorService` is shut down alongside `executorService` in `stop()`.

### `Args` (`framework`)
- Extracted a shared `getInetSockerAddress(config, path)` helper that calls `InetUtil.resolveInetSocketAddressList`, used by both `getInetSocketAddress` (seed/active/passive nodes) and `getInetAddress`. Unresolvable domain entries in seed lists are silently dropped; unresolvable backup member entries throw `TronError(PARAMETER_INIT)` at startup via the new `checkBackupMembers()` validator.

### Dependency
- Switched `libp2p` from `io.github.tronprotocol:libp2p:2.2.7` to `io.github.tronprotocol:libp2p:2.2.8`, which exposes the `LookUpTxt.lookUpIp` API used for DNS resolution. 

## Behaviour Summary

| Config field | Unresolvable domain | Domain IP change |
|---|---|---|
| Seed / active / passive / fastForward | Entry skipped silently | Re-read at next startup |
| `node.backup.members` | `TronError` thrown at startup, skip when BackupManager.init() as it will not be triggered until catch up | Auto-refreshed every 60 s |

## Test

- **`InetUtilTest`** (18 tests): `resolveInetSocketAddressList` — IPv4/IPv6 literals, domain resolution, IPv6 fallback, input order preserved, unresolvable entries dropped. `resolveInetAddress` — same coverage for the no-port variant.
- **`BackupManagerTest`** (6 new tests): `init()` domain-to-IP mapping and local-IP exclusion; `refreshMemberIps()` — IP changed, IP unchanged, DNS failure retains old IP.
- **`ArgsTest`** (3 new tests): `checkBackupMembers()` — IP members pass, unresolvable domain throws `TronError(PARAMETER_INIT)`, resolvable domain passes.


framework/src/main/java/org/tron/core/config/args/InetUtil.java
```diff
@@ -0,0 +1,171 @@
...
+      InetSocketAddress parsed = NetUtil.parseInetSocketAddress(item);
+      parsedMap.put(item, parsed);
+      if (!isIpLiteral(parsed.getHostString())) {
+        domainEntries.add(item);
+      }
+    }
+
+    // Resolve domain names: spin up a thread pool only when there are multiple domains.
+    Map<String, InetSocketAddress> resolvedDomains = new HashMap<>();
+    if (domainEntries.size() > 1) {
```
xxo1shine: [SHOULD] It should be `domainEntries.size() > 0`.

317787106: When `domainEntries.size() == 1`, we use the main thread directly instead of creating  thread pool to save resources. Parallel execution is only used when the length is greater than 1.

framework/src/main/java/org/tron/core/config/args/InetUtil.java
```diff
@@ -0,0 +1,171 @@
...
+          logger.warn("Failed to resolve address, skip: {}", entry);
+        } catch (TimeoutException e) {
+          logger.warn("DNS lookup timed out after {}s, skip: {}", DNS_LOOKUP_TIMEOUT_SECONDS,
+              entry);
+        }
+      }
+      ExecutorServiceManager.shutdownAndAwaitTermination(dnsPool, DNS_POOL_NAME);
+    } else if (domainEntries.size() == 1) {
+      String entry = domainEntries.get(0);
+      resolvedDomains.put(entry, resolveInetSocketAddress(entry));
```
SeriousCoding789: **Inconsistent reliability between single-domain and multi-domain paths**

`resolveInetSocketAddressList()` has two code paths for DNS resolution: when there are 2+ domain entries, lookups go through a thread pool with a 10-second per-lookup timeout via `Future.get()`; when there is exactly 1 domain entry, it calls `resolveInetSocketAddress()` synchronously with no timeout protection.

Suggest:
- Removing the single-domain fast path and always going through the bounded path.

common/build.gradle
```diff
@@ -21,7 +21,8 @@ dependencies {
     api 'org.aspectj:aspectjrt:1.9.8'
     api 'org.aspectj:aspectjweaver:1.9.8'
     api 'org.aspectj:aspectjtools:1.9.8'
-    api group: 'io.github.tronprotocol', name: 'libp2p', version: '2.2.7',{
+    api group: 'com.github.tronprotocol', name: 'libp2p', version: 'release-v2.2.8-SNAPSHOT',{
```
SeriousCoding789, line 24: Is this `release-v2.2.8-SNAPSHOT` a correct usage? 
Why change from `io.github.tronprotocol` to `com.github.tronprotocol`

317787106: It's a gitpack temporary dependency now. After the libp2p v2.2.8 releases, it will be updated as 
```
api group: 'io.github.tronprotocol', name: 'libp2p', version: '2.2.8'
```

framework/src/main/java/org/tron/core/config/args/InetUtil.java
```diff
@@ -0,0 +1,171 @@
...
+   *   <li>{@code [fe80::48ff:fe00:1122]:18888}
+   *   <li>{@code example.com:18888}
+   *   <li>{@code hostname:18888}
+   * </ul>
+   *
+   * @param ipOrDomainWithPortList list of address strings in {@code ipOrDomain:port} format,
+   *     may mix IP literals and domain names
+   * @return resolved addresses in the same order as the input, omitting unresolvable entries
+   */
+  public static List<InetSocketAddress> resolveInetSocketAddressList(
```
SeriousCoding789, line 55: Suggest this function rewrite as 
```
private static Map<String, InetSocketAddress> resolveDomainsInParallel(
    List<String> domainEntries) {
  Map<String, InetSocketAddress> resolved = new HashMap<>();
  if (domainEntries.isEmpty()) {
    return resolved;
  }

  int poolSize = StrictMath.min(domainEntries.size(), DNS_POOL_MAX_SIZE);
  ExecutorService dnsPool = ExecutorServiceManager
      .newFixedThreadPool(DNS_POOL_NAME, poolSize, true);

  try {
    LinkedHashMap<String, CompletableFuture<InetSocketAddress>> futures =
        new LinkedHashMap<>();
    for (String entry : domainEntries) {
      futures.put(entry, CompletableFuture.supplyAsync(
          () -> resolveInetSocketAddress(entry), dnsPool));
    }

    // Single global deadline for all lookups combined.
    try {
      CompletableFuture
          .allOf(futures.values().toArray(new CompletableFuture[0]))
          .get(DNS_LOOKUP_TIMEOUT_SECONDS, TimeUnit.SECONDS);
    } catch (TimeoutException e) {
      logger.warn("DNS lookup budget {}s exceeded, dropping unresolved entries",
          DNS_LOOKUP_TIMEOUT_SECONDS);
    } catch (InterruptedException e) {
      Thread.currentThread().interrupt();
    } catch (ExecutionException ignored) {
      // per-entry exceptions handled below
    }

    // Collect whatever finished; drop pending/failed entries.
    for (Map.Entry<String, CompletableFuture<InetSocketAddress>> e : futures.entrySet()) {
      CompletableFuture<InetSocketAddress> f = e.getValue();
      if (f.isDone() && !f.isCompletedExceptionally()) {
        InetSocketAddress addr = f.getNow(null);
        if (addr != null) {
          resolved.put(e.getKey(), addr);
        }
      } else {
        logger.warn("DNS unresolved or timed out, skip: {}", e.getKey());
      }
    }
  } finally {
    ExecutorServiceManager.shutdownAndAwaitTermination(dnsPool, DNS_POOL_NAME);
  }
  return resolved;
}
```

framework/src/main/java/org/tron/common/backup/BackupManager.java
```diff
@@ -162,4 +190,28 @@ public enum BackupStatusEnum {
...
+  private void refreshMemberIps() {
+    for (Map.Entry<String, String> entry : domainIpCache.entrySet()) {
+      String domain = entry.getKey();
+      String oldIp = entry.getValue();
+      InetAddress inetAddress = resolveInetAddress(domain);
+      if (inetAddress == null) {
+        logger.warn("DNS refresh: failed to re-resolve backup member domain {}, keep it", domain);
+        continue;
+      }
+      String newIp = inetAddress.getHostAddress();
```
SeriousCoding789, line 206: Is there a need to do the following check?
```
if (localIp.equals(newIp) {
        continue;
}
```

317787106: You are right, there is no need to add this check. I will drop it.

framework/src/main/java/org/tron/core/config/args/InetUtil.java
317787106: Use parallel although `domainEntries.size() == 1` to simplify the logic.

framework/src/main/java/org/tron/core/config/args/InetUtil.java
317787106: Accept it, always going through the bounded path.

framework/src/main/java/org/tron/core/config/args/InetUtil.java
317787106: Using `CompletableFuture` is a good idea, accept it. Thanks for your review.

framework/src/main/java/org/tron/common/backup/BackupManager.java
```diff
@@ -78,10 +82,20 @@ public void init() {
...
+    for (String ipOrDomain : parameter.getBackupMembers()) {
+      InetAddress inetAddress = resolveInetAddress(ipOrDomain);
+      if (inetAddress == null) {
+        logger.warn("Failed to resolve backup member domain: {}", ipOrDomain);
+        continue;
+      }
+      String ip = inetAddress.getHostAddress();
+      if (localIp.equals(ip)) {
+        continue;
+      }
```
xxo1shine, line 94: [SHOULD] Currently, the master and backup servers mainly use external IP addresses, so we can also add external IP address filtering logic.

framework/src/main/java/org/tron/core/config/args/InetUtil.java
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Resolves a {@code ipOrDomain:port} config string to an {@link InetSocketAddress} via DNS.
+   *
+   * <p>The host is looked up first over IPv4, then over IPv6 as a fallback. Returns {@code null}
+   * if DNS resolution fails for both address families.
+   *
+   * @param ipOrDomainWithPort address string in {@code ipOrDomain:port} format
+   * @return resolved {@link InetSocketAddress}, or {@code null} if the host cannot be resolved
+   */
+  private static InetSocketAddress resolveInetSocketAddress(String ipOrDomainWithPort) {
```
SeriousCoding789, line 151: Could you rename `resolveInetSocketAddress` to `resolveInetAddressWithPort`, as it will be more clear?

317787106: Class `InetSocketAddress` already contains InetAddress and Port.

Sunny6889 (approved):

xxo1shine (approved):

framework/src/main/java/org/tron/common/backup/BackupManager.java
317787106: Maybe, we can run master and backup in the same node. So, there maybe no need to ignore external IP.

PR #6728, opened 2026-04-28
feat(jsonrpc): add resource restrict for jsonrpc

### LINKED ISSUE #6632: [Feature] Introduce resource limits for JSON-RPC (batch size, response size, address size, timeout)
## Problem Statement
The current JSON-RPC service in java-tron lacks unified constraints on both request scale and response payload size. While this “unbounded” design works in simple scenarios, it introduces significant stability risks when the node is exposed to public traffic or operates under high concurrency.

Specifically:

- For batch array requests, it may contain a large array of sub-requests. A single request can occupy processing threads for an extended period, degrading overall throughput.
- Certain APIs (e.g., `eth_getLogs`, `eth_getFilterChanges`) may return extremely large result sets. Without response size limits, this can lead to excessive memory consumption and even trigger OOM conditions.
- Some requests may become long-running due to I/O latency, holding node resources for prolonged periods.
- Log-related APIs (`eth_getLogs`, `eth_newFilter`) currently allow an unbounded number of address filters, significantly increasing query complexity and directly impacting node performance.

These issues are further amplified when handling untrusted external requests. Therefore, it is necessary to introduce fundamental constraint mechanisms at the JSON-RPC layer.

## Proposed Solution

The implementation prioritizes low-overhead approaches focusing on early validation and resource control.

- For all JSON-RPC requests, if its json type is array, the length should be validated during the parsing phase to enable early rejection, avoiding unnecessary downstream processing.
- For `eth_getLogs` and `eth_newFilter`, the address array should be validated before entering query execution, preventing expensive database or index scans.
- Response size control should not rely on constructing the full payload in memory before validation. Instead, size tracking should be performed incrementally during serialization, using streaming or chunked output to enforce limits and reduce memory pressure.
- Request execution time should be controlled via asynchronous execution. Requests exceeding a configured timeout should be interrupted.

Error handling should remain consistent with the JSON-RPC specification and align with Ethereum conventions for better client interoperability. Different constraint violations should map to clear error codes:

- Address limit exceeded → -32602 (JsonRpcInvalidParamsException)
- Batch size exceeded → -32005
- Response too large → -32003
- Request timeout → -32002

### Specification

Introduce a set of configurable limits for the JSON-RPC service to constrain request size and response payload, with validation performed as early as possible in the request lifecycle to improve system stability and predictability.

The following configuration items are proposed:

1. `node.jsonrpc.maxBatchSize`: Limits the maximum number of sub-requests in a batch JSON-RPC call. Default: 1. Requests exceeding this limit should be rejected immediately after parsing to avoid further processing. Note: values greater than 1 may bypass existing per-method QPS controls.

2.  `node.jsonrpc.maxAddressSize`: Limits the number of address filters in log-related APIs (`eth_getLogs`, `eth_newFilter`). Default: 1000. Validation must occur before query execution to avoid high-cost operations. This follows the same design pattern as the existing node.jsonrpc.maxSubTopics.

3. `node.jsonrpc.maxResponseSize`: Limits the maximum size (in bytes) of a JSON-RPC response. Default: 25MB. The size should be tracked incrementally during serialization. Once the limit is exceeded, the response should be terminated immediately instead of continuing to build the full payload.

4. `node.jsonrpc.maxRequestTimeout`: Limits the execution time of a JSON-RPC request. Default: 30s. Requests exceeding the timeout should be interrupted and return an error.

All configurations should be integrated into the existing configuration system and follow the current JSON-RPC configuration conventions.

**Configuration Changes** 

Add the following configuration items:

- node.jsonrpc.maxBatchSize
- node.jsonrpc.maxResponseSize
- node.jsonrpc.maxAddressSize
- node.jsonrpc.maxRequestTimeout

## Scope of Impact
- API/RPC

**Breaking Changes**
These constraints effectively reduce the impact of oversized or abnormal requests on node resources, mitigate memory risks, and improve stability under high-load conditions. They also enhance the predictability of the JSON-RPC service, making it more suitable for production environments exposed to external users.

**Backward Compatibility**
There are no changes to API fields, so protocol-level compatibility is preserved. However, additional constraints on request inputs and response outputs may require minor client-side adaptations.

## Implementation

Yes

Yes

**Estimated Complexity**
Medium (moderate changes)

## Testing Strategy

Focus primarily on boundary conditions, ensuring that exceptions are triggered as early as possible in the request lifecycle.

**Test Scenarios**

Since request timeout enforcement introduces asynchronous control and may shift execution toward a more serialized model, there could be an impact on performance. Changes in QPS should be evaluated and benchmarked accordingly.

317787106:
**What does this PR do?**

Adds configurable resource limits to the JSON-RPC endpoint to prevent memory exhaustion and abuse from oversized requests or responses. Closes https://github.com/tronprotocol/java-tron/issues/6632

**Changes:**

1. **Batch size limit (`node.jsonrpc.maxBatchSize`, default: 100)**
   - Validates the array length of batch JSON-RPC requests before dispatching.
   - Requests exceeding the limit are rejected with error code `-32005` (exceed limit).
   - The check is skipped when `maxBatchSize ≤ 0` (no limit).

2. **Empty batch rejection**
   - An empty batch array `[]` is now rejected with error code `-32600` (Invalid Request) per [JSON-RPC 2.0 §6](https://www.jsonrpc.org/specification#batch): *"the response from the Server MUST be a single Response object"* when the input is not an array with at least one value.
   - The response is a single error object (not an array), matching the spec requirement.

3. **Response size limit (`node.jsonrpc.maxResponseSize`, default: 25 MB)**
   - Introduces `BufferedResponseWrapper`: intercepts `getOutputStream()` and `getWriter()` writes into an in-memory buffer. When a write would exceed the configured limit, it sets an `overflow` flag and resets the buffer instead of continuing to accumulate bytes, bounding worst-case memory usage to at most `maxResponseSize`.
   - Introduces `CachedBodyRequestWrapper`: replays the pre-read request body via both `getInputStream()` and `getReader()`, so the body can be inspected before being forwarded to `JsonRpcServer`.
   - After the handler returns, the servlet checks `isOverflow()` and — if set — discards the partial buffer and returns error code `-32003` (response too large).

4. **Address list limit (`node.jsonrpc.maxAddressSize`, default: 1000)**
   - In `LogFilter`, validates the `address` array length in `eth_getLogs` / `eth_newFilter` requests.
   - Requests exceeding the limit are rejected with `JsonRpcInvalidParamsException`.

5. **Structured JSON-RPC error responses**
   - `writeJsonRpcError` uses `ObjectMapper` to build error responses safely, avoiding JSON injection from error messages.
   - Error codes follow the JSON-RPC 2.0 spec: 
   -   `-32700` parse error
   -   `-32600` invalid request
   -   `-32603` internal error
   -   `-32005` exceed limit
   -   `-32003` response too large

- Without limits, a client can send an arbitrarily large batch, trigger an expensive query with many addresses, or force the node to serialize a massive response — all of which cause unbounded memory growth.
- The response buffer caps worst-case allocation to `maxResponseSize` and fails fast rather than buffering the entire response before checking.
- Rejecting `[]` closes a spec compliance gap: previously the empty batch fell through to `JsonRpcServer`, whose behavior for an empty array is undefined by the spec.

**Configuration**

```hocon
node {
  jsonrpc {
    # Allowed batch size, default: 100, <=0 means no limit
    maxBatchSize = 100
    # Allowed max response byte size, default: 26214400 (25 MB), <=0 means no limit
    maxResponseSize = 26214400
    # Allowed maximum number for newFilter, <=0 means no limit
    maxLogFilterNum = 20000
  }
}
```

**Tests**

| Test class | Coverage |
|---|---|
| `JsonRpcServletTest` | Parse error, empty body, empty batch `[]` → `-32600`, batch size limit, response overflow, internal error, normal path |
| `BufferedResponseWrapperTest` | Write, overflow, reset, `getWriter` delegation |
| `CachedBodyRequestWrapperTest` | Body replay via `getInputStream` and `getReader` |
| `JsonRpcTest.testLogFilterAddressSizeLimit` | Address list at limit (passes), at limit+1 (throws with `"exceed max addresses:"`), limit=0 disabled |


common/src/main/java/org/tron/core/config/args/NodeConfig.java
```diff
@@ -302,6 +302,9 @@ public void setHttpPBFTPort(int v) {
     private int maxBlockRange = 5000;
     private int maxSubTopics = 1000;
     private int maxBlockFilterNum = 50000;
+    private int maxBatchSize = 100;
+    private int maxResponseSize = 25 * 1024 * 1024;
```
bladehan1, line 315: [SHOULD] Use a memory-size config type for maxResponseSize

`private int maxResponseSize = 25 * 1024 * 1024` is a byte-quantity field, but it is read as a raw `int` so the config file has to spell out `26214400` instead of a human-readable `25M` / `25MiB`. The project's config conventions call for `getMemorySize()` for size-class settings — keeping `int` here makes the value error-prone for operators (the inline comment `// 25 MB = 25 * 1024 * 1024 B` in `config.conf` is an early symptom). `maxBatchSize` and `maxAddressSize` are count-class and `int` is fine for them.

Suggestion: change `maxResponseSize` to a `String` field and parse it with `getMemorySize()`, so HOCON values like `25M` work; keep the count-class fields as `int`.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +84,71 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+      writeJsonRpcError(resp, JsonRpcError.EXCEED_LIMIT,
+          "Batch size " + rootNode.size() + " exceeds the limit of " + batchSize, null);
+      return;
+    }
+
+    CachedBodyRequestWrapper cachedReq = new CachedBodyRequestWrapper(req, body);
+    BufferedResponseWrapper bufferedResp = new BufferedResponseWrapper(
+        resp, parameter.getJsonRpcMaxResponseSize());
+
+    try {
```
bladehan1, line 134: [SHOULD] Wrapping every handler exception as IOException breaks the JSON-RPC over HTTP 200 contract

`catch (Exception e) throw new IOException("RPC execution failed", e)` rethrows every RuntimeException from `rpcServer.handle`. The parent `RateLimiterServlet.service` only catches `ServletException | IOException` and re-throws, so the servlet container emits an HTTP 500 with no JSON-RPC body. jsonrpc4j's `ErrorResolver` would normally convert internal exceptions into a structured error response on HTTP 200 — that contract is now lost. ETH-compatible clients (web3.js / ethers / web3j) treat HTTP 500 as a transport failure and will retry, amplifying load on the node under stress.

Suggestion: drop the catch (let the original IOException path from `rpcServer.handle` propagate so jsonrpc4j's structured error path stays intact); or log the cause and emit `-32603 Internal error` via `writeJsonRpcError` so the HTTP 200 + JSON-RPC error contract is preserved.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +84,71 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
   protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
-    rpcServer.handle(req, resp);
+    CommonParameter parameter = CommonParameter.getInstance();
+
+    byte[] body;
+    JsonNode rootNode;
+    try {
+      body = readBody(req.getInputStream());
+      rootNode = MAPPER.readTree(body);
+    } catch (IOException e) {
```
bladehan1: [SHOULD] Empty or whitespace-only body can make readTree return null and NPE the next line

For zero-length / whitespace-only input, `MAPPER.readTree(byte[])` can return `null` (depending on Jackson version and parser feature flags) rather than a `MissingNode`. Line 99 then dereferences `rootNode.isArray()` and throws NullPointerException. The NPE is not caught by the `IOException` clause at line 95 — it bubbles into the `catch (Exception e)` at line 109, gets wrapped as IOException, and the client sees HTTP 500 instead of the `-32700 Parse error` the parse path was supposed to return.

Suggestion: after `MAPPER.readTree(body)`, treat `rootNode == null || rootNode.isMissingNode()` as a parse error and emit `-32700` via `writeJsonRpcError`.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +84,71 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+      writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "Parse error", null);
+      return;
+    }
+    int batchSize = parameter.getJsonRpcMaxBatchSize();
+    if (rootNode.isArray() && batchSize > 0 && rootNode.size() > batchSize) {
+      writeJsonRpcError(resp, JsonRpcError.EXCEED_LIMIT,
+          "Batch size " + rootNode.size() + " exceeds the limit of " + batchSize, null);
+      return;
```
bladehan1, line 117: [SHOULD] Batch-size rejection returns a single error object, not a JSON-RPC 2.0 batch array

When `rootNode.isArray()` is true and the batch exceeds `maxBatchSize`, the current code calls `writeJsonRpcError(...)` which writes a single object response. JSON-RPC 2.0 requires that a batch request be answered with a JSON array of responses. Standard ETH-compatible clients (web3.js, ethers.js, web3j) parsing a batch response as an array will fail or silently drop the entire result, instead of surfacing the structured `-32005` error to the caller.

Suggestion: when rejecting an over-sized batch, write a single-element array `[{jsonrpc:"2.0", error:{code:-32005, message:...}, id:null}]` so it round-trips through batch-aware clients. Same applies to the response-too-large path on line 117 if the original request was an array.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
```diff
@@ -0,0 +1,133 @@
...
+
+  @Override
+  public void setContentLengthLong(long len) {
+    if (maxBytes > 0 && len > maxBytes) {
+      markOverflow();
+    }
+  }
+
+  @Override
+  public void setStatus(int sc) {
```
bladehan1, line 113: [SHOULD] Override getStatus and intercept setHeader/addHeader for Content-Length

Header capture currently only covers `setStatus`, `setContentType`, `setContentLength(int|long)`. Two gaps:

1. **`getStatus()` is not overridden.** Inherited `HttpServletResponseWrapper.getStatus()` returns the underlying response's status (still `SC_OK` until `commitToResponse` runs). Any logging filter / metrics interceptor that reads status via the wrapper before commit will see a stale value.

2. **`setHeader(name, value)` / `addHeader(name, value)` pass through to the underlying response.** jsonrpc4j currently uses `setContentLength` so this is latent — but any downstream filter or library upgrade that writes Content-Length via `setHeader` would commit a Content-Length to the actual response before the size check runs.

Suggestion: override `getStatus()` to return `this.status`; intercept `setHeader` / `addHeader` for `Content-Length` (case-insensitive) so they go through the same buffering / overflow check as `setContentLength`.

common/src/main/java/org/tron/common/parameter/CommonParameter.java
```diff
@@ -459,6 +459,15 @@ public class CommonParameter {
   @Getter
   @Setter
   public int jsonRpcMaxBlockFilterNum = 50000;
+  @Getter
+  @Setter
+  public int jsonRpcMaxBatchSize = 100;
```
bladehan1, line 497: [SHOULD] Validate non-negative range for the new size-limit fields at config load

The three new fields (`jsonRpcMaxBatchSize`, `jsonRpcMaxResponseSize`, `jsonRpcMaxAddressSize`) are read via `Args.applyNodeConfig` with no range validation. The `> 0` guards in the call sites mean a negative value silently becomes a permanent 'no limit' state — that is fine if `<= 0` is the documented contract, but neither `reference.conf` nor `config.conf` says so explicitly, only `> 0 otherwise no limit`. Operators reading the comment may assume only `0` disables the limit; setting `-1` (a common 'unset' sentinel) silently has the same effect, while `Integer.MIN_VALUE` is also accepted with no warning.

Suggestion: validate `value >= 0` in `Args.applyNodeConfig` (reject startup with a clear error on negative values), and update the reference/config comments to spell out the exact 'disabled' semantics — e.g. `# 0 disables the limit; negative values are rejected at startup`.

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+  private final byte[] body;
+
+  public CachedBodyRequestWrapper(HttpServletRequest request, byte[] body) {
+    super(request);
+    this.body = body;
+  }
+
+  @Override
+  public ServletInputStream getInputStream() {
```
bladehan1, line 47: [SHOULD] getInputStream() and getReader() should be mutually exclusive per servlet spec

Servlet 3.1 spec (§ 5.4 / § 5.5) requires that once one of `getInputStream()` / `getReader()` has been called on a request, the other must throw `IllegalStateException`. This wrapper returns a fresh stream/reader from the cached byte array on every call and allows arbitrary interleaving. jsonrpc4j only calls one today, so the divergence is latent — but any future filter that reads the body through the other accessor would silently double-read with no error, which is exactly the kind of bug the spec wants to prevent.

Suggestion: track which accessor was used first (boolean field) and throw `IllegalStateException` on the second.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
```diff
@@ -0,0 +1,133 @@
...
+ * Buffers the response body without writing to the underlying response,
+ * so the caller can replay it after the handler returns.
+ *
+ * <p>If {@code maxBytes > 0} and the response would exceed that limit, the
+ * {@link #isOverflow()} flag is set instead of throwing. The caller should check this flag after
+ * the handler returns and write its own error response when true.
+ *
+ * <p>Header-mutating methods ({@code setStatus}, {@code setContentType}) are buffered here and
+ * only forwarded to the real response via {@link #commitToResponse()}, preventing a timed-out
+ * handler thread from racing with the timeout error writer.
```
bladehan1: [NIT] Class javadoc references a timeout race that has no implementation

The class javadoc says headers are buffered to prevent 'a timed-out handler thread from racing with the timeout error writer' — but this PR has no timeout / cancellation logic, and `overflow` is not `volatile`. The comment implies a concurrency guarantee that the code does not provide, which is misleading for future maintainers who might rely on it.

Suggestion: drop the timeout-race wording, or convert the implication into an actual constraint (single-threaded handler assumption documented; or `volatile` flag if multi-threaded use is intended).

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +84,71 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
-    rpcServer.handle(req, resp);
+    CommonParameter parameter = CommonParameter.getInstance();
+
+    byte[] body;
+    JsonNode rootNode;
+    try {
+      body = readBody(req.getInputStream());
+      rootNode = MAPPER.readTree(body);
+    } catch (IOException e) {
+      writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "Parse error", null);
```
bladehan1: [NIT] Don't collapse transport IOException with JSON parse errors

The single `catch (IOException e)` at line 95 maps every IO failure to `-32700 Parse error`. `readBody` can throw IOException for legitimate transport issues (client aborted, socket reset, read timeout); none of those are parse errors. Only Jackson's `JsonProcessingException` (an IOException subclass) should map to `-32700`. Mixing them makes server-side logs less useful for diagnosing real client/network issues.

Suggestion: catch `JsonProcessingException` separately for `-32700`, and either let other IOExceptions propagate or map them to a distinct code (and log).

framework/src/main/resources/config.conf
```diff
@@ -366,15 +366,18 @@ node {
...
     maxBlockRange = 5000
-
-    # The maximum number of allowed topics within a topic criteria, default value is 1000,
-    # should be > 0, otherwise means no limit.
+    # Allowed max address count in filter request, default: 1000, >0 otherwise no limit
+    maxAddressSize = 1000
+    # The maximum number of allowed topics within a topic criteria, default: 1000, >0 otherwise no limit
     maxSubTopics = 1000
-    # Allowed maximum number for blockFilter
+    # Allowed maximum number for blockFilter, default: 50000, >0 otherwise no limit
```
bladehan1: [NIT] Two small follow-ups on the new jsonrpc config comments

A pair of small documentation issues introduced in the new jsonrpc block:

1. **Duplicated `default: 100`.** The `maxBatchSize` line reads `# Allowed batch size, default: 100, default: 100, >0 otherwise no limit` — the `default: 100` clause is repeated.

2. **Inline math comment is easy to misread.** `maxResponseSize = 26214400 // 25 MB = 25 * 1024 * 1024 B` — operators skimming this line may read the value as `25` rather than `26214400`. Also, the surrounding keys all use `#` comments; `//` is the only one of its kind in this file.

Suggestion: drop the duplicated `default: 100`; convert `//` to `#`; consider aligning all three new keys' comments with the `<=0 means no limit` wording used by neighbours like `maxBlockRange`.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
```diff
@@ -0,0 +1,133 @@
...
+  public ServletOutputStream getOutputStream() {
+    return outputStream;
+  }
+
+  @Override
+  public PrintWriter getWriter() {
+    return writer;
+  }
+
+  public void commitToResponse() throws IOException {
```
bladehan1, line 158: [NIT] Make commitToResponse idempotent or fail-fast on second call

After `commitToResponse()`, the wrapper still holds the buffered bytes; calling it a second time would write the same body twice. The current call site only commits once so there's no live bug, but the contract is implicit and a future refactor could trip on it.

Suggestion: either clear the buffer at the end of `commitToResponse`, or set a `committed` flag and throw `IllegalStateException` on a second call.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -14,15 +19,28 @@
...
+import org.tron.core.services.filter.CachedBodyRequestWrapper;
 import org.tron.core.services.http.RateLimiterServlet;
 
 @Component
 @Slf4j(topic = "API")
 public class JsonRpcServlet extends RateLimiterServlet {
 
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  enum JsonRpcError {
```
bladehan1: [NIT] Make JsonRpcError enum visibility explicit

`enum JsonRpcError { ... final int code; }` has no explicit visibility modifier; both the enum and `code` default to package-private. If there's no reason to expose this enum outside the class, tighten it to `private`. If tests in the same package will assert against `JsonRpcError.RESPONSE_TOO_LARGE.code`, keep package-private but add a one-line comment so future readers know it's intentional.

Suggestion: mark the enum and `code` `private`, or document the package-private decision in a one-line comment.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
bladehan1, line 134: [NIT] Document the user-visible behavior change of wrapping rpcServer.handle exceptions

Independent of the structural concern (separate [SHOULD] comment about HTTP 200 vs 500), the wrap-as-IOException change also bypasses the existing `RateLimiterServlet.service`'s `catch (Exception unexpected) { logger.error(...) }` path, so the standard `Http Api {}, Method:{}` error log line no longer fires. This is a quiet observability regression worth at least a log line and a PR-description bullet.

Suggestion: log the original cause inside the catch before rethrowing, and add a one-line note to the PR description about the new error-mapping.

common/src/main/java/org/tron/core/config/args/NodeConfig.java
317787106: Using `getMemorySize()` increases the cognitive burden for users; using explicit integer values better conveys the intended meaning.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
317787106: Good catch, remove time-out related doc. I will add volatile to `overflow` though it's not necessary.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
317787106: Add a varible `boolean committed` to specify whether it has been write though writing twice will never happen in jsonrpc.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Good catch, use a single-element array when batch exceeds `maxBatchSize` or response is too large.

framework/src/main/resources/config.conf
317787106: Drop the duplicated comments. Using `<=0 means no limit`. Thanks for your careful cancern.

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
317787106: At present, jsonrpc4j only invokes one of them; this is a potential issue rather than an existing bug. Adding the relevant checks is actually redundant, but i will try it.

common/src/main/java/org/tron/common/parameter/CommonParameter.java
317787106: The comment specifies it already, but I can optimze it as `<=0 means no limit` in config.conf.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
317787106: Thanks four your review: 
- `getStatus()` is overridden;
- Add addtional check of `content-length` for `setHeader` and `addHeader`. Really, it will be overrideen by `actual.setContentLength(buffer.size());` so there is little neceesay.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Add `-32603 Internal error` for RuntimeException, other IOException will be rethrown.

317787106: Add the log by using `logger.error("RPC execution failed", e);`. But actually there may be too many error stack if the node is acctacked.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: There is not related testcase now, adding the modifier `private` is OK

framework/src/main/java/org/tron/core/services/jsonrpc/filters/LogFilter.java
```diff
@@ -50,6 +50,10 @@ public LogFilter(FilterRequest fr) throws JsonRpcInvalidParamsException {
       withContractAddress(addressToByteArray((String) fr.getAddress()));
 
     } else if (fr.getAddress() instanceof ArrayList) {
+      int maxAddressSize = Args.getInstance().getJsonRpcMaxAddressSize();
+      if (maxAddressSize > 0 && ((ArrayList<?>) fr.getAddress()).size() > maxAddressSize) {
```
0xbigapple, line 54: [SHOULD] LogFilter `maxAddressSize` branch has no test coverage
**Suggestion**: extend (or add) a `LogFilterTest` with two cases — list exactly at the limit (passes) and list at limit+1 (throws `JsonRpcInvalidParamsException` whose message contains `exceed max addresses:`).

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public BufferedReader getReader() {
+    if (accessor == BodyAccessor.STREAM) {
+      throw new IllegalStateException("getInputStream() has already been called on this request");
+    }
+    accessor = BodyAccessor.READER;
+    String encoding = getCharacterEncoding();
+    Charset charset = encoding != null ? Charset.forName(encoding) : StandardCharsets.UTF_8;
```
0xbigapple: [NIT] CachedBodyRequestWrapper.getReader does not handle malformed charset
`Charset.forName` raises an unchecked `IllegalCharsetNameException` / `UnsupportedCharsetException` for malformed values. A client sending `Content-Type: application/json; charset=foo!` will crash inside `getReader()` instead of letting the upstream JSON parser produce a clean `-32700`. `getCharacterEncoding()` returns whatever the client put on the wire — untrusted input.

**Suggestion**:
```java
Charset charset;
try {
  charset = encoding != null ? Charset.forName(encoding) : StandardCharsets.UTF_8;
} catch (IllegalCharsetNameException | UnsupportedCharsetException ex) {
  charset = StandardCharsets.UTF_8;
}
```

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
```diff
@@ -0,0 +1,168 @@
...
+    public boolean isReady() {
+      return true;
+    }
+
+    @Override
+    public void setWriteListener(WriteListener writeListener) {
+    }
+  };
+
+  private final PrintWriter writer = new PrintWriter(outputStream, true);
```
0xbigapple: [NIT] BufferedResponseWrapper PrintWriter uses platform default charset

**Suggestion**:
```java
private final PrintWriter writer =
    new PrintWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8), true);
```

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +87,87 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+        return;
+      }
+    } catch (JsonProcessingException e) {
+      writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "Parse error", null, false);
+      return;
+    }
+
+    boolean isBatch = rootNode.isArray();
+    int batchSize = parameter.getJsonRpcMaxBatchSize();
+    if (isBatch && batchSize > 0 && rootNode.size() > batchSize) {
```
0xbigapple, line 114: [NIT] Empty batch `[]` is not rejected — violates JSON-RPC 2.0

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Optimize it like this:
```
JsonNode rootNode;
    try {
      rootNode = MAPPER.readTree(body);
      if (rootNode == null || rootNode.isMissingNode()) {
        writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "Parse error", null, false);
        return;
      }
    } catch (JsonProcessingException e) {
      writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "Parse error", null, false);
      return;
    }
```

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Good catch, add more check:
```
    if (isBatch && rootNode.isEmpty()) {
      writeJsonRpcError(resp, JsonRpcError.INVALID_REQUEST, "Invalid Request", null, false);
      return;
    }
```

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
317787106: Accept it.

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
317787106: Accept it.

framework/src/main/java/org/tron/core/services/jsonrpc/filters/LogFilter.java
317787106: Good coverage, add testcase `JsonRpcTest.testLogFilterAddressSizeLimit` to verify maxAddressSize.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Fix it.

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
```diff
@@ -0,0 +1,79 @@
...
+import java.nio.charset.IllegalCharsetNameException;
+import java.nio.charset.StandardCharsets;
+import java.nio.charset.UnsupportedCharsetException;
+import javax.servlet.ReadListener;
+import javax.servlet.ServletInputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+
+/**
+ * Wraps a request and replays a pre-read body from a byte array.
```
bladehan1: [NIT] Document the supported scope of CachedBodyRequestWrapper in the class javadoc

The wrapper currently works correctly for what `JsonRpcServlet.doPost` needs — synchronous JSON-body POST, body read via `getInputStream()` / `getReader()`. It is, however, placed in `services/filter/` (alongside `Filter` implementations) and named generically, which invites future reuse. Several `HttpServletRequest` capabilities are silently incompatible with a cached `byte[]` body and are easy to overlook:

- `application/x-www-form-urlencoded` — the container consumes the original stream when `getParameter*` is called; once the body is cached, those accessors return null/empty.
- `multipart/*` — `getPart()` / `getParts()` read from the original (already-consumed) stream.
- async non-blocking I/O — `setReadListener` cannot be honored over a `ByteArrayInputStream`.
- request dispatch / forward chains — the cached body does not propagate through dispatchers in a meaningful way.

Making the scope explicit in the class javadoc is the single cheapest hedge against this wrapper being moved to a generic filter or reused on a different content-type later. Suggested wording (feel free to adjust):

```java
/**
 * Wraps a request to replay a pre-read body from a byte array,
 * allowing the body to be read more than once.
 *
 * <p><b>Scope:</b> designed for synchronous, raw-body POST endpoints
 * (e.g. JSON-RPC). It is NOT compatible with:
 * <ul>
 *   <li>{@code application/x-www-form-urlencoded} — cached body cannot back
 *       {@code getParameter*}.</li>
 *   <li>multipart — {@code getPart()/getParts()} read from the original
 *       (already-consumed) stream.</li>
 *   <li>async non-blocking I/O — see {@code setReadListener}.</li>
 *   <li>request dispatch / forward chains.</li>
 * </ul>
 *
 * <p>Multiple calls to {@code getInputStream()} (or {@code getReader()})
 * are allowed and each returns a fresh stream over the same cached body —
 * a deliberate extension of the standard servlet contract.
 */
```

Not blocking. Pure documentation, no behavior change.

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
```diff
@@ -0,0 +1,79 @@
...
+        return bais.available() == 0;
+      }
+
+      @Override
+      public boolean isReady() {
+        return true;
+      }
+
+      @Override
+      public void setReadListener(ReadListener readListener) {
```
bladehan1, line 75: [NIT] Make setReadListener fail-fast instead of a silent no-op

A cached `byte[]` body cannot drive async non-blocking I/O — the registered `ReadListener` would never receive `onDataAvailable()` callbacks against a `ByteArrayInputStream` snapshot. The current empty implementation accepts the listener silently, so a future caller that turns this servlet into an async path would hang with no visible error.

jsonrpc4j today is fully synchronous, so this is latent rather than active. Throwing fail-fast here costs one line and surfaces a future regression immediately at the right place.

```java
@Override
public void setReadListener(ReadListener readListener) {
  throw new UnsupportedOperationException(
      "async I/O is not supported on cached body");
}
```

The existing `isReady()` returning `true` is fine — for an in-memory stream that is the honest answer.

317787106: Thanks for your very attentive. I will add it.

framework/src/main/java/org/tron/core/services/filter/CachedBodyRequestWrapper.java
317787106: Add the documentation for `CachedBodyRequestWrapper`, thanks.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +89,156 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+      try {
+        responseNode = MAPPER.readTree(responseBytes);
+      } catch (IOException e) {
+        writeJsonRpcError(resp, JsonRpcError.INTERNAL_ERROR, "Internal error", null, true);
+        return;
+      }
+      batchResult.add(responseNode);
+    }
+
+    byte[] finalBytes = MAPPER.writeValueAsBytes(batchResult);
```
bladehan1, line 221: [NIT] Add a final size check on the serialized batch result

`accumulatedSize` is incremented by `subOutput.toByteArray().length + (separator? 1 : 0)`, which is an estimate. The final `MAPPER.writeValueAsBytes(batchResult)` re-serializes parsed `JsonNode`s, and the resulting byte length can drift from the estimate (whitespace, unicode escaping, ObjectNode key ordering). The drift is tiny in practice but means the `maxResponseSize` hard cap is approximate, not exact, for the batch path.

Suggestion: after computing `finalBytes`, do one more cheap check before writing:

```java
byte[] finalBytes = MAPPER.writeValueAsBytes(batchResult);
if (maxResponseSize > 0 && finalBytes.length > maxResponseSize) {
  writeJsonRpcError(resp, JsonRpcError.RESPONSE_TOO_LARGE,
      "Response exceeds the limit of " + maxResponseSize + " bytes", null, true);
  return;
}
```

This turns the per-iteration accumulation into a fast short-circuit and lets the final check enforce the hard cap exactly.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +89,156 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+      accumulatedSize += addition;
+
+      JsonNode responseNode;
+      try {
+        responseNode = MAPPER.readTree(responseBytes);
+      } catch (IOException e) {
+        writeJsonRpcError(resp, JsonRpcError.INTERNAL_ERROR, "Internal error", null, true);
+        return;
+      }
+      batchResult.add(responseNode);
```
bladehan1, line 211: [NIT] All-notification batch returns `[]` instead of an empty body

When every sub-request in a batch is a notification (no `id`), `batchResult` ends up empty and the servlet writes `[]` with HTTP 200. JSON-RPC 2.0 § 6 says: "If there are no Response objects contained within the Response array as it is to be sent to the client, the server MUST NOT return an empty Array and should return nothing at all." The current `JsonRpcServletTest.batchLimitDisabled_largeBatchAllowed` test asserts `body.size() == 0` — i.e. the test acknowledges and pins this behavior.

Most ETH-compatible clients tolerate `[]`, but this is a low-effort spec compliance gap. Suggestion: either (a) flip to a spec-compliant empty response (`resp.setStatus(204); resp.setContentLength(0);` and skip the write) when `batchResult.isEmpty()`, or (b) document this intentional deviation in the class javadoc / PR description.

framework/src/test/java/org/tron/core/services/filter/CachedBodyRequestWrapperTest.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test
+  public void getInputStream_emptyBody_isFinishedImmediately() {
+    CachedBodyRequestWrapper w = new CachedBodyRequestWrapper(new MockHttpServletRequest(),
+        new byte[0]);
+    assertTrue(w.getInputStream().isFinished());
+  }
+
+  @Test
```
bladehan1: [NIT] Add a unit test pinning setReadListener as fail-fast

The new `setReadListener` now throws `UnsupportedOperationException` (great catch on the fail-fast). There is no test that asserts this — a future refactor that re-introduces a silent no-op (the very case this change was protecting against) would not be caught by CI.

Suggestion (one line, in `CachedBodyRequestWrapperTest`):

```java
@Test(expected = UnsupportedOperationException.class)
public void setReadListener_throws() throws java.io.IOException {
  new CachedBodyRequestWrapper(req, BODY).getInputStream().setReadListener(null);
}
```

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
bladehan1, line 221: [NOTE] Batch path peak memory ≈ ~3x maxResponseSize — worth documenting for ops

Not a defect — just an operational observation that helps deployers size the JVM heap correctly.

At this line, the batch path holds three coexisting representations of the response:

1. `batchResult` (ArrayNode of parsed `JsonNode`s) — usually ~1.5–2x the underlying byte size due to object overhead and unicode string expansion.
2. `finalBytes` — the freshly serialized byte[] (~1x the byte size).
3. The just-written response on the way to the socket buffer.

So per concurrent batch the transient peak is roughly ~3x the configured `maxResponseSize` (default 25 MB → ~75 MB peak per batch; 100 concurrent batches → ~7.5 GB). Operators sizing `Xmx` from the upper-bound config alone would underestimate.

Suggestion (no code change required, doc-only):
- Add a line to the `node.json-rpc.max-response-size` comment in `reference.conf`: e.g. "actual peak heap per concurrent batch ≈ 3x this value due to JsonNode tree + serialized bytes coexisting briefly."
- Or an implementation note on `handleBatch`'s javadoc.

This is strictly informational; feel free to skip if it's already covered in your deploy docs.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +89,156 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+  }
+
+  private void handleSingle(HttpServletRequest req, HttpServletResponse resp,
+      JsonNode rootNode, byte[] body, int maxResponseSize) throws IOException {
+    CachedBodyRequestWrapper cachedReq = new CachedBodyRequestWrapper(req, body);
+    BufferedResponseWrapper bufferedResp = new BufferedResponseWrapper(
+        resp, maxResponseSize);
+
+    try {
+      rpcServer.handle(cachedReq, bufferedResp);
```
waynercheung, line 135: [MUST] `maxResponseSize` does not bound memory on the single-request path. jsonrpc4j's `handle(HttpServletRequest, HttpServletResponse)` uses an internal `ByteArrayOutputStream` before writing to the servlet response, so `BufferedResponseWrapper` observes the response too late. Please call `handleRequest(new ByteArrayInputStream(body), boundedOutput)` or equivalent so the size limit is enforced during serialization.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +89,156 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+    for (int i = 0; i < rootNode.size(); i++) {
+      byte[] subBody;
+      try {
+        subBody = MAPPER.writeValueAsBytes(rootNode.get(i));
+      } catch (JsonProcessingException e) {
+        writeJsonRpcError(resp, JsonRpcError.INTERNAL_ERROR, "Internal error", null, true);
+        return;
+      }
+
+      ByteArrayOutputStream subOutput = new ByteArrayOutputStream();
```
waynercheung, line 179: [MUST] Batch sub‑responses are buffered without a hard limit. `subOutput` is a plain `ByteArrayOutputStream`, so one large sub‑response can allocate beyond `maxResponseSize` before line 183 checks the size. Please replace this with a bounded output stream and stop the batch as soon as overflow is detected.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +89,156 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+      // comma separator between array elements
+      int addition = responseBytes.length + (!batchResult.isEmpty() ? 1 : 0);
+      if (maxResponseSize > 0 && accumulatedSize + addition > maxResponseSize) {
+        writeJsonRpcError(resp, JsonRpcError.RESPONSE_TOO_LARGE,
+            "Response exceeds the limit of " + maxResponseSize + " bytes", null, true);
+        return;
+      }
+      accumulatedSize += addition;
+
+      JsonNode responseNode;
```
waynercheung, line 204: [SHOULD] This stores all sub‑responses in `batchResult` and serializes the whole batch again at line 200. That creates another unbounded in‑memory copy before writing the response. Please accumulate serialized sub‑response bytes into a bounded batch buffer instead of building a full `ArrayNode`.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
```diff
@@ -0,0 +1,171 @@
...
+      }
+      buffer.write(b);
+    }
+
+    @Override
+    public void write(byte[] b, int off, int len) {
+      if (overflow) {
+        return;
+      }
+      if (maxBytes > 0 && buffer.size() + len > maxBytes) {
```
waynercheung, line 54: [MUST] `buffer.size() + len` is unsafe int arithmetic and can overflow before the limit check. Please use checked or long arithmetic, e.g. `long nextSize = (long) buffer.size() + len`, and mark overflow when it exceeds `maxBytes`.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -66,6 +89,156 @@ public Integer getJsonRpcCode(int httpStatusCode) {
...
+        return;
+      }
+
+      byte[] responseBytes = subOutput.toByteArray();
+      if (responseBytes.length == 0) {
+        continue; // notification — no response
+      }
+
+      // comma separator between array elements
+      int addition = responseBytes.length + (!batchResult.isEmpty() ? 1 : 0);
```
waynercheung, line 194: [MUST] `responseBytes.length + ...` and `accumulatedSize + addition` are unsafe int additions. Please use checked or long arithmetic for response‑size accounting so overflow cannot bypass `maxResponseSize`.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
```diff
@@ -0,0 +1,171 @@
...
+  public void commitToResponse() throws IOException {
+    if (committed) {
+      throw new IllegalStateException("commitToResponse() already called");
+    }
+    committed = true;
+    if (contentType != null) {
+      actual.setContentType(contentType);
+    }
+    actual.setStatus(status);
+    actual.setContentLength(buffer.size());
```
waynercheung, line 174: [SHOULD] `commitToResponse()` should flush the wrapper writer before reading `buffer.size()`
`commitToResponse()` reads `buffer.size()` and calls `actual.setContentLength(buffer.size())`, but it never flushes the cached `PrintWriter`. `PrintWriter(autoFlush=true)` does not flush on plain `print()` / `write()`, so bytes written through `getWriter()` can remain buffered in the writer/encoder and never reach the internal `ByteArrayOutputStream` before the content length/body are committed.

The current test `BufferedResponseWrapperTest.writeViaWriter_commitToResponse_flushesBody` only covers the explicitly‑flushed case because it calls `w.getWriter().flush()` before `commitToResponse()`. Please make `commitToResponse()` flush the writer before calculating `buffer.size()`, then check `overflow` again because the flush itself can trip the size cap. Also add a regression test for `getWriter().print("hello"); commitToResponse();` without a manual flush.

Latent today because jsonrpc4j 1.6 writes through `getOutputStream()`, but this wrapper exposes `getWriter()` as part of its public servlet API.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: There is no need to tell user who only concerns the result.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Wo can only check the memory's size after single-reques is finished. No need for such low-level refactoring.

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
317787106: There is little possibility to int overflow, so stays.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: The possibility of integer overflow is extremely low, so we'll leave it unchanged for now.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Thanks for your review, add it :
```
    // JSON-RPC 2.0 §6: MUST NOT return an empty Array when there are no response objects.
    if (batchResult.isEmpty()) {
      resp.setStatus(HttpServletResponse.SC_OK);
      resp.setContentLength(0);
      return;
    }
```

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: We can't satisfy both requirements at the same time:

1. If the accumulated size of the first few items does not exceed the threshold, return them normally, while marking the subsequent ones as overflowed.
2. When the overall result overflows, return only a single overflow indicator.

bladehan1 (approved):

framework/src/main/java/org/tron/core/services/filter/BufferedResponseWrapper.java
317787106: Optimize it like this:
```
  public void commitToResponse() throws IOException {
    if (committed) {
      throw new IllegalStateException("commitToResponse() already called");
    }
    committed = true;
    // Flush the PrintWriter's OutputStreamWriter encoder into our ByteArrayOutputStream.
    // PrintWriter(autoFlush=true) only auto-flushes on println/printf/format, not print/write,
    // so bytes can remain buffered in the encoder until an explicit flush.
    writer.flush();
    if (overflow) {
      return;
    }
    if (contentType != null) {
      actual.setContentType(contentType);
    }
    actual.setStatus(status);
    actual.setContentLength(buffer.size());
    buffer.writeTo(actual.getOutputStream());
    actual.getOutputStream().flush();
  }
```

0xbigapple (approved):

PR #6730, opened 2026-04-29
feat(metrics): refine block_transaction_count buckets, add changelog

warku123:
**What does this PR do?**

Two changes related to the metrics added in #6624:

1. Refines the default histogram buckets for `tron:block_transaction_count` from `[0, 10, 50, 100, 200, 500, 1000, 2000, 5000, 10000]` to `[0, 20, 50, 80, 100, 120, 140, 160, 180, 200, 230, 260, 300, 500, 2000]`, densifying the 0–300 range so percentile interpolation in the typical per-SR TPS range is more accurate. Per @Sunny6889's review on #6624.
2. Introduces `METRICS_CHANGELOG.md` to track Prometheus metric additions, changes, and removals going forward, with a Pre-4.8.2 baseline snapshot of currently emitted metrics. Closes one of the post-merge follow-ups noted by @halibobo1205 on #6624.

The bucket change makes the histogram useful for the workload it was designed to observe — most blocks fall under 300 tx, and the previous bucket layout was too coarse there to interpolate percentiles meaningfully. The changelog gives operators a single in-repo place to see what metrics each release adds or removes; today this information is scattered across release notes and the tron-docker `metric_monitor/README.md`.

Related to #6590 (closed when #6624 merged).

- Unit Tests

**Follow up**

- Update tron-docker `metric_monitor/README.md` to mirror the new metrics and the bucket change.

**Extra details**

The existing `PrometheusApiServiceTest` only queries the `le="0.0"` bucket (empty-block detection), which exists in both the old and new bucket arrays, so no test changes are needed.

### DIFF
--- a/METRICS_CHANGELOG.md
+++ b/METRICS_CHANGELOG.md
@@ -1,0 +1,85 @@
+Metrics Changelog
+=================
+
+This file tracks Prometheus metric additions, changes, and removals in java-tron. For the full set of metrics emitted today, see the references at the bottom.
+
+**4.8.2**
+
+### New Metrics
+
+#### Core
+
+- `tron:block_transaction_count` (Histogram, label `miner`) — per-block transaction count, sampled at the entry of `Manager#pushBlock` before any early return so duplicate, stale, and fork-switched pushes are observed alongside applied blocks. Primary use cases: empty-block detection per super representative, and per-SR TPS / throughput percentile interpolation. Default buckets `[0, 20, 50, 80, 100, 120, 140, 160, 180, 200, 230, 260, 300, 500, 2000]` are densified around 0–300 so percentile interpolation in the typical TPS range is more accurate. ([#6624](https://github.com/tronprotocol/java-tron/pull/6624))
+
+#### Consensus
+
+- `tron:sr_set_change` (Counter, labels `action`, `witness`) — incremented once per witness whenever the active SR set rotates at a maintenance boundary. `action` is one of `add` / `remove`. Cardinality grows with the number of distinct witnesses that have ever entered or left the active set, not with the active set size at any given moment. ([#6624](https://github.com/tronprotocol/java-tron/pull/6624))
+
+**Pre-4.8.2 Baseline**
+
+Snapshot of metrics emitted prior to this changelog. Per-version provenance is not tracked here; consult `git log` on [`common/src/main/java/org/tron/common/prometheus/`](common/src/main/java/org/tron/common/prometheus/) for exact origin of each metric.
+
+### Existing Metrics
+
+#### Core (block / transaction processing)
+
+- `tron:header_height` (Gauge) — latest block height on this node.
+- `tron:header_time` (Gauge) — latest block timestamp on this node.
+- `tron:block_push_latency_seconds` (Histogram) — `Manager#pushBlock` latency.
+- `tron:block_process_latency_seconds` (Histogram, label `sync`) — `TronNetDelegate#processBlock` latency.
+- `tron:block_generate_latency_seconds` (Histogram, label `address`) — block generation latency per producer.
+- `tron:block_fetch_latency_seconds` (Histogram) — block fetch latency.
+- `tron:block_receive_delay_seconds` (Histogram) — `receiveTime - blockTime`.
+- `tron:block_fork` (Counter, label `type`) — fork events by type.
+- `tron:lock_acquire_latency_seconds` (Histogram, label `type`) — DB / chain lock acquisition latency.
+- `tron:miner` (Counter, labels `miner`, `type`) — blocks produced by an SR.
+- `tron:miner_latency_seconds` (Histogram, label `miner`) — block mining latency per producer.
+- `tron:miner_delay_seconds` (Histogram, label `miner`) — `actualTime - planTime` for block production.
+- `tron:txs` (Counter, labels `type`, `detail`) — transaction counts.
+- `tron:process_transaction_latency_seconds` (Histogram, labels `type`, `contract`) — transaction processing latency.
+- `tron:verify_sign_latency_seconds` (Histogram, label `type`) — signature verification latency for transactions and blocks.
+- `tron:tx_cache` (Gauge, label `type`) — transaction cache stats.
+- `tron:manager_queue_size` (Gauge, label `type`) — `Manager` queue sizes (pending / popped / queued / repush).
+
+#### Net (P2P)
+
+- `tron:peers` (Gauge, label `type`) — peer counts.
+- `tron:p2p_error` (Counter, label `type`) — P2P error events.
+- `tron:p2p_disconnect` (Counter, label `type`) — P2P disconnect events.
+- `tron:ping_pong_latency_seconds` (Histogram) — peer ping-pong RTT.
+- `tron:message_process_latency_seconds` (Histogram, label `type`) — peer message processing latency.
+- `tron:tcp_bytes` (Histogram, label `type`) — TCP traffic.
+- `tron:udp_bytes` (Histogram, label `type`) — UDP traffic.
+
+#### API
+
+- `tron:http_service_latency_seconds` (Histogram, label `url`) — HTTP endpoint latency.
+- `tron:http_bytes` (Histogram, labels `url`, `status`) — HTTP traffic.
+- `tron:grpc_service_latency_seconds` (Histogram, label `endpoint`) — gRPC endpoint latency.
+- `tron:jsonrpc_service_latency_seconds` (Histogram, label `method`) — JSON-RPC method latency.
+- `tron:internal_service_latency_seconds` (Histogram, labels `class`, `method`) — internal service-call latency.
+- `tron:internal_service_fail` (Counter, labels `class`, `method`) — internal service-call failure count.
+
+#### DB
+
+- `tron:db_size_bytes` (Gauge, labels `type`, `db`, `level`) — LevelDB storage size.
+- `tron:db_sst_level` (Gauge, labels `type`, `db`, `level`) — LevelDB SST files per compaction level.
+- `tron:guava_cache_hit_rate`, `tron:guava_cache_request`, `tron:guava_cache_eviction_count` — Guava cache stats, registered via `GuavaCacheExports` for caches that opt in.
+
+#### System
+
+Emitted by `OperatingSystemExports` (no labels):
+
+- `system_available_cpus`, `system_cpu_load`, `system_load_average`, `system_total_physical_memory_bytes`, `system_free_physical_memory_bytes`, `system_total_swap_spaces_bytes`, `system_free_swap_spaces_bytes`.
+
+#### JVM / process
+
+Auto-emitted by the Prometheus client library via `DefaultExports.initialize()` (`simpleclient_hotspot`). The full list is owned by the upstream library and not enumerated here; see the [client_java](https://github.com/prometheus/client_java) docs. Common ones: `jvm_memory_bytes_*`, `jvm_gc_collection_seconds_*`, `jvm_threads_*`, `process_cpu_seconds_total`, `process_open_fds`, `process_resident_memory_bytes`.
+
+---
+
+**References**
+
+- [Official metrics documentation](https://tronprotocol.github.io/documentation-en/using_javatron/metrics/) — descriptions, configuration, and example queries.
+- [tron-docker `metric_monitor/README.md`](https://github.com/tronprotocol/tron-docker/blob/main/metric_monitor/README.md) — operator-oriented overview with deployment guidance.
+- [java-tron-server (Grafana dashboard 16567)](https://grafana.com/grafana/dashboards/16567-java-tron-server/) — reference dashboard template.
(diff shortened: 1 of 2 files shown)


METRICS_CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+
+This file tracks Prometheus metric additions, changes, and removals in java-tron. For the full set of metrics emitted today, see the references at the bottom.
+
+**4.8.2**
+
+### New Metrics
+
+#### Core
+
+- `tron:block_transaction_count` (Histogram, label `miner_address`) — per-block transaction count, sampled at the entry of `Manager#pushBlock` before any early return so duplicate, stale, and fork-switched pushes are observed alongside applied blocks. Primary use cases: empty-block detection per super representative, and per-SR throughput observation. Default buckets: `[0, 1, 10, 50, 100, 200, 500, 1000, 2000, 5000]`. ([#6624](https://github.com/tronprotocol/java-tron/pull/6624))
```
SeriousCoding789: Suggest change the default buckets to `[0, 20, 50, 80, 100, 120, 140, 160, 180, 200, 230, 260, 300, 500, 2000]`, as it will increase the calculation accuracy.

SeriousCoding789: (Histogram, label `miner`)

METRICS_CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+
+#### Consensus
+
+- `tron:sr_set_change` (Counter, labels `witness`, `event`) — incremented once per witness whenever the active SR set rotates at a maintenance boundary. `event` is one of `added` / `removed`. Cardinality grows with the number of distinct witnesses that have ever entered or left the active set, not with the active set size at any given moment. ([#6624](https://github.com/tronprotocol/java-tron/pull/6624))
```
SeriousCoding789: Please check the below more accurate version:

> `tron:sr_set_change` (Counter, labels `action`, `witness`) — incremented once per witness whenever the active SR set rotates at a maintenance boundary. `action` is one of `add` / `remove`. Cardinality grows with the number of distinct witnesses that have ever entered or left the active set, not with the active set size at any given moment. ([#6624](https://github.com/tronprotocol/java-tron/pull/6624))

Changes from the original:
- `event` → `action` (label name)
- `added` / `removed` → `add` / `remove` (label values)

METRICS_CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+
+Auto-emitted by the Prometheus client library via `DefaultExports.initialize()` (`simpleclient_hotspot`). The full list is owned by the upstream library and not enumerated here; see the [client_java](https://github.com/prometheus/client_java) docs. Common ones: `jvm_memory_bytes_*`, `jvm_gc_collection_seconds_*`, `jvm_threads_*`, `process_cpu_seconds_total`, `process_open_fds`, `process_resident_memory_bytes`.
+
+---
+
+**References**
+
+- [Official metrics documentation](https://tronprotocol.github.io/documentation-en/using_javatron/metrics/) — descriptions, configuration, and example queries.
+- [tron-docker `metric_monitor/README.md`](https://github.com/tronprotocol/tron-docker/blob/main/metric_monitor/README.md) — operator-oriented overview with deployment guidance.
+- [java-tron-server (Grafana dashboard 16567)](https://grafana.com/grafana/dashboards/16567-java-tron-server/) — reference dashboard template.
```
SeriousCoding789: How about refer https://github.com/tronprotocol/tron-docker/blob/main/metric_monitor/grafana_dashboard/java-tron-server.json as it will maintained and updated in time.

METRICS_CHANGELOG.md
warku123: Done in 425288821 — applied the new bucket array to both `MetricsHistogram.java` and the changelog entry, and reworded the rationale to call out densifying around 0–300 for percentile interpolation in the typical per-SR TPS range. Thanks for the suggestion!

warku123: Done in 425288821 — fixed `miner_address` → `miner`.

METRICS_CHANGELOG.md
warku123: Done in 425288821 — adopted the rewrite verbatim. Thanks for catching the drift from the actual code (`event`→`action`, `added`/`removed`→`add`/`remove`).

METRICS_CHANGELOG.md
warku123: Good call — switching to the in-repo JSON in 61d19934e. As you noted, `grafana.com/dashboards/16567` isn't being updated anymore, and the in-org JSON tracks the current panels. Note that the panels for the two new metrics from #6624 are pending in tronprotocol/tron-docker#118; once that lands the dashboard JSON will reflect them too.

Sunny6889 (approved):

METRICS_CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+
+#### Consensus
+
+- `tron:sr_set_change` (Counter, labels `action`, `witness`) — incremented once per witness whenever the active SR set rotates at a maintenance boundary. `action` is one of `add` / `remove`. Cardinality grows with the number of distinct witnesses that have ever entered or left the active set, not with the active set size at any given moment. ([#6624](https://github.com/tronprotocol/java-tron/pull/6624))
+
+**Pre-4.8.2 Baseline**
+
+Snapshot of metrics emitted prior to this changelog. Per-version provenance is not tracked here; consult `git log` on [`common/src/main/java/org/tron/common/prometheus/`](common/src/main/java/org/tron/common/prometheus/) for exact origin of each metric.
+
+### Existing Metrics
```
bladehan1, line 24: [NIT] Pre-4.8.2 baseline omits `tron:error_info`

The baseline claims to be a complete snapshot of metrics emitted prior to this changelog, but it does not list `tron:error_info` (Counter, labels `topic`, `type`), which is registered in `common/src/main/java/org/tron/common/prometheus/InstrumentedAppender.java:12-17` and emitted on every ERROR-level log line. The metric is already actively asserted in `framework/src/test/java/org/tron/core/metrics/prometheus/PrometheusApiServiceTest.java:114`.

Leaving it out of the baseline means a future PR that renames or removes it will have no documented baseline to compare against, which is the exact failure mode this changelog is meant to prevent.

Suggestion: Add a one-line entry under one of the existing subsections (Core or a new "Logging" subsection) describing `tron:error_info` (Counter, labels `topic`, `type`) — emitted on every ERROR-level log line by `InstrumentedAppender`.

METRICS_CHANGELOG.md
```diff
@@ -0,0 +1,85 @@
...
+- `tron:http_service_latency_seconds` (Histogram, label `url`) — HTTP endpoint latency.
+- `tron:http_bytes` (Histogram, labels `url`, `status`) — HTTP traffic.
+- `tron:grpc_service_latency_seconds` (Histogram, label `endpoint`) — gRPC endpoint latency.
+- `tron:jsonrpc_service_latency_seconds` (Histogram, label `method`) — JSON-RPC method latency.
+- `tron:internal_service_latency_seconds` (Histogram, labels `class`, `method`) — internal service-call latency.
+- `tron:internal_service_fail` (Counter, labels `class`, `method`) — internal service-call failure count.
+
+#### DB
+
+- `tron:db_size_bytes` (Gauge, labels `type`, `db`, `level`) — LevelDB storage size.
```
bladehan1: [NIT] DB metrics are not LevelDB-only — RocksDB also emits these

`tron:db_size_bytes` and `tron:db_sst_level` are emitted from `chainbase/src/main/java/org/tron/common/storage/metric/DbStat.java:18-19`, where the first label value comes from `DbStat#getEngine()`. Both `LevelDbDataSourceImpl#getEngine()` (returns `LEVELDB`) and `RocksDbDataSourceImpl#getEngine()` (returns `ROCKSDB`) extend `DbStat`, so on a node configured with the RocksDB backend these metrics are emitted with `engine="ROCKSDB"`.

Describing them as "LevelDB storage size" / "LevelDB SST files per compaction level" will mislead operators of RocksDB-backend nodes into thinking the metrics are missing, or into writing PromQL that implicitly filters for LevelDB and drops half their data.

Suggestion: Reword as engine-aware, e.g. "Storage size in bytes per (engine, db, level); engine ∈ {LEVELDB, ROCKSDB}." The first label is named `type` in code (`MetricsGauge.java:19-20`) but its values are storage engines — calling that out in the description avoids the confusion.

common/src/main/java/org/tron/common/prometheus/MetricsHistogram.java
```diff
@@ -51,7 +51,7 @@ public class MetricsHistogram {
 
     init(MetricKeys.Histogram.BLOCK_TRANSACTION_COUNT,
         "Distribution of transaction counts per block.",
-        new double[]{0, 10, 50, 100, 200, 500, 1000, 2000, 5000, 10000},
+        new double[]{0, 20, 50, 80, 100, 120, 140, 160, 180, 200, 230, 260, 300, 500, 2000},
```
bladehan1: [DISCUSS] Consider keeping 5000/10000 as safety-net buckets above 2000

The densified 0–300 range is a real improvement, but capping the array at 2000 trades away an observability property the previous layout had: visibility into outlier blocks. Because buckets are cumulative and the client library always appends `+Inf`, any block with > 2000 transactions now lands in `+Inf` only — you can read "how many exceeded 2000" as `+Inf - 2000`, but you can't tell whether those samples are 2001 or 50000. The previous array `[..., 2000, 5000, 10000]` did give that resolution, even if its lower-range buckets were too coarse for the typical workload.

A hybrid layout keeps both properties:

```
[0, 20, 50, 80, 100, 120, 140, 160, 180, 200,
 230, 260, 300, 500, 2000, 5000, 10000]
```

- The 0–300 densification stays intact for typical-TPS percentile interpolation.
- 5000 and 10000 act as safety-net buckets for rare bursts (pathological pushes, repush storms, fork-switched blocks observed before early-return at `Manager.java:1273-1277`).
- True outliers above 10000 still fall into `+Inf`, which is appropriate — at that point it's an incident signal, not something quantiles should smooth over.

Adding two buckets has no cardinality impact (buckets are series-internal, not label values) and negligible per-`observe()` cost. It strictly subsumes the alerting hook proposed in the other comment: the `+Inf` ratio alert still works as a tripwire, and operators get richer detail when it does fire.

Happy to defer this if there's a deliberate reason for the 2000 cap (e.g. aligning with an external dashboard's expected max), but didn't see one in the PR description.

METRICS_CHANGELOG.md
warku123: Added a dedicated **Logging** subsection to the Pre-4.8.2 Baseline with `tron:error_info` (Counter, labels `topic`, `type`) — incremented on every ERROR-level log line by `InstrumentedAppender`.

Also took the opportunity to do a full audit of all metric registration sites (`MetricsCounter`, `MetricsGauge`, `MetricsHistogram`, `InstrumentedAppender`, `OperatingSystemExports`, `GuavaCacheExports`). Found one more gap: `process_cpu_load` (defined in `OperatingSystemExports`) was missing from the System section — added that in the same commit.

METRICS_CHANGELOG.md
warku123: Rewrote both entries to be engine-aware:
- `tron:db_size_bytes` — "storage size in bytes per engine, database, and level; `type` is the storage engine (`LEVELDB` or `ROCKSDB`) depending on node configuration."
- `tron:db_sst_level` — "SST files per compaction level per engine and database; `type` is the storage engine (`LEVELDB` or `ROCKSDB`) depending on node configuration."

Also noticed the Guava cache entries were missing their `type` label (cache name) — expanded those into individual descriptions with the label noted.

common/src/main/java/org/tron/common/prometheus/MetricsHistogram.java
warku123: Added an Operational note under `tron:block_transaction_count` recommending a `+Inf` overflow ratio alert as the signal to re-tune the upper bound.

On the bucket cap itself: real-world TRON blocks average around 300–350 tx, and even at the observed peak TPS (~272) a 3-second block tops out around ~800 tx — well under 2000. The 2000 cap covers both normal and peak operation; the `+Inf` ratio alert in the operational note serves as the tripwire for pathological cases (repush storms, fork-switched blocks). Keeping the array short with the alert as the escalation path seems sufficient for now, but happy to add 5000/10000 back if you feel the extra resolution on those outlier events is worth it.

bladehan1: In stress testing scenarios, the number will exceed 2000.

warku123: Good point on the stress-test / repush scenario. Added 5000 and 10000 back as safety-net buckets — the array is now `[0, 20, 50, 80, 100, 120, 140, 160, 180, 200, 230, 260, 300, 500, 2000, 5000, 10000]`. The 0–300 densification for typical-TPS percentiles is unchanged, and the `+Inf` operational note is updated to reference 10000 as the new effective upper bound.

bladehan1 (approved):
LGTM

Sunny6889 (approved):

PR #6732, opened 2026-04-29
fix(jsonrpc): enforce log filter cap and improve match efficiency

### LINKED ISSUE #6510: Parallelizing eth_newFilter event matching
# Background

The JSON-RPC event subscription API `eth_newFilter` is a stateful service designed for efficiently querying newly generated transaction events. This mechanism works by matching user-defined filter criteria against the Bloom filters of block events, and returning the matched results through the `eth_getFilterChanges` API.

Currently, the event-matching process is executed in a single-threaded, sequential manner. Under high concurrency with a large number of active subscriptions, the matching latency increases significantly, causing the event processing throughput to fall behind the block synchronization speed. As a result, users experience noticeable delays when retrieving the latest events.

To address this issue, we propose optimizing the existing sequential matching mechanism into a parallel matching model to improve overall throughput and processing efficiency. In addition, an upper limit will be enforced on the number of concurrently created subscriptions. When this limit is exceeded, the system will return a clear and explicit error message to prevent resource exhaustion and further improve user experience.

# Rationale

To improve user experience by preventing delayed delivery of subscribed events.

### What are the use-cases?

The combination of `eth_newFilter` and `eth_getFilterChanges` can be used to build event-driven WebSocket services for real-time event subscriptions.

# Specification

After a FullNode finishes processing a block, all transaction events generated within that block are written into a FIFO queue named `filterCapsuleQueue`. A background thread, `filterProcessLoop`, continuously consumes events from this queue and processes them block by block. For each block, the `handleLogsFilter` method is invoked to match the events against the subscription conditions created via `eth_newFilter`. Matched events are temporarily stored in a cache and can later be retrieved by users through `eth_getFilterChanges`.

If a subscription is not accessed within 300 seconds, it is considered expired, and its associated cached data will be automatically cleaned up.

In the current implementation, if there are m active subscriptions, each of the n events in a block must be matched against all subscriptions, resulting in a time complexity of O(m × n). When the number of subscriptions reaches approximately 200,000, the matching time for a single block can take up to 20 seconds, which significantly exceeds the block production interval (3 seconds). This leads to severe event processing lag: existing subscribers experience substantial delays, while newly created subscriptions starting from the latest block may fail to receive any events for an extended period of time.

To resolve this issue, the sequential matching model between block events and subscriptions is optimized by partitioning subscriptions into groups and matching them in parallel against the event list of a block. At the same time, the ordering of events within each individual subscription is strictly preserved.

Since event matching is a CPU-intensive operation, the level of parallelism is intentionally limited to 2–4 threads to avoid negatively impacting block synchronization and other core processing workflows.

Furthermore, to prevent uncontrolled matching latency under extreme subscription volumes, a configuration parameter `node.jsonrpc.maxLogFilterNum` is introduced to cap the maximum number of log filters that can be created. When the number of active subscriptions exceeds this threshold, new subscription requests will be rejected with a clear error message: “exceed max log filters“.

Service operators can adjust the default maximum subscription limit based on their hardware resources and system load.

# Test Specification

Stress testing can be used to evaluate block-level event matching latency and CPU utilization. Test results will be provided in a follow-up update.

# Scope of Impact

This change only affects the JSON-RPC `eth_newFilter` interface.

# Implementation

Yes.

Yes.

317787106:
## Problem

Three issues exist in the JSON-RPC log filter subsystem (`eth_newFilter` / `eth_getLogs`), reported in tronprotocol/java-tron#6510:

1. **Unbounded memory growth** — `eth_newFilter` imposes no limit on the number of active filters held in memory. A client can register filters indefinitely and eventually exhaust heap on the node.

2. **Correctness bug** — `LogMatch.matchBlockOneByOne` evaluated the `MAX_RESULT` guard *after* `addAll`, so the result list could transiently contain more than `MAX_RESULT` entries before the exception was thrown.

3. **Performance bottleneck under high filter count** — `handleLogsFilter` iterated the filter map with an `Iterator` and called `result.add()` once per element, which is slow and contention-prone when thousands of filters are active.

---

### 1. Enforce a configurable cap on active log filters

Adds `node.jsonrpc.maxLogFilterNum` (default: `20000`). When the cap is reached, `eth_newFilter` immediately returns `JsonRpcExceedLimitException` (JSON-RPC code `-32005`) instead of growing without bound.

```hocon
node {
  jsonrpc {
    # Maximum number of concurrent eth_newFilter registrations (0 = unlimited)
    maxLogFilterNum = 20000
  }
}
```

### 2. Fix the MAX_RESULT boundary check in `LogMatch.matchBlockOneByOne` (correctness)

Moves the `size + matchedLog.size() > MAX_RESULT` guard **before** `addAll`. The result list now never exceeds `MAX_RESULT` entries regardless of how many logs a single block contributes.

### 3. Optimize `handleLogsFilter` for large filter maps

| Condition | Before | After |
|-----------|--------|-------|
| filter count ≤ 10,000 (common case) | `Iterator` loop, `result.add()` per element | `ConcurrentHashMap.forEach` with local list + single `addAll` |
| filter count > 10,000 | same slow path | bounded `ForkJoinPool(3)` + `parallelStream` |

Key improvements:
- **Bounded parallelism**: `logsFilterPool` is a `ForkJoinPool(3)` created once per `TronJsonRpcImpl` instance and shut down with it, avoiding unbounded thread creation under spike load.
- **Observability**: a `logger.debug` timing line records dispatch cost and filter-map size.

### 4. Decouple filter state from static class scope

The four filter maps (`eventFilter2ResultFull`, `blockFilter2ResultFull`, `eventFilter2ResultSolidity`, `blockFilter2ResultSolidity`), `logsFilterPool`, and `filterParallelThreshold` are now **instance fields**. `handleBLockFilter`, `handleLogsFilter`, and `processLogFilterEntry` are now **instance methods**.

Motivation: with static shared state, any test that constructs a `TronJsonRpcImpl` instance would silently read and write the same filter maps used by the running node, making test isolation impossible. Switching to instance state means each instance owns its own filter maps; tests create a lightweight instance with no Spring dependencies (`new TronJsonRpcImpl(null, null, null)`) and operate on an isolated map.

As a consequence, `BlockFilterCapsule` and `LogsFilterCapsule` now receive the `TronJsonRpcImpl` instance at construction time and call `handleBLockFilter` / `handleLogsFilter` on it directly, eliminating the static imports that were the only way to invoke those methods before.

### 5. Behavor changes when node.jsonrpc.maxBlockFilterNum = 0 

It forbid all new block filters when maxBlockFilterNum = 0 in release-4.8.1, now, it is not limited. 

It has the same meaning as node.jsonrpc.maxLogFilterNum =0 for log filter number. 

---

## Testing

- **Unit tests** (new / updated):
  - `HandleLogsFilterTest` — 10 cases covering event dispatch, block-range filtering, expired-filter removal, solidified/non-solidified routing, and **3 parallel-path cases** (all-filters receive events, expired eviction, solidified routing — each with `filterParallelThreshold` lowered via reflection to keep the test fast)
  - `LogMatchOverLimitTest` — 4 cases covering under-limit, exact-limit, exceed-limit (verifies exception is thrown *before* `addAll`), and empty-block skip
  - `WalletCursorTest.testNewFilter_exceedsCapThrowsException` — verifies the cap throws the correct exception with the limit value in the message; fixed to populate the map on the **same instance** that calls `newFilter` (the original test was populating the static map then constructing a new instance, so it never actually tested the cap)
  - `BlockFilterCapsuleTest`, `LogsFilterCapsuleTest`, `ConcurrentHashMapTest` — updated to pass a `TronJsonRpcImpl` instance to constructors after the static-to-instance refactoring
- Manual testing

Closes tronprotocol/java-tron#6510


common/src/main/resources/reference.conf
```diff
@@ -400,6 +400,9 @@ node {
 
     # Maximum number for blockFilter
     maxBlockFilterNum = 50000
```
alan-eth, line 397: Really nice to see a default-on safeguard added — and using `0` as a sentinel for "unlimited" is a clean pattern.

Minor doc accuracy: the description here says "Maximum number of log entries returned by eth_getLogs / eth_newFilter", but the value actually caps the number of *concurrent filter registrations*, not the result size. The framework `config.conf` describes it correctly ("Allowed maximum number for newFilter, >0 otherwise no limit"). Could we align the wording in `reference.conf` to avoid operators misreading it as a per-query cap?

Suggestion:
```hocon
# Maximum number of concurrent eth_newFilter registrations (>0; 0 means unlimited)
maxLogFilterNum = 20000
```

317787106: Thanks for your suggestion, I fix it.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -169,6 +171,8 @@ public enum RequestSource {
   private static final String NO_BLOCK_HEADER_BY_HASH = "header for hash not found";
 
   private static final String ERROR_SELECTOR = "08c379a0"; // Function selector for Error(string)
+  private static final int FILTER_PARALLEL_THRESHOLD = 10000;
+  private static final ForkJoinPool LOGS_FILTER_POOL = new ForkJoinPool(2);
```
0xbigapple: [SHOULD]`LOGS_FILTER_POOL` is declared as a static (class-level) `ForkJoinPool`, meaning it is shared across all instances within the JVM. However, it is being shut down inside an instance-level close() method. Once close() is invoked, the shared pool transitions to a terminated state and cannot accept new tasks. Any subsequent usage (including from other instances or test cases) will result in RejectedExecutionException.
This is especially problematic in unit tests, where one test invoking close() can unintentionally affect others due to the shared static resource.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -1440,7 +1462,7 @@ public String newBlockFilter() throws JsonRpcMethodNotFoundException,
     } else {
       blockFilter2Result = blockFilter2ResultSolidity;
     }
-    if (blockFilter2Result.size() >= maxBlockFilterNum) {
+    if (maxBlockFilterNum > 0 && blockFilter2Result.size() >= maxBlockFilterNum) {
```
0xbigapple, line 1475: `newBlockFilter` semantics changed silently. 
**Old:** maxBlockFilterNum=0 blocked all calls. 
**New:** means "unlimited". 
Worth calling out in the PR description; also update the `reference.conf `comment for `maxBlockFilterNum` — currently only `maxLogFilterNum` documents the new semantics.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchOverLimitTest.java
```diff
@@ -0,0 +1,156 @@
...
+    Assert.assertEquals(MAX_RESULT, results.length);
+  }
+
+  /**
+   * Verifies the fix: when the second block would push the total over MAX_RESULT,
+   * {@link JsonRpcTooManyResultException} is thrown BEFORE {@code addAll}.
+   * The result list must contain only the logs from block 1 (not the partial block-2 logs).
+   */
+  @Test
+  public void testExceedsLimit_throwsBeforeAddAll()
```
0xbigapple: [NIT] `testExceedsLimit_throwsBeforeAddAll` doesn't actually verify "before addAll". The pre-fix code also threw  `JsonRpcTooManyResultException`

framework/src/test/java/org/tron/core/jsonrpc/LogMatchOverLimitTest.java
```diff
@@ -0,0 +1,156 @@
...
+      throws BadItemException, ItemNotFoundException, JsonRpcInvalidParamsException {
+    // block 1: MAX_RESULT - 1 logs, block 2: 2 logs → 9999 + 2 = 10001 > MAX_RESULT
+    Manager manager = buildMockManager(
+        1L, buildTxList(MAX_RESULT - 1),
+        2L, buildTxList(2));
+    LogMatch logMatch = buildLogMatch(Arrays.asList(1L, 2L), manager);
+
+    try {
+      logMatch.matchBlockOneByOne();
+      Assert.fail("Expected JsonRpcTooManyResultException");
```
0xbigapple: [NIT] Use assertThrows instead of try/fail/catch for expected-exception tests

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -115,6 +116,7 @@ public enum RequestSource {
   private static final String FILTER_NOT_FOUND = "filter not found";
   public static final int EXPIRE_SECONDS = 5 * 60;
   private static final int maxBlockFilterNum = Args.getInstance().getJsonRpcMaxBlockFilterNum();
+  private static final int maxLogFilterNum = Args.getInstance().getJsonRpcMaxLogFilterNum();
```
0xbigapple: [NIT] `maxLogFilterNum` is snapshotted into static final at class load. That's why the cap test has to populate 20K entries — Args.set...()
  from tests has no effect. Reading via Args.getInstance() on each call (singleton, cheap) would let tests inject cap=5.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchOverLimitTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.tron.core.services.jsonrpc.TronJsonRpc.FilterRequest;
+import org.tron.core.services.jsonrpc.TronJsonRpc.LogFilterElement;
+import org.tron.core.services.jsonrpc.filters.LogBlockQuery;
+import org.tron.core.services.jsonrpc.filters.LogFilterWrapper;
+import org.tron.core.services.jsonrpc.filters.LogMatch;
+import org.tron.protos.Protocol.TransactionInfo;
+import org.tron.protos.Protocol.TransactionInfo.Log;
+
+/**
+ * Verifies the over-limit check in {@link LogMatch#matchBlockOneByOne()} introduced in PR #71.
```
0xbigapple: [NIT] Javadoc reference to "PR #71" in LogMatchOverLimitTest is fork-local, should drop it.

framework/src/test/java/org/tron/core/jsonrpc/HandleLogsFilterTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.tron.common.runtime.vm.DataWord;
+import org.tron.common.runtime.vm.LogInfo;
+import org.tron.core.exception.jsonrpc.JsonRpcInvalidParamsException;
+import org.tron.core.services.jsonrpc.TronJsonRpc.FilterRequest;
+import org.tron.core.services.jsonrpc.TronJsonRpcImpl;
+import org.tron.core.services.jsonrpc.filters.FilterResult;
+import org.tron.core.services.jsonrpc.filters.LogFilterAndResult;
+import org.tron.protos.Protocol.TransactionInfo;
+
+public class HandleLogsFilterTest {
```
0xbigapple, line 21: [SHOULD] Parallel branch has zero test coverage.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -57,6 +57,8 @@
 @Slf4j(topic = "API")
 public class JsonRpcApiUtil {
 
+  static SecureRandom random = new SecureRandom();
```
0xbigapple: [NIT] Hoisting random to a static field is a nice perf win👍. One small follow-up: consider tightening it to private static final:
`  private static final SecureRandom random = new SecureRandom();`
As written (package-private, non-final), anything in `org.tron.core.services.jsonrpc` can reassign it. A well-meaning test or perf experiment could quietly do: `JsonRpcApiUtil.random = new Random(42);// compiles, no warning`
 …and filter IDs silently become predictable, which breaks the unguessability assumption clients rely on. private final closes that path at compile time, and also gives JMM safe-publication guarantees for free.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchOverLimitTest.java
317787106: Delete it, thanks.

framework/src/test/java/org/tron/core/jsonrpc/HandleLogsFilterTest.java
317787106: I have add more testcases, thanks for your suggestion.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchOverLimitTest.java
317787106: Fix it.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: ForkJoinPool has been fixed as local varible, not static. Thanks for your suggesction.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: `maxLogFilterNum` has been optimized as local varibles. Thanks for your suggestion.

framework/src/test/java/org/tron/core/jsonrpc/LogMatchOverLimitTest.java
317787106: Godd suggsetion, optimize it.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Good suggestion, I have updated two config files and PR description.

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -276,6 +278,10 @@ public class Manager {
   @Autowired
   private RewardViCalService rewardViCalService;
 
+  @Lazy
+  @Autowired
+  private TronJsonRpcImpl tronJsonRpcImpl;
```
bladehan1, line 284: [SHOULD] Avoid @Lazy circular dependency from Manager into TronJsonRpcImpl

`Manager` is the core db / block-processing component of java-tron. Its dependency graph is supposed to be inbound-only (other components depend on Manager). This PR reverses that by injecting `TronJsonRpcImpl` (an API-layer @Component) into Manager, forming a cycle that can only be resolved with `@Lazy`. The cycle exists because the new `BlockFilterCapsule` / `LogsFilterCapsule` carry a `TronJsonRpcImpl` reference, so Manager has to obtain that reference before constructing each capsule.

Why this matters even though it compiles:
- `@Lazy` is a band-aid for circular dependencies, not a fix. It defers bean resolution to first method call and hides container init-order problems.
- Spring 6 / Spring Boot 3 default `spring.main.allow-circular-references=false`. Any Spring upgrade can fail on this exact cycle.
- Once Manager accepts a reverse dependency on the API layer, future capsule types follow the same path — the breach is contagious.
- `postBlockFilter` is on a hot path; if `TronJsonRpcImpl` is mid-initialization (e.g. during a restart), the proxy can throw `BeanCurrentlyInCreationException`.

Two options to resolve before merge:

**Option A — drop the reverse dependency (smallest change, matches the `hotfix/fix_newFilter` branch name):**
- Revert capsules to plain POJOs (no `TronJsonRpcImpl` field, original ctor signature).
- Drop `@Lazy @Autowired TronJsonRpcImpl` from Manager.
- Keep the three real fixes — `maxLogFilterNum` cap, `LogMatch.matchBlockOneByOne` MAX_RESULT check, and `forEach` + `ForkJoinPool` performance — none of them require static→instance.
- For test isolation, expose a static `clearAllFilters()` helper (or use `@Before` / `@After` to clear the four maps).
- Pros: minimal blast radius for a hotfix; production cache stays cross-instance; no `@Lazy`.
- Cons: test isolation depends on test discipline; multi-instance scenarios still share state (not an issue today since the bean is a singleton).

**Option C — Visitor / Dispatcher (cleanest, may warrant splitting from this PR):**
- Keep capsules as POJOs (no service reference).
- Dispatch in `Manager.filterProcessLoop` directly, e.g.:
  ```
  FilterTriggerCapsule c = filterCapsuleQueue.poll(...);
  if (c instanceof LogsFilterCapsule)  tronJsonRpcImpl.handleLogsFilter((LogsFilterCapsule) c);
  else if (c instanceof BlockFilterCapsule) tronJsonRpcImpl.handleBLockFilter((BlockFilterCapsule) c);
  ```
  Or — preferred — introduce `@Component FilterTriggerDispatcher` that holds `TronJsonRpcImpl` (forward dep). Manager `@Autowired`s the dispatcher (also forward), so all edges Manager → Dispatcher → JsonRpcImpl are forward. Cycle gone.
- Pros: capsules become genuinely independent, testable, serializable; Manager has no API-layer awareness; future capsule types just extend the dispatcher.
- Cons: slightly more code than the current PR (one extra bean or one `instanceof` block).

Suggestion: pick A or C. If A, this PR is lighter than today; if C, the architecture is durable and the `@Lazy` ban is enforced naturally.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -1426,7 +1453,7 @@ public String newBlockFilter() throws JsonRpcMethodNotFoundException,
     } else {
       blockFilter2Result = blockFilter2ResultSolidity;
     }
-    if (blockFilter2Result.size() >= maxBlockFilterNum) {
+    if (maxBlockFilterNum > 0 && blockFilter2Result.size() >= maxBlockFilterNum) {
```
bladehan1, line 1475: [SHOULD] Highlight maxBlockFilterNum=0 as a behavior reversal in release notes

The new check is `maxBlockFilterNum > 0 && size >= maxBlockFilterNum`. Previously it was `size >= maxBlockFilterNum`, so `maxBlockFilterNum = 0` rejected every `eth_newBlockFilter` call (effectively a kill switch). After this change, `0` means unlimited. Operators who set `0` to disable the endpoint will silently flip from "forbid all" to "allow all" on upgrade.

The PR description and reference.conf already mention `0 = unlimited`, but neither flags this as a backward-incompatible change.

Suggestion: Add an explicit "behavior change" note in the PR body / release notes ("was: forbid all, now: unlimited"), and consider `grep`'ing known deployment configs for `maxBlockFilterNum = 0` before release.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -166,6 +168,13 @@ public enum RequestSource {
   private static final String NO_BLOCK_HEADER_BY_HASH = "header for hash not found";
 
   private static final String ERROR_SELECTOR = "08c379a0"; // Function selector for Error(string)
+  private final int filterParallelThreshold = 10000;
+  /**
+   * Using the default maxLogFilterNum of 20,000, a 3-thread pool can keep up with log event
+   * processing for each block within the 3-second BLOCK_PRODUCED_INTERVAL. Increasing the thread
+   * pool size too much may affect the performance of the main block processing thread.
+   */
+  private final ForkJoinPool logsFilterPool = new ForkJoinPool(3);
```
bladehan1: [SHOULD] Create logsFilterPool through ExecutorServiceManager for consistent thread naming

Every other thread pool in the project is built via `ExecutorServiceManager` so threads carry a recognizable name (`query-section-0`, `event-trigger`, etc.). `new ForkJoinPool(3)` here produces the JDK default `ForkJoinPool-N-worker-M` names, making thread dumps and logs harder to correlate to the JSON-RPC subsystem. The shutdown path already routes through `ExecutorServiceManager.shutdownAndAwaitTermination` at line 1589, so creation is the only place that diverges.

Suggestion: Add a `newForkJoinPool(name, parallelism, isDaemon)` factory in `ExecutorServiceManager` that supplies a `ForkJoinWorkerThreadFactory` setting `setName(name + "-N")`, or pass a custom factory inline so the worker threads identify themselves.

framework/src/main/java/org/tron/common/logsfilter/capsule/BlockFilterCapsule.java
```diff
@@ -19,19 +18,21 @@ public class BlockFilterCapsule extends FilterTriggerCapsule {
   @Setter
   private boolean solidified;
 
-  public BlockFilterCapsule(BlockCapsule block, boolean solidified) {
-    blockHash = block.getBlockId().toString();
-    this.solidified = solidified;
+  private final TronJsonRpcImpl jsonRpc;
```
bladehan1: [SHOULD] Exclude jsonRpc from Lombok @ToString

The class is annotated `@ToString`, which now includes the new `jsonRpc` field. `TronJsonRpcImpl` has no `toString` of its own, so the generated representation will include `org.tron.core.services.jsonrpc.TronJsonRpcImpl@<hash>` — useless in logs and a future risk if `TronJsonRpcImpl` ever gets `@ToString`, since it would recurse through capsules.

Suggestion: Annotate the field with `@ToString.Exclude` (and `@EqualsAndHashCode.Exclude` if the class adds `@EqualsAndHashCode` later).

framework/src/main/java/org/tron/common/logsfilter/capsule/LogsFilterCapsule.java
```diff
@@ -33,18 +32,22 @@ public class LogsFilterCapsule extends FilterTriggerCapsule {
   @Setter
   private boolean removed;
 
+  private final TronJsonRpcImpl jsonRpc;
```
bladehan1: [SHOULD] Exclude jsonRpc from Lombok @ToString

The class is annotated `@ToString`, which now includes the new `jsonRpc` field. `TronJsonRpcImpl.toString()` will fall back to `Object#toString` (`org.tron...TronJsonRpcImpl@<hash>`), polluting any log line that prints the capsule and risking recursive `toString` if `TronJsonRpcImpl` later gets `@ToString` on its own.

Suggestion: Annotate the field with `@ToString.Exclude`.

framework/src/test/java/org/tron/core/jsonrpc/WalletCursorTest.java
```diff
@@ -136,9 +143,41 @@ public void testEnableInFullNode() {
...
+  @Test
+  public void testNewFilter_exceedsCapThrowsException() throws Exception {
+    int cap = Args.getInstance().getJsonRpcMaxLogFilterNum();
+    FilterRequest fr = new FilterRequest();
+    TronJsonRpcImpl tronJsonRpc = new TronJsonRpcImpl(nodeInfoService, wallet, dbManager);
+    Map<String, LogFilterAndResult> map = tronJsonRpc.getEventFilter2ResultFull();
+    List<String> addedKeys = new ArrayList<>();
+
+    try {
+      for (int i = 0; i < cap; i++) {
```
bladehan1, line 173: [SHOULD] Reduce cap before testing newFilter limit instead of populating 20000 entries

The test reads `Args.getInstance().getJsonRpcMaxLogFilterNum()` (default 20000) and pre-populates the map with that many `LogFilterAndResult` instances — each carrying its own `LogFilterWrapper` and `LinkedBlockingQueue`. That is roughly 20 MB of heap allocated only to drive a `>= cap` boundary check, and it makes this single test responsible for a large fraction of the test class's runtime and GC pressure.

Suggestion: Either inject a smaller cap before constructing the instance (e.g. via `Args.setParam` or by setting `maxLogFilterNum` reflectively to 5, mirroring the `setParallelThreshold` helper in `HandleLogsFilterTest`), or expose the field as test-writable. The boundary semantics being verified are independent of the actual cap value.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -218,54 +227,69 @@ public static void handleBLockFilter(BlockFilterCapsule blockFilterCapsule) {
...
+      try {
+        // compare with hashcode() first, then with equals(). If not exist, put it.
+        cachedElement = logElementCache.get(element, () -> element);
+      } catch (ExecutionException e) {
+        logger.error("Getting/loading LogFilterElement from cache fails", e); // never happen
+        cachedElement = element;
       }
+      localResults.add(cachedElement);
     }
+    logFilterAndResult.getResult().addAll(localResults);
```
bladehan1, line 306: [NIT] Local list + addAll does not batch on LinkedBlockingQueue

The PR description claims the new pattern reduces lock contention by "collecting matched elements into a local list, then inserting with a single `addAll` on the shared `LinkedBlockingQueue`." `LinkedBlockingQueue` does not override `addAll`, so it inherits `AbstractCollection.addAll`, which calls `add(e)` in a loop — N invocations of `putLock`, identical to the previous inline `add` per element. The intermediate `ArrayList` adds an allocation without changing the lock count.

Suggestion: Update the PR description to say the local buffer is for keeping per-thread state local in the parallel path (not for batched insertion), or remove the buffer in the serial path and `add` directly to the queue.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -1407,7 +1431,10 @@ public String newFilter(FilterRequest fr) throws JsonRpcInvalidParamsException,
     } else {
       eventFilter2Result = eventFilter2ResultSolidity;
     }
-
+    if (maxLogFilterNum > 0 && eventFilter2Result.size() >= maxLogFilterNum) {
```
bladehan1, line 1453: [NIT] Cap is best-effort under concurrent newFilter

The `size() >= cap` check and the subsequent `put` are not atomic, so concurrent `eth_newFilter` calls can land between the check and the insert. In the worst case the map can briefly hold `cap + concurrent_requests` entries. This matches the pre-existing `maxBlockFilterNum` pattern, so it is not a regression, but the security framing of the cap (DoS hardening) implies hard limits.

Suggestion: Document the cap as approximate in `reference.conf` ("may be exceeded by concurrent registrations"). A stricter implementation would track size with an `AtomicInteger` and CAS-increment before `put`.

framework/src/test/java/org/tron/core/jsonrpc/HandleLogsFilterTest.java
```diff
@@ -0,0 +1,285 @@
...
+    jsonRpc.handleLogsFilter(capsule);
+
+    Assert.assertTrue(filterAndResult.getResult().isEmpty());
+  }
+
+  // -------------------------------------------------------------------------
+  // Parallel path tests (filterParallelThreshold lowered to 2 via reflection)
+  // -------------------------------------------------------------------------
+
+  private void setParallelThreshold(int value) throws Exception {
```
bladehan1: [NIT] Reflective write of a private final int is fragile

`setParallelThreshold` reflectively sets `private final int filterParallelThreshold`. Non-static final fields can be written via `setAccessible(true)` on JDK 8/11/17, but the pattern is brittle:
- if the field becomes `static final` it requires the (deprecated, JDK 17-broken) `modifiers` hack;
- JIT inlining of final fields can race with the reflective write in some scenarios;
- adding `--add-opens` is a moving target across JDK upgrades.

Suggestion: Drop `private final` to package-private (or expose a `@VisibleForTesting` setter) so tests can override the threshold without reflection.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -218,54 +227,69 @@ public static void handleBLockFilter(BlockFilterCapsule blockFilterCapsule) {
...
+        logsFilterCapsule.isSolidified() ? "Solidity" : "Full", t2 - t1, eventFilterMap.size());
+  }
+
+  private void processLogFilterEntry(
+      Map.Entry<String, LogFilterAndResult> entry,
+      Map<String, LogFilterAndResult> eventFilterMap,
+      LogsFilterCapsule logsFilterCapsule) {
+    LogFilterAndResult logFilterAndResult = entry.getValue();
+    if (logFilterAndResult.isExpire()) {
+      eventFilterMap.remove(entry.getKey());
```
bladehan1, line 273: [NIT] Use ConcurrentMap.remove(key, value) for expired-filter eviction

The parallel path replaces the old `Iterator.remove()` with `eventFilterMap.remove(entry.getKey())`. `Iterator.remove()` only removed the snapshot mapping the iterator was looking at; `map.remove(key)` will remove any current mapping for that key, including a newly-inserted one. Filter IDs are 16 random bytes so a collision is astronomically unlikely, but the CAS variant matches the original semantic precisely.

Suggestion: `eventFilterMap.remove(entry.getKey(), entry.getValue())` — only delete the entry the caller actually saw.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -117,13 +118,14 @@ public enum RequestSource {
 
   private static final String FILTER_NOT_FOUND = "filter not found";
   public static final int EXPIRE_SECONDS = 5 * 60;
-  private static final int maxBlockFilterNum = Args.getInstance().getJsonRpcMaxBlockFilterNum();
-  private static final Cache<LogFilterElement, LogFilterElement> logElementCache =
+  private final int maxBlockFilterNum = Args.getInstance().getJsonRpcMaxBlockFilterNum();
+  private final int maxLogFilterNum = Args.getInstance().getJsonRpcMaxLogFilterNum();
+  private final Cache<LogFilterElement, LogFilterElement> logElementCache =
```
bladehan1: [NIT] logElementCache and blockHashCache do not need instance scope

The motivation for the static→instance change is to make filter state isolatable in tests. That argument applies to the four `*Filter2Result*` maps but not to these two caches: they are LRU dedup caches with no per-test-driven semantics, and moving them to instance scope means every `new TronJsonRpcImpl(null,null,null)` allocates two Guava caches (300k + 60k entries) plus their internal book-keeping.

Suggestion: Keep `logElementCache` and `blockHashCache` `static` to limit the blast radius of the refactor and reduce per-test cost. The four filter maps remain instance-scoped as needed.

framework/src/main/java/org/tron/common/logsfilter/capsule/BlockFilterCapsule.java
317787106: Thanks for your suggestion, I have removes `jsonRpc`.

framework/src/main/java/org/tron/core/db/Manager.java
317787106: Very good suggestion. To keep durable, `@lazy` is removed and use `instanceof`.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
317787106: `static` is added.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Restore static to `logElementCache` and `blockHashCache`. Adding it or not has no additional impact.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Use `ExecutorServiceManager.newForkJoinPool` to replace `new ForkJoinPool`

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Generally, there are almost maxLogFilterNum Filter IDs at the same, they shoule never be same. Remove it by key is more clear.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Thanks for your correction. To keep the meaning of several numeric parameters under JSON-RPC consistent, a value of 0 is uniformly treated as having no size limit. The PR description has been revised with emphasis.

framework/src/test/java/org/tron/core/jsonrpc/HandleLogsFilterTest.java
317787106: Use following:
```
  @Setter
  private int filterParallelThreshold = 10000;
```

framework/src/test/java/org/tron/core/jsonrpc/WalletCursorTest.java
317787106: Good suggestion, use following code instead to reduce memory allocated:
```
int cap = 5;
Args.getInstance().setJsonRpcMaxLogFilterNum(cap);
```

framework/src/main/java/org/tron/common/logsfilter/capsule/LogsFilterCapsule.java
317787106: `jsonRpc` has been removed.

framework/src/main/java/org/tron/core/db/Manager.java
317787106: ### Summary
It don't works if remove `@lazy` so i add it back.   FullNode.java:50 disables Spring circular dependency resolution:
```
beanFactory.setAllowCircularReferences(false);
```
After removing `@Lazy`, Spring detects the following dependency cycle during startup and rejects it directly:

```
Manager → TronJsonRpcImpl (required by constructor ↓)
               NodeInfoService → @Autowired Manager  ← cycle!
               Wallet          → @Autowired Manager  ← cycle!
```
Why `FilterTriggerDispatcher` Alone Does Not Break the Cycle ?                                                                                                                                                                  
                                                                                                                                                                                                                                   
  ### The Proposed Dependency Graph        
 ```                                                                                                                                                                                        
  Manager  ──field──▶  FilterTriggerDispatcher  ──field──▶  TronJsonRpcImpl                                                                                                                                                        
  ```                           
  At first glance, all three edges point in one direction — no cycle.
                                                                                                                                                                                                                                   
  ### The Hidden Transitive Dependencies
                                                                                                                                                                                                                                   
  `TronJsonRpcImpl` uses **constructor injection** for two dependencies:

  ```java
  @Autowired
  public TronJsonRpcImpl(NodeInfoService nodeInfoService, Wallet wallet) { … }
                                                                                                                                                                                                                                   
  Both of those beans carry a hard dependency back to Manager:
                                                                                                                                                                                                                                   
  // NodeInfoService.java    
  @Autowired private Manager dbManager;

  // Wallet.java
  @Autowired private Manager dbManager;
```
  Expanding the full graph:                                                                                                                                                                                                        
  ``` 
  Manager                                                                                                                                                                                                                          
    └─field──▶ FilterTriggerDispatcher
                 └─field──▶ TronJsonRpcImpl
                              ├─ctor──▶ NodeInfoService ──field──▶ Manager  ✗
                              └─ctor──▶ Wallet          ──field──▶ Manager  ✗                                                                                                                                                      
   ```
  ### Why Spring Rejects This                                                                                                                                                                                                          
                             
  FullNode.java explicitly disables circular-reference resolution:
  ```
  beanFactory.setAllowCircularReferences(false);   // FullNode.java:50
  ```                                                                                                                                                                                                                                
  With this flag, Spring does not expose an early singleton reference for beans that are
  still being created. The creation sequence becomes:                                                                                                                                                                              
                             
  1. Spring begins creating Manager → added to "currently in creation" set.                                                                                                                                                        
  2. Manager needs FilterTriggerDispatcher → Spring creates it.
  3. FilterTriggerDispatcher needs TronJsonRpcImpl → Spring creates it.                                                                                                                                                            
  4. TronJsonRpcImpl's constructor requires NodeInfoService → Spring creates it.
  5. NodeInfoService needs Manager → already in "currently in creation", no early reference available.                                                                                                                             
  6. Spring attempts to create Manager again → BeanCurrentlyInCreationException.
                                                                                                                                                                                                                                   
  The new class does not appear anywhere in the cycle; the cycle runs entirely through
  TronJsonRpcImpl → NodeInfoService/Wallet → Manager.                                                                                                                                                                              
                             
  ### What Would Actually Fix It

  The FilterTriggerDispatcher design only eliminates the cycle if one of the following is                                                                                                                                          
  also true:
                                                                                                                                                                                                                                   
```  ┌────────────────────────────────────────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────┐
  │                                         Additional Change                                          │                                          Effect                                          │
  ├────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
  │ FilterTriggerDispatcher injects TronJsonRpcImpl with @Lazy                                         │ Defers creation past Manager's init; equivalent to the current @Lazy fix, just relocated │
  ├────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
  │ TronJsonRpcImpl switches NodeInfoService and Wallet to setter injection                            │ Constructor no longer forces eager creation of Manager-dependent beans                   │                                
  ├────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
  │ FilterTriggerDispatcher looks up TronJsonRpcImpl via ApplicationContext.getBean() at dispatch time │ Bypasses Spring DI for this edge entirely                                                │                                
  └────────────────────────────────────────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────┘                                
   ```
  Without one of these additions, introducing FilterTriggerDispatcher changes only the                                                                                                                                             
  name of the path through which the cycle travels, not the cycle itself.

### What to do next ?
If a more systematic architectural refactor is planned in the future, FilterTriggerDispatcher is worth tracking in a separate issue and addressing together with changing NodeInfoService / Wallet to setter injection.

File: framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java (hunk 16 above)
Comment (317787106) [reply framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java:306]: Remove some specification about lock contention in PR description.

File: framework/src/main/java/org/tron/core/db/Manager.java (hunk 10 above)
Comment (bladehan1) [reply framework/src/main/java/org/tron/core/db/Manager.java:284]: Thanks for the detailed analysis — you are right that introducing `FilterTriggerDispatcher` alone does not break the cycle, because the cycle actually runs through `TronJsonRpcImpl → NodeInfoService / Wallet → Manager` regardless of whether a Dispatcher sits in between. My original suggestion was incomplete on that point.

But I think the conversation we're having so far is still framed at the wrong level. The three workarounds you listed (`@Lazy` on the Dispatcher edge, switching `NodeInfoService` / `Wallet` to setter injection, `ApplicationContext.getBean()` lookup) are all techniques to **make Spring tolerate the cycle**. None of them remove the underlying design problem, which is:

> **`Manager` should not depend on `TronJsonRpcImpl` at all.**

The reason this dependency exists today is that `Manager` is hosting both the producer side (`postBlockFilter` / `postLogsFilter` → `queue.offer(...)`) **and** the consumer side (`filterProcessLoop` → calls into `TronJsonRpcImpl`) of the same queue. The consumer side is where the reverse edge comes from. Move it out and the entire cycle disappears — no `@Lazy`, no setter-injection rewrite, no `ApplicationContext.getBean()`.

## Concrete plan: move `filterProcessLoop` into `TronJsonRpcImpl`

**Manager** keeps only the producer role:

```java
@Component
public class Manager {
    private final BlockingQueue<FilterTriggerCapsule> filterCapsuleQueue =
        new LinkedBlockingQueue<>();

    public BlockingQueue<FilterTriggerCapsule> getFilterCapsuleQueue() {
        return filterCapsuleQueue;
    }

    private void postBlockFilter(...) { filterCapsuleQueue.offer(new BlockFilterCapsule(blk, sol)); }
    private void postLogsFilter(...)  { filterCapsuleQueue.offer(new LogsFilterCapsule(...));    }

    // delete: @Lazy @Autowired private TronJsonRpcImpl tronJsonRpcImpl;
    // delete: filterProcessLoop Runnable
    // delete: filterEs / filterEsName / isRunFilterProcessThread
}
```

**TronJsonRpcImpl** owns the consumer role:

```java
@Component
public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
    private final Manager manager;  // back to ctor injection (forward edge)
    private volatile boolean running = true;
    private ExecutorService filterEs;

    @Autowired
    public TronJsonRpcImpl(NodeInfoService nodeInfoService, Wallet wallet, Manager manager) { ... }

    @PostConstruct
    public void start() {
        if (!CommonParameter.getInstance().isJsonRpcFilterEnabled()) return;
        filterEs = ExecutorServiceManager.newSingleThreadExecutor("filter-process");
        ExecutorServiceManager.submit(filterEs, this::filterProcessLoop);
    }

    private void filterProcessLoop() {
        BlockingQueue<FilterTriggerCapsule> queue = manager.getFilterCapsuleQueue();
        while (running) {
            try {
                FilterTriggerCapsule c = queue.poll(1, TimeUnit.SECONDS);
                if (c instanceof LogsFilterCapsule)        handleLogsFilter((LogsFilterCapsule) c);
                else if (c instanceof BlockFilterCapsule)  handleBLockFilter((BlockFilterCapsule) c);
            } catch (InterruptedException e) { Thread.currentThread().interrupt(); }
              catch (Throwable t) { logger.error("filterProcessLoop error", t); }
        }
    }

    @Override public void close() {
        running = false;
        ExecutorServiceManager.shutdownAndAwaitTermination(filterEs, "filter-process");
        // existing logsFilterPool / sectionExecutor shutdown
    }
}
```

Resulting Spring graph (all edges forward, no cycle):

```
TronJsonRpcImpl ──► Manager
TronJsonRpcImpl ──► NodeInfoService ──► Manager
TronJsonRpcImpl ──► Wallet          ──► Manager
Manager ──► (nothing in jsonrpc)
```

## Why this cycle showed up *now*, and why it is worth fixing properly

Before this PR, `BlockFilterCapsule` / `LogsFilterCapsule` called `TronJsonRpcImpl.handleBLockFilter` / `handleLogsFilter` via `import static`. Those are Java static dispatches, not Spring bean references — Spring's container had no idea Manager depended on `TronJsonRpcImpl`. The cycle existed in the *call graph* the whole time; Spring just couldn't see it.

This PR turning those methods into instance methods (which is the right thing to do for test isolation of the filter maps) forces the dependency to be expressed as `@Autowired`, and only then does Spring complain. So the cycle isn't introduced here — it's *exposed* here. Adding `@Lazy` papers over an existing layering issue that the previous static-import pattern was hiding.

If we ship `@Lazy` as the final state, the next time someone does the same cleanup on another piece of jsonrpc state (a counter, a metrics object, another cache), they'll add another `@Lazy` field for the same reason, and the layering breach compounds.

## Suggestion

For this PR: consider Plan X above. The diff is small (~30 lines net), the cycle is genuinely removed (no `@Lazy`), and the fix matches the existing branch name (`hotfix/fix_newFilter`).

If the PR is already too large in scope to add this: leave the current `@Lazy` and open a follow-up issue framed as **"Move `filterProcessLoop` into `TronJsonRpcImpl` so `Manager` no longer depends on jsonrpc"** rather than "switch `NodeInfoService` / `Wallet` to setter injection". The latter only quiets Spring; the former actually fixes the layering.

317787106: Very good suggestion. The PR is enough large now, maybe need to refactor `TronJsonRpcImpl` ( including remove `@Lazy`) in next ISSUE or PR. The main goal of removing modifier `static` from four ConcurrentHashMap is to solve 10 testcase failure that happens in parallel test introduced by PR `t.maxParallelForks ` https://github.com/tronprotocol/java-tron/pull/6447/changes, so `TronJsonRpcImpl` needs to be optimized in current PR.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -166,25 +169,34 @@ public enum RequestSource {
...
   /**
    * thread pool of query section bloom store
    */
   private final ExecutorService sectionExecutor;
   private final NodeInfoService nodeInfoService;
   private final Wallet wallet;
-  private final Manager manager;
+  @Setter
+  @Autowired
+  private Manager manager;
```
bladehan1, line 190: [NIT] Restrict @Setter visibility on the manager field to package-private

The refactor moved `Manager manager` from constructor injection to field injection with `@Setter @Autowired private Manager manager;`. Lombok's default `@Setter` is public, so production callers can now invoke `tronJsonRpc.setManager(other)` and silently swap the core db reference. The setter only exists for tests (which need `tronJsonRpc.setManager(dbManager)` after the constructor stops accepting Manager).

Suggestion: `@Setter(AccessLevel.PACKAGE)` or `@VisibleForTesting` (Guava) so production code cannot mutate the wired Manager. Tests in the same package still get access.

framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -332,8 +338,10 @@ public class Manager {
         while (isRunFilterProcessThread) {
           try {
             FilterTriggerCapsule filterCapsule = filterCapsuleQueue.poll(1, TimeUnit.SECONDS);
-            if (filterCapsule != null) {
-              filterCapsule.processFilterTrigger();
+            if (filterCapsule instanceof LogsFilterCapsule) {
+              tronJsonRpcImpl.handleLogsFilter((LogsFilterCapsule) filterCapsule);
+            } else if (filterCapsule instanceof BlockFilterCapsule) {
+              tronJsonRpcImpl.handleBLockFilter((BlockFilterCapsule) filterCapsule);
             }
```
bladehan1, line 346: [NIT] Log unknown FilterTriggerCapsule subtypes instead of dropping silently

The new dispatcher uses an `instanceof` chain that only handles `LogsFilterCapsule` and `BlockFilterCapsule`. Any future `FilterTriggerCapsule` subtype will be polled from the queue and silently dropped — no warning, no exception. The original polymorphic `processFilterTrigger()` covered new subtypes for free.

Suggestion: Add a trailing `else if (filterCapsule != null) logger.warn("Unknown FilterTriggerCapsule type: {}", filterCapsule.getClass().getName());`. A `Map<Class<?>, Consumer<FilterTriggerCapsule>>` registered at startup would make extension explicit.

common/src/main/java/org/tron/common/es/ExecutorServiceManager.java
```diff
@@ -35,6 +38,17 @@ public static ScheduledExecutorService newSingleThreadScheduledExecutor(String n
         new ThreadFactoryBuilder().setNameFormat(name).setDaemon(isDaemon).build());
   }
 
+  public static ForkJoinPool newForkJoinPool(String name, int parallelism) {
```
bladehan1, line 41: [NIT] Add isDaemon overload to newForkJoinPool to match sibling factories

All the other factories in this class come in pairs — `newSingleThreadExecutor(name)` / `newSingleThreadExecutor(name, isDaemon)`, same for `newSingleThreadScheduledExecutor` and `newFixedThreadPool`. The new `newForkJoinPool(String, int)` only has the single-arg form, with no way to opt into daemon threads.

Suggestion: Add `newForkJoinPool(String name, int parallelism, boolean isDaemon)` that calls `thread.setDaemon(isDaemon)` inside the worker factory; have the two-arg form delegate with `isDaemon=false` (matching `newFixedThreadPool`'s default).

framework/src/test/java/org/tron/core/jsonrpc/WalletCursorTest.java
```diff
@@ -132,13 +142,48 @@ public void testEnableInFullNode() {
...
   }
 
+  /**
+   * When the active filter count reaches the configured cap (node.jsonrpc.maxLogFilterNum),
+   * eth_newFilter must throw JsonRpcExceedLimitException instead of growing without bound.
+   */
+  @Test
+  public void testNewFilter_exceedsCapThrowsException() throws Exception {
+    int cap = 5;
+    Args.getInstance().setJsonRpcMaxLogFilterNum(cap);
```
bladehan1, line 165: [NIT] Restore Args singleton in @After to avoid cross-test pollution

`Args.getInstance().setJsonRpcMaxLogFilterNum(5)` mutates a JVM-wide singleton. The test does not restore the original value on exit, so any later test (in this class or — when surefire does not fork — in any other test class) that reads `jsonRpcMaxLogFilterNum` will see `5` instead of the configured default (20000). This is the exact pattern the project's `knowledge/test-singleton-pollution.md` warns against.

Suggestion: Save the original cap before the mutation and restore it in `finally` (or in an `@After` method): `int original = Args.getInstance().getJsonRpcMaxLogFilterNum(); ... finally { Args.getInstance().setJsonRpcMaxLogFilterNum(original); }`.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -166,25 +169,34 @@ public enum RequestSource {
   private static final String NO_BLOCK_HEADER_BY_HASH = "header for hash not found";
 
   private static final String ERROR_SELECTOR = "08c379a0"; // Function selector for Error(string)
```
bladehan1, line 172: [NIT] Restrict @Setter visibility on filterParallelThreshold to package-private

Replacing reflection with `@Setter` is a clear improvement, but Lombok's default `@Setter` is public, so production callers can change the threshold at runtime. The setter is only needed by tests (which already live in the same package).

Suggestion: `@Setter(AccessLevel.PACKAGE)` or annotate with `@VisibleForTesting`.

framework/src/main/java/org/tron/core/db/Manager.java
317787106: There are only two types of FilterTriggerCapsule. There is no need to log it.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Yes, the `maxLogFilterNum` is not matched exactly, it's an approximate value. Generally, its effect is little, overly detailed configuration files may increase operational overhead; comments can be added in the code instead.

common/src/main/java/org/tron/common/es/ExecutorServiceManager.java
317787106: Add following in `ExecutorServiceManager`:
```
public static ForkJoinPool newForkJoinPool(String name, int parallelism, boolean isDaemon) {...}
``` 
`isDaemon` is `false` default.

framework/src/test/java/org/tron/core/jsonrpc/WalletCursorTest.java
317787106: Restored it.  Really, `Args.clearParams()` will restore it automatically, there is no need to restore it manually.

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Remove `@setter`, add following code instead:
```
  @VisibleForTesting
  public void setFilterParallelThreshold(int filterParallelThreshold) {
    this.filterParallelThreshold = filterParallelThreshold;
  }
```

framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
317787106: Add following code instead:
```
  @VisibleForTesting
  public void setManager(Manager manager) {
    this.manager = manager;
  }
```
Not all the testcase invoking it are belong to package `org.tron.core.services.jsonrpc`, so cannot restrict `@Setter` visibility on the manager field to package-private

0xbigapple (approved):
LGTM

bladehan1 (approved):

PR #6734, opened 2026-04-30
feat(framework): support solidity conditional shutdown

### LINKED ISSUE #6610: [Feature] SolidityNode supports conditional shutdown
# Background

A SolidityNode is designed to provide query services for blocks and events that are finalized. Its core value lies in stability, determinism, and suitability for production workloads. Since it does not need to perform signature verification for transactions and blocks, its synchronization speed is over 30% faster than a FullNode.

However, compared to a FullNode, SolidityNode lacks certain functionalities, which introduces inconvenience for users. For example, it currently does not support conditional shutdown, which is critical for scenarios such as verifying the global state at a specific block height.

# Rationale

- To enable the SolidityNode to stop automatically based on specified conditions, allowing users to query and verify the state at a target block.

### What are the use cases?
- Synchronizing a specific block range more quickly than FullNode
- Performing state consistency verification across nodes

# Specification

The following configuration allows customizing the automatic shutdown conditions. Only one of the three options can be enabled at a time:
```
node.shutdown = {
#  BlockTime  = "54 59 08 * * ?"  # Trigger shutdown when block header time matches
#  BlockHeight = 33350800         # Trigger shutdown when block height matches
#  BlockCount = 12                # Trigger shutdown after syncing a number of blocks since startup
}
```
## Design Context

To verify global execution consistency across heterogeneous architectures (e.g., x86 vs.  ARM) during full sync, the blockchain is partitioned into segments based on snapshots for parallel processing. Data continuity is validated by ensuring the Final State of a preceding segment perfectly matches the Initial State of the subsequent one. While SolidityNodes accelerate this synchronization, their current inability to halt at a specific block height necessitates further enhancements.

## Implementation Approach

We introduce these three condition checks from `TronNetDelegate` into the `org.tron.program.SolidityNode` class.

Additionally: Force `p2pdisable` = true, ensure that:
- Core block processing logic remains unchanged
- Shutdown conditions can still be triggered correctly

No modification is required for the existing configuration format.

# Test Specification

Set different conditions to verify whether the node shutdowns as expected. 

# Scope of Impact

This change only affects the Solidity Node.

# Implementation

Yes

Yes

317787106:
## Summary

Closes tronprotocol/java-tron#6610.

Refactors `SolidityNode` into a Spring-managed component with proper lifecycle support, enabling it to respond to conditional shutdown signals (block height, block time, block count) already implemented in `TronNetDelegate.processBlock`. The previous static `start()` approach bypassed Spring entirely, so the shutdown conditions could never be reached during solidity sync.

## Problem

`SolidityNode` previously initialized its own `TronApplicationContext`, created a `SolidityNode` instance manually, and called `appT.blockUntilShutdown()` in a separate JVM entry path. This architecture had two consequences:

1. **Shutdown conditions were unreachable.** The conditional-shutdown logic lives in `TronNetDelegate.processBlock` (checking `latestSolidityNumShutDown`), but `SolidityNode` called `Manager.pushVerifiedBlock -> Manager.pushBlock` — a path that never passes through `TronNetDelegate`. The shutdown signal was therefore silently ignored.
2. **No graceful resource cleanup.** Thread pools and gRPC connections were abandoned on JVM exit rather than being released through Spring's `@PreDestroy` lifecycle.

### `SolidityNode` (framework)
- Converted from a static utility class to a Spring `@Component` with `@Conditional(SolidityCondition.class)` — only registered when `--solidity` flag is set.
- Implements `ApplicationListener<ContextClosedEvent>`: sets `flag = false` at the very start of context close, **before** any `@PreDestroy` methods run on other beans, preventing block pushes into a partially-shut-down system.
- `@PostConstruct init()`: initialises named `ExecutorService` thread pools via `ExecutorServiceManager`. gRPC client and `remoteBlockNum` are initialised in `run()` to keep init lightweight.
- `@PreDestroy shutdown()`: sets `flag = false` (safety net), shuts down both executors via `ExecutorServiceManager.shutdownAndAwaitTermination`, and closes the gRPC client cleanly.
- `processSolidityBlock()` uses `blockQueue.poll(timeout)` instead of `take()`, and handles `InterruptedException` explicitly so thread interruption during executor shutdown is clean.
- `getBlockByNum()`: loop changed to `while (flag)`; throws `RuntimeException("SolidityNode is closing.")` when flag is false — avoids `NullPointerException` from `blockQueue.put(null)`.
- `getLastSolidityBlockNum()`: loop changed to `while (flag)`; returns `0` when flag is false — lets the caller's `while (flag)` loop exit quietly without a misleading error log.
- `loopProcessBlock()` now calls `tronNetDelegate.pushVerifiedBlock()` instead of `dbManager.pushVerifiedBlock()`, routing blocks through the shutdown-condition check in `processBlock`.

### `TronNetDelegate` (framework)
- Added `pushVerifiedBlock(BlockCapsule)`: sets `generatedByMyself = true` and delegates to `processBlock(block, true)`, which already contains the conditional-shutdown logic. Includes timing log consistent with the removed `Manager` method.

### `Manager` (framework)
- Removed `pushVerifiedBlock()`. Block push for solidity sync now goes through `TronNetDelegate`.

### `FullNode` (framework)
- In solidity mode, `trustNodeAddr` validation and `p2pDisable = true` are now enforced before the Spring context is created.
- After `appT.startup()`, retrieves the `SolidityNode` bean from the context and calls `run()` — reusing the same application lifecycle rather than spawning a second context.

## Tests

- `testSolidityGrpcCall` — verifies gRPC connectivity through the running `RpcApiService`.
- `testSolidityNodeHttpApiService` — verifies HTTP service start/stop idempotency.
- `testExecutorsInitializedOnStartup` — `@PostConstruct` creates both executors before `run()` is called.
- `testOnApplicationEventSetsFlagFalse` — `ContextClosedEvent` sets `flag = false` before any `@PreDestroy` runs.
- `testGetBlockByNumThrowsWhenClosed` — `RuntimeException` (not `null`) is thrown when `flag = false`.
- `testGetLastSolidityBlockNumReturnsZeroWhenClosed` — returns `0` when `flag = false` for a quiet loop exit.
- `testSolidityConditionMatchesWhenSolidityFlagSet` — `SolidityCondition.matches()` returns `true` with `--solidity`.


framework/src/main/java/org/tron/program/FullNode.java
```diff
@@ -28,19 +29,24 @@ public static void main(String[] args) {
...
-    logger.info("Full node running.");
-    if (Args.getInstance().isDebug()) {
-      logger.info("in debug mode, it won't check energy time");
+    if (parameter.isSolidityNode()) {
+      logger.info("Solidity node is running.");
+      if (ObjectUtils.isEmpty(parameter.getTrustNodeAddr())) {
+        throw new TronError(new IllegalArgumentException("Trust node is not set."),
+            TronError.ErrCode.SOLID_NODE_INIT);
+      }
+      parameter.setP2pDisable(true);
```
halibobo1205: What is the intended purpose of forcing `parameter.setP2pDisable(true)` for SolidityNode here? Since `ApplicationImpl.startup()` already skips `tronNetService.start()` when `Args.getInstance().isSolidityNode()` is true, this does not seem necessary to prevent P2P startup. It also changes behavior for code paths that check `isP2pDisable()`, such as `Manager.initAutoStop()`, where `shutdownBlockHeight == headNum` will no longer exit immediately in solidity mode. Is this flag being set for another side effect, or should the existing `isSolidityNode()` checks be sufficient?

317787106: Good catch, there is no need to set p2pDisable = false explictly as it skip the `tronNetService.start()` for `SolidityNode`, I will remove it.

framework/src/main/java/org/tron/program/SolidityNode.java
```diff
@@ -1,227 +1,223 @@
...
+        logger.error("Failed to get block: {}, reason: {}.", blockNum, e.getMessage());
+        sleep(exceptionSleepTime);
+      }
+    }
+    //throw RuntimeException instead of return null to avoid NullPointException
+    throw new RuntimeException("SolidityNode is closing.");
+  }
+
+  private long getLastSolidityBlockNum() {
+    while (flag) {
```
xxo1shine: [SHOULD] Suggest `while (flag && !tronNetDelegate.isHitDown())`.  It is recommended to encapsulate this logic into a function, and each `while(flag)` loop can be replaced.

framework/src/main/java/org/tron/program/SolidityNode.java
```diff
@@ -1,227 +1,223 @@
...
+      } catch (Exception e) {
+        logger.error("Failed to process block {}.", new BlockCapsule(block), e);
+        sleep(exceptionSleepTime);
+        block = getBlockByNum(blockNum);
+      }
+    }
+  }
+
+  private Block getBlockByNum(long blockNum) {
+    while (flag) {
```
xxo1shine: [SHOULD] Suggest `while (flag && !tronNetDelegate.isHitDown())`.

SeriousCoding789:
**PR description is out of sync with the code.** The description says "p2pDisable = true is now enforced before the Spring context is created", but commit `c02c3e2a1a` removed `parameter.setP2pDisable(true)` from `FullNode.java`. The current code does not set this flag for solidity nodes.

This isn't just a doc nit — removing it changes behavior at two places that still rely on `isP2pDisable()`:

- [`[Manager.java:765](https://claude.ai/epitaxy/framework/src/main/java/org/tron/core/db/Manager.java:765)`](framework/src/main/java/org/tron/core/db/Manager.java:765): `if (exitHeight == headNum && !isP2pDisable())` → solidity nodes will now hit this startup-exit branch instead of skipping it.
- [`[Wallet.java:2747](https://claude.ai/epitaxy/framework/src/main/java/org/tron/core/Wallet.java:2747)`](framework/src/main/java/org/tron/core/Wallet.java:2747): `listNodes()` will now call `TronNetService.getP2pService().getConnectableNodes()` on solidity nodes, where the P2p service was never `start()`-ed.

Could you:
1. Update the PR description to remove the stale "p2pDisable = true" line.
2. Explain in the commit message why `setP2pDisable(true)` was removed.
3. Confirm whether the behavior changes at `Manager.java:765` and `Wallet.java:2747` are intentional. If not, those sites should switch from `!isP2pDisable()` to `!isSolidityNode()` (or add the latter as an additional guard).

xxo1shine (approved):

halibobo1205 (approved):

framework/src/main/java/org/tron/program/SolidityNode.java
317787106: Yes, I miss this logic when merge PR that exists.

PR #6637, opened 2026-04-02
feat(plugins): migrate keystore CLI from FullNode to Toolkit

barbatos2011:
## What

Migrate the `--keystore-factory` interactive REPL from FullNode.jar to Toolkit.jar as picocli subcommands, extract the keystore core library from `framework` to `crypto` module, and add new capabilities (list, update, SM2 support, JSON output, password-file support).

## Why

**The current `--keystore-factory` has several limitations:**

1. **Coupled to FullNode.jar** — operators must have the full node binary and configuration to manage keystore files, even though keystore operations have zero dependency on the node runtime
2. **Interactive REPL only** — cannot be scripted, piped, or used in CI/CD pipelines; no support for `--password-file` or structured output
3. **Security gaps** — `Scanner`-based password input echoes to terminal; no `--key-file` option forces private keys to be typed interactively or passed as command-line arguments (visible in `ps`, shell history)
4. **Module coupling** — the keystore library (`Wallet.java`, `WalletUtils.java`) lives in `framework` and calls `Args.getInstance()`, preventing reuse by `plugins` (Toolkit.jar) without pulling in the entire framework dependency chain

### Commit 1: `refactor(crypto): extract keystore library from framework module`

Extract the keystore package from `framework` to `crypto` module to break the framework dependency:

| File | Change |
|------|--------|
| `Wallet.java` | Moved to `crypto/`, removed `Args` import, added `boolean ecKey` parameter to `decrypt()`, changed MAC comparison to constant-time `MessageDigest.isEqual()` |
| `WalletUtils.java` | Moved to `crypto/`, removed `Args` import, added `boolean ecKey` parameter to `generateNewWalletFile()`, `generateFullNewWalletFile()`, `generateLightNewWalletFile()`, `loadCredentials()` |
| `Credentials.java`, `WalletFile.java` | Moved to `crypto/` unchanged |
| `WitnessInitializer.java` | Updated to pass `Args.getInstance().isECKeyCryptoEngine()` to `loadCredentials()` |
| `KeystoreFactory.java` | Updated to pass `CommonParameter.getInstance().isECKeyCryptoEngine()` to keystore calls |
| `plugins/build.gradle` | Added `implementation project(":crypto")` with exclusions (libp2p, prometheus, aspectj, httpcomponents) |
| `CredentialsTest.java` | Merged two test files (fixed `keystroe` typo directory) into `crypto` module |
| `WalletFileTest.java` | Moved to `crypto` module |

[...]

PR #6738, opened 2026-05-01
fix(plugins): exclude transitive leveldbjni-all from :crypto

barbatos2011:
## Summary

x86_64 builds of `Toolkit.jar` since #6637 throw `NoSuchMethodError` on `org.iq80.leveldb.Options.maxBatchSize(I)` when running commands such as `db archive` / `--rebuild-manifest`:

```
Exception in thread "main" java.lang.NoSuchMethodError:
  org.iq80.leveldb.Options.maxBatchSize(I)Lorg/iq80/leveldb/Options;
    at org.tron.plugins.DbArchive$ArchiveManifest.newDefaultLevelDbOptions(DbArchive.java:137)
    ...
```

This PR fixes the root cause and adds CI smoke tests so the same class of bug cannot reach a release artifact again.

## Why

#6637 introduced `implementation(project(":crypto"))` to the `plugins` module without excluding `org.fusesource.leveldbjni:leveldbjni-all`, which `:crypto` pulls in transitively via `:common -> :platform`. The direct `:platform` dependency in `plugins/build.gradle` already excludes `leveldbjni-all` (so the TRON-fork `io.github.tronprotocol:leveldbjni-all:1.18.2` can be used instead), but the new transitive path through `:crypto` does not. The result on x86 is two jars carrying `org/iq80/leveldb/Options.class` on the runtime classpath:

- `org.fusesource.leveldbjni:leveldbjni-all:1.8` — does NOT have `Options.maxBatchSize(int)`
- `io.github.tronprotocol:leveldbjni-all:1.18.2` — DOES have `Options.maxBatchSize(int)`

`:plugins:dependencyInsight` confirms the regression path:

```
org.fusesource.leveldbjni:leveldbjni-all:1.8
\--- project :platform
     +--- runtimeClasspath
     \--- project :common
          \--- project :crypto
               \--- runtimeClasspath
```

## Why tests didn't catch it

- `:plugins:test` keeps source jars separate; the JVM classloader returns the first match in classpath order, which is `1.18.2` (declared directly by `plugins/build.gradle`). Tests pass.
- The fat-jar `binaryRelease` task merges class entries with last-write-wins (Gradle Zip default `INCLUDE`), so the surviving `Options.class` can be the `1.8` copy. Same classpath, opposite winner.
- ARM64 declares only `:platform` directly (no `1.18.2` fork), so there is no conflict; ARM64 also excludes Archive tests.
- The CI matrix builds the fat jar but never executes it — Toolkit-specific runtime errors are not exercised.

## Fix

1. **`plugins/build.gradle`** — exclude `org.fusesource.leveldbjni:leveldbjni-all` from the `:crypto` dependency, mirroring the existing exclusion on the direct `:platform` dependency. This collapses the x86 runtimeClasspath to a single `leveldbjni-all` (`1.18.2`).
2. **`.github/workflows/pr-build.yml`** — after each platform build (macos / ubuntu-arm / rockylinux-x86 / debian11-x86), run a short smoke test against the freshly built `Toolkit.jar`:
    ```
    java -jar Toolkit.jar help
    java -jar Toolkit.jar db --help
    java -jar Toolkit.jar db archive -h     # directly exercises the failing path
    java -jar Toolkit.jar keystore --help   # exercises the :crypto path added by #6637
    ```
    Each smoke step adds ~10–15s; jobs run in parallel so PR critical path increases by at most ~15s.

## Verification

Reproduced on x86_64 + JDK 8 in `eclipse-temurin:8-jdk` (clean build):

- Before fix: `java -jar Toolkit.jar db archive -d <empty>` → `NoSuchMethodError`
- After fix: `java -jar Toolkit.jar db archive -d <empty>` → exits 0, reports "directory does not contain any database"
- `:plugins:dependencies --configuration runtimeClasspath` no longer lists `org.fusesource.leveldbjni:leveldbjni-all:1.8`
- `./gradlew checkstyleMain checkstyleTest` passes
- `./gradlew :plugins:test` passes (KeystoreUpdate, DbLite, RocksDb tests)

## Test plan

- [ ] CI passes on all four platform builds
- [ ] New "Toolkit jar smoke test" step is green on each platform
- [ ] Reviewer manually confirms `java -jar Toolkit.jar db archive -d <db>` works on x86 with this branch's artifact

## Notes

- No production Java changes; only `plugins/build.gradle` (+6 lines) and `.github/workflows/pr-build.yml` (+36 lines).
- ARM64 path is unaffected: the `:platform` direct dependency on ARM64 has no exclusion (it never had the conflict), and the new `:crypto` exclusion is a no-op there.
- The two `org/iq80/leveldb/Options.class` entries that still exist in the fat jar (from `leveldbjni-all:1.18.2` uber and `leveldb-api:1.18.2` standalone) are pre-existing and harmless because both are the same `1.18.2` version with `maxBatchSize` available.

### DIFF
--- a/plugins/build.gradle
+++ b/plugins/build.gradle
@@ -39,12 +39,17 @@ dependencies {
         exclude group: 'io.prometheus'
         exclude group: 'org.aspectj'
         exclude group: 'org.apache.httpcomponents'
-        // :crypto -> :common -> :platform transitively pulls leveldbjni-all 1.8,
-        // which conflicts with the io.github.tronprotocol leveldbjni-all 1.18.2
-        // declared below on x86. Both jars contain org/iq80/leveldb/Options.class;
-        // in the fat jar the duplicate-entry write order can leave the 1.8 copy,
-        // which lacks Options.maxBatchSize(int) and breaks `db archive` at runtime.
+        // :crypto -> :common -> :platform transitively pulls in artifacts that
+        // the direct :platform dependency below already excludes on x86. Mirror
+        // those excludes here so both paths to :platform agree, otherwise the
+        // transitive copies leak into the fat jar. In particular leveldbjni-all
+        // 1.8 conflicts with io.github.tronprotocol leveldbjni-all 1.18.2: both
+        // jars carry org/iq80/leveldb/Options.class; the duplicate-entry write
+        // order in the fat jar can leave the 1.8 copy, which lacks
+        // Options.maxBatchSize(int) and breaks `db archive` at runtime.
         exclude group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all'
+        exclude group: 'io.github.tronprotocol', module: 'zksnark-java-sdk'
+        exclude group: 'commons-io', module: 'commons-io'
     }
     implementation group: 'info.picocli', name: 'picocli', version: '4.6.3'
     implementation group: 'com.typesafe', name: 'config', version: '1.3.2'
--- a/plugins/build.gradle
+++ b/plugins/build.gradle
@@ -39,6 +39,12 @@ dependencies {
         exclude group: 'io.prometheus'
         exclude group: 'org.aspectj'
         exclude group: 'org.apache.httpcomponents'
+        // :crypto -> :common -> :platform transitively pulls leveldbjni-all 1.8,
+        // which conflicts with the io.github.tronprotocol leveldbjni-all 1.18.2
+        // declared below on x86. Both jars contain org/iq80/leveldb/Options.class;
+        // in the fat jar the duplicate-entry write order can leave the 1.8 copy,
+        // which lacks Options.maxBatchSize(int) and breaks `db archive` at runtime.
+        exclude group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all'
     }
     implementation group: 'info.picocli', name: 'picocli', version: '4.6.3'
     implementation group: 'com.typesafe', name: 'config', version: '1.3.2'
--- a/plugins/build.gradle
+++ b/plugins/build.gradle
@@ -141,7 +141,13 @@ def binaryRelease(taskName, jarName, mainClass) {
         from(sourceSets.main.output) {
             include "/**"
         }
-        dependsOn (project(':protocol').jar, project(':platform').jar) // explicit_dependency
+        // Fat jar zips up runtimeClasspath, which includes the jar outputs of
+        // every project dependency. Declare them all explicitly so Gradle does
+        // not warn about implicit_dependency and disable execution optimizations
+        // (and so partial / parallel builds cannot run binaryRelease before the
+        // dependency jars exist).
+        dependsOn (project(':protocol').jar, project(':platform').jar,
+                project(':crypto').jar, project(':common').jar) // explicit_dependency
         from {
             configurations.runtimeClasspath.collect { // https://docs.gradle.org/current/userguide/upgrading_version_6.html#changes_6.3
                 it.isDirectory() ? it : zipTree(it)
--- a/plugins/build.gradle
+++ b/plugins/build.gradle
@@ -39,20 +39,22 @@ dependencies {
         exclude group: 'io.prometheus'
         exclude group: 'org.aspectj'
         exclude group: 'org.apache.httpcomponents'
-        // :crypto -> :common -> :platform transitively pulls leveldbjni-all
-        // 1.8, which conflicts with the io.github.tronprotocol fork
-        // leveldbjni-all 1.18.2 declared below on x86. Both jars carry
-        // org/iq80/leveldb/Options.class; the duplicate-entry write order in
-        // the fat jar can leave the 1.8 copy, which lacks
+        // x86 declares io.github.tronprotocol:leveldbjni-all:1.18.2 below.
+        // :crypto -> :common -> :platform also transitively pulls
+        // org.fusesource.leveldbjni:leveldbjni-all:1.8; both jars carry
+        // org/iq80/leveldb/Options.class and the duplicate-entry write order
+        // in the fat jar can leave the 1.8 copy, which lacks
         // Options.maxBatchSize(int) and breaks `db archive` at runtime.
+        // Drop the 1.8 transitive on x86 only; ARM64 has a single copy via
+        // its direct :platform declaration and no conflict.
         //
-        // The direct :platform dependency below also excludes
-        // zksnark-java-sdk and commons-io, but those are *not* mirrored here:
-        // the plugins test runtime (DbLiteTest -> Spring -> ZksnarkInitService)
-        // actually loads classes from both, so :crypto must keep transitively
-        // providing them. The :platform-side excludes are deduplication, not
-        // a "kept out of plugins" intent.
-        exclude group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all'
+        // zksnark-java-sdk and commons-io are deliberately NOT mirrored from
+        // the :platform-direct excludes below: the plugins test runtime
+        // (DbLiteTest -> Spring -> ZksnarkInitService) loads classes from
+        // both at runtime, so :crypto must keep transitively providing them.
+        if (!rootProject.archInfo.isArm64) {
+            exclude group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all'
+        }
     }
     implementation group: 'info.picocli', name: 'picocli', version: '4.6.3'
     implementation group: 'com.typesafe', name: 'config', version: '1.3.2'
--- a/plugins/build.gradle
+++ b/plugins/build.gradle
@@ -39,17 +39,20 @@ dependencies {
         exclude group: 'io.prometheus'
         exclude group: 'org.aspectj'
         exclude group: 'org.apache.httpcomponents'
-        // :crypto -> :common -> :platform transitively pulls in artifacts that
-        // the direct :platform dependency below already excludes on x86. Mirror
-        // those excludes here so both paths to :platform agree, otherwise the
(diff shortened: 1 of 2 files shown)


plugins/build.gradle
```diff
@@ -39,6 +39,12 @@ dependencies {
         exclude group: 'io.prometheus'
         exclude group: 'org.aspectj'
         exclude group: 'org.apache.httpcomponents'
+        // :crypto -> :common -> :platform transitively pulls leveldbjni-all 1.8,
+        // which conflicts with the io.github.tronprotocol leveldbjni-all 1.18.2
+        // declared below on x86. Both jars contain org/iq80/leveldb/Options.class;
+        // in the fat jar the duplicate-entry write order can leave the 1.8 copy,
+        // which lacks Options.maxBatchSize(int) and breaks `db archive` at runtime.
+        exclude group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all'
```
halibobo1205: The fix itself is correct, but the excludes aren't aligned.

 Two ways to make this consistent — pick one:
 - **A.** Add the other two excludes on `:crypto` — if `plugins` should NOT bundle `zksnark` / `commons-io`.
 - **B.** Remove the other two excludes on `:platform` — if `plugins` actually NEEDS `zksnark` / `commons-io`.

plugins/build.gradle
halibobo1205: The `binaryRelease` task in `plugins/build.gradle` declares

 ```gradle
 dependsOn (project(':protocol').jar, project(':platform').jar)
 ```

 but its body consumes outputs from `:crypto:jar` and `:common:jar` via `runtimeClasspath`. Gradle prints an `implicit_dependency` warning and disables execution optimizations to compensate; under parallel / incremental / partial-build scenarios this can lead to the fat jar being assembled before `:crypto:jar` or `:common:jar` is up-to-date, producing broken artifacts or build failures. Future Gradle versions will likely promote this to a hard error.

 Recommend adding `project(':crypto').jar` and `project(':common').jar` to the `dependsOn` list.

The log
```log
> Task :plugins:buildToolkitJar
Execution optimizations have been disabled for task ':plugins:buildToolkitJar' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/boson/IdeaProjects/java-tron/crypto/build/libs/crypto-1.0.0.jar'. Reason: Task ':plugins:buildToolkitJar' uses this output of task ':crypto:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.6.4/userguide/validation_problems.html#implicit_dependency for more details about this problem.
  - Gradle detected a problem with the following location: '/Users/boson/IdeaProjects/java-tron/common/build/libs/common-1.0.0.jar'. Reason: Task ':plugins:buildToolkitJar' uses this output of task ':common:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.6.4/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

plugins/build.gradle
barbatos2011: Good catch — verified via `:plugins:dependencyInsight` that both `zksnark-java-sdk` and `commons-io` were indeed leaking through `:crypto -> :common -> :platform`. Went with **option A** to preserve the existing `:platform` exclusion intent. Pushed `69d7f2156` aligning all three excludes on `:crypto`. Re-verified on x86_64 + JDK 8: runtimeClasspath no longer contains any of the three, smoke still passes.

plugins/build.gradle
barbatos2011: Confirmed locally — `./gradlew clean && ./gradlew :plugins:buildToolkitJar` immediately fails with `Cannot expand ZIP '.../crypto/build/libs/crypto-1.0.0.jar'`, exactly as you predicted. Pushed `0aebaffb3` adding `:crypto:jar` and `:common:jar` to the `dependsOn` list. After the fix the same command builds clean with `--warning-mode all` and no `implicit_dependency` warning. Thanks for catching this.

plugins/build.gradle
```diff
@@ -39,12 +39,17 @@ dependencies {
...
-        // which lacks Options.maxBatchSize(int) and breaks `db archive` at runtime.
+        // :crypto -> :common -> :platform transitively pulls in artifacts that
+        // the direct :platform dependency below already excludes on x86. Mirror
+        // those excludes here so both paths to :platform agree, otherwise the
+        // transitive copies leak into the fat jar. In particular leveldbjni-all
+        // 1.8 conflicts with io.github.tronprotocol leveldbjni-all 1.18.2: both
+        // jars carry org/iq80/leveldb/Options.class; the duplicate-entry write
+        // order in the fat jar can leave the 1.8 copy, which lacks
+        // Options.maxBatchSize(int) and breaks `db archive` at runtime.
         exclude group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all'
```
halibobo1205: Maybe exclude just for x86_64? 
**pre :platform** 
```gradle
    if (rootProject.archInfo.isArm64) {
        testRuntimeOnly group: 'org.fusesource.hawtjni', name: 'hawtjni-runtime', version: '1.18' // for test
        implementation project(":platform")
    } else {
        implementation project(":platform"), {
            exclude(group: 'org.fusesource.leveldbjni', module: 'leveldbjni-all')
            exclude(group: 'io.github.tronprotocol', module: 'zksnark-java-sdk')
            exclude(group: 'commons-io', module: 'commons-io')
        }
        implementation 'io.github.tronprotocol:leveldbjni-all:1.18.2'
        implementation 'io.github.tronprotocol:leveldb:1.18.2'
    }
```

plugins/build.gradle
barbatos2011: Update: option A worked for `leveldbjni-all` but **broke `:plugins:test` on x86_64** (rockylinux CI just failed). `DbLiteTest` boots a Spring context that loads `ZksnarkInitService`, which actually pulls in classes from both `commons-io` (`FileUtils`) and `zksnark-java-sdk` (`LibrustzcashWrapper`) at runtime. Once both `:platform` and `:crypto` excluded those two, x86 `testRuntimeClasspath` lost them entirely (ARM64 was fine because the ARM branch declares `:platform` without excludes).

So the `:platform`-side excludes for `zksnark` and `commons-io` are deduplication, not a 'kept out of plugins' intent — the artifacts must keep arriving via `:crypto -> :common -> :platform`. Reverted those two in `320da62c3` and added a comment explaining why the asymmetry is intentional. `leveldbjni-all` remains mirrored because that one is the actual classpath conflict. `:plugins:test` passes again on x86 docker and smoke is still green.

plugins/build.gradle
barbatos2011: Good call — applied in `26f463b72`. Wrapped the `leveldbjni-all` exclude on `:crypto` in `if (!rootProject.archInfo.isArm64)` to mirror the existing arch split on the `:platform`-direct declaration. Behavior is preserved (ARM64 still resolves `leveldbjni-all 1.8` via `:platform` direct; x86 still resolves only the `io.github.tronprotocol` 1.18.2 fork) and the comment now reads as 'this fix is x86-specific' rather than a generic policy. Verified `:plugins:test` and the Toolkit smoke on both arches.

plugins/build.gradle
halibobo1205: remove those from `project(":platform")` to keep the same for project(":crypto")?  It's looking a bit messy right now.
Additionally, since the test classpath loading order is the opposite of the fat jar packaging order, CI fails to catch certain errors — will the loading order be fixed in a follow-up?

barbatos2011: Done in `ee3b8686e` — `zksnark-java-sdk` and `commons-io` are removed from the `:platform` direct dep too, so both project-dep blocks now exclude only `leveldbjni-all` on x86. The blocks are mirror-symmetric and the previous asymmetry note in the `:crypto` comment is dropped. Verified `:plugins:test` and the Toolkit smoke on both arches.

On the loading-order question: yes, planned as a follow-up. The cleanest fix is migrating `binaryRelease` to the **Gradle Shadow plugin** (`com.github.johnrengelman.shadow`), which provides explicit `DuplicatesStrategy`, `mergeServiceFiles()` for `META-INF/services/*` (so we don't have to dance around grpc `LoadBalancerProvider` etc.), and proper relocation if needed. That's a non-trivial build-system change with its own surface area (Shadow integrates with the `application` plugin in a specific way and changes how start scripts/dist artifacts are built), so I'd rather land it as a focused PR after this fix merges. The CI smoke step added in this PR already catches the same shape of bug end-to-end, so we have a safety net in the meantime.

halibobo1205 (approved):

Sunny6889 (approved):
[NIT]
The 7-line smoke test block is duplicated verbatim across all four platform jobs (macos / ubuntu-arm / rockylinux / debian11). Consider extracting it into a composite action at `.github/actions/toolkit-smoke-test/action.yml`, then replacing each block with a one-liner `uses: ./.github/actions/toolkit-smoke-test`. That way the command list lives in one place — adding a new check later (e.g. `db lite -h`) is a single-file change instead of four.

Sketch:
```yaml
# .github/actions/toolkit-smoke-test/action.yml
name: 'Toolkit jar smoke test'
description: 'Run a quick CLI smoke test against the freshly built Toolkit.jar'
inputs:
  jar:
    description: 'Path to Toolkit.jar'
    required: false
    default: 'plugins/build/libs/Toolkit.jar'
runs:
  using: 'composite'
  steps:
    - shell: bash
      run: |
        set -e
        JAR="${{ inputs.jar }}"
        java -jar "$JAR" help
        java -jar "$JAR" db --help
        java -jar "$JAR" db archive -h
        java -jar "$JAR" keystore --help
```
Not a blocker — happy to see it land if works smoothly.

PR #6755, opened 2026-05-09
fix(config): optimizes config binding for node 

317787106:
**What does this PR do?**

1. bug Fixes

  - treat "null" as unset for `discovery.external.ip`; bind via a flat externalIP.
  - fix the proprity of some config items, do not default `allowShieldedTransactionApi` in reference.conf; use it when set, else fall back to `fullNodeAllowShieldedTransaction` with a deprecation warning.

2. migration
  - remove unused node bindings: `node.connection.timeout`, `node.channel.read.timeout`, `node.tcpNettyWorkThreadNum`, `node.udpNettyWorkThreadNum`
  
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/common/src/main/java/org/tron/core/config/args/NodeConfig.java
+++ b/common/src/main/java/org/tron/core/config/args/NodeConfig.java
@@ -5,7 +5,6 @@ import static org.tron.core.exception.TronError.ErrCode.PARAMETER_INIT;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigBeanFactory;
-import com.typesafe.config.ConfigFactory;
 import com.typesafe.config.ConfigValueFactory;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


common/src/main/java/org/tron/core/config/args/NodeConfig.java
```diff
@@ -403,14 +377,23 @@ public static NodeConfig fromConfig(Config config) {
       nc.maxConnectionsWithSameIp = section.getInt("maxActiveNodesWithSameIp");
     }
 
+    nc.externalIP = getString(section, "discovery.external.ip", "");
```
halibobo1205, line 380: [SHOULD] Please add the necessary unit tests for this fix(1 and 2). Ignore this if they have already been added.

317787106: There are testcases in `NodeConfigTest` already.

halibobo1205 (approved):

kuny0707 (approved):

PR #6756, opened 2026-05-10
fix(config): restore --es behavior when event.subscribe.enable=false

halibobo1205:
**What does this PR do?**

The previous flow built EventPluginConfig/FilterQuery only when the config bean had enable=true, so passing the deprecated `--es` CLI flag against a config with enable=false set eventSubscribe=true but left the plugin config null, causing "Failed to load eventPlugin" at startup.
Split applyConfigParams into atomic-bind (eventSubscribe, contractParseSwitch) and buildEventDerivatives (EPC + FilterQuery). The derivative build now runs once at the end of setParam, after CLI overrides have settled, so --es produces a valid EPC and --contract-parse-enable is not clobbered.

    restore` --es` behavior when `event.subscribe.enable`=false
- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -178,6 +178,14 @@ public class Args extends CommonParameter {
 
     // 5. Init witness (depends on CLI witness flag)
     initLocalWitnesses(config, cmd);
+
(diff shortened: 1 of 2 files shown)


halibobo1205:
use #6757

PR #6760, opened 2026-05-11
fix(event): reject incompatible event-plugin versions below 3.0.0

halibobo1205:
**What does this PR do?**

Enforce a minimum `Plugin-Version` at startup in `EventPluginLoader.startPlugin` using `pf4j's VersionManager` (semver). When the descriptor version is below `3.0.0`, log a clear upgrade hint and return false; the existing call chain in `Manager.startEventSubscribing` wraps it in a `TronError(EVENT_SUBSCRIBE_INIT)` and aborts node startup instead of silently degrading.

Pre-3.0.0 event-plugin builds still link against `com.alibaba.fastjson`, which was removed from java-tron in #6701. When such a plugin is loaded, the `NoClassDefFoundError` surfaces inside the plugin's own worker thread and is swallowed by its catch(Throwable) handler. The node keeps running while silently dropping every trigger, leaving operators without any indication that the event subscription is broken.

- Unit Tests: Add a unit test covering accept/reject boundaries and empty/null versions.
- Manual Testing

**Follow up**
   - https://github.com/tronprotocol/event-plugin/pull/54
**Extra details**
```log
09:06:03.792 ERROR [main] [o.t.c.l.EventPluginLoader](EventPluginLoader.java:499) event-plugin 'kafka' version 1.0.0 is older than required 3.0.0, please upgrade event-plugin
09:07:41.179 ERROR [main] [o.t.c.l.EventPluginLoader](EventPluginLoader.java:499) event-plugin 'mongodb' version 1.0.0 is older than required 3.0.0, please upgrade event-plugin
```

### DIFF
--- a/framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
+++ b/framework/src/main/java/org/tron/common/logsfilter/EventPluginLoader.java
@@ -3,6 +3,7 @@ package org.tron.common.logsfilter;
 import com.beust.jcommander.internal.Sets;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Strings;
 import java.io.File;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


lvs0075 (approved):

317787106 (approved):

waynercheung (approved):
LGTM

lvs0075 (approved):

PR #6762, opened 2026-05-13
refactor(config,protocol,ci): optimize config binding

317787106:
**What does this PR do?**

Cleans up config bean binding across `CommitteeConfig`, `NodeConfig`, `EventConfig`, `StorageConfig`, and `VmConfig`, removes the dead `Overlay` class, drops the unused `google/api/annotations.proto` import, and deletes the stale `codecov.yml`.

1. **CommitteeConfig** — `allowPBFT` / `pBFTExpireNum` use non-standard casing that JavaBean Introspector cannot derive correctly. Previously these were excluded from `ConfigBeanFactory` and read field-by-field manually. Now a `normalizeNonStandardKeys()` helper remaps them to standard camelCase (`allowPbft` / `pbftExpireNum`) before binding, so `ConfigBeanFactory` handles them automatically. The manual getter overrides (`getAllowPBFT`, `getPBFTExpireNum`) are replaced by standard Lombok-generated getters.

2. **NodeConfig** — Three improvements:
   - `isOpenFullTcpDisconnect`: the `is`-prefix in the config key conflicts with JavaBean boolean-getter naming. The field is renamed to `openFullTcpDisconnect` (Lombok auto-generates the `isOpenFullTcpDisconnect()` getter) and `normalizeNonStandardKeys()` remaps the legacy config key before binding.
   - `discovery.external.ip`: previously a standalone `externalIP` field on `NodeConfig` read manually after binding. Now it is modelled as `DiscoveryConfig.ExternalConfig.ip` and auto-bound by `ConfigBeanFactory`. `normalizeNonStandardKeys()` normalises both HOCON `null` and the string sentinel `"null"` to `""` before binding.
   - `nativeQueue` / `topics` in **EventConfig**: removed redundant `@Getter(NONE)` + hand-written getters; `@Setter(NONE)` is kept so `ConfigBeanFactory` skips the keys that require special handling, which are still read explicitly in `fromConfig()`.

3. **EventConfig** — Replaced a 30-line field-by-field manual binding loop for `nativeQueue` and `topics` with `ConfigBeanFactory.create()` calls. A `TOPIC_DEFAULTS` fallback config fills optional `TopicConfig` fields (`solidified`, `ethCompatible`, `redundancy`) so each topic item can be bound uniformly.

4. **StorageConfig** — Removed `@Getter(NONE)` + hand-written getters for `rawStorageConfig`, `defaultDbOption`, `defaultMDbOption`, and `defaultLDbOption`; standard Lombok getters now apply. `@Setter(NONE)` is retained on all four to prevent `ConfigBeanFactory` from requiring their keys.

5. **VmConfig** — `constantCallTimeoutMs` carried `@Setter(AccessLevel.NONE)` to block `ConfigBeanFactory` from touching it; the value was bound manually in `postProcess()` via `vmSection.hasPath()` + `vmSection.getLong()`, and zero was treated as a misconfiguration. This workaround is removed:
   - `constantCallTimeoutMs = 0` is added to `reference.conf` so the key is always present and `ConfigBeanFactory` binds it automatically.
   - `@Setter(AccessLevel.NONE)` is dropped; the field becomes a plain Lombok-managed field.
   - The manual `hasPath` / `getLong` binding is replaced by a simple range check in `postProcess()`: negative values and values that overflow microsecond conversion are rejected; zero is accepted as the valid default meaning "share the block-processing deadline".
   - The `CONSTANT_CALL_TIMEOUT_MS_KEY` and `MAX_CONSTANT_CALL_TIMEOUT_MS` constants are removed.

6. **`google/api/annotations.proto`** — Removed the unused `import "google/api/annotations.proto"` from `api.proto`. No RPC method in the file carries an `option (google.api.http)` annotation, so the import was dead weight that pulled in an unnecessary dependency.

7. **`codecov.yml`** — Deleted the stale Codecov configuration file. Codecov integration is no longer active; coverage is handled by JaCoCo + `madrapps/jacoco-report` in `pr-build.yml`. The file had already been marked deprecated in a comment but was never removed.

8. **Overlay** — Deleted the unused `Overlay` class and its sole test. The corresponding `overlay` field is removed from `CommonParameter`.

The previous code used `@Getter(lombok.AccessLevel.NONE)` + `@Setter(lombok.AccessLevel.NONE)` + hand-written getter/setter pairs as a workaround for config keys that `ConfigBeanFactory` could not bind automatically (reserved words, non-standard casing, optional keys). This pattern is verbose, error-prone, and scatters binding logic across fields and `fromConfig()` bodies. The `normalizeNonStandardKeys()` pattern and `@Setter(NONE)`-only approach eliminate the boilerplate while preserving the same runtime behaviour.

- Unit Tests

**Follow up**

**Extra details**


protocol/src/main/protos/api/api.proto
```diff
@@ -2,7 +2,6 @@ syntax = "proto3";
 package protocol;
 
 import "core/Tron.proto";
-import "google/api/annotations.proto";
```
halibobo1205, line 5: [MUST] Deleting the import `google/api/annotations.proto` is out of scope for this PR. The change is unrelated to config bean binding — it touches a gRPC service definition, not any of the **Config beans or ConfigBeanFactory wiring**. Please split it into a separate PR (e.g. 
 chore(proto): remove unused google.api.annotations import) So it can be reviewed and reverted independently.

halibobo1205 (approved):

protocol/src/main/protos/api/api.proto
317787106: This is a little change, only including code clear up. Modified the PR title and description to align with the changes.

common/src/main/java/org/tron/core/config/args/CommitteeConfig.java
```diff
@@ -85,32 +67,30 @@ public class CommitteeConfig {
...
 
+  // "allowPBFT" and "pBFTExpireNum" use non-standard casing that JavaBean Introspector
+  // cannot derive correctly (setPBFTExpireNum -> property "PBFTExpireNum", not "pBFTExpireNum").
+  // Remap them to standard camelCase keys so ConfigBeanFactory binds them normally.
+  // Config is immutable; withValue() returns a new object.
+  private static Config normalizeNonStandardKeys(Config section) {
+    if (section.hasPath("allowPBFT")) {
+      ConfigValue v = section.getValue("allowPBFT");
+      section = section.withValue("allowPbft", v); // rename allowPBFT -> allowPbft
+    }
```
bladehan1, line 86: [NIT] normalizeNonStandardKeys cleanliness: withoutPath + hasPath guard

Two small cleanups apply to both `CommitteeConfig.normalizeNonStandardKeys` (this file, allowPBFT/pBFTExpireNum) and the sibling `NodeConfig.normalizeNonStandardKeys` (isOpenFullTcpDisconnect / discovery.external.ip):

1. `section.withValue(newKey, ...)` adds the new key but does not remove the legacy one. After normalize, both keys coexist in the Config tree. This is harmless for `ConfigBeanFactory` (only the canonical name binds), but any downstream diagnostic that walks the raw Config (debug dumps, future read paths) will see two competing entries with no documented winner. Chaining `.withoutPath(oldPath)` after the rename keeps the post-normalize tree canonical.

2. `NodeConfig.normalizeNonStandardKeys` calls `section.getIsNull("discovery.external.ip")`, which throws `ConfigException.Missing` if the path itself is absent (it only returns true when the path exists with a null value). The current code relies on the `reference.conf` default at `external.ip = ""` always populating this path. If a user explicitly writes `node.discovery = null` or `node.discovery.external = null` to disable a sub-tree, normalize blows up with a stack trace pointing into its internals rather than at the offending config path. A `hasPath` guard or an explicit javadoc precondition (`section` must be merged with reference.conf fallback) makes the implicit contract traceable.

Neither is a correctness bug given current `reference.conf` content; treat as nits.

common/src/test/java/org/tron/core/config/args/VmConfigTest.java
```diff
@@ -117,19 +119,6 @@ public void testConstantCallTimeoutAcceptsAnyPositiveValue() {
...
-      VmConfig.fromConfig(withRef("vm { constantCallTimeoutMs = 0 }"));
-      org.junit.Assert.fail("expected IllegalArgumentException for explicit 0");
-    } catch (IllegalArgumentException ex) {
-      org.junit.Assert.assertTrue(ex.getMessage(),
-          ex.getMessage().contains("constantCallTimeoutMs"));
-    }
-  }
-
   @Test
   public void testConstantCallTimeoutNegativeRejected() {
```
bladehan1, line 125: [NIT] Replace try/fail/catch with assertThrows for clarity

`testConstantCallTimeoutNegativeRejected` and `testConstantCallTimeoutOverflowRejected` both use the legacy `try { ... Assert.fail(...); } catch (IllegalArgumentException ex) { ... }` idiom. Two issues with this pattern:

1. If `fromConfig` ever throws a different exception type (e.g. `ConfigException`), the catch block does not match, the test method propagates the exception, and the failure message no longer points at the expected behaviour.
2. The assertion on `ex.getMessage()` runs inside the catch block, so a mismatch reports the catch line rather than the assertion line.

`assertThrows(IllegalArgumentException.class, () -> ...)` plus `assertThat(ex.getMessage(), containsString(...))` is the standard JUnit 4.13+ form and matches `knowledge/junit-exception-idiom.md` in this repo.

Suggestion: switch to `assertThrows` for both test methods; consider the same change for `CommitteeConfigTest.testAllowOldRewardOptWithoutPrerequisites` which uses `@Test(expected = ...)`.

common/src/test/java/org/tron/core/config/args/VmConfigTest.java
```diff
@@ -107,6 +107,8 @@ public void testConstantCallTimeoutDefaultWhenAbsent() {
```
bladehan1, line 109: [NIT] Comment about reference.conf state is stale

The javadoc just above this method (line 102) says "No path in the config, no entry in reference.conf -> default 0L kept", but this PR adds `constantCallTimeoutMs = 0` to reference.conf at line 703. The assertion still passes because the reference.conf default of 0 happens to equal the Java field default, but a future maintainer reading this comment will be misled about where the default comes from.

Suggestion: reword to "reference.conf default is 0; absence of a user override keeps that default; no validation triggered."

framework/src/main/resources/config.conf
```diff
@@ -750,7 +750,7 @@ vm = {
   # Omit the property entirely to keep the default behaviour of sharing the
   # block-processing deadline. Migration note: if previously running --debug
   # to extend constant calls, switch to this option (--debug also extends
-  # block-processing, which is unsafe; see issue #6266).
+  # block-processing, which is unsafe; see issue #6266). Default: 0.
   # constantCallTimeoutMs = 100
```
bladehan1, line 754: [NIT] Document the new semantics of constantCallTimeoutMs = 0 in the sample config

This PR makes `constantCallTimeoutMs = 0` a valid configuration meaning "share the block-processing deadline" — previously the same value was rejected. The sample config still shows the commented hint `# constantCallTimeoutMs = 100` without explaining the special semantics of zero. Operators who previously set this to a positive value as a workaround for the old rejection won't know what `0` now implies after the upgrade.

Suggestion: add an inline comment in config.conf clarifying that `0` means "use the block-processing deadline", so the meaning of the default is discoverable from the sample alone.

bladehan1 (approved):

ouy95917 (approved):
LGTM

yanghang8612 (approved):

common/src/test/java/org/tron/core/config/args/VmConfigTest.java
317787106: Update this comment.

common/src/test/java/org/tron/core/config/args/VmConfigTest.java
317787106: Fix it as Suggestion.

waynercheung (approved):
LGTM

PR #6763, opened 2026-05-14
fix(jsonrpc): make the JSON-RPC output more compliant with specification

317787106:
## What does this PR do?

Fixes three JSON-RPC spec-compliance issues and one parser-hardening gap in `JsonRpcServlet`:

### 1. `Content-Type` response header — remove spurious `; charset=utf-8`

The JSON-RPC 1.0/2.0 spec and Ethereum JSON-RPC both use `application/json` or `application/json-rpc` without a charset suffix. The extra `; charset=utf-8` suffix was non-standard and has been removed.

### 2. Non-object/non-array root node → `Invalid Request`

Sending a JSON primitive as the request body (e.g. `null`, `true`, `123`) was silently forwarded to `jsonrpc4j`, which produced a malformed response with `"id":"null"` (string) instead of `"id":null` (JSON null). Now these cases are caught before dispatch and return a proper `-32600 Invalid Request` response with `"id":null`.

**Before:**
```
POST /jsonrpc  body: null
→ {"jsonrpc":"2.0","id":"null","error":{"code":-32600,"message":"invalid request"}}
```

**After:**
```
POST /jsonrpc  body: null
→ {"jsonrpc":"2.0","id":null,"error":{"code":-32600,"message":"Invalid Request"}}
```

### 3. Non-object element inside a batch → `Invalid Request`

A batch containing a non-object element (e.g. `[[]]`) was forwarded per-element to `jsonrpc4j`, which echoed the inner array back unchanged, producing `[[]]` instead of an error. Non-object sub-requests are now rejected with `-32600 Invalid Request` before dispatch.

**Before:**
```
POST /jsonrpc  body: [[]]
→ [[]]
```

**After (Ethereum-compatible):**
```
POST /jsonrpc  body: [[]]
→ [{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid Request"},"id":null}]
```

### 4. Apply `StreamReadConstraints` to the JSON-RPC parser

`JsonRpcServlet` used a bare `new ObjectMapper()`, so the `node.http.maxNestingDepth` and `node.http.maxTokenCount` settings introduced in #6701 had no effect on JSON-RPC requests. The mapper is now built with a `JsonFactory` that reads those two limits from `CommonParameter`, consistent with how `JSON.java` constructs its mapper.

Without this fix, a ~4 MB request body such as `[0,0,...,0]` (~2 M tokens) could bypass the configured `maxTokenCount` (default 100,000) and force Jackson to allocate ~2 M `JsonNode` objects, causing GC pressure (Eden/Survivor saturation → premature promotion → Full GC). With the fix, the parser throws after the configured token limit is reached, well before the 4 MB body-size cap.

- Issues 1–3: bring responses in line with the JSON-RPC 2.0 specification and Ethereum JSON-RPC behaviour.
- Issue 4: make the `node.http.maxNestingDepth` / `node.http.maxTokenCount` config knobs actually effective for the JSON-RPC endpoint, closing a token-density amplification vector.

## This PR has been tested by

- Unit Tests
- Manual Testing

## Extra details

Issue 4 shares the same root cause identified in the review comment on this PR: the `MAPPER` in `JsonRpcServlet` was not wired to the `StreamReadConstraints` configured via `node.http.*`.

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
@@ -247,6 +247,7 @@ public class JsonRpcServlet extends RateLimiterServlet {
 
     // JSON-RPC 2.0 §6: MUST NOT return an empty Array when there are no response objects.
     if (batchResult.isEmpty()) {
+      resp.setContentType("application/json-rpc");
       resp.setStatus(HttpServletResponse.SC_OK);
       resp.setContentLength(0);
       return;
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
@@ -177,18 +177,28 @@ public class JsonRpcServlet extends RateLimiterServlet {
     for (int i = 0; i < rootNode.size(); i++) {
       JsonNode subRequest = rootNode.get(i);
 
-      if (!subRequest.isObject()) {
-        batchResult.add(buildErrorNode(JsonRpcError.INVALID_REQUEST, "Invalid Request", null));
-        continue;
-      }
-
       if (overflow) {
-        // Notifications (no "id") do not get a response even on overflow.
-        if (subRequest.has("id")) {
+        if (!subRequest.isObject()) {
+          batchResult.add(buildErrorNode(JsonRpcError.INVALID_REQUEST, "Invalid Request", null));
+        } else if (subRequest.has("id")) {
+          // Notifications (no "id") do not get a response even on overflow.
           batchResult.add(buildErrorNode(JsonRpcError.RESPONSE_TOO_LARGE,
               "Response exceeds the limit of " + maxResponseSize + " bytes",
               subRequest.get("id")));
         }
+        continue;
+      }
+
+      if (!subRequest.isObject()) {
+        ObjectNode errNode = buildErrorNode(JsonRpcError.INVALID_REQUEST, "Invalid Request", null);
+        byte[] errBytes = MAPPER.writeValueAsBytes(errNode);
+        int addition = errBytes.length + (!batchResult.isEmpty() ? 1 : 0);
+        if (maxResponseSize > 0 && accumulatedSize + addition > maxResponseSize) {
+          overflow = true;
+        } else {
+          accumulatedSize += addition;
+        }
+        batchResult.add(errNode);
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


0xbigapple (approved):
LGTM

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
waynercheung, line 240: [MUST] `MAPPER` is constrained by `node.http.maxTokenCount/maxNestingDepth`, but it is also used to parse server-generated batch sub-responses. A valid response below `node.jsonrpc.maxResponseSize` can exceed `maxTokenCount` and be converted into `-32603 Internal error`.

Please use the constrained mapper only for inbound request parsing, and use an unconstrained/internal mapper for response assembly and `readTree(responseBytes)`. Add a regression test where a batch sub-response is below `maxResponseSize` but above `maxTokenCount`.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
```diff
@@ -97,11 +111,16 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
     try {
       rootNode = MAPPER.readTree(body);
       if (rootNode == null || rootNode.isMissingNode()) {
-        writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "Parse error", null, false);
+        writeJsonRpcError(resp, JsonRpcError.PARSE_ERROR, "JSON parse error", null, false);
```
waynercheung, line 114: [MUST] This changes the observable JSON-RPC error response, not just code style. JSON-RPC 2.0 defines the standard message for code -32700 as `Parse error`, and the PR is explicitly moving responses toward spec compliance. Please keep the previous `Parse error` message here.

bladehan1 (approved):
LGTM

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: Thank you for your suggestion. The default maxTokenCount(100_000)/maxNestingDepth(100) is enough for normal request. So using the same MAPPER for inbound or batch is enough.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcServlet.java
317787106: This keeps the same as v4.8.1, it outputs `JSON parse error`.

PR #6775, opened 2026-05-18
fix(crypto): bind burn cipher nonce to nullifier

Federico2014:
## What does this PR do?

Hardens shielded TRC-20 burn encryption by binding the burn AEAD nonce to the spend nullifier, the burn amount, and the transparent recipient address; storing the nonce in the burn record; making the burn-record version explicit in the trailing reserved bytes; and tightening input validation on the burn-cipher setter exposed by `ShieldedTRC20ParametersBuilder`.

The original implementation encrypted every burn payload under the same long-lived `ovk` with a static all-zero ChaCha20-Poly1305 nonce. Reusing `(key, nonce)` pairs breaks the AEAD security assumptions and leaks keystream relationships across burns.

This PR gives each burn a nonce derived from `(nf, amount, addr)` — all three of which are public on chain — and records enough metadata on chain for scanners to distinguish legacy and upgraded burn records without relying on ambiguous zero-nonce heuristics. Binding amount and address in addition to the nullifier prevents a single off-chain `nf` from producing multiple `(amount, addr)` variants that would reuse the same AEAD nonce before one of them lands on chain.

### Nonce derivation with domain separation

```text
nonce = SHA3("Ztron_BurnNonce" || nf || amount32 || addr21)[:12]
```

- `nf` is the 32-byte spend nullifier.
- `amount32` is the 32-byte big-endian burn amount that the on-chain `TokenBurn` event logs.
- `addr21` is the 21-byte `prefix || transparent recipient (20B)` that the same event logs.

All three inputs are visible on chain in the burn log, so any scanner can recompute the nonce from public state. The fixed domain tag prevents collisions with any other SHA3-of-nullifier use. Since `nf` is unique per spend, and `(amount, addr)` further partitions any off-chain re-encryption of the same input note, the resulting `(ovk, nonce)` pair is unique per burn payload.

### 96-byte burn record

`encryptBurnMessageByOvk` returns a 96-byte record that occupies the same calldata / log slot as the previous `cipher(80) || zero-padding(16)` layout:

```text
record (96B) = cipher(80) || nonce(12) || reserved/version(4)
```

Current version marker:

```text
v2 marker = 0x00000001
```

### Explicit legacy/v2 discrimination for scanners

`decryptBurnMessageByOvk(ovk, cipher, nonceFromLog, reservedFromLog, nf, amount32, addr21)` now treats the trailing 4 reserved bytes as an explicit version marker:

- `reserved = 0x00000000` and `nonce = 0x000...000` -> legacy v1 path
- `reserved = 0x00000001` -> v2 path; the scanner must verify `nonce == SHA3(domain || nf || amount32 || addr21)[:12]`
- any other `reserved` value -> reject

This removes the previous ambiguity where an all-zero nonce could be interpreted both as a valid new record and as the legacy sentinel.

### Defensive input validation on the builder

`ShieldedTRC20ParametersBuilder.setBurnCiphertext(...)` is no longer the Lombok-generated setter. The hand-written replacement rejects `null` and any length other than `BURN_CIPHER_RECORD_SIZE = 96`, and stores a defensive copy. This closes the only external surface through which a malformed-length burn ciphertext could reach `burnParamsToHexString` and produce silently corrupted calldata.

### Encryption moved into the builder

BURN encryption now runs inside `ShieldedTRC20ParametersBuilder.build()` after `generateSpendProof(...)`, once the nullifier is available. Callers only need to provide `ovk`; if omitted, the builder falls back to `spend.expsk.getOvk()`.

## Compatibility

| Dimension | Impact |
|---|---|
| Consensus / hard fork | None — wallet / scanner / trigger-input only, no protocol-state transition |
| On-chain burn log layout | Still 160B total (`toAddress 32 + value 32 + burn record 96`) |
| New burn record semantics | `cipher(80) + nonce(12) + reserved/version(4)` replaces `cipher(80) + zero padding(16)` for newly generated burns |
| Legacy burns already on chain | Still scannable via the explicit legacy branch (`reserved=0 && nonce=0`) |
| `getTriggerInputForShieldedTRC20Contract` | **Breaking change for legacy callers**: 80-byte burn ciphers are now deprecated and rejected; callers must send the new 96-byte burn record |
| `setBurnCiphertext` | **Behavior change**: now throws `IllegalArgumentException` for null or non-96-byte input, where the previous Lombok setter accepted any byte array |
| gRPC / HTTP / JSON-RPC shape | Field names are unchanged, but acceptable burn payload semantics are narrower because legacy 80-byte burn ciphers are rejected |
| Config / DB schema | None |
| Third-party SDKs | SDKs that still build or submit 80-byte burn ciphers must be updated to emit the 96-byte burn record |
| Old java-tron / old SDK scanners | One-way compatibility: the new scanner reads both legacy and v2 records, but old scanners interpret a v2 record as `cipher(80) + zero-padding(16)` and attempt zero-nonce decryption, which fails. Scanners that need to read v2 burns must be upgraded in lockstep |

## Upgrade guidance

The deployed shielded TRC-20 contract accepts an opaque `bytes32[3]` burn payload and does not interpret its layout, so this PR cannot reject v1 calldata on-chain. The v1→v2 migration is therefore a coordinated wallet/SDK/scanner upgrade. To get the privacy guarantees of `(nf, amount, addr)`-bound nonces, all of the following must be upgraded:

1. **Node operators** — upgrade java-tron to the release that includes this PR. Old nodes still serve `getTriggerInputForShieldedTRC20Contract` with v1 semantics and old scanners cannot read v2 burn records (they interpret them as `cipher(80) + zero-padding(16)` and fail decryption).
2. **Wallet / SDK developers who hand-build burn calldata** — emit the 96-byte v2 record: `cipher(80) + nonce(12) + reserved=0x00000001(4)`, where `nonce = SHA3("Ztron_BurnNonce" || nf || amount32 || addr21)[:12]`. The on-chain contract does not validate this layout, so 80-byte calldata continues to confirm, but any burn assembled that way loses nf/amount/addr-binding and falls back to the legacy zero-nonce decryption path with the known `(ovk, nonce=0)` reuse weakness.
3. **End users** — until their wallet/SDK is upgraded, burns generated through that wallet retain the original v1 weakness (static `(ovk, nonce=0)` reused across every burn under the same `ovk`). Privacy improves only after the wallet itself emits v2 records.

Fully disallowing v1 calldata on-chain is an application-layer change: it requires redeploying the shielded TRC-20 contract with explicit layout validation (e.g. reject burn payloads whose trailing 16 bytes are all zero). That is out of scope for this PR.

## Key changes

| File | Change |
|---|---|
| `NoteEncryption.java` | Adds `BURN_CIPHER_RECORD_SIZE=96`, explicit reserved/version marker handling, nonce derivation from `(nf, amount, addr)`, and strict v1/v2 burn-record decryption rules |
| `ShieldedTRC20ParametersBuilder.java` | Moves BURN encryption into the builder after spend proof generation so the nullifier is available; emits the 96-byte burn record; replaces the Lombok `@Setter` on `burnCiphertext` with a hand-written setter that validates non-null + 96-byte length and stores a defensive copy |
| `Wallet.java` | Parses burn logs as `cipher(80)+nonce(12)+reserved/version(4)`; scanner now uses explicit version markers and verifies the v2 nonce against `(nf, amount, addr)` reconstructed from the log; 80-byte trigger-input burn ciphers throw a clear deprecation/rejection error |
| `BurnCipherTest.java` | Covers 96-byte record structure, reserved/version marker behavior, v2 `(nf, amount, addr)`-required decryption, wrong-amount / wrong-addr rejection, unknown marker rejection, and malformed input handling |
| `NoteEncDecryTest.java` | Covers legacy zero-nonce compatibility, 96-byte trigger-input acceptance, 80-byte trigger-input rejection, burn-log round trip, burn-log truncation, missing-nf rejection, and same-tx cursor pairing |

## Testing

- `./gradlew framework:test --tests "org.tron.core.zen.note.BurnCipherTest"`
- `./gradlew framework:test --tests "org.tron.core.zksnark.NoteEncDecryTest"`
- `./gradlew framework:test --tests org.tron.core.zksnark.NoteEncDecryTest.testGetTriggerInputBurn80ByteCipherRejected`

### DIFF
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -3909,7 +3909,9 @@ public class Wallet {
             int index = 0;
             for (TransactionInfo.Log log : logList) {
               int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress);
-              if (logType > 0) {
+              // Only note-producing log types (1..3) advance the note index;
+              // TokenBurn (4) and NoteSpent (5) do not emit a leaf.
+              if (logType > 0 && logType < 4) {
                 noteBuilder = DecryptNotesTRC20.NoteTx.newBuilder();
                 noteBuilder.setTxid(ByteString.copyFrom(txId));
                 noteBuilder.setIndex(index);
@@ -4307,17 +4309,45 @@ public class Wallet {
         parameterType);
     if (parametersBuilder.getShieldedTRC20ParametersType() == ShieldedTRC20ParametersType.BURN) {
       byte[] burnCiper = ByteArray.fromHexString(shieldedTRC20Parameters.getTriggerContractInput());
-      if (!ArrayUtils.isEmpty(burnCiper)
-          && burnCiper.length == NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE) {
-        parametersBuilder.setBurnCiphertext(burnCiper);
-      } else if (!ArrayUtils.isEmpty(burnCiper) && burnCiper.length == 80) {
-        throw new ZksnarkException(
-            "legacy 80-byte burn cipher is deprecated and rejected; expected "
-                + NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE + "-byte burn record");
-      } else {
+      if (ArrayUtils.isEmpty(burnCiper)
+          || burnCiper.length != NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE) {
+        if (!ArrayUtils.isEmpty(burnCiper) && burnCiper.length == 80) {
+          throw new ZksnarkException(
+              "legacy 80-byte burn cipher is deprecated and rejected; expected "
+                  + NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE + "-byte burn record");
+        }
         throw new ZksnarkException(
             "invalid shielded TRC-20 contract parameters for burn trigger input");
       }
+      // v2-only: length alone would accept a legacy all-zero suffix and bypass
+      // the nf-bound nonce. Require reserved==v2 marker and nonce==derive(nf).
+      byte[] reserved = Arrays.copyOfRange(burnCiper,
+          NoteEncryption.Encryption.BURN_RESERVED_OFFSET,
+          NoteEncryption.Encryption.BURN_RESERVED_OFFSET
+              + NoteEncryption.Encryption.BURN_RESERVED_LEN);
+      if (!Arrays.equals(reserved, NoteEncryption.Encryption.getBurnRecordV2Marker())) {
+        throw new ZksnarkException(
+            "burn trigger input must be v2 (reserved=0x00000001); legacy/unknown markers rejected");
+      }
+      if (shieldedTRC20Parameters.getSpendDescriptionList().size() != 1) {
+        throw new ZksnarkException(
+            "burn trigger input requires exactly one spendDescription for nf-bound nonce");
+      }
+      byte[] nf = shieldedTRC20Parameters.getSpendDescription(0).getNullifier().toByteArray();
+      if (nf.length != 32) {
+        throw new ZksnarkException(
+            "burn trigger input requires 32-byte spendDescription.nullifier");
+      }
+      byte[] nonceFromInput = Arrays.copyOfRange(burnCiper,
+          NoteEncryption.Encryption.BURN_NONCE_OFFSET,
+          NoteEncryption.Encryption.BURN_NONCE_OFFSET
+              + NoteEncryption.Encryption.BURN_NONCE_LEN);
+      byte[] expectedNonce = NoteEncryption.Encryption.deriveNonceFromNf(nf);
+      if (!Arrays.equals(nonceFromInput, expectedNonce)) {
+        throw new ZksnarkException(
+            "burn trigger input nonce does not match nf-bound nonce");
+      }
+      parametersBuilder.setBurnCiphertext(burnCiper);
     }
     String input = parametersBuilder
         .getTriggerContractInput(shieldedTRC20Parameters, spendAuthoritySignature, value, false,
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -270,6 +270,8 @@ public class Wallet {
       "BurnNewLeaf(uint256,bytes32,bytes32,bytes32,bytes32[21])"));
   private static final byte[] SHIELDED_TRC20_LOG_TOPICS_BURN_TOKEN = Hash.sha3(ByteArray
       .fromString("TokenBurn(address,uint256,bytes32[3])"));
+  private static final byte[] SHIELDED_TRC20_LOG_TOPICS_NOTE_SPENT = Hash.sha3(ByteArray
+      .fromString("NoteSpent(bytes32)"));
   private static final String BROADCAST_TRANS_FAILED = "Broadcast transaction {} failed, {}.";
 
   @Getter
@@ -3672,9 +3674,7 @@ public class Wallet {
         builder.setTransparentToAddress(transparentToAddressTvm);
         builder.setTransparentToAmount(toAmount);
 
-        Optional<byte[]> cipher = NoteEncryption.Encryption
-            .encryptBurnMessageByOvk(ovk, toAmount, transparentToAddress);
-        cipher.ifPresent(builder::setBurnCiphertext);
+        builder.setOvk(ovk);
 
         ExpandedSpendingKey expsk = new ExpandedSpendingKey(ask, nsk, ovk);
         GrpcAPI.SpendNoteTRC20 spendNote = shieldedSpends.get(0);
@@ -3799,9 +3799,7 @@ public class Wallet {
         System.arraycopy(transparentToAddress, 1, transparentToAddressTvm, 0, 20);
         builder.setTransparentToAddress(transparentToAddressTvm);
         builder.setTransparentToAmount(toAmount);
-        Optional<byte[]> cipher = NoteEncryption.Encryption
-            .encryptBurnMessageByOvk(ovk, toAmount, transparentToAddress);
-        cipher.ifPresent(builder::setBurnCiphertext);
+        builder.setOvk(ovk);
         GrpcAPI.SpendNoteTRC20 spendNote = shieldedSpends.get(0);
         buildShieldedTRC20InputWithAK(builder, spendNote, ak, nsk);
         if (receiveSize == 1) {
@@ -3838,6 +3836,8 @@ public class Wallet {
         return 3;
       } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_BURN_TOKEN)) {
         return 4;
+      } else if (Arrays.equals(topicsBytes, SHIELDED_TRC20_LOG_TOPICS_NOTE_SPENT)) {
+        return 5;
       }
     }
     return 0;
@@ -4001,7 +4001,8 @@ public class Wallet {
 
   private Optional<DecryptNotesTRC20.NoteTx> getNoteTxFromLogListByOvk(
       DecryptNotesTRC20.NoteTx.Builder builder,
-      TransactionInfo.Log log, byte[] ovk, int logType) throws ZksnarkException {
+      TransactionInfo.Log log, byte[] ovk, int logType, byte[] pendingNf)
+      throws ZksnarkException {
     byte[] logData = log.getData().toByteArray();
     if (!ArrayUtils.isEmpty(logData)) {
       if (logType > 0 && logType < 4) {
@@ -4040,18 +4041,32 @@ public class Wallet {
           }
         }
       } else if (logType == 4) {
-        //Data = toAddress(32) + value(32) + ciphertext(80) + padding(16)
+        // Data = toAddress(32) + value(32) + cipher(80) + nonce(12) + reserved/version(4)
+        if (logData.length < 64 + NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE) {
+          return Optional.empty();
+        }
         byte[] logToAddress = ByteArray.subArray(logData, 12, 32);
         byte[] logAmountArray = ByteArray.subArray(logData, 32, 64);
         byte[] cipher = ByteArray.subArray(logData, 64, 144);
+        byte[] nonceFromLog = ByteArray.subArray(logData, 144,
+            144 + NoteEncryption.Encryption.BURN_NONCE_LEN);
+        byte[] reservedFromLog = ByteArray.subArray(logData,
+            144 + NoteEncryption.Encryption.BURN_NONCE_LEN,
+            144 + NoteEncryption.Encryption.BURN_NONCE_LEN
+                + NoteEncryption.Encryption.BURN_RESERVED_LEN);
         BigInteger logAmount = ByteUtil.bytesToBigInteger(logAmountArray);
         byte[] plaintext;
         byte[] amountArray = new byte[32];
         byte[] decryptedAddress = new byte[20];
+
         Optional<byte[]> decryptedText = NoteEncryption.Encryption
-            .decryptBurnMessageByOvk(ovk, cipher);
+            .decryptBurnMessageByOvk(ovk, cipher, nonceFromLog, reservedFromLog, pendingNf);
+
         if (decryptedText.isPresent()) {
           plaintext = decryptedText.get();
+          if (plaintext[32] != Wallet.getAddressPreFixByte()) {
+            return Optional.empty();
+          }
           System.arraycopy(plaintext, 0, amountArray, 0, 32);
           System.arraycopy(plaintext, 33, decryptedAddress, 0, 20);
           BigInteger decryptedAmount = ByteUtil.bytesToBigInteger(amountArray);
@@ -4091,15 +4106,24 @@ public class Wallet {
           if (!Objects.isNull(logList)) {
             Optional<DecryptNotesTRC20.NoteTx> noteTx;
             int index = 0;
+            byte[] pendingNf = null;
             for (TransactionInfo.Log log : logList) {
               int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress);
-              if (logType > 0) {
+              if (logType == 5) {
+                byte[] logData = log.getData().toByteArray();
+                if (logData.length >= 32) {
+                  pendingNf = ByteArray.subArray(logData, 0, 32);
+                }
+              } else if (logType > 0) {
                 noteBuilder = DecryptNotesTRC20.NoteTx.newBuilder();
                 noteBuilder.setTxid(ByteString.copyFrom(txid));
                 noteBuilder.setIndex(index);
                 index += 1;
-                noteTx = getNoteTxFromLogListByOvk(noteBuilder, log, ovk, logType);
+                noteTx = getNoteTxFromLogListByOvk(noteBuilder, log, ovk, logType, pendingNf);
                 noteTx.ifPresent(builder::addNoteTxs);
+                if (logType == 4) {
+                  pendingNf = null;
+                }
               }
             }
           }
@@ -4283,8 +4307,13 @@ public class Wallet {
         parameterType);
     if (parametersBuilder.getShieldedTRC20ParametersType() == ShieldedTRC20ParametersType.BURN) {
       byte[] burnCiper = ByteArray.fromHexString(shieldedTRC20Parameters.getTriggerContractInput());
-      if (!ArrayUtils.isEmpty(burnCiper) && burnCiper.length == 80) {
+      if (!ArrayUtils.isEmpty(burnCiper)
+          && burnCiper.length == NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE) {
         parametersBuilder.setBurnCiphertext(burnCiper);
+      } else if (!ArrayUtils.isEmpty(burnCiper) && burnCiper.length == 80) {
+        throw new ZksnarkException(
+            "legacy 80-byte burn cipher is deprecated and rejected; expected "
+                + NoteEncryption.Encryption.BURN_CIPHER_RECORD_SIZE + "-byte burn record");
       } else {
         throw new ZksnarkException(
             "invalid shielded TRC-20 contract parameters for burn trigger input");
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -4061,8 +4061,12 @@ public class Wallet {
         byte[] amountArray = new byte[32];
         byte[] decryptedAddress = new byte[20];
 
+        byte[] addr21FromLog = new byte[21];
+        addr21FromLog[0] = Wallet.getAddressPreFixByte();
+        System.arraycopy(logToAddress, 0, addr21FromLog, 1, 20);
         Optional<byte[]> decryptedText = NoteEncryption.Encryption
-            .decryptBurnMessageByOvk(ovk, cipher, nonceFromLog, reservedFromLog, pendingNf);
+            .decryptBurnMessageByOvk(ovk, cipher, nonceFromLog, reservedFromLog, pendingNf,
+                logAmountArray, addr21FromLog);
 
         if (decryptedText.isPresent()) {
           plaintext = decryptedText.get();
@@ -4342,10 +4346,14 @@ public class Wallet {
           NoteEncryption.Encryption.BURN_NONCE_OFFSET,
           NoteEncryption.Encryption.BURN_NONCE_OFFSET
               + NoteEncryption.Encryption.BURN_NONCE_LEN);
-      byte[] expectedNonce = NoteEncryption.Encryption.deriveNonceFromNf(nf);
+      byte[] amount32 = ByteUtil.bigIntegerToBytes(value, 32);
+      byte[] addr21 = new byte[21];
+      addr21[0] = Wallet.getAddressPreFixByte();
+      System.arraycopy(transparentToAddressTvm, 0, addr21, 1, 20);
+      byte[] expectedNonce = NoteEncryption.Encryption.deriveBurnNonce(nf, amount32, addr21);
       if (!Arrays.equals(nonceFromInput, expectedNonce)) {
         throw new ZksnarkException(
-            "burn trigger input nonce does not match nf-bound nonce");
+            "burn trigger input nonce does not match nonce bound to (nf, amount, addr)");
       }
       parametersBuilder.setBurnCiphertext(burnCiper);
     }
--- a/framework/src/main/java/org/tron/core/zen/ShieldedTRC20ParametersBuilder.java
+++ b/framework/src/main/java/org/tron/core/zen/ShieldedTRC20ParametersBuilder.java
@@ -63,8 +63,16 @@ public class ShieldedTRC20ParametersBuilder {
   private BigInteger transparentToAmount;
   @Setter
   private byte[] ovk;
-  @Setter
   private byte[] burnCiphertext = new byte[Encryption.BURN_CIPHER_RECORD_SIZE];
+
+  public void setBurnCiphertext(byte[] burnCiphertext) {
+    if (burnCiphertext == null
+        || burnCiphertext.length != Encryption.BURN_CIPHER_RECORD_SIZE) {
+      throw new IllegalArgumentException(
+          "burnCiphertext must be " + Encryption.BURN_CIPHER_RECORD_SIZE + " bytes");
+    }
+    this.burnCiphertext = burnCiphertext.clone();
+  }
 
   public ShieldedTRC20ParametersBuilder() {
 
--- a/framework/src/main/java/org/tron/core/zen/ShieldedTRC20ParametersBuilder.java
+++ b/framework/src/main/java/org/tron/core/zen/ShieldedTRC20ParametersBuilder.java
@@ -30,6 +30,7 @@ import org.tron.core.zen.address.ExpandedSpendingKey;
 import org.tron.core.zen.address.PaymentAddress;
 import org.tron.core.zen.note.Note;
 import org.tron.core.zen.note.NoteEncryption;
+import org.tron.core.zen.note.NoteEncryption.Encryption;
 import org.tron.core.zen.note.OutgoingPlaintext;
 import org.tron.protos.contract.ShieldContract;
 import org.tron.protos.contract.ShieldContract.ReceiveDescription;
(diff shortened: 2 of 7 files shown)


framework/src/main/java/org/tron/core/Wallet.java
waynercheung: [MUST] `NoteSpent` is now returned as `logType == 5`, but this IVK scanner increments `index` for every `logType > 0`. `NoteSpent` does not produce a note; in transfer logs it is emitted before `TransferNewLeaf`, so returned note indexes are shifted. Please only increment for note-producing log types (`1..3`) and add a regression test with `NoteSpent` before `TransferNewLeaf`.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -4283,8 +4307,13 @@ public BytesMessage getTriggerInputForShieldedTRC20Contract(
         parameterType);
     if (parametersBuilder.getShieldedTRC20ParametersType() == ShieldedTRC20ParametersType.BURN) {
       byte[] burnCiper = ByteArray.fromHexString(shieldedTRC20Parameters.getTriggerContractInput());
-      if (!ArrayUtils.isEmpty(burnCiper) && burnCiper.length == 80) {
+      if (!ArrayUtils.isEmpty(burnCiper)
```
waynercheung: [MUST] Length-only validation still accepts a 96-byte legacy burn record (`cipher(80) + zero nonce(12) + reserved=0`). That lets external-signature callers bypass the new nf-bound nonce requirement by appending 16 zero bytes to an old burn cipher. Please validate burn trigger input as v2 only: `reserved == 0x00000001` and `nonce == SHA3("Ztron_BurnNonce" || spendDescription.nullifier)[:12]`; reject legacy and unknown markers for newly submitted trigger input. The current test that accepts an all-zero 96-byte record should also be updated.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
```diff
@@ -246,47 +258,115 @@ public static Optional<OutPlaintext> attemptOutDecryption(
...
+      }
+
+      byte[] effectiveNonce;
+      if (isAllZero(reservedFromLog)) {
+        if (!isAllZero(nonceFromLog)) {
+          return Optional.empty();
+        }
+        effectiveNonce = nonceFromLog;
+      } else if (Arrays.equals(reservedFromLog, BURN_RECORD_V2_MARKER)) {
+        if (nf == null) {
```
waynercheung: [SHOULD] The v2 decrypt path should also require `nf.length == 32`. The nonce derivation is defined over a Sapling nullifier, so accepting arbitrary-length `nf` makes malformed caller input derive a different nonce instead of being rejected explicitly.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Fixed in the latest commit: the IVK scanner now only advances `index` when `logType` is 1..3, matching the OVK path. Added `WalletMockTest#scanShieldedTRC20NotesByIvkSkipsNoteSpentIndex` which lays out `[NoteSpent, TransferNewLeaf]` and asserts the returned note ends up at index 0 — verified to fail before the fix (`expected:<0> but was:<1>`) and pass after.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
```diff
@@ -246,47 +258,115 @@ public static Optional<OutPlaintext> attemptOutDecryption(
...
+          || reservedFromLog == null || reservedFromLog.length != BURN_RESERVED_LEN) {
+        return Optional.empty();
+      }
+
+      byte[] effectiveNonce;
+      if (isAllZero(reservedFromLog)) {
+        if (!isAllZero(nonceFromLog)) {
+          return Optional.empty();
+        }
+        effectiveNonce = nonceFromLog;
```
3for, line 358: [SHOULD] Since the on-chain anonymous contract side is not upgraded, its `bytes32[3] calldata burnCipher` parameter still accepts arbitrary `bytes32[3]` values and emits them into events. Although the Java scanner introduced a new v2 marker, `reserved=0 && nonce=0` still falls back to legacy decryption: `NoteEncryption.java` (line 329). An attacker can therefore bypass the new Wallet API and directly submit a contract call that injects a forged legacy payload into `burnCipher`, which may still be accepted by the new scanner.

A better fix would be to add an activation-height/time cutoff for legacy decryption, or reject legacy payloads by default and only allow them in a dedicated “historical scan mode” based on block height.

3for:
[SHOULD] The PR states that the change is “wallet / scanner / trigger-input only”. However, from the perspective of the old scanner, the new record is interpreted as `cipher(80) + padding(16)`, and the legacy logic attempts decryption with a zero nonce. Decryption of the v2 cipher will therefore fail, because encryption now uses a derived nonce.

So this is effectively a one-way compatibility upgrade:

* the new scanner can read both new and legacy records,
* while the old scanner can only read legacy records.

Suggested addition to the Compatibility table:

```text
┌────────────────────────────────┬──────────────────────────────────────────────────┐
│ Dimension                      │ Impact                                           │
├────────────────────────────────┼──────────────────────────────────────────────────┤
│ Old java-tron / old SDK scanner│ Cannot parse upgraded v2 burn records;           │
│                                │ synchronized upgrade is required                 │
└────────────────────────────────┴──────────────────────────────────────────────────┘
```

> ## Compatibility
> Dimension	Impact
> Consensus / hard fork	None — wallet / scanner / trigger-input only, no protocol-state transition

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Tightened `getTriggerInputForShieldedTRC20Contract` (Wallet.java:4310): for `BURN` we now require `reserved == 0x00000001`, exactly one 32-byte `spendDescription.nullifier`, and `nonce == SHA3("Ztron_BurnNonce" || nf)[:12]`. The all-zero 96-byte legacy fallback is rejected with `burn trigger input must be v2 (reserved=0x00000001); legacy/unknown markers rejected`, and a forged nonce is rejected with `burn trigger input nonce does not match nf-bound nonce`. Replaced the old all-zero-record test with `testGetTriggerInputBurnV2Accepted` + `…Legacy96ByteRecordRejected` + `…UnknownReservedRejected` + `…NonceMismatchRejected` in `NoteEncDecryTest`.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
Federico2014: Tightened the v2 decrypt branch (NoteEncryption.java:340) to require `nf.length == 32`, matching the Sapling nullifier size. Now malformed-length `nf` is rejected explicitly at the length check rather than implicitly via nonce mismatch. Added `BurnCipherTest#testDecryptWithWrongNfLengthFailsForV2` covering 0, 31, and 33 bytes.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
Federico2014: The threat model here looks asymmetric to us. An attacker emitting a forged `reserved=0 nonce=0` `burnCipher` on-chain doesn't gain anything: without the target's OVK they can't produce a cipher that decrypts to meaningful plaintext, so the forged log is just noise.

The real risk v2 addresses is on the **emit** side — a legitimate user producing burns with a predictable zero nonce lets an off-chain observer correlate multiple burns from the same sender. That's a self-harm scenario, not a cross-user attack.

We've already closed the emit path in r3256689153: `getTriggerInputForShieldedTRC20Contract` rejects anything that isn't v2, so any burn going through java-tron's Wallet API emits v2. Decrypt-side v1 support exists so users can still scan their own pre-upgrade burns; rejecting it would break that without preventing any cross-user attack.

Happy to add an activation-height cutoff if there's a concrete cross-user vector we're missing — but as we read it, the trigger-input check is the actual defense; the decrypt-side fallback is not security-critical.

Federico2014:
@3for  Good point — the asymmetry deserves to be called out explicitly. Updated the Compatibility table with a row for old scanners:

| Old java-tron / old SDK scanners | One-way compatibility: the new scanner reads both legacy and v2 records, but old scanners interpret a v2 record as `cipher(80) + zero-padding(16)` and attempt zero-nonce decryption, which fails. Scanners that need to read v2 burns must be upgraded in lockstep |

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
3for: At the moment, legacy nodes using the old zero-nonce format can still successfully get transactions on-chain. Even on upgraded nodes, `Wallet.java:4309-4317` now causes the `getShieldedTRC20TriggerContractParams` gRPC API to throw a `ZksnarkException` when receiving an 80-byte cipher, but this only prevents users from “assembling transactions through the upgraded node’s gRPC”.

If the caller:

* directly uses the gRPC of an unupgraded node, or
* manually constructs the calldata (`80B cipher + 16B 0`) and then calls `triggersmartcontract`

they can still bypass this restriction and get the transaction included in a block.

In other words, this fix cannot prevent an adversary from actively crafting zero-nonce transactions. It only guarantees that:

* (a) burns generated by the new wallet always include nf-binding;
* (b) the new wallet scanner rejects forged nf-mismatch v2 records.

[DISCUSS] To make “disallowing v1” an enforceable rule, the validation must be added at the consensus/contract layer (via a fork), e.g. by checking the last 16 bytes of the cipher, rather than merely rejecting it at the wallet layer.

[MUST] OR, the release notes should emphasize the importance of upgrading nodes for anonymous-transaction privacy protection. In addition, any self-developed tools that manually construct burn calldata transactions also need to be upgraded accordingly.

Federico2014: Thanks — the boundary you've drawn is accurate. This PR is intentionally wallet/scanner/trigger-input only, so the trigger-input check is a defense against accidental v1 emission by upgraded callers, not an on-chain ban on v1.

**On [DISCUSS] — consensus/contract-layer enforcement**

Agreed it's the only way to truly disallow v1 calldata on-chain, but pushing the check into the contract (or a consensus rule like "reject burn calldata whose last 16 bytes are all zero") would:

- turn this into a hard-fork PR, which is out of scope for the current "wallet-only, no protocol-state transition" framing in the Compatibility table;
- and impose a layout assumption on a user-supplied `bytes32[3]` calldata field that the contract doesn't currently interpret.

I'd rather track this as a follow-up TIP if the community wants v1 fully banned on-chain, instead of expanding this PR's scope.

**On [MUST] — upgrade guidance**

Fully agreed. I'll add an "Upgrade guidance" section to the PR body covering:

1. Node operators must upgrade java-tron to get the v2 scanner; old scanners can't read v2 burn records.
2. Wallet / SDK developers who hand-build burn calldata must emit the 96-byte v2 record (`cipher(80) + nf-bound nonce(12) + reserved=0x00000001(4)`); 80-byte calldata is still accepted on-chain but the resulting burn loses nf-binding and is decryptable only via the legacy zero-nonce path.
3. Until upgraded, users of un-upgraded wallets that still emit v1 burns receive weaker privacy guarantees (same `(ovk, nonce=0)` reuse across burns).

I'll mirror this into the release notes when 4.8.2 is cut.

waynercheung (approved):
LGTM

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
```diff
@@ -246,47 +262,115 @@ public static Optional<OutPlaintext> attemptOutDecryption(
...
-    public static Optional<byte[]> decryptBurnMessageByOvk(byte[] ovk, byte[] ciphertext)
-        throws ZksnarkException {
+    public static byte[] deriveNonceFromNf(byte[] nf) {
+      byte[] tagged = new byte[BURN_NONCE_DOMAIN.length + nf.length];
+      System.arraycopy(BURN_NONCE_DOMAIN, 0, tagged, 0, BURN_NONCE_DOMAIN.length);
+      System.arraycopy(nf, 0, tagged, BURN_NONCE_DOMAIN.length, nf.length);
+      byte[] hash = Hash.sha3(tagged);
+      byte[] nonce = new byte[BURN_NONCE_LEN];
+      System.arraycopy(hash, 0, nonce, 0, BURN_NONCE_LEN);
+      return nonce;
```
3for: [SHOULD] `nonce` is currently fully determined by `nf`, so repeated encryption with the same `nf` can still reuse the same `(key, nonce)` pair. This breaks AEAD requirements: different burn plaintexts under the same `ovk + nf` will reuse the ChaCha20-Poly1305 keystream.

This is realistically triggerable before anything lands on-chain: repeated `getTriggerInputForShieldedTRC20Contract()` calls with different `(amount, transparentToAddress)` values (user retries, frontend reconstruction, multiple signing attempts, etc.) will produce different plaintexts under the same deterministic nonce.

The chain only guarantees at most one successful burn is committed. It does not guarantee that multiple ciphertexts were never exposed off-chain (API responses, frontend↔node traffic, unbroadcast txs, etc.). Even if practical leakage is somewhat limited because `amount/toAddress` later appear in calldata/logs, this still violates the PR’s **“unique nonce per burn”** claim and is cryptographically unsafe AEAD usage.

Suggestion: include plaintext fields in nonce derivation, e.g.

```text
nonce = SHA3("Ztron_BurnNonce" ‖ nf ‖ amount ‖ addr21)[:12]
```

The scanner already has `amount/toAddress` from log data, so it can still verify:

```text
nonceFromLog == derive(nf, amount, addr)
```

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -4091,15 +4108,24 @@ public DecryptNotesTRC20 scanShieldedTRC20NotesByOvk(long startNum, long endNum,
...
+              } else if (logType > 0) {
                 noteBuilder = DecryptNotesTRC20.NoteTx.newBuilder();
                 noteBuilder.setTxid(ByteString.copyFrom(txid));
                 noteBuilder.setIndex(index);
                 index += 1;
-                noteTx = getNoteTxFromLogListByOvk(noteBuilder, log, ovk, logType);
+                noteTx = getNoteTxFromLogListByOvk(noteBuilder, log, ovk, logType, pendingNf);
                 noteTx.ifPresent(builder::addNoteTxs);
+                if (logType == 4) {
+                  pendingNf = null;
```
3for, line 4131: [NIT] `pendingNf` cleanup is incomplete.

`Wallet.java:4111-4128`:

```java
byte[] pendingNf = null;
for (log : logList) {
  if (logType == 5) { pendingNf = ...; }       // set
  else if (logType > 0) {
    ...
    if (logType == 4) { pendingNf = null; }    // only cleared after burn
  }
}
```

Current contract behavior happens to avoid issues:

* In the transfer path, `NoteSpent` is followed by `TransferNewLeaf` (type 2), so `pendingNf` remains set across iterations. Type 2 currently does not read `pendingNf`, so there is no immediate bug.
* However, this relies on an implicit invariant. Any future contract change that allows `TokenBurn` ordering to shift relative to `NoteSpent`, or allows transfer + burn within the same tx, could consume the wrong `nf`.

There is also a related fragility:

* `transfer()` with two inputs emits two consecutive `NoteSpent` logs. The second `nf` overwrites the first one, so the first `nf` is silently lost.
* This is currently harmless because transfer burn ciphers do not depend on `nf`, but the logic is brittle.

Suggested fix:

* Clear `pendingNf` whenever the next iteration is not type 4 after a type 5.
* Or more robustly: only hold `pendingNf` locally inside the type-4 handling path and always clear it after processing (success or failure).

framework/src/main/java/org/tron/core/zen/ShieldedTRC20ParametersBuilder.java
```diff
@@ -61,7 +62,9 @@ public class ShieldedTRC20ParametersBuilder {
   @Setter
   private BigInteger transparentToAmount;
   @Setter
-  private byte[] burnCiphertext = new byte[80];
+  private byte[] ovk;
+  @Setter
+  private byte[] burnCiphertext = new byte[Encryption.BURN_CIPHER_RECORD_SIZE];
```
3for: [SHOULD] `burnParamsToHexString()` currently relies on external setters to provide a correct 96-byte ciphertext.

`ShieldedTRC20ParametersBuilder.java:67`:

```java
private byte[] burnCiphertext = new byte[Encryption.BURN_CIPHER_RECORD_SIZE];
```

The default value is 96 zero bytes (legacy v1 zero-padding behavior). In the normal `build()` → `BURN` flow, this gets replaced with a proper v2 record. However:

* `setBurnCiphertext(80B)` can currently inject an invalid-length array. Later `ByteUtil.merge(...)` does not validate lengths and silently produces malformed calldata.
* Calling `burnParamsToHexString(...)` directly without going through `build()` has the same issue.
* The generated Lombok `@Setter` performs no length validation.

Suggested fix: replace `@Setter` with a manual setter enforcing `BURN_CIPHER_RECORD_SIZE`, e.g.

```java
public void setBurnCiphertext(byte[] burnCiphertext) {
  if (burnCiphertext == null
      || burnCiphertext.length != Encryption.BURN_CIPHER_RECORD_SIZE) {
    throw new IllegalArgumentException("burnCiphertext must be 96 bytes");
  }
  this.burnCiphertext = burnCiphertext.clone();
}
```

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
```diff
@@ -246,47 +262,115 @@ public static Optional<OutPlaintext> attemptOutDecryption(
...
+        }
+        effectiveNonce = nonceFromLog;
+      } else if (Arrays.equals(reservedFromLog, BURN_RECORD_V2_MARKER)) {
+        if (nf == null || nf.length != 32) {
+          return Optional.empty();
+        }
+        byte[] derived = deriveNonceFromNf(nf);
+        if (!Arrays.equals(nonceFromLog, derived)) {
+          return Optional.empty();
+        }
```
3for: [SHOULD] This is mainly an observability / diagnosability issue, not a cryptographic vulnerability.

The v2 nonce is redundant since the scanner can already derive it from `nf + amount + address`. However, because the contract does not validate burn payloads, malformed records such as:

```text
reserved = 0x00000001
nonce != derive(nf)
```

can still be committed on-chain successfully.

The scanner then silently drops the record:

```java
if (!Arrays.equals(nonceFromLog, derived)) {
  return Optional.empty();
}
```

Result: the burn succeeds, but OVK scanning cannot recover the record. This can be caused by buggy or malicious SDKs and makes debugging / wallet support difficult.

Suggestion:

* Emit debug/warn logs for malformed v2 records (nonce mismatch, missing `nf`, unknown reserved value, etc.).
* Avoid logging sensitive plaintexts; `txid/index/logType` should be enough.
* Ideally expose malformed/rejected status instead of only returning `Optional.empty()`.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
Federico2014: Good catch, agreed — addressed in the latest commit.

  Change: `deriveNonceFromNf(nf) → deriveBurnNonce(nf, amount32, addr21)`, deriving as:
  
  `nonce = SHA3("Ztron_BurnNonce" ‖ nf ‖ amount32 ‖ addr21)[:12]`

  This restores AEAD nonce uniqueness when the same input note is used to produce multiple burn trigger inputs off-chain with differing `(amount, addr)`.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
Federico2014: Thanks, but I'd prefer to leave this silent.

Builder-side validation in `getTriggerInputForShieldedTRC20Contract` (Wallet.java:4344, commit 61e6a891a) already rejects malformed v2 records before signing. The only remaining way to put one on-chain is hand-crafted calldata that intentionally skips the library — a self-DoS where the sender's own burn becomes unrecoverable. Their cipher is under their own OVK, not ours, so it never reaches our nonce check anyway.

Adding warn logs would mostly give operators a noisy, non-actionable signal that's easy to amplify by spamming junk records. Happy to revisit if a real debugging scenario shows up.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Thanks for catching this. After looking at the actual call graph, I decided to leave the existing `if (logType == 4)` cleanup as-is:

- `pendingNf` is only ever *read* by `getNoteTxFromLogListByOvk` when `logType == 4`. Types 1/2/3 ignore it entirely.
- The current actuator never interleaves output ciphers (1/2/3) with burn (4) inside a single transaction: burn emits `[5, 4]`, transfer emits `[5..., 1/2/3...]`, mint emits `[1/2/3...]`. So the stale-`pendingNf` window the comment describes isn't reachable today.
- `pendingNf` is reinitialized to `null` at the top of every transaction's log loop, so there's no cross-transaction leak either.

If the actuator ever changes to interleave types in a single tx, the scanner would need a re-review anyway, so I'd rather not add defensive code without a corresponding real failure mode. Happy to revisit if the protocol evolves.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -3909,7 +3909,9 @@ private DecryptNotesTRC20 queryTRC20NoteByIvk(long startNum, long endNum,
             int index = 0;
             for (TransactionInfo.Log log : logList) {
               int logType = getShieldedTRC20LogType(log, shieldedTRC20ContractAddress);
-              if (logType > 0) {
+              // Only note-producing log types (1..3) advance the note index;
+              // TokenBurn (4) and NoteSpent (5) do not emit a leaf.
+              if (logType > 0 && logType < 4) {
```
3for, line 3914: [NIT] The `ivk` scan index semantic change is a silent breaking change. `Wallet.java:3914`

```java
// Before:
if (logType > 0)

// After:
if (logType > 0 && logType < 4)
```

For the same transaction containing a burn event, the old version would consume one index slot for `TokenBurn` (`logType = 4`) — still returning empty, resulting in a sparse index. The new version no longer consumes a slot. Any SDK or indexer relying on cross-version index stability will break.

Recommend explicitly documenting this in the release notes, e.g.:

`scanShieldedTRC20NotesByIvk` / `scanAndMark...` `NoteTx.index` now counts only note-producing events; `TokenBurn` and `NoteSpent` no longer consume index slots. Indexers relying on the previous log-ordinal behavior need to adjust accordingly.

3for (approved):

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Thanks for flagging — I went back and checked the cross-version behavior and I don't think this actually changes any observable index. Two reasons:

1. `NoteSpent` was not recognized by `getShieldedTRC20LogType` on `develop` — it falls through to `return 0`, so the old `logType > 0` guard already skipped it. The new code recognizes it as `5` and the new `logType > 0 && logType < 4` guard still skips it. No change.

2. `TokenBurn` is always the last log in any given shielded burn tx (the shielded TRC-20 contract emits it at the end and there are no further note-producing logs after it). Under the old guard the cursor did `index += 1` for that final burn log, but since no subsequent log consumed that incremented value, the index returned for every preceding note is identical on both sides.

So this is purely an internal cleanup — the same set of `(txid, index)` pairs comes out for any tx that exists on chain today. Let me know if you can think of a log ordering where the two versions would diverge; otherwise I'd prefer not to add a release-notes warning that doesn't correspond to a real behavior change.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -4091,15 +4112,24 @@ public DecryptNotesTRC20 scanShieldedTRC20NotesByOvk(long startNum, long endNum,
           if (!Objects.isNull(logList)) {
             Optional<DecryptNotesTRC20.NoteTx> noteTx;
             int index = 0;
+            byte[] pendingNf = null;
```
waynercheung, line 4115: [SHOULD] This v2 burn scanner now depends on an implicit per-transaction log ordering invariant: the corresponding NoteSpent event must be observed before TokenBurn so that pendingNf is populated for burn decryption. If that ordering ever changes, scanShieldedTRC20NotesByOvk will silently stop decoding v2 burns. Please either document that invariant here or add a scanner-level test that feeds a full log list containing NoteSpent + TokenBurn and proves the pairing behavior.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
```diff
@@ -246,47 +262,137 @@ public static Optional<OutPlaintext> attemptOutDecryption(
     }
 
     /**
-     * encrypt the message by ovk used for scanning
+     * encrypt burn message with nonce bound to (nf, amount, addr21), returns a 96B record:
+     * cipher(80) + nonce(12) + reserved/version(4).
      */
     public static Optional<byte[]> encryptBurnMessageByOvk(byte[] ovk, BigInteger toAmount,
```
waynercheung, line 268: [SHOULD] encryptBurnMessageByOvk(...) validates ovk / addr / nf, but not toAmount itself. Current Wallet entry points validate amount upstream, so this is not a user-facing RPC bug today; however, as a public crypto helper this method should still reject null, negative, and >256-bit amounts explicitly instead of relying on ByteUtil.bigIntegerToBytes(..., 32), which can otherwise re-encode invalid values into a different 32-byte unsigned representation.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Thanks for the careful read. The ordering (`NoteSpent` immediately before `TokenBurn` in the same tx) is fixed by the shielded TRC20 contract itself — it's part of the on-chain emission contract, not a property of the scanner. If that order ever changed, it would be a hard fork of the shielded contract, not a silent scanner regression, and the existing v1 burn path (`logType == 4` with `pendingNf == null`) would still decode correctly.

Given that, I'd prefer to leave this PR scoped to the nonce-binding fix and not add documentation/tests for an invariant that lives in the contract layer. Happy to revisit separately if we want to harden the scanner against hypothetical contract changes.

framework/src/main/java/org/tron/core/zen/note/NoteEncryption.java
Federico2014: Fair point on defensive coding for a public helper. The reason I'd prefer to leave this PR scoped: the only caller is `ShieldedTRC20ParametersBuilder`, which gets `toAmount` from the protobuf builder where it's already constrained to a non-negative `uint256` by `checkBurnAmountValid()` / `checkBurnAmount()` upstream. Null can't reach here either — the protobuf path produces a `BigInteger.ZERO`-or-greater value.

So today the guards would be belt-and-suspenders rather than a fix. If we want to make this helper safe for future callers outside the shielded TRC20 pipeline, I'd rather do it in a small follow-up that adds the `toAmount` range check together with similar hardening for other public crypto helpers, instead of mixing it into the nonce-binding fix.

PR #6776, opened 2026-05-18
fix(crypto): strengthen signature length validation

Federico2014:
## Summary

- Tighten signature-length validation across transaction, block, relay, and PBFT verification paths.
- Centralize the bounds in `ChainConstant.MIN_SIGNATURE_SIZE` (65) and `ChainConstant.MAX_SIGNATURE_SIZE` (68), and centralize the predicate in `SignUtils.isValidLength(int size, boolean checkMaxSignatureSize)`.
- Gate the consensus/PBFT upper-bound checks with `ForkController.signatureMaxSizeChecked()`, which passes only after `VERSION_4_8_2` fork stats meet the required threshold.
- Unconditionally enforce the upper bound on admission paths that ingest fresh, non-historical signatures (`Wallet.broadcastTransaction`, `TransactionsMsgHandler.check`, `RelayService.checkHelloMessage`).
- Preserve compatibility for read-only transaction introspection by keeping `Wallet.getTransactionApprovedList` on a lower-bound-only check, so historical transactions with padded signatures remain inspectable.

### Why use `[65, 68]` instead of strict `65`?

After the fork gate is active, the accepted signature-length window is `[65, 68]` bytes rather than strict `== 65`.

This is a deliberate compatibility trade-off:

- A canonical ECDSA signature is exactly 65 bytes (`r || s || v`).
- Historical on-chain data contains signatures with a small amount of trailing padding that ECDSA recovery silently ignores.
- `MAX_SIGNATURE_SIZE = 68` closes the previously unbounded encoding window while remaining compatible with historical data already on chain.
- The long-term goal is to tighten the rule to strict `== 65` once the compatibility window is no longer needed.

Because the bounds are centralized in `ChainConstant` and the predicate is centralized in `SignUtils.isValidLength`, tightening the rule later will be straightforward.

### Why gate consensus paths with `ForkController.signatureMaxSizeChecked()`?

Consensus and block-validation paths must be able to replay history. The upper bound on these paths is enforced only after `VERSION_4_8_2` passes through the existing fork-stats mechanism:

- `SignUtils.isValidLength(size, false)` enforces only the lower bound.
- `SignUtils.isValidLength(size, true)` enforces both the lower and upper bounds.
- Block witness validation, transaction permission checks, and PBFT signature analysis use `ForkController.signatureMaxSizeChecked()`.
- `ForkController.signatureMaxSizeChecked()` delegates to `ForkController.instance().pass(ForkBlockVersionEnum.VERSION_4_8_2)`.
- The read-only lookup path in `Wallet.getTransactionApprovedList` intentionally passes `false` so historical padded signatures remain readable even after the fork is active.

### Why enforce both bounds unconditionally on admission paths?

Admission paths see freshly created signatures, not historical chain state, so there is no need to keep a permissive window for replay:

- `Wallet.broadcastTransaction` - local RPC entry; clients producing oversized signatures are misbehaving regardless of fork state.
- `TransactionsMsgHandler.check` - P2P transaction ingress; an oversized signature here cannot come from already-on-chain data.
- `RelayService.checkHelloMessage` - peer handshake signature, generated on demand from a fresh timestamp.

Enforcing `[65, 68]` unconditionally on these paths blocks malformed traffic at the edge without affecting historical chain replay.

## Compatibility

This PR tightens validation behavior, but it does not apply the new upper bound uniformly to every path:

- Before `VERSION_4_8_2` passes, consensus/block/PBFT validation paths enforce only the lower bound (`>= 65`).
- After `VERSION_4_8_2` passes, consensus/block/PBFT validation paths enforce both bounds (`[65, 68]`).
- Admission paths (`Wallet.broadcastTransaction`, `TransactionsMsgHandler.check`, `RelayService.checkHelloMessage`) enforce both bounds (`[65, 68]`) unconditionally.
- `Wallet.getTransactionApprovedList` remains intentionally more permissive and continues to skip the upper bound so historical transactions can still be inspected.

This means admission policy can be stricter than pre-fork consensus replay. That is intentional because admission only handles fresh inbound data, while consensus replay must preserve historical compatibility.

## Breaking Changes

The following public method signatures changed:

| Method | Before | After |
|---|---|---|
| `TransactionCapsule.checkWeight` | `(Permission, List<ByteString>, byte[], List<ByteString>)` | `(Permission, List<ByteString>, byte[], List<ByteString>, boolean)` |

All in-tree callers were updated.

External JVM consumers of the `chainbase` artifact, such as SDKs, signing services, and third-party tools, must update these call sites. Existing callers will fail at compile time rather than silently misbehave at runtime.

## Main Code Changes

| File | Change |
|------|--------|
| `common/.../config/Parameter.java` | Add `MIN_SIGNATURE_SIZE`, `MAX_SIGNATURE_SIZE`, and `VERSION_4_8_2`/`BLOCK_VERSION = 35` |
| `crypto/.../SignUtils.java` | Add `isValidLength(int, boolean)` |
| `chainbase/.../ForkController.java` | Add `signatureMaxSizeChecked()` as the `VERSION_4_8_2` fork gate |
| `chainbase/.../DynamicPropertiesStore.java` | Remove the temporary `ALLOW_TVM_OSAKA`-based `signatureMaxSizeChecked()` helper |
| `chainbase/.../TransactionCapsule.java` | Thread max-size checking through `checkWeight`; `validateSignature` uses the fork gate, while `addSign` validates existing signatures with an unconditional max-size check |
| `chainbase/.../BlockCapsule.java` | Apply fork-aware signature-size validation to witness signatures in `validateSignature` |
| `framework/.../Wallet.java` | Add unconditional sig-length check in `broadcastTransaction`; keep the read-only approved-list path lower-bound-only |
| `framework/.../TransactionsMsgHandler.java` | Add unconditional sig-length check in `check()` for the P2P transaction ingress path |
| `framework/.../RelayService.java` | Apply unconditional sig-length validation in `checkHelloMessage` |
| `framework/.../PbftDataSyncHandler.java` | Apply fork-aware signature-size validation in `ValidPbftSignTask` |
| `framework/.../PbftMsgHandler.java` | Use the shared fork-aware PBFT signature check |
| `consensus/.../PbftBaseMessage.java` | Apply signature-size validation in `analyzeSignature` |
| `consensus/.../PbftMessageHandle.java` | Use the shared fork-aware PBFT signature check |
| `actuator/.../TransactionUtil.java` | Keep read-only `checkWeight` usage lower-bound-only |

## Test Coverage Added or Updated

- `ForkControllerTest`
  - Verify `ForkController.signatureMaxSizeChecked()` uses real `VERSION_4_8_2` fork stats: false below threshold, true once 80% stats pass.
- `TransactionCapsuleTest`
  - Reject short signatures.
  - Reject padded signatures when max-size checking is enabled.
  - Preserve padded-signature acceptance when max-size checking is disabled.
- `BlockCapsuleTest`
  - Reject short witness signatures.
  - Reject padded witness signatures when max-size checking is enabled.
  - Preserve padded witness-signature acceptance when max-size checking is disabled.
- `TransactionExpireTest`
  - Keep padded signatures readable through `getTransactionApprovedList` after Osaka.
  - Reject short signatures on the read-only path.
  - Preserve padded-signature readability before Osaka.
- `TransactionUtilTest`
  - Reject short signatures in `getTransactionSignWeight`.
  - Accept padded signatures on the read-only weight path after Osaka.
- `WalletMockTest`
  - Reject short, oversized, and empty signatures on the `broadcastTransaction` admission path; accept `[65, 68]`.
- `TransactionsMsgHandlerTest`
  - Reject short and oversized signatures on the P2P ingress path; accept `[65, 68]`.
- `RelayServiceTest`
  - Reject oversized hello-message signatures unconditionally.
- `PbftDataSyncHandlerTest`
  - Reject padded PBFT signatures when max-size checking is enabled.
- `PbftMsgHandlerTest`
  - Reject padded PBFT message signatures when max-size checking is enabled.

## Verification

- `./gradlew framework:test --tests org.tron.core.ForkControllerTest`
- `./gradlew framework:test --tests org.tron.core.capsule.BlockCapsuleTest --tests org.tron.core.capsule.TransactionCapsuleTest --tests org.tron.core.net.messagehandler.PbftDataSyncHandlerTest --tests org.tron.core.net.messagehandler.PbftMsgHandlerTest`
- `./gradlew framework:checkstyleMain framework:checkstyleTest`
- `git diff --check`


chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
```diff
@@ -185,9 +185,14 @@ private Sha256Hash getRawHash() {
   public boolean validateSignature(DynamicPropertiesStore dynamicPropertiesStore,
       AccountStore accountStore) throws ValidateSignatureException {
     try {
+      ByteString witnessSig = block.getBlockHeader().getWitnessSignature();
+      if (!SignUtils.isValidLength(witnessSig.size(),
+          dynamicPropertiesStore.isAllowTvmOsaka())) {
+        throw new ValidateSignatureException(
+            "Witness signature size is " + witnessSig.size());
+      }
```
SeriousCoding789, line 194: suggest a more concise way
```
if (dynamicPropertiesStore.isAllowTvmOsaka()) {
    SignUtils.isValidLength(witnessSig.size() 
}
```
Then inside` SignUtils.isValidLength` if not validate just 
```
throw new ValidateSignatureException("Witness signature size is " + witnessSig.size());
```

SeriousCoding789: In above way, all below similar logics will looks short and clear.

Federico2014: Thanks for the suggestion. I'd prefer to keep the current shape here:

- `isValidLength` enforces both `MIN_SIGNATURE_SIZE = 65` and (conditionally) `MAX_SIGNATURE_SIZE = 68`. The 65-byte lower bound is a physical property of ECDSA signatures and is unrelated to the Osaka fork, so gating the whole call behind `if (isAllowTvmOsaka())` would let sub-65-byte witness signatures bypass the size guard pre-Osaka — a regression in the existing check.
- `SignUtils` lives in the `crypto` module as a pure utility, while `ValidateSignatureException` belongs to `chainbase/core`. There are 6 call sites today (`BlockCapsule`, `TransactionCapsule`, `PbftBaseMessage`, `Wallet`, `PbftDataSyncHandler`, `RelayService`), each throwing the exception appropriate to its layer; moving the `throw` into the utility would either pin one exception type on everyone or pull multiple core exceptions back into `crypto`.
- `isValidLength` returning a boolean keeps it a pure, easily-testable predicate, consistent with its name.

Keeping it as-is for now — happy to revisit if you see a concrete issue with this layering.

framework/src/main/java/org/tron/core/net/messagehandler/PbftDataSyncHandler.java
```diff
@@ -173,14 +173,18 @@ private class ValidPbftSignTask implements Callable<Boolean> {
...
+            chainBaseManager.getDynamicPropertiesStore().isAllowTvmOsaka())) {
+          throw new SignatureException("PBFT signature size is " + sign.size());
+        }
         byte[] srAddress = ECKey.signatureToAddress(dataHash,
             TransactionCapsule.getBase64FromByteString(sign));
         if (!srSet.contains(ByteString.copyFrom(srAddress))) {
           logger.error("valid sr signature fail,error sr address:{}",
               ByteArray.toHexString(srAddress));
           return false;
         }
```
3for: `PbftDataSyncHandler.validPbftSign` should not reject the entire message just because a single task fails. It should only count valid SR addresses and accept the message once the quorum threshold is met. Otherwise, after Osaka, an attacker could reject an otherwise valid multi-signature PBFT message simply by injecting one invalid signature.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -644,7 +645,8 @@ public TransactionApprovedList getTransactionApprovedList(Transaction trx) {
           byte[] hash = Sha256Hash.hash(CommonParameter
               .getInstance().isECKeyCryptoEngine(), trx.getRawData().toByteArray());
           for (ByteString sig : trx.getSignatureList()) {
-            if (sig.size() < 65) {
+            if (!SignUtils.isValidLength(sig.size(),
+                chainBaseManager.getDynamicPropertiesStore().isAllowTvmOsaka())) {
```
3for:
```text
gettransactionbyid(txid)
        ↓
retrieve full transaction
        ↓
getapprovedlist(transaction)
        ↓
resolve signer addresses
```

When users follow the above workflow, querying/evaluating a pre-Osaka historical transaction with a 69-byte signature after Osaka will now return a signature format error. However, when replayed against the historical block height, it should not fail due to signature length.

Since these APIs are able to support historical objects, signature validation should be performed against the proposal state corresponding to the transaction’s block height/timestamp. Otherwise, the documentation should explicitly state that validation is performed using the current rules.

`getTransactionSignWeight` has a similar issue as well.

framework/src/main/java/org/tron/core/net/messagehandler/PbftDataSyncHandler.java
Federico2014: Good catch, thanks. You're right — this is a pre-existing DoS vector (one bad address-mapping was already enough to take down the whole message), but the new size check makes it cheaper to trigger, so it's fair to fix it here.

Done in the latest push:

1. **`validPbftSign` outer loop** no longer early-exits on a single task failure. It now consumes all `Future`s and lets the post-loop `validSrAddressSet.size() < agreeNodeCount` quorum check decide. Each task still contributes to `validSrAddressSet` only on success, so a malformed/unknown-SR signature simply doesn't count toward quorum.
2. **`ValidPbftSignTask` size-check** now `return false` directly (with a log line) instead of throwing a `SignatureException` only to catch it one block below — keeps it consistent with the existing `!srSet.contains(...)` path. The `catch (SignatureException e)` is kept for the real `ECKey.signatureToAddress` failure path.

Added a regression test `testValidPbftSignAcceptsQuorumDespiteOneBadSig`: 3 valid SR sigs + 1 injected 64-byte garbage sig with `agreeNodeCount=3`, expects `true`. The pre-fix code would have rejected the whole message.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Good catch — fixed.

The fork-state lookup would be the strictly-correct fix, but adding a block-height → proposal-state path just for these two read-only APIs is more machinery than the problem warrants. Instead I dropped the upper-bound check on the read-only path entirely:

- `Wallet.getTransactionApprovedList` now only enforces `sig.size() >= MIN_SIGNATURE_SIZE` (the lower bound is a structural invariant, not a fork rule, so it stays).
- `TransactionUtil.getTransactionSignWeight` calls a new `TransactionCapsule.checkWeight(..., boolean osakaAllowed)` overload with `false`, so historical 69-byte sigs introspect cleanly.

The post-Osaka upper bound still gates *new submissions* via the consensus path (`validateSignature`, `TransactionUtil` validation, PBFT sigs), which is what the proposal is actually meant to do. A read-only API exposing a previously-committed transaction is not a place that should second-guess the chain.

New tests cover both directions: `getTransactionApprovedList`/`getTransactionSignWeight` accept a 69-byte sig post-Osaka, and still reject sub-65-byte sigs. Done in the latest push.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
```diff
@@ -230,7 +232,18 @@ public static long getWeight(Permission permission, byte[] address) {
    *  @see ForkController#init(org.tron.core.ChainBaseManager)
    */
   public static long checkWeight(Permission permission, List<ByteString> sigs, byte[] hash,
-      List<ByteString> approveList)
+      List<ByteString> approveList, DynamicPropertiesStore dynamicPropertiesStore)
+      throws SignatureException, PermissionException, SignatureFormatException {
+    Objects.requireNonNull(dynamicPropertiesStore, "dynamicPropertiesStore");
+    return checkWeight(permission, sigs, hash, approveList,
+        dynamicPropertiesStore.isAllowTvmOsaka());
```
SeriousCoding789: Since inner `checkWeight` only need `dynamicPropertiesStore.isAllowTvmOsaka()`, how about just pass this variable instead of `DynamicPropertiesStore dynamicPropertiesStore`

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
```diff
@@ -230,7 +232,18 @@ public static long getWeight(Permission permission, byte[] address) {
...
+  }
+
+  // Read-only callers (e.g. getTransactionSignWeight) should pass osakaAllowed=false
+  // so historical transactions with sigs above MAX_SIGNATURE_SIZE are still
+  // introspectable; the upper bound is meant to gate new submissions post-Osaka.
+  public static long checkWeight(Permission permission, List<ByteString> sigs, byte[] hash,
+      List<ByteString> approveList, boolean osakaAllowed)
```
SeriousCoding789: Better rename `osakaAllowed` to `signatureMaxSizeChecked` in this function, as `osakaAllowed` is just a reused flag in `dynamicPropertiesStore`, should not keep it passing through.

crypto/src/main/java/org/tron/common/crypto/SignUtils.java
```diff
@@ -5,9 +5,15 @@
 import org.tron.common.crypto.ECKey.ECDSASignature;
 import org.tron.common.crypto.sm2.SM2;
 import org.tron.common.crypto.sm2.SM2.SM2Signature;
+import org.tron.core.config.Parameter.ChainConstant;
 
 public class SignUtils {
 
+  public static boolean isValidLength(int size, boolean osakaAllowed) {
+    return size >= ChainConstant.MIN_SIGNATURE_SIZE
+        && (!osakaAllowed || size <= ChainConstant.MAX_SIGNATURE_SIZE);
```
SeriousCoding789: Rename `osakaAllowed` to something like `signatureMaxSizeChecked`

framework/src/main/java/org/tron/core/net/service/relay/RelayService.java
```diff
@@ -152,10 +152,15 @@ public boolean checkHelloMessage(HelloMessage message, Channel channel) {
 
     boolean flag;
     try {
+      ByteString signature = msg.getSignature();
+      if (!SignUtils.isValidLength(signature.size(),
+          chainBaseManager.getDynamicPropertiesStore().isAllowTvmOsaka())) {
```
SeriousCoding789: [MUST] This signature size check is newly added, please add a `logger` following the above styles.

framework/src/main/java/org/tron/core/net/messagehandler/PbftDataSyncHandler.java
```diff
@@ -122,31 +123,28 @@ private void processPBFTCommitMessage(PbftCommitMessage pbftCommitMessage) {
   private boolean validPbftSign(Raw raw, List<ByteString> srSignList,
       List<ByteString> currentSrList) {
     //valid sr list
+    if (srSignList.size() < Param.getInstance().getAgreeNodeCount()) {
```
SeriousCoding789: Put this line above` if (srSignList.size() != 0) {`, when `srSignList.size() == 0` old logic will return true, while new logic will return false.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -644,7 +645,10 @@ public TransactionApprovedList getTransactionApprovedList(Transaction trx) {
           byte[] hash = Sha256Hash.hash(CommonParameter
               .getInstance().isECKeyCryptoEngine(), trx.getRawData().toByteArray());
           for (ByteString sig : trx.getSignatureList()) {
-            if (sig.size() < 65) {
+            // Read-only introspection: only enforce the lower bound so a tx
+            // committed pre-Osaka with a sig above MAX_SIGNATURE_SIZE is still
+            // resolvable; the upper bound only gates new submissions post-Osaka.
+            if (sig.size() < ChainConstant.MIN_SIGNATURE_SIZE) {
```
SeriousCoding789: Although here only check min size, for future easy update, I suggest still use `SignUtils.isValidLength` with a hardcode `false`. In this way, all signature length check logic will be uniformly handled by that function, and easy to refactor later.

framework/src/test/java/org/tron/core/capsule/BlockCapsuleTest.java
```diff
@@ -196,4 +203,86 @@ public void testConcurrentToString() throws InterruptedException {
...
+    when(dps.isAllowTvmOsaka()).thenReturn(true);
+    AccountStore accountStore = mock(AccountStore.class);
+
+    try {
+      paddedBlock.validateSignature(dps, accountStore);
+      Assert.fail("Expected ValidateSignatureException for padded witness sig post-Osaka");
+    } catch (ValidateSignatureException e) {
+      Assert.assertTrue("Error must mention the sig size",
+          e.getMessage().contains("69"));
+    }
```
SeriousCoding789, line 355: Suggest a more standard syntax used in JUnit 4.13+
```
ValidateSignatureException e = Assert.assertThrows(
      ValidateSignatureException.class,
      () -> paddedBlock.validateSignature(dps, accountStore));
Assert.assertTrue("Error must mention the sig size",
      e.getMessage().contains("69"));
```

framework/src/test/java/org/tron/core/net/messagehandler/PbftMsgHandlerTest.java
```diff
@@ -120,4 +122,52 @@ public void testPbft() throws Exception {
...
+    peer.setNeedSyncFromPeer(false);
+
+    DynamicPropertiesStore dynamicPropertiesStore = context.getBean(DynamicPropertiesStore.class);
+    dynamicPropertiesStore.saveAllowPBFT(1);
+    dynamicPropertiesStore.saveAllowTvmOsaka(1);
+    try {
+      context.getBean(PbftMsgHandler.class).processMessage(peer, pbftMessage);
+      Assert.fail("Padded PBFT signature must be rejected post-Osaka");
+    } catch (SignatureException e) {
+      Assert.assertTrue(e.getMessage().contains("PBFT signature size is 69"));
```
SeriousCoding789:
```
SignatureException e = Assert.assertThrows(
      SignatureException.class,
      () -> context.getBean(PbftMsgHandler.class).processMessage(peer, pbftMessage));
  Assert.assertTrue("Padded PBFT signature must be rejected post-Osaka",
      e.getMessage().contains("PBFT signature size is 69"));
```

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
Federico2014: Good catch, done.

framework/src/main/java/org/tron/core/net/messagehandler/PbftDataSyncHandler.java
Federico2014: This is intentional. An empty `srSignList` cannot meet the `2/3 + 1` quorum, so returning `true` in that path was a latent bug: the old code accepted a PBFT commit message with zero attached signatures as if it had passed consensus. The new guard rejects it consistently with the size-based quorum check.

actuator/src/main/java/org/tron/core/utils/TransactionUtil.java
```diff
@@ -223,9 +223,13 @@ public TransactionSignWeight getTransactionSignWeight(Transaction trx) {
         tswBuilder.setPermission(permission);
         if (trx.getSignatureCount() > 0) {
           List<ByteString> approveList = new ArrayList<>();
+          // Read-only introspection: pass osakaAllowed=false so a historical tx
```
SeriousCoding789: Modify the comments `osakaAllow` to `checkMaxSignatureSize`, just delete`committed pre-Osaka`.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -160,6 +160,7 @@
 import org.tron.core.capsule.VotesCapsule;
 import org.tron.core.capsule.WitnessCapsule;
 import org.tron.core.capsule.utils.MarketUtils;
+import org.tron.core.config.Parameter.ChainConstant;
```
bladehan1: [should] remove Unused ChainConstant import

Sunny6889 (approved):

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -505,6 +505,16 @@ public GrpcAPI.Return broadcastTransaction(Transaction signedTransaction) {
     trx.setTime(System.currentTimeMillis());
     Sha256Hash txID = trx.getTransactionId();
     try {
+      for (ByteString sig : signedTransaction.getSignatureList()) {
+        if (!SignUtils.isValidLength(sig.size(), true)) {
+          String info = "Signature size is " + sig.size();
+          logger.warn("Broadcast transaction {} has failed, {}.", txID, info);
+          return builder.setResult(false).setCode(response_code.SIGERROR)
```
3for, line 512: [DISCUSS] Now, if a node upgrades before VERSION_4_8_2 is activated, the networking layer will immediately and unconditionally reject oversized-signature transactions, while the consensus layer will only start rejecting them after VERSION_4_8_2 is activated. Was this inconsistency between the networking-layer and consensus-layer rejection conditions intentional by design?

Federico2014: Yes — this is intentional. The asymmetry follows from what each path actually sees:

- **Consensus / block / PBFT validation** must be able to replay history. Some historical on-chain signatures carry trailing padding within `[66, 68]` bytes that ECDSA recovery silently ignores. If we unconditionally enforced `[65, 68]` on these paths, a node that upgraded before activation would reject otherwise-valid historical blocks and stall. That's why the upper bound is gated by `ForkController.signatureMaxSizeChecked()` (i.e. `VERSION_4_8_2` fork stats), so the network reaches the new rule together.
- **Admission paths** (`Wallet.broadcastTransaction`, `TransactionsMsgHandler.check`, `RelayService.checkHelloMessage`) only see freshly-produced signatures from clients and peers — never historical chain state. A client emitting a 69-byte signature today is misbehaving regardless of fork state, so there is no replay risk in rejecting it at the edge.

Net effect on an upgraded-but-not-yet-activated node: it stops accepting freshly-broadcast oversized signatures, but still replays historical blocks correctly. This is documented in the "Compatibility" section of the PR description; happy to fold a shorter version into the code comment near the admission check if that would make the intent more discoverable.

PR #6777, opened 2026-05-18
fix(db): re-verify block signature during fork replay

xxo1shine:
**What does this PR do?**
- Manager.switchFork now re-validates each replayed block's witness signature before applying. The witness account's permission can change between forks (via permission-update transactions), so a signature that was valid on the original chain may no longer be valid on the replay path.
- Use `if (!validateSignature(...)) throw new ValidateSignatureException` rather than discarding the boolean return: validateSignature only throws on malformed signature bytes; an attacker-supplied valid-format signature with a wrong-signer address returns false. Discarding the return would let that attack through.
- The existing switchFork catch list already includes ValidateSignatureException, so the new throw is wired into the existing switchback path with no additional handling.
- Add three BlockCapsule.validateSignature contract tests pinning the two failure modes the fix relies on: signer-mismatch returns false, signer-match returns true, and a 65-byte malformed signature throws ValidateSignatureException.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1140,6 +1140,11 @@ public class Manager {
         Exception exception = null;
         // todo  process the exception carefully later
         try (ISession tmpSession = revokingStore.buildSession()) {
+          if (!item.getBlk().validateSignature(
+              getDynamicPropertiesStore(), getAccountStore())) {
+            throw new ValidateSignatureException(
+                "switch fork: block " + item.getBlk().getNum() + " signature invalid");
+          }
           applyBlock(item.getBlk().setSwitch(true));
           tmpSession.commit();
         } catch (AccountResourceInsufficientException
--- a/framework/src/test/java/org/tron/core/db/ManagerMockTest.java
+++ b/framework/src/test/java/org/tron/core/db/ManagerMockTest.java
@@ -5,12 +5,14 @@ import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.mockConstruction;
 import static org.mockito.Mockito.mockStatic;
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import com.google.protobuf.Any;
@@ -22,6 +24,7 @@ import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedList;
 import java.util.List;
 
 import lombok.SneakyThrows;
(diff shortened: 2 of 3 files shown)


framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -1140,6 +1140,11 @@ private void switchFork(BlockCapsule newHead)
         Exception exception = null;
         // todo  process the exception carefully later
         try (ISession tmpSession = revokingStore.buildSession()) {
+          if (!item.getBlk().validateSignature(
```
waynercheung, line 1143: [MUST] In normal chain state a witness account should exist, but fork replay re-validates the block under a different state from the one used when the block was first accepted into KhaosDB. If the witness account/permission only existed on the old branch and is gone after rollback to the common ancestor, `accountStore.get(witnessAccountAddress)` can return null and `validateSignature()` throws NPE. This unchecked exception skips the existing switchback path. Please convert this case to `false` / `ValidateSignatureException` or otherwise route it through the same rollback handling.

framework/src/test/java/org/tron/core/db/ManagerMockTest.java
```diff
@@ -564,4 +569,157 @@ public void testPostContractTriggerSwallowsThrowable() throws Exception {
...
+    when(khaosDb.getBranch(any(BlockCapsule.BlockId.class), any(Sha256Hash.class)))
+        .thenReturn(new Pair<>(first, value));
+
+    Method switchFork = Manager.class.getDeclaredMethod("switchFork", BlockCapsule.class);
+    switchFork.setAccessible(true);
+
+    // The throw fires before the finally's switchback runs. Switchback's applyBlock
+    // may surface another exception due to partial mocks; we tolerate any throwable
+    // here because the new code's throw has already been executed (line covered).
+    try {
```
waynercheung, line 637: [MUST] This test still passes if production code calls `validateSignature()` but ignores the `false` return value, because it catches any `Throwable` and only verifies the method was invoked. Please assert the reflected invocation fails specifically with `ValidateSignatureException`, and verify `setSwitch(true)` / apply path is not reached for the invalid block.

framework/src/main/java/org/tron/core/db/Manager.java
xxo1shine: 1. When a block arrives from the network, it first goes through `TronNetDelegate.validBlock` — which already invokes `block.validateSignature(dynamicPropertiesStore, accountStore)` once.
2. If `accountStore.get(witnessAccountAddress)` had returned null at that point, that `validateSignature` call would return false due to address mismatch before any NullPointerException could fire, and `validBlock` would raise it as `BLOCK_SIGN_INVALID` and reject the block — so it could never have entered KhaosDB, and would never reach `switchFork`.

framework/src/test/java/org/tron/core/db/ManagerMockTest.java
xxo1shine: Thanks for the heads-up. This was considered — the boolean return contract of `validateSignature` is already pinned independently in `BlockCapsuleTest` by three contract tests; this `ManagerMockTest` test is here to pin that the switchFork replay path actually reaches `validateSignature`. The two layers together already cover the regression. Adding a cause assertion plus `verify(never()).setSwitch(true)` here would couple the test to the implementation detail that `applyBlock` calls `setSwitch` via its argument expression — a future refactor would break this test for the wrong reason. Prefer to keep the current form.

halibobo1205 (approved):

waynercheung (approved):
LGTM

PR #6778, opened 2026-05-19
fix(tvm): canonicalize ModExp zero modulus output (TIP-871)

yanghang8612:
## Summary

- gate MODEXP zero-modulus output canonicalization behind `ALLOW_TVM_OSAKA`
- reset CREATE2 call return-data state behind the same Osaka gate
- add direct Osaka MODEXP tests plus a Solidity 0.8.25 VM-level reproduction for MODEXP and CREATE2 observable behavior

## Motivation

Related to tronprotocol/tips#871. MODEXP currently returns empty output for an all-zero modulus even when `modLen > 0`; after Osaka it should return `modLen` zero bytes. CREATE2 also needs to match CREATE by clearing stale return-data state before execution once the Osaka behavior is active.

## Testing

```bash
./gradlew :framework:test --tests org.tron.common.runtime.vm.AllowTvmOsakaTest --tests org.tron.common.runtime.vm.TvmIssueVerifierTest
```


yanghang8612:
Closing to reopen a clean PR with the finalized title.

PR #6782, opened 2026-05-19
fix(net): restrict admission signature length

Federico2014:
**What does this PR do?**

Adds a signature-length range check at broadcast / P2P admission entrypoints, before any signature recovery work is done. Accepted sizes are `[PER_SIGN_LENGTH, MAX_PER_SIGN_LENGTH]` = `[65, 68]`:

- `Wallet.broadcastTransaction` rejects the transaction with `SIGERROR` when any signature in the list falls outside `[65, 68]`.
- `TransactionsMsgHandler.check` throws `P2pException(BAD_TRX)` when any signature in a `TransactionsMessage` falls outside `[65, 68]`, causing the peer to be disconnected with `BAD_TX`.
- `RelayService.checkHelloMessage` returns `false` when the `HelloMessage` signature from a fast-forward peer falls outside `[65, 68]`.

The predicate is centralized in `SignUtils.isValidLength(int size)`; all three call sites now share that helper. The bounds live in `Constant.PER_SIGN_LENGTH` (= 65) and `Constant.MAX_PER_SIGN_LENGTH` (= 68).

A canonical TRON ECDSA signature is encoded as 65 bytes (`r` 32 + `s` 32 + `v` 1). The affected entrypoints currently pass malformed admission payloads into `SignUtils.signatureToAddress` / signature recovery first, which wastes CPU before the request is rejected or fails later.

This PR adds a cheap length check at the network and broadcast boundaries so invalid admission payloads fail before crypto recovery is attempted. This is intentionally scoped to admission handling only.

**Why `[65, 68]` instead of strict `== 65`?**

A canonical signature is 65 bytes. Historical on-chain data, however, contains transactions with a small amount of trailing padding (up to 3 bytes) that ECDSA recovery silently ignores. Using `MAX_PER_SIGN_LENGTH = 68` closes the previously unbounded encoding window at admission while remaining tolerant of clients that may still produce the legacy padded form.

The long-term goal is to tighten the rule once the compatibility window is no longer needed. Because the bounds are centralized in `Constant` and the predicate is centralized in `SignUtils.isValidLength`, that tightening is a one-line change.

**Consensus compatibility**

This PR does not change transaction validation for blocks or any shared consensus validation path.

In particular, `TransactionCapsule.checkWeight` keeps the looser `sig.size() < 65` check so historical signatures with trailing padding remain valid; longer signatures continue to be parsed by the existing `Rsv.fromSignature` behavior using the first 65 bytes. Tightening that shared validation path would be a consensus-impacting behavior change and is intentionally not part of this PR.

To keep this rule visible to future readers, `SignUtils.isValidLength` carries a javadoc that names the admission-vs-consensus split and points at `TransactionCapsule.checkWeight` so the next reviewer who greps either operator lands on the explanation.

The observable behavior change is limited to local broadcast / P2P admission:

- Clients broadcasting transactions with signatures outside `[65, 68]` now receive `SIGERROR` before downstream checks.
- Peers sending transaction messages with signatures outside `[65, 68]` are rejected at message admission with `BAD_TX`.
- Fast-forward hello messages with signatures outside `[65, 68]` are rejected before signature recovery.

- Unit Tests
  - `WalletMockTest#testBroadcastTxInvalidSigLength`: 64 / 69 / empty signatures return `SIGERROR`; 65-byte and 68-byte (upper bound) signatures fall through.
  - `TransactionsMsgHandlerTest#testInvalidSigLength`: 64 / 69 signatures raise `P2pException(BAD_TRX)` via `assertThrows`; 65-byte and 68-byte signatures pass the new length check.
  - `RelayServiceTest#testCheckHelloMessage`: extended to assert that 64 / 69 / empty `HelloMessage` signatures return `false`; the existing 65-byte case still returns `true`.
- Manual Testing
  - `./gradlew :framework:checkstyleMain :framework:checkstyleTest`
  - `./gradlew :framework:test --tests "org.tron.core.WalletMockTest.testBroadcastTxInvalidSigLength" --tests "org.tron.core.net.messagehandler.TransactionsMsgHandlerTest.testInvalidSigLength" --tests "org.tron.core.net.services.RelayServiceTest.testCheckHelloMessage"`

**Follow up**

None.

### DIFF
--- a/common/src/main/java/org/tron/core/Constant.java
+++ b/common/src/main/java/org/tron/core/Constant.java
@@ -19,7 +19,8 @@ public class Constant {
   public static final long MAXIMUM_TIME_UNTIL_EXPIRATION = 24 * 60 * 60 * 1_000L; //one day
   public static final long TRANSACTION_DEFAULT_EXPIRATION_TIME = 60 * 1_000L; //60 seconds
   public static final long TRANSACTION_FEE_POOL_PERIOD = 1; //1 blocks
-  public static final long PER_SIGN_LENGTH = 65L;
+  public static final int PER_SIGN_LENGTH = 65;
+  public static final int MAX_PER_SIGN_LENGTH = 68;
   public static final long MAX_CONTRACT_RESULT_SIZE = 2L;
 
   // Smart contract / Energy
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -24,7 +24,6 @@ import static org.tron.common.math.Maths.max;
 import static org.tron.common.utils.Commons.getAssetIssueStoreFinal;
 import static org.tron.common.utils.Commons.getExchangeStoreFinal;
 import static org.tron.common.utils.WalletUtil.isConstant;
-import static org.tron.core.Constant.PER_SIGN_LENGTH;
 import static org.tron.core.capsule.utils.TransactionUtil.buildInternalTransaction;
 import static org.tron.core.config.Parameter.ChainConstant.BLOCK_PRODUCED_INTERVAL;
 import static org.tron.core.config.Parameter.ChainConstant.TRX_PRECISION;
@@ -507,7 +506,7 @@ public class Wallet {
     Sha256Hash txID = trx.getTransactionId();
     try {
       for (ByteString sig : signedTransaction.getSignatureList()) {
-        if (sig.size() != PER_SIGN_LENGTH) {
+        if (!SignUtils.isValidLength(sig.size())) {
           String info = "Signature size is " + sig.size();
           logger.warn("Broadcast transaction {} has failed, {}.", txID, info);
           return builder.setResult(false).setCode(response_code.SIGERROR)
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -24,6 +24,7 @@ import static org.tron.common.math.Maths.max;
 import static org.tron.common.utils.Commons.getAssetIssueStoreFinal;
 import static org.tron.common.utils.Commons.getExchangeStoreFinal;
 import static org.tron.common.utils.WalletUtil.isConstant;
+import static org.tron.core.Constant.PER_SIGN_LENGTH;
 import static org.tron.core.capsule.utils.TransactionUtil.buildInternalTransaction;
 import static org.tron.core.config.Parameter.ChainConstant.BLOCK_PRODUCED_INTERVAL;
 import static org.tron.core.config.Parameter.ChainConstant.TRX_PRECISION;
@@ -505,6 +506,16 @@ public class Wallet {
     trx.setTime(System.currentTimeMillis());
     Sha256Hash txID = trx.getTransactionId();
     try {
+      for (ByteString sig : signedTransaction.getSignatureList()) {
+        if (sig.size() != PER_SIGN_LENGTH) {
+          String info = "Signature size is " + sig.size();
+          logger.warn("Broadcast transaction {} has failed, {}.", txID, info);
+          return builder.setResult(false).setCode(response_code.SIGERROR)
+              .setMessage(ByteString.copyFromUtf8("Validate signature error: " + info))
+              .build();
+        }
+      }
+
       if (tronNetDelegate.isBlockUnsolidified()) {
         logger.warn("Broadcast transaction {} has failed, block unsolidified.", txID);
         return builder.setResult(false).setCode(response_code.BLOCK_UNSOLIDIFIED)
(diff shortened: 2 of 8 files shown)


xxo1shine (approved):

SeriousCoding789:
Small suggestion: all three new callsites — `Wallet.broadcastTransaction`, `TransactionsMsgHandler.check`, `RelayService.checkHelloMessage` — are doing the same `sig.size() != PER_SIGN_LENGTH` length check, only differing in the error envelope they wrap around it. Could we factor the predicate into a single helper, e.g. `SignUtils.isStrictLength(int size)`?

Each callsite would then read:

```java
if (!SignUtils.isStrictLength(sig.size())) {
  // site-specific error: SIGERROR / BAD_TRX / return false
}
```

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -505,6 +506,16 @@ public GrpcAPI.Return broadcastTransaction(Transaction signedTransaction) {
     trx.setTime(System.currentTimeMillis());
     Sha256Hash txID = trx.getTransactionId();
     try {
+      for (ByteString sig : signedTransaction.getSignatureList()) {
+        if (sig.size() != PER_SIGN_LENGTH) {
```
bladehan1: [SHOULD] Encode the `!= 65` (admission) vs `< 65` (consensus) rationale in the code, not just the PR description

The PR description's "Consensus compatibility" section explains nicely why admission entry-points use `!= PER_SIGN_LENGTH` (strict) while `TransactionCapsule.checkWeight` keeps `sig.size() < 65` (lenient, to preserve consensus behavior). That rationale is currently invisible to anyone reading the code: the next reviewer who greps `PER_SIGN_LENGTH` or `sig.size()` will see two different operators and ask the same question again.

Suggestion: add a short javadoc on `Constant.PER_SIGN_LENGTH` stating "used by admission entry-points with strict `!=` matching; `TransactionCapsule.checkWeight` keeps `< 65` for consensus backward compatibility", or leave an equivalent one-line comment at each of the three new check sites. One central location is preferred.

bladehan1 (commented):
reviewed

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -505,6 +505,16 @@ public GrpcAPI.Return broadcastTransaction(Transaction signedTransaction) {
     trx.setTime(System.currentTimeMillis());
     Sha256Hash txID = trx.getTransactionId();
     try {
+      for (ByteString sig : signedTransaction.getSignatureList()) {
+        if (!SignUtils.isValidLength(sig.size())) {
```
3for, line 509: [SHOULD] The `for` loop is skipped entirely when `getSignatureList()` is empty, which means “zero-signature transactions” are not blocked by this gate. `checkWeight` will handle it later, but since the PR description positions this as an admission gate, it may make sense to add an explicit `signatureCount > 0` check outside the loop as well, to align with the semantics of rejecting unsigned transactions.

common/src/main/java/org/tron/core/Constant.java
```diff
@@ -20,6 +20,7 @@ public class Constant {
   public static final long TRANSACTION_DEFAULT_EXPIRATION_TIME = 60 * 1_000L; //60 seconds
   public static final long TRANSACTION_FEE_POOL_PERIOD = 1; //1 blocks
   public static final long PER_SIGN_LENGTH = 65L;
+  public static final long MAX_PER_SIGN_LENGTH = 68L;
```
317787106: [SHOULD] These two variables should be `int`.

framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
```diff
@@ -337,6 +337,91 @@ public void testDuplicateTransactionRejected() throws Exception {
     }
   }
 
+  @Test
+  public void testInvalidSigLength() throws Exception {
+    TransactionsMsgHandler handler = new TransactionsMsgHandler();
+    handler.init();
+    try {
+      PeerConnection peer = Mockito.mock(PeerConnection.class);
+
```
317787106, line 346: [NIT] Add one testcase to verify the transaction with empty-sig.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Thanks for the catch. I'd prefer to keep the scope strictly length-only here — empty signature lists are already rejected downstream by `validateSignature` / `checkWeight` (weight 0 vs. a positive permission threshold), so an explicit `getSignatureCount() == 0` guard at this gate would be redundant.

The "admission gate" framing in the PR description is specifically about closing the unbounded length-encoding window before recovery; the count-validity check already has a single owner downstream and I'd rather not duplicate it.

framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
Federico2014: Thanks — I'd prefer to skip this one. `TransactionsMsgHandler.check` only validates signature *length*, not *count*; empty-list rejection is handled downstream in `validatePubSignature` / `checkWeight`. Adding an empty-sig test here would assert a no-op (the for-loop is trivially skipped), which doesn't exercise behavior that lives in this handler. The empty-sig contract is covered where it's enforced, not here.

317787106 (approved):

3for (approved):

PR #6788, opened 2026-05-20
refactor(config): merge test config files

317787106:
**What does this PR do?**

Consolidates test configuration files by deleting four redundant config files (`args-test.conf`, `config-test-index.conf`, `config-test-mainnet.conf`, `config-test-storagetest.conf`) and standardizing tests to use a single shared `config-test.conf`. Also renames `config-localtest.conf` to `config-shield.conf` to better reflect its purpose (shielded-transfer tests), and removes the hardcoded seed node list from `config-test.conf` since unit tests have no need for P2P connectivity.

Specifically:
- Delete 4 redundant test config files (~940 lines removed)
- Rename `config-localtest.conf` → `config-shield.conf`
- Remove `NET_CONF`, `MAINNET_CONF`, `LOCAL_CONF`, `STORAGE_CONF`, `INDEX_CONF` from `TestConstants`; add `SHIELD_CONF`
- Update all affected test classes to reference `TestConstants.TEST_CONF` or `TestConstants.SHIELD_CONF` instead of bare string literals
- Clear seed node IPs from `config-test.conf`; adjust `ArgsTest` seed-node assertion from 11 → 0. Does not slim the unified `config-test.conf`, Per-test inline overrides via ConfigFactory.parseString are a follow-up.

The previous test suite maintained five separate `.conf` files that largely duplicated each other, making it hard to know which config to use for a new test and causing drift over time (e.g., `config-test-mainnet.conf` contained mainnet genesis/seed data that is irrelevant and potentially misleading in a unit-test context). Centralizing to one default test config (`config-test.conf`) and one purpose-specific config (`config-shield.conf`) reduces maintenance overhead and eliminates accidental coupling to live-network parameters in tests.

- Unit Tests

**Extra details**

`StorageTest` gains additional coverage to compensate for the removal of `config-test-storagetest.conf`-dependent tests that were folded into the unified config.


framework/src/test/java/org/tron/core/config/args/StorageTest.java
```diff
@@ -16,21 +16,124 @@
...
+   *       writeBufferSize = 3,
+   *       cacheSize = 4,
+   *       maxOpenFiles = 5
+   *     },
+   *   ]
+   */
+  private static void setupStorage() {
+    StorageConfig sc = new StorageConfig();
+    try {
+      setPrivateField(sc, "defaultDbOption", makeOverride(50));
```
barbatos2011: The reflection-based `setPrivateField` here bypasses the explicit `@Setter(AccessLevel.NONE)` on `StorageConfig.defaultDbOption` / `defaultMDbOption` / `defaultLDbOption`. Those annotations document an encapsulation contract — those fields are bound by `ConfigBeanFactory` and should be immutable post-binding — which this code silently breaks.

Three concrete downsides:

1. **Refactor-fragile**: the field name is a string literal (`"defaultDbOption"`). A future field rename via IDE will not touch this string, the project still compiles, and the test only fails at runtime with `NoSuchFieldException`.
2. **Integration coverage lost**: the deleted `config-test-storagetest.conf` exercised the full `ConfigFactory → ConfigBeanFactory → @Setter(NONE) fields` path. Bypassing it with reflection turns this into a unit test of utility functions; a future regression in `ConfigBeanFactory` binding of `@Setter(NONE)` fields would not be caught here.
3. **Anti-pattern propagation**: once `setPrivateField` exists in the test tree, future tests are likely to copy it instead of pursuing cleaner solutions.

**Suggested alternative** — use `ConfigFactory.parseString` to exercise the real binding path with zero reflection and refactor-safety:

```java
private static void setupStorage() {
  Config cfg = ConfigFactory.parseString(
      "storage.defaultDbOption.maxOpenFiles = 50\n" +
      "storage.defaultMDbOption.maxOpenFiles = 500\n" +
      "storage.defaultLDbOption.maxOpenFiles = 1000\n" +
      "storage.properties = [\n" +
      "  { name = account, path = storage_directory_test, maxOpenFiles = 100, ... },\n" +
      "  { name = account-index, path = storage_directory_test, maxOpenFiles = 100, ... },\n" +
      "  { name = test_name, path = test_path, createIfMissing = false, blockSize = 2, ... }\n" +
      "]")
      .withFallback(ConfigFactory.defaultReference());
  StorageConfig sc = StorageConfig.fromConfig(cfg);
  storage.setDefaultDbOptions(sc);
  storage.setPropertyMapFromBean(sc.getProperties());
}
```

If `@Setter(NONE)` truly needs to be relaxed for testability, prefer `@Setter(AccessLevel.PACKAGE)` or an explicit `@VisibleForTesting` setter on `StorageConfig` — both are type-safe and document the intent.

framework/src/test/java/org/tron/core/config/args/ArgsTest.java
```diff
@@ -50,8 +50,7 @@ public class ArgsTest {
 
   @Test
   public void get() {
-    Args.setParam(new String[] {"-c", TestConstants.TEST_CONF, "--keystore-factory"},
-        TestConstants.NET_CONF);
+    Args.setParam(new String[] {"-c", TestConstants.TEST_CONF, "--keystore-factory"}, "");
```
barbatos2011: Passing `""` as the second argument here is a code smell — `Configuration.getByFileName` explicitly rejects blank input with `IllegalArgumentException("Configuration path is required!")`, so an empty string is documented as invalid.

The test currently works because the `-c TestConstants.TEST_CONF` flag short-circuits the fallback path (in `Args.setParam` line ~166: `isNoneBlank(shellConfFileName) ? shellConfFileName : confFileName`). But this leaves a fragile dependency: if a future refactor removes the `-c` flag from this call site (e.g. switching to a different mechanism for selecting the test config), the empty-string second arg silently flows into `getByFileName` and the test fails with an opaque `IllegalArgumentException`.

Two safer options:

1. **Keep TestConstants.NET_CONF as the fallback constant** (re-add it, semantically harmless since the body of the new `config-test.conf` covers what NET_CONF used to point at).
2. **Add an `Args.setParam(String[])` overload** that resolves `confFileName` internally (or accepts `Optional<String>`) — eliminates the misleading placeholder pattern across the test suite.

Not blocking; just worth tracking before this pattern gets copy-pasted into other tests.

barbatos2011:
Follow-up observation, not blocking: although this PR deletes four redundant config files (-996 lines, well-deserved cleanup), the consolidated `config-test.conf` itself drops only **18 lines** (387 → 369), and the renamed `config-shield.conf` drops 17. So the consolidation eliminates **file drift risk** but doesn't reduce **per-test config-load surface** — each of the ~21 affected tests still loads a 369-line conf, mostly boilerplate it doesn't read.

That's a real, but smaller, win. Worth being explicit in the PR description:

> Note: this PR consolidates files but does not slim the unified `config-test.conf` (~369 lines) itself. Per-test inline overrides via `ConfigFactory.parseString` are a follow-up.

Otherwise reviewers (and future contributors) may expect the consolidation step alone to address the boilerplate problem, when in fact the boilerplate just got concentrated in one place.

Concrete follow-up path, if someone wants to pursue it later: tests with narrow needs (e.g. `AssetUpdateHelperTest`, `StorageTest`, `ApiUtilTest`) can move to inline `ConfigFactory.parseString(...).withFallback(ConfigFactory.defaultReference())` and shed their dependency on `TEST_CONF` entirely. `reference.conf` already provides the safety net thanks to PR #6615.

framework/src/test/java/org/tron/core/config/args/StorageTest.java
317787106: Good suggestion, use the real config string instead.

framework/src/test/java/org/tron/core/config/args/ArgsTest.java
317787106: Thanks for your suggestion, use the standard method instead although it doesn't matter much:
```
Args.setParam(new String[] {"--keystore-factory"}, TestConstants.TEST_CONF);
```

framework/src/test/resources/config-shield.conf
```diff
@@ -169,18 +169,7 @@ node {
 
 
 seed.node = {
```
halibobo1205, line 164: [NIT] docs still mention deleted test resource config-localtest.conf in implement-a-customized-actuator-{en/cn}.md

317787106: Delete it.

halibobo1205 (approved):

halibobo1205 (approved):

bladehan1 (approved):

PR #6792, opened 2026-05-20
feat(ci): enforce lowerCamelCase and max depth in reference.conf

bladehan1:
**What does this PR do?**

Adds a CI gate that validates `common/src/main/resources/reference.conf` on every PR / push:
- Every dot-separated segment of every key must match `^[a-z][a-zA-Z0-9]*$`: starts with a lowercase ASCII letter, then ASCII letters/digits only. Acronym runs after position 1 (e.g. `httpPBFTEnable`, `openHistoryQueryWhenLiteFN`, `allowShieldedTRC20Transaction`) are accepted — only the first character is constrained, which matches what `java.beans.Introspector` / `ConfigBeanFactory` actually require for bean-property auto-binding.
- Total path depth must be ≤ 5 (set exactly at the current max in tree — no buffer; new deeper keys must explicitly bump `MAX_DEPTH` via a reviewed change).
- Seven legacy keys are grandfathered via an explicit ALLOWLIST (4 PBFT + 3 `node.shutdown.*`).
- Array steps count toward depth (each `[]` = +1 level); nested arrays (HOCON list-of-list) are supported.

Implementation: `.github/scripts/check_reference_conf.py` using `pyhocon` (pinned via `.github/scripts/requirements.txt`), wired into the existing `checkstyle` job in `.github/workflows/pr-check.yml`. `actions/setup-python` enables pip caching keyed on the requirements file. Violations produce per-line stdout output plus one consolidated `::error file=...,title=reference.conf::...` GHA annotation so failures show up in the PR check summary.

`reference.conf` is the single source of default values for every config key in java-tron, and key names must auto-bind to `XxxConfig.java` bean fields via Typesafe Config's `ConfigBeanFactory`. This gate is stricter than typical Java/Spring projects (Spring Boot's `Binder` / Quarkus / Micronaut tolerate kebab-case, snake_case, UPPER_CASE, and camelCase as equivalent forms) because `ConfigBeanFactory` uses `java.beans.Introspector` and requires the HOCON key to match the JavaBean property name exactly — no case normalization. Without a CI gate, non-conforming keys silently fail to bind and accumulate. Today the file has drifted: 4 PBFT keys violate the rule and the file header documents legacy exceptions handled via manual normalization.

A depth ceiling also prevents deeply nested hierarchies from creeping in. The gate is set exactly at the current max (5) with no buffer: silent drift is unacceptable for a mature project, so any new key that needs to go deeper has to bump `MAX_DEPTH` explicitly and be reviewed.

- Manual Testing: ran the script against the real `reference.conf` — passes with 294 keys, max depth 5.
- Built a local fixture suite (not committed) covering: empty / simple / dotted-form / at-max-depth / object array / scalar array / allowlist / format violations (Pascal / snake / digit / array-internal / non-allowlisted acronym) / depth violations (dotted + in-array) / combined + dedup invariant / env errors (parse failure, missing file). 20 assertions PASS.
- Verified: parse failure (invalid HOCON) and missing file both exit 2 with `::error` annotation; a single user-declared deep key produces exactly one depth violation line (leaf-only filtering).

**Follow up**

- Sunset legacy ALLOWLIST entries via a rename + deprecation cycle (out of scope here — would break existing user `config.conf` files).
- Bean-field ↔ key parity + default-value drift gate (Gate 2) is the highest-leverage follow-up — would have caught the dead keys removed in #6794 and the default drift fixed in #6755 / #6790. Scope and implementation (`Class.getDeclaredFields()` reflection vs Java-AST parsing; default-value comparison rules across types; what counts as a config bean) warrant a separate PR.

**Extra details**

- Allowlist exempts 7 legacy keys to keep user `config.conf` compatible: `node.http.PBFTEnable`, `node.http.PBFTPort`, `node.rpc.PBFTEnable`, `node.rpc.PBFTPort`, plus `node.shutdown.BlockTime/BlockHeight/BlockCount` (PascalCase exceptions handled manually in `NodeConfig.fromConfig`; currently commented out in `reference.conf` — pre-listed so the gate stays green if ever uncommented with defaults).
- Library choice: `pyhocon` over Typesafe Config because the gate runs before Gradle/JDK setup in CI (~3 steps: setup-python + pip install + invoke). Both implement the HOCON spec; outputs are equivalent.


.github/workflows/pr-check.yml
```diff
@@ -103,6 +103,20 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
```
warku123, line 112: [SHOULD] Consider pinning the `pyhocon` version and enabling pip caching, so CI is not exposed to upstream releases and avoids re-downloading on every PR. Suggested change:

```yaml
      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: '3.11'
          cache: 'pip'
          cache-dependency-path: .github/scripts/requirements.txt

      - name: Install pyhocon
        run: pip install --quiet -r .github/scripts/requirements.txt
```

And a new `.github/scripts/requirements.txt`:

```
pyhocon==0.3.61
```

Rationale: (1) a future pyhocon release that changes API or behavior would silently break unrelated PRs on this step; (2) `cache: 'pip'` saves the install round-trip on every PR once warm; (3) a requirements file also makes it trivial to reproduce the CI environment locally via `pip install -r`.

bladehan1: Agree,Thanks for the suggestion!

PR #6797, opened 2026-05-22
feat(net): normalize inbound messages

halibobo1205:
### Summary

Add `sanitize()` to `BlockCapsule` and `TransactionCapsule` that normalizes the protobuf representation by stripping outer unknown fields, leaving all consensus-hashed and signed regions byte-identical.

Invocation:
- `BlockMsgHandler.processMessage` calls `blockMessage.sanitize()` after the existing size/time checks and before sync/adv dispatch. `BlockMessage.sanitize()` updates both the capsule and the message's wire `data` field (if it contains unknown fields).
- `Manager.processTransaction` calls `trxCap.sanitize()` after signature validation, only for mempool transactions (`!trxCap.isInBlock()`).
- `Wallet.broadcastTransaction` constructs the outbound `TransactionMessage` after `pushTransaction` so it reflects the post-sanitize state.

Each `sanitize()` short-circuits when there are no unknown fields, so the common path is allocation-free.

### Compatibility

- Consensus state is unchanged.
- Existing size checks, error paths, response codes, and log lines are untouched.
- No consensus impact.

### DIFF
--- a/framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java
+++ b/framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java
@@ -29,8 +29,9 @@ public class BlockMessage extends TronMessage {
   }
 
   public void sanitize() {
-    this.block.sanitize();
-    this.data = this.block.getData();
+    if (this.block.sanitize()) {
+      this.data = this.block.getData();
+    }
   }
 
   public BlockId getBlockId() {
--- a/framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java
+++ b/framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java
@@ -28,6 +28,11 @@ public class BlockMessage extends TronMessage {
     this.block = block;
   }
 
+  public void sanitize() {
+    this.block.sanitize();
+    this.data = this.block.getData();
+  }
+
   public BlockId getBlockId() {
     return getBlockCapsule().getBlockId();
   }
--- a/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
@@ -329,11 +329,11 @@ public class BlockCapsule implements ProtoCapsule<Block> {
     return !getInstance().getBlockHeader().getWitnessSignature().isEmpty();
   }
 
-  public void sanitize() {
+  public boolean sanitize() {
     boolean blockHasUnknown = !this.block.getUnknownFields().asMap().isEmpty();
     boolean headerHasUnknown = !this.block.getBlockHeader().getUnknownFields().asMap().isEmpty();
     if (!blockHasUnknown && !headerHasUnknown) {
-      return;
+      return false;
     }
     UnknownFieldSet empty = UnknownFieldSet.getDefaultInstance();
     Block.Builder builder = this.block.toBuilder();
@@ -346,6 +346,7 @@ public class BlockCapsule implements ProtoCapsule<Block> {
           .build());
     }
     this.block = builder.build();
+    return true;
   }
 
   @Override
--- a/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/BlockCapsule.java
@@ -21,6 +21,7 @@ import com.google.common.primitives.Longs;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.CodedInputStream;
 import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.UnknownFieldSet;
 import java.security.SignatureException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -326,6 +327,25 @@ public class BlockCapsule implements ProtoCapsule<Block> {
 
   public boolean hasWitnessSignature() {
     return !getInstance().getBlockHeader().getWitnessSignature().isEmpty();
+  }
+
+  public void sanitize() {
+    boolean blockHasUnknown = !this.block.getUnknownFields().asMap().isEmpty();
+    boolean headerHasUnknown = !this.block.getBlockHeader().getUnknownFields().asMap().isEmpty();
+    if (!blockHasUnknown && !headerHasUnknown) {
+      return;
+    }
+    UnknownFieldSet empty = UnknownFieldSet.getDefaultInstance();
+    Block.Builder builder = this.block.toBuilder();
+    if (blockHasUnknown) {
+      builder.setUnknownFields(empty);
+    }
+    if (headerHasUnknown) {
+      builder.setBlockHeader(this.block.getBlockHeader().toBuilder()
+          .setUnknownFields(empty)
+          .build());
+    }
+    this.block = builder.build();
   }
 
   @Override
--- a/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
@@ -28,6 +28,7 @@ import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.GeneratedMessageV3;
 import com.google.protobuf.Internal;
 import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.UnknownFieldSet;
 import java.io.IOException;
 import java.security.SignatureException;
 import java.util.ArrayList;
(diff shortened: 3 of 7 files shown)


framework/src/main/java/org/tron/core/net/message/adv/TransactionMessage.java
```diff
@@ -28,6 +40,39 @@ public TransactionMessage(Transaction trx) {
...
 
+  /**
+   * Mirror Manager.validateCommon's size check. Static helper, also invoked
+   * from {@link TransactionsMessage} for each tx in a batch.
+   */
+  static void checkSize(Transaction transaction) throws P2pException {
+    TransactionCapsule capsule = new TransactionCapsule(transaction);
+    capsule.removeRedundantRet();
+    long generalBytesSize = capsule.getInstance().toBuilder().clearRet().build().getSerializedSize()
+        + Constant.MAX_RESULT_SIZE_IN_TX + Constant.MAX_RESULT_SIZE_IN_TX;
```
317787106: [Question] Why do you add `Constant.MAX_RESULT_SIZE_IN_TX` twice ? If needed, please specify the reason.

framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java
```diff
@@ -1,23 +1,44 @@
...
+  // previously enforced in BlockMsgHandler (BLOCK_SIZE + ONE_THOUSAND).
+  static final int MAX_RAW_SIZE = BLOCK_SIZE + Constant.ONE_THOUSAND;
+
   private BlockCapsule block;
 
   public BlockMessage(byte[] data) throws Exception {
-    super(data);
+    if (data.length > MAX_RAW_SIZE) {
+      throw new P2pException(TypeEnum.BAD_MESSAGE, "block size over limit");
+    }
```
317787106: Observability regression in BlockMsgHandler:
  
  The removed block size check had a detailed log:
```
  // removed:
  logger.error("Receive bad block {} from peer {}, block size over limit",
      blockMessage.getBlockId(), peer.getInetSocketAddress());
 ```     
  The new check fires in the constructor before `blockMessage` is constructed, so `processException` logs `msg = null`. Operators lose the block id and peer address in the error path. Consider logging the raw byte count or peer info from the handler even when the message object couldn't be constructed.

File: framework/src/main/java/org/tron/core/net/message/adv/TransactionMessage.java (hunk 1 above)
Comment (halibobo1205) [reply framework/src/main/java/org/tron/core/net/message/adv/TransactionMessage.java]: <p>The Manager check has to stay because it also guards a path that the new network-layer check does <strong>not</strong> cover: every transaction carried inside an inbound <code>BlockMessage</code>. <code>BlockMessage(byte[])</code> sanitizes the <code>Block</code> / <code>BlockHeader</code> outer layer but does not re-decode each nested transaction through <code>TransactionMessage(byte[])</code> — those txs reach <code>Manager.validateCommon</code> directly via <code>pushBlock → processTransaction</code>, and the size cap there is the only safeguard for them.</p>
<p>BTW, if we removed the Manager check, a post-PR node and a pre-PR node could disagree on the same block: a block carrying an oversize tx would be rejected by the pre-PR node (Manager rejects, block fails) and accepted by the post-PR node (no in-block size check left), producing a consensus split during rolling upgrade.</p>
<p>So the two checks are complementary rather than redundant:</p>
<ul>
<li>The new constructor-level check is an early-reject fast path for the mempool entry points (<code>TransactionMessage</code> / <code>TransactionsMessage</code>, plus <code>Wallet.broadcastTransaction</code>'s internal <code>new TransactionMessage(byte[])</code>).</li>
<li>The Manager-level check is the consensus-layer guarantee that every tx, including those that arrive inside a block, is bounded by <code>TRANSACTION_MAX_BYTE_SIZE</code>.</li>
</ul>

File: framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java (hunk 2 above)
Comment (halibobo1205) [reply framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java]: Thanks for raising this — let me lay out the trade-off explicitly so we can pick the path that matches the project's preference.

 **Current state (intentional)**

 The size check fires before `Block.parseFrom` so we never allocate / decode multi-MB payloads we already know exceed the cap. Because the throw happens before `TronMessageFactory.create(packed)` finishes returning, `msg` in `processException(peer, msg, e)` is `null`. To clarify one detail in the original comment: `peer.getInetSocketAddress()` itself **is still logged** — `P2pEventHandlerImpl` holds the `peer` reference directly. What we lose is:

 - the block id (cannot be computed pre-parse), and
 - the `msg` placeholder, which now renders as `null` in the error line

 Exception detail (`"block size over limit"`) is intentionally identical to the original handler string for grep-compat.

 **Two improvement options, each with a downside**

 1. **Enrich the exception detail with the actual byte count and threshold**, e.g.:

    ```java
    throw new P2pException(TypeEnum.BAD_MESSAGE,
        "block size " + data.length + " over limit " + MAX_RAW_SIZE);
    ```

    Pro: operators see the actual size and the cap in `processException`'s log.
    Con: the exact string `"block size over limit"` is no longer present verbatim — operators relying on an exact-phrase grep need to switch to a substring match (`"block size"` / `"over limit"` still match). Block id remains unavailable (still pre-parse).

 2. **Move the size check back to `BlockMsgHandler.processMessage`** (post-parse) and restore the original `logger.error("Receive bad block {} from peer {}, block size over limit", blockId, peer)`.

>   Pro: log line, exception detail and exception type all stay byte-identical to the previous release; block id and peer are both in the dedicated log.
    Con: we lose the early-reject benefit — every oversize payload pays a full `Block.parseFrom` + `toBuilder/build` pass before being rejected. Not a correctness issue (libp2p frame cap bounds the absolute worst case), but it does violate the "reject as early as possible" principle and adds avoidable CPU / GC pressure on the rejection path.

 Happy to take either direction — could you share which trade-off the project prefers? My slight lean is option 1 (richer detail, keep early-reject), but I want to defer to the maintainer convention on log format compatibility.

Comment (317787106) [reply framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java]: Option 1 is more simple.

File: framework/src/main/java/org/tron/core/net/message/adv/TransactionMessage.java (hunk 1 above)
Comment (317787106) [reply framework/src/main/java/org/tron/core/net/message/adv/TransactionMessage.java]: Get it. Suggest to add comment like this "keep the same check of transaction size as validateCommon".

File: framework/src/main/java/org/tron/core/net/message/adv/BlockMessage.java [hunk 3]
```diff
@@ -28,6 +28,11 @@ public BlockMessage(BlockCapsule block) {
     this.block = block;
   }
 
+  public void sanitize() {
+    this.block.sanitize();
+    this.data = this.block.getData();
+  }
```
317787106, line 35: Good improvement, this is more simple now!  There maybe little performance issue, `BlockMessage.sanitize()` unconditionally re-serializes on every inbound block:
```
  public void sanitize() {
      this.block.sanitize();            // short-circuits and allocates nothing when no unknown fields
      this.data = this.block.getData(); // always calls toByteArray() — even when nothing changed
  }
```
`BlockCapsule.sanitize()` correctly early-returns when there are no unknown fields, so it is allocation-free on the common path. However, `BlockMessage.sanitize()` always overwrites `this.data` by calling `this.block.getData()` (which invokes `toByteArray()` and re-serializes the entire block proto), regardless of whether the capsule was actually modified. Since nearly all inbound blocks in normal traffic carry no unknown fields, this triggers a full protobuf serialization on every received block for no benefit.
  
  Suggested fix — have `BlockCapsule.sanitize()` return a boolean indicating whether any change was made, and let `BlockMessage.sanitize()` skip the assignment on a no-op:
```
  // BlockCapsule.java
  public boolean sanitize() {
      boolean blockHasUnknown = !this.block.getUnknownFields().asMap().isEmpty();
      boolean headerHasUnknown = !this.block.getBlockHeader().getUnknownFields().asMap().isEmpty();
      if (!blockHasUnknown && !headerHasUnknown) {
          return false;
      }
      // ... strip unknown fields ...
      this.block = builder.build();
      return true;
  }

  // BlockMessage.java
  public void sanitize() {
      if (this.block.sanitize()) {
          this.data = this.block.getData();
      }
  }
```
  With this change, the common path (clean block) remains allocation-free end to end, matching the intent described in the PR summary.

halibobo1205: Done with https://github.com/tronprotocol/java-tron/pull/6797/changes/38c52982bae83a1d146f32b0ee245a92a5222f0b

xxo1shine (approved):

317787106 (approved):

lvs0075 (approved):

PR #6800, opened 2026-05-25
fix(vm): write TIP-2935 parent hash after witness permission check

yanghang8612:
**What does this PR do?**

Move the TIP-2935 parent-hash write (`HistoryBlockHashUtil.write`) in `Manager#generateBlock` to after the witness-permission check, so a failed permission check (which returns `null`) never writes the parent hash. `accountStateCallBack.preExecute` keeps its original position before the guard; only the write moves.

`generateBlock` can `return null` from the witness-permission check (when `ALLOW_MULTI_SIGN` is on and the producing key does not match the witness permission address). The TIP-2935 write previously ran before that guard, so on a permission failure the parent hash was still written into the current revoking session. The failure path returns without `session.reset()`, so the stale write survives into the session reused by the next `pushTransaction` (the `!session.valid()` branch is skipped), polluting the pending-transaction preview until the next reset. It never reaches disk, but it is an avoidable dirty write. Moving the write below the guard removes it from the abort path; the write still runs before the tx loop, so SR/validator parity is unchanged. `processBlock`'s write is unaffected — it already runs after `consensus.validBlock(...)`.

- Unit Tests — `HistoryBlockHashIntegrationTest` (16 cases, all passing). `generateBlockWritesParentHashBeforeTxLoop` now snapshots the slot at `executeGenerateFinish` (the last callback before `session.reset()`), since `preExecute` now precedes the write.

**Follow up**

N/A

**Extra details**

Follow-up to #6686 (TIP-2935). One-statement reorder.

### DIFF
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1625,7 +1625,6 @@ public class Manager {
     session.reset();
     session.setValue(revokingStore.buildSession());
 
-    HistoryBlockHashUtil.write(this, blockCapsule);
     accountStateCallBack.preExecute(blockCapsule);
 
     if (getDynamicPropertiesStore().getAllowMultiSign() == 1) {
@@ -1637,6 +1636,8 @@ public class Manager {
         return null;
       }
     }
+
+    HistoryBlockHashUtil.write(this, blockCapsule);
 
     Set<String> accountSet = new HashSet<>();
     AtomicInteger shieldedTransCounts = new AtomicInteger(0);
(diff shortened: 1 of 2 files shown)


CodeNinjaEvan (approved):

liang3885 (approved):

framework/src/main/java/org/tron/core/db/Manager.java
lvs0075, line 1635: Why not simply call `session.reset()` here?

yanghang8612: Our intent here is to keep the change limited to the newly-added TIP-2935 code and not touch the existing logic — so we relocated the `write` below the guard rather than adding `session.reset()` to the pre-existing branch.

aiden3885 (approved):

PR #6801, opened 2026-05-25
fix(log): suppress spurious error logs when solidity node exits

317787106:
### What does this PR do?

- Suppress spurious `ERROR ... reason: null` logs when the solidity node shuts down.

### Root cause

`close()` calls `getBlockExecutor.shutdownNow()` while the `get-block` thread may be blocked in `blockQueue.put()`. That throws `InterruptedException`, whose `getMessage()` returns `null`. The original `catch (Exception e)` logged it as an ERROR with `reason: null`.

The original `sleep()` had the same issue: it caught `Exception` instead of `InterruptedException`, logged a null message, and never restored the interrupt flag.

### Changes

| Location | Change |
|---|---|
| `getBlock()` | Add `catch (InterruptedException e)` — restore interrupt flag, log INFO, return |
| `getBlock()` | Add `!flag` guard in `catch (Exception e)` — log INFO instead of ERROR on shutdown |
| `getBlockByNum()` | Null-check gRPC response to prevent NPE during channel teardown |
| `getLastSolidityBlockNum()` | Add `!flag` guard — same pattern as `getBlock()` |
| `sleep()` | Catch `InterruptedException` specifically and restore interrupt status |

### DIFF
--- a/framework/src/main/java/org/tron/program/SolidityNode.java
+++ b/framework/src/main/java/org/tron/program/SolidityNode.java
@@ -117,7 +117,15 @@ public class SolidityNode implements ApplicationListener<ContextClosedEvent> {
         Block block = getBlockByNum(blockNum);
         blockQueue.put(block);
         blockNum = ID.incrementAndGet();
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        logger.info("getBlock interrupted, exiting.");
+        return;
       } catch (Exception e) {
+        if (!flag) {
+          logger.info("getBlock stopped during shutdown, last block: {}.", blockNum);
+          return;
+        }
         logger.error("Failed to get block {}, reason: {}.", blockNum, e.getMessage());
         sleep(exceptionSleepTime);
       }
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/tron/program/SolidityNode.java
```diff
@@ -167,6 +175,11 @@ private Block getBlockByNum(long blockNum) {
       try {
         long time = System.currentTimeMillis();
         Block block = databaseGrpcClient.getBlock(blockNum);
+        if (block == null) {
```
halibobo1205: Looking at `DatabaseGrpcClient.getBlock()`, it delegates to `databaseBlockingStub.getBlockByNum(...)` — a gRPC blocking stub. Per gRPC semantics, blocking stubs don't return `null`: success returns a response message (the default instance is still a non-null object), and channel teardown/cancellation/deadline expiry surfaces as `StatusRuntimeException` (`CANCELLED` / `UNAVAILABLE` / `DEADLINE_EXCEEDED`), not null.

 A few questions:
 1. Have we observed this null path being hit in production? If so, what's the server-side implementation that produces it?
 2. If this is purely defensive (against mocks or future stub changes), could we add an inline comment? Future readers will spend time looking for "when does this return null."

317787106: You're right, it is defensive. Per gRPC semantics, a blocking stub call never returns null — on channel teardown it throws `StatusRuntimeException` (e.g. CANCELLED), which is already caught by the outer `catch (Exception e)` block and logged as INFO via the `!flag` guard. Removed the unreachable null check.

halibobo1205 (approved):

bladehan1 (approved):
LGTM

PR #6803, opened 2026-05-29
test(config): add reference.conf to bean parity gate

bladehan1:
**What does this PR do?**

Adds a single build-time parity gate between `reference.conf` and the `*Config` beans bound by `ConfigBeanFactory`. Two new files under `common/src/test/java/org/tron/core/config/args/`:

- `ConfigParityGateTest` — JUnit class holding the `Section` registry (path → bean → per-section allowlists) and the `@Test` methods that iterate every registered section.
- `ConfigParityCheck` — shared helpers: recursive HOCON / bean walkers, scalar type dispatcher, allowlist plumbing, and cross-section aggregate logging.

Five sub-gates run for every entry in the registry:

1. `hoconKeysAreBound` — every HOCON key under `<section>.*` has a writable bean property (recursing into nested `*Config` beans) or is on the per-section HOCON-orphan allowlist with an inline rationale.
2. `beanPropertiesHaveHoconKeys` — every writable bean property has a matching HOCON key (recursing the same way) or is on the per-section bean-orphan allowlist.
3. `defaultValuesMatch` — every supported-type bean property's default value equals the `reference.conf` value. Nested `*Config` beans are recursed into; types outside the scalar matrix hard-fail with no allowlist escape.
4. `allowlistEntriesAreLive` — every per-section allowlist entry (`hoconOrphans` / `beanOrphans` / `divergent`) still resolves to a live HOCON path or writable bean property. Catches the rot mode where a key/property is renamed or removed but the grandfathering entry is left behind — without this gate, the rationale comment would silently outlive the underlying anomaly (the long-term failure mode of Cassandra-style `PROPERTIES_TO_IGNORE` blacklists).

A meta-gate (`everyReferenceConfTopLevelKeyIsCovered`) ensures every top-level `reference.conf` key is either covered by a registered `Section` or explicitly listed in `TOP_LEVEL_NON_BEAN` with a rationale (`crypto`, `enery`, `localwitness`, `net`, `seed`, `trx`).

Registered sections: `block`, `committee`, `event.subscribe`, `genesis.block`, `node`, `node.metrics`, `rate.limiter`, `storage`, `vm`.

`ConfigBeanFactory.create(...)` silently drops HOCON keys that have no matching setter, and silently retains bean defaults that diverge from `reference.conf`. A unit-test gate is the cheapest place to catch this drift at PR time rather than at process startup (or worse, in production).

The gate's class Javadoc also documents two naming conventions required for new keys to bind without a `normalizeNonStandardKeys()` hook:

- **Pure lowerCamelCase** — treat acronyms (PBFT, HTTP, JSON, RPC, …) as words: `allowPbft`, not `allowPBFT`; `httpUrl`, not `HTTPURL`. Avoids the "first two characters preserved on decapitalize" trap that produced `pBFTExpireNum`.
- **Avoid `isXxx` boolean keys** — prefer `enableXxx` or `xxxSwitch`. The `is` prefix forces a manual rename hook.

The v4.8.2 baseline (acronym casing, `isOpenFullTcpDisconnect`, `event.subscribe.native`/`topics`) is grandfathered via per-section allowlists with inline rationale; the policy is "do not extend without explicit reason."

- Unit Tests:
  - End-to-end gate — `./gradlew :common:test --tests "*ConfigParityGateTest*"` → all five sub-gates green (`hoconKeysAreBound`, `beanPropertiesHaveHoconKeys`, `defaultValuesMatch`, `allowlistEntriesAreLive`, `everyReferenceConfTopLevelKeyIsCovered`). The `[parity-summary]` block verifies the two coverage invariants:
    - HOCON alignment: `file-hoconKey = checkSection + cantCheckSection`
    - Bean alignment: `registry-beanKey = parity-bean = parity-default`
  - Full `:common:test` — 108 tests across 12 suites green, including all per-section `*ConfigTest` behavioural tests alongside the new parity gate.
  - Helper meta-tests — the four `(label, Config, …)` overloads that the external runner exercised have been removed (no in-repo callers); the scenario inventory is retained as a local reference for future re-introduction if needed.
  - Checkstyle — `config/checkstyle/checkStyleAll.xml` passes on both new files, 0 violations.
- Manual Testing — N/A (test-only change; no runtime code touched).

**Follow up**

- If a future `*Config` bean introduces a Map/enum/Duration/MemorySize field, re-introduce a `Section.typeSkip` field and the corresponding `assertDefaultValuesMatch` parameter following the comment block in `Section`.
- **Inline comment coverage gate on leaf HOCON keys** (planned). Two ordered steps: (a) complete the inline `// ...` comments on `reference.conf` so every leaf key carries a one-line description of intent / units / interaction; (b) add a sub-gate that fails when a registered-section leaf key has no preceding comment. Step (b) is held until step (a) lands — gating an under-commented file would force vacuous one-liners and defeat the point.

**Extra details**

Design tradeoffs — five gate ideas were considered. We deliberately ship only the four shipped here and reject / defer the rest:

| #  | Idea | Decision | Reason |
|----|------|----------|--------|
| 2  | Bean ↔ key parity + default-value drift (this PR) | **Ship** | The drift modes here are silent and have already produced operator-visible incidents. The check is cheap (reflection + `Config.get*`) and the contract surface is small and explicit. |
| 3a | `applyXxxConfig` assigns every bean field to `Parameter` | Reject | Business logic — eyeballable at review time. AST-pattern-matching this is form-fitting, not value-adding. |
| 3b | Every `getXxx` has a production consumer | Reject | Equivalent to unused-symbol lint, already handled by IDE/SpotBugs/jdtls. Periodic dead-code cleanup, not a gate. |
| 4  | Inline comment coverage on leaf HOCON keys | Deferred (follow-up) | Gating today would force vacuous one-liners — `reference.conf` is commented by logical block, not per-leaf, so a coverage gate over the current file is noise. Plan is to first complete the inline comments (every leaf key gets a one-line description of intent / units / interaction), then land the gate. Tracked under **Follow up** above. |
| 5  | Every bean field is referenced by at least one assertion in `*ConfigTest` | Reject | Measures test shape, not test value. Would catalyze `assertEquals(x, x)` boilerplate. Coverage belongs to JaCoCo + review. |

Scope discipline that fell out of this:

- The gate validates the **contract layer** (key exists, field exists, default matches). It does **not** validate the **behavior layer** (the bound value flows to the right downstream consumer). That stays with `*ConfigTest` behavioural cases (alias fallback, clamp, `fromConfig` outputs) plus integration tests.
- Manual-read configs (`MiscConfig`, `LocalWitnessConfig`) are intentionally out of scope. Their HOCON paths are scattered top-level keys with no 1:1 bean field, so any "parity gate" would mean asserting a hand-written mapping that duplicates the very thing it's supposed to protect, and could not catch the actual failure mode (a new manual-read key being added without sync). `TOP_LEVEL_NON_BEAN` accounts for them as "explicitly out-of-scope with rationale" instead.
- Unsupported scalar types (Map, enum, `Duration`, `MemorySize`, custom classes) are a hard failure today rather than an allowlistable skip. None of the current `*Config` beans use such types; the comment in `Section` documents the path for re-introducing a `typeSkip` field if a future bean genuinely cannot be value-compared.

Implementation notes:

- Registry-driven: one `Section` row per (path, bean, hoconOrphans, beanOrphans, divergent). New `*Config` beans add a row here — no per-section parity test file.
- Recursion: descends into a child `*Config` bean iff the property's Java type has a default constructor, lives in `org.tron.*`, and the HOCON value is an OBJECT. Lists / primitives / JDK types never recurse.
- Logging: emits `[parity-hocon]`, `[parity-bean]`, `[parity-default]`, `[parity-sweep]` per section, then a final `[parity-summary]` block with the file-coverage equation `file-hoconKey = checkSection + cantCheckSection` and the bean-coverage triple-equality `registry = parity-bean = parity-default`. When the gate runs without booting a tron main, the lines land in `common/build/test-results/test/TEST-...ConfigParityGateTest.xml` under `<system-out>`; once a main has touched logback in the same fork, they route to `logs/` on disk — class Javadoc spells this out.

Allowlists preserved (do not extend):

- HOCON orphans: `committee.{allowPBFT,pBFTExpireNum}`, `node.{isOpenFullTcpDisconnect,metrics}`, `event.subscribe.{native,topics}`.
- Bean orphans: `committee.{allowPbft,pbftExpireNum}`, `node.openFullTcpDisconnect`.
- Divergent defaults: `genesis.block.{timestamp,parentHash,assets,witnesses}`, `node.fastForward`, `event.subscribe.filter.{contractAddress,contractTopic}`.

### DIFF
--- a/common/src/test/java/org/tron/core/config/args/ConfigParityCheck.java
+++ b/common/src/test/java/org/tron/core/config/args/ConfigParityCheck.java
@@ -1,0 +1,713 @@
+package org.tron.core.config.args;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigObject;
+import com.typesafe.config.ConfigValue;
+import com.typesafe.config.ConfigValueType;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Shared helpers for reference.conf &lt;-&gt; {@code *Config} bean parity tests.
+ * Asserts that every HOCON key under a section binds to a writable bean
+ * property and matches the bean's default. Drift fails the build at PR time
+ * instead of waiting for {@code ConfigBeanFactory} to throw at startup.
+ * <p>
+ * {@code [parity-*]} audit log lines land in the JUnit XML {@code <system-out>}
+ * when the gate runs in isolation; if mixed with tests that boot a tron main,
+ * production logback may redirect them to {@code logs/} on disk.
+ */
+@Slf4j(topic = "test")
+final class ConfigParityCheck {
+
+  private ConfigParityCheck() {
+
+  }
+
+  private static Map<String, PropertyDescriptor> writablePropertyDescriptors(Class<?> beanClass) {
+    try {
+      Map<String, PropertyDescriptor> m = new TreeMap<>();
+      for (PropertyDescriptor pd :
+          Introspector.getBeanInfo(beanClass, Object.class).getPropertyDescriptors()) {
+        if (pd.getWriteMethod() != null) {
+          m.put(pd.getName(), pd);
+        }
+      }
+      return m;
+    } catch (java.beans.IntrospectionException e) {
+      throw new AssertionError("Introspector failed on " + beanClass.getName(), e);
+    }
+  }
+
+  /**
+   * {@code shapeMismatches}: HOCON key matches a bean property of nested
+   * {@code *Config} type but the HOCON value is not OBJECT — walker cannot
+   * recurse, and downstream binding would throw {@code WrongType}.
+   */
+  private static final class OrphanCounters {
+    int total;
+    int bound;
+    final Set<String> orphans = new TreeSet<>();
+    final Set<String> allowlisted = new TreeSet<>();
+    final Set<String> shapeMismatches = new TreeSet<>();
+  }
+
+  /**
+   * Fails when reference.conf has keys under {@code sectionPath} (recursively
+   * through nested {@code *Config} sub-sections) that the bean cannot bind.
+   * Recurses into a sub-config when the property type satisfies
+   * {@link #isRecursiveConfigBean} AND the HOCON value is an OBJECT.
+   */
+  static void assertNoHoconOrphans(
+      String sectionPath, Class<?> beanClass, Set<String> allowedHoconOrphans) {
+    Config section = ConfigFactory.defaultReference().getConfig(sectionPath);
+    OrphanCounters c = walkAndLogHoconOrphans(
+        sectionPath, section, beanClass, allowedHoconOrphans);
+    AGGREGATES.hoconKey += c.total;
+    AGGREGATES.hoconBound += c.bound;
+    AGGREGATES.hoconAllowlisted += c.allowlisted.size();
+    AGGREGATES.beans.add(beanClass);
+    failOnHoconOrphans(sectionPath, beanClass, c);
+  }
+
+  /** Overload for meta-tests: walks the supplied Config directly, skips AGGREGATES. */
+  static void assertNoHoconOrphans(
+      String label, Config section, Class<?> beanClass,
+      Set<String> allowedHoconOrphans) {
+    OrphanCounters c = walkAndLogHoconOrphans(
+        label, section, beanClass, allowedHoconOrphans);
+    failOnHoconOrphans(label, beanClass, c);
+  }
+
+  private static OrphanCounters walkAndLogHoconOrphans(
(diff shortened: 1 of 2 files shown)


halibobo1205 (approved):

317787106:
[SHOULD] `ConfigParityCheck` ships four `(String label, Config section, Class, Set…)` overloads (the `assertNoHoconOrphans` / `assertNoBeanOrphans` / `assertDefaultValuesMatch` / `assertAllowlistEntriesAreLive` meta-test variants). Their only callers are the 37 helper meta-tests that — per the PR description — live outside `src/` and are copied in by an external runner, so they are **not committed**. Within this PR, `ConfigParityGateTest` invokes only the 3-arg/5-arg production variants, leaving these overloads with no in-repo caller (dead code per the "no unused methods / YAGNI" rule), and the branch-coverage proof for the walkers invisible to the repo, CI, and JaCoCo.

Recommendation: commit the meta-tests — they are the strongest evidence the reflection walkers are correct, and that aligns with this PR's own goal of making config drift visible in-repo. If they must stay external, drop the four overloads and have the external harness call the production overloads with a `defaultReference()`-derived section instead.

bladehan1:
Good catch. We went with option 2: the four `(label, Config, …)` overloads have been dropped in the latest commit — they had no in-repo callers and the meta-test path is external to CI. For now we're not wiring in additional unit coverage for the walkers; if this area changes in the future, targeted tests can be added at that point.

317787106 (approved):

PR #6804, opened 2026-05-29
fix(net): exit solidity node block sync on shutdown

317787106:
**What does this PR do?**

1. **`TronNetDelegate`** — wrap `LockSupport.park()` in a `while (!hitDown && !Thread.currentThread().isInterrupted())` loop so that spurious unparks (e.g. from `Thread.interrupt()` in `close()`) cannot trigger the `System.exit(0)` path prematurely, and so the hit-thread exits cleanly when the Spring context is destroyed.

2. **`SolidityNode`** — fix a shutdown race in `getBlockByNum()`: when a gRPC exception is thrown while the node is shutting down (`!flag || isHitDown()`), break out immediately instead of sleeping `exceptionSleepTime` (1 s) and retrying.

3. **`ShieldedReceiveTest.pushSameSkAndScanAndSpend`** — fix a test isolation bug and a performance regression:
   - The test relied on the genesis bypass in `validBlock()` (`latestBlockHeaderNumber == 0 → return true`) to skip DPoS schedule validation. When prior tests cause the consensus background task to produce blocks, this bypass no longer applies and the block push fails with `ValidateScheduleException`.
   - Replace hardcoded `System.currentTimeMillis()` timestamps with `nextScheduledTime()`, which looks up the nearest slot where the witness is actually scheduled according to `DposSlot`.
   - Stop `ConsensusService` before manual block pushes: `DposTask` shares the same `localwitness` key and would race to produce blocks at the same slot, triggering `switchFork` and making the test slow.

4. **`reference.conf`** — remove the stale "Keys that cannot auto-bind" comment block; those implementation details belong in the source classes, not in the config template.

5. **`README.md`** — surface `reference.conf` alongside `config.conf` in the configuration-file list so users know a built-in defaults template exists (added in v4.8.2).

- The single `LockSupport.park()` call could be woken by `Thread.interrupt()` from `close()`, causing the process to reach `System.exit(0)` before `hitDown` was ever set. With the while-loop fix, `close()` calling `interrupt()` caused a busy-spin because `park()` returns immediately when interrupted but the loop never exited — adding the `isInterrupted()` guard lets the thread exit cleanly.
- During shutdown, a transient gRPC exception in `getBlockByNum()` triggered a 1-second sleep and an unnecessary retry loop, delaying clean exit and producing noisy error logs.
- `pushSameSkAndScanAndSpend` passed in isolation because the chain started at genesis (block 0), bypassing DPoS validation. In a full test run, the consensus background task produces blocks before this test runs, advancing the head past 0 and exposing the missing schedule-aware timestamp logic.
- The `reference.conf` comment described internal normalisation details that have since been refactored into the Java classes.
- `reference.conf` was introduced in v4.8.2 but not mentioned in the README.

- Unit Tests
  - `testGetBlockByNumNoErrorOnExceptionDuringShutdown`: verifies the shutdown race fix — method breaks in < 500 ms; without the fix a 1-second sleep fires first.
  - `pushSameSkAndScanAndSpend`: now stable in both single-test and full-suite runs.
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/TronNetDelegate.java
+++ b/framework/src/main/java/org/tron/core/net/TronNetDelegate.java
@@ -111,7 +111,9 @@ public class TronNetDelegate {
   @PostConstruct
   public void init() {
     hitThread =  new Thread(() -> {
-      LockSupport.park();
+      while (!hitDown && !Thread.currentThread().isInterrupted()) {
+        LockSupport.park();
+      }
       // to Guarantee Some other thread invokes unpark with the current thread as the target
       if (hitDown && exit) {
         System.exit(0);
--- a/framework/src/main/java/org/tron/program/SolidityNode.java
+++ b/framework/src/main/java/org/tron/program/SolidityNode.java
@@ -122,7 +122,7 @@ public class SolidityNode implements ApplicationListener<ContextClosedEvent> {
         logger.info("getBlock interrupted, exiting.");
         return;
       } catch (Exception e) {
-        if (!flag) {
+        if (!flag || tronNetDelegate.isHitDown()) {
           logger.info("getBlock stopped during shutdown, last block: {}.", blockNum);
           return;
         }
--- a/framework/src/main/java/org/tron/program/SolidityNode.java
+++ b/framework/src/main/java/org/tron/program/SolidityNode.java
@@ -185,6 +185,10 @@ public class SolidityNode implements ApplicationListener<ContextClosedEvent> {
           sleep(exceptionSleepTime);
         }
       } catch (Exception e) {
+        if (!flag || tronNetDelegate.isHitDown()) {
+          logger.info("getBlockByNum stopped during shutdown, block: {}.", blockNum);
+          break;
+        }
         logger.error("Failed to get block: {}, reason: {}.", blockNum, e.getMessage());
         sleep(exceptionSleepTime);
       }
--- a/framework/src/main/java/org/tron/program/SolidityNode.java
+++ b/framework/src/main/java/org/tron/program/SolidityNode.java
@@ -206,7 +206,7 @@ public class SolidityNode implements ApplicationListener<ContextClosedEvent> {
             blockNum, remoteBlockNum, System.currentTimeMillis() - time);
         return blockNum;
       } catch (Exception e) {
-        if (!flag) {
+        if (!flag || tronNetDelegate.isHitDown()) {
           logger.info("getLastSolidityBlockNum stopped during shutdown.");
           return 0;
         }
--- a/framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
+++ b/framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
@@ -8,6 +8,7 @@ import com.google.common.primitives.Bytes;
 import com.google.protobuf.Any;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
+import java.lang.reflect.Field;
 import java.security.SignatureException;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -46,6 +47,7 @@ import org.tron.common.zksnark.LibrustzcashParam.IvkToPkdParams;
 import org.tron.common.zksnark.LibrustzcashParam.OutputProofParams;
 import org.tron.common.zksnark.LibrustzcashParam.SpendSigParams;
 import org.tron.consensus.dpos.DposSlot;
+import org.tron.consensus.dpos.DposTask;
 import org.tron.core.Wallet;
 import org.tron.core.actuator.Actuator;
 import org.tron.core.actuator.ActuatorCreator;
@@ -140,6 +142,8 @@ public class ShieldedReceiveTest extends BaseTest {
   private static final String URL = "https://tron.network";
   @Resource
   private ConsensusService consensusService;
+  @Resource
+  private DposTask dposTask;
   @Resource
   private Wallet wallet;
   @Resource
@@ -2534,6 +2538,11 @@ public class ShieldedReceiveTest extends BaseTest {
       boolean ok2 = dbManager.pushTransaction(transactionCap2);
       Assert.assertTrue(ok2);
     } finally {
+      // DposTask.init() does not reset isRunning (it stays false after stop()), so force it back
+      // to true via reflection before restarting.
+      Field isRunning = DposTask.class.getDeclaredField("isRunning");
+      isRunning.setAccessible(true);
+      isRunning.set(dposTask, true);
       consensusService.start();
     }
   }
--- a/framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
+++ b/framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
@@ -45,6 +45,7 @@ import org.tron.common.zksnark.LibrustzcashParam.CheckSpendParams;
 import org.tron.common.zksnark.LibrustzcashParam.IvkToPkdParams;
 import org.tron.common.zksnark.LibrustzcashParam.OutputProofParams;
 import org.tron.common.zksnark.LibrustzcashParam.SpendSigParams;
+import org.tron.consensus.dpos.DposSlot;
 import org.tron.core.Wallet;
 import org.tron.core.actuator.Actuator;
 import org.tron.core.actuator.ActuatorCreator;
@@ -142,12 +143,12 @@ public class ShieldedReceiveTest extends BaseTest {
   @Resource
   private Wallet wallet;
   @Resource
-  private TransactionUtil transactionUtil;
+  private DposSlot dposSlot;
 
   private static boolean init;
 
   static {
-    Args.setParam(new String[]{"--output-directory", dbPath(), "-w"}, SHIELD_CONF);
+    Args.setParam(new String[] {"--output-directory", dbPath(), "-w"}, SHIELD_CONF);
     ADDRESS_ONE_PRIVATE_KEY = getRandomPrivateKey();
     FROM_ADDRESS = getHexAddressByPrivateKey(ADDRESS_ONE_PRIVATE_KEY);
   }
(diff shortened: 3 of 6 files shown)


framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
```diff
@@ -2402,10 +2405,13 @@ public void pushSameSkAndScanAndSpend() throws Exception {
     assert ecKey != null;
     byte[] witnessAddress = ecKey.getAddress();
     WitnessCapsule witnessCapsule = new WitnessCapsule(ByteString.copyFrom(witnessAddress));
+    // Stop the consensus task before modifying the witness schedule: DposTask uses the same
+    // localwitness key and would otherwise race to produce blocks at the same slot,
+    // triggering fork resolution and making the test slow.
+    consensusService.stop();
```
3for, line 2413: [SHOULD] `@Before` only starts the consensus service once when `init == false` (ShieldedReceiveTest.java:167). However, `pushSameSkAndScanAndSpend()` calls `consensusService.stop()` (ShieldedReceiveTest.java:2411) without restoring it in a `finally` block and without resetting `init`.

Since JUnit does not guarantee test execution order, subsequent tests in the same suite may inherit a stopped consensus service and become affected if they rely on background consensus tasks.

Consider restoring the service in a `try/finally` block, or isolating this test in a dedicated fixture / Spring context to avoid cross-test contamination.

317787106: Good catch. Fixed in b93f7d4564 — wrapped the test body in `try/finally` so `consensusService.start()` is guaranteed to run even if an assertion throws, preventing the stopped state from leaking into subsequent tests.

framework/src/test/java/org/tron/program/SolidityNodeTest.java
```diff
@@ -329,6 +330,46 @@ public void testGetBlockByNumWhenClosed() throws Exception {
...
+   */
+  @Test(timeout = 5000)
+  public void testGetBlockByNumNoErrorOnExceptionDuringShutdown() throws Exception {
+    Method m = SolidityNode.class.getDeclaredMethod("getBlockByNum", long.class);
+    m.setAccessible(true);
+    Field clientField = getField("databaseGrpcClient");
+    Object origClient = clientField.get(solidityNode);
+    try {
+      DatabaseGrpcClient mockClient = mock(DatabaseGrpcClient.class);
+      // flag races to false inside the gRPC call — exact close() race
```
3for, line 348: [SHOULD] `testGetBlockByNumNoErrorOnExceptionDuringShutdown()` does not explicitly set `flag = true` before invoking the method under test. Instead, it relies on previous tests restoring the state via `finally { setFlag(true); }`.

This makes the test order-dependent: if executed in isolation while `flag` is already `false`, the `while` loop would not execute and `verify(..., times(1))` would fail. The test currently passes because `BaseTest` leaves `flag` initialized to `true`, but it would be more robust to set the required precondition explicitly within the test.

317787106: Good point. Fixed in cc4f0784b1 — added `setFlag(true)` as an explicit precondition at the start of the test so it no longer relies on execution order to guarantee `flag=true` when the `while` loop is entered.

framework/src/main/java/org/tron/core/net/TronNetDelegate.java
```diff
@@ -111,7 +111,9 @@ public class TronNetDelegate {
   @PostConstruct
   public void init() {
     hitThread =  new Thread(() -> {
-      LockSupport.park();
+      while (!hitDown && !Thread.currentThread().isInterrupted()) {
+        LockSupport.park();
+      }
```
3for, line 116: [NIT] Consider adding a dedicated test for the `close() → interrupt() → hitDown == false` path. The scenario this fix targets currently lacks direct test coverage.

framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
```diff
@@ -2402,128 +2403,153 @@ public void pushSameSkAndScanAndSpend() throws Exception {
...
+      byte[] memo3 = org.tron.keystore.Wallet.generateRandomBytes(512);
+      builder2.addOutput(expsk2.getOvk(), paymentAddress3,
+          (1000 * 1000000L - wallet.getShieldedTransactionFee()) / 2 - wallet
+              .getShieldedTransactionFee(), memo3);
+
+      TransactionCapsule transactionCap2 = builder2.build();
+      boolean ok2 = dbManager.pushTransaction(transactionCap2);
+      Assert.assertTrue(ok2);
+    } finally {
+      consensusService.start();
```
3for, line 2546: [SHOULD] `pushSameSkAndScanAndSpend` still appears to leak state into subsequent tests through the consensus background task.

`ShieldedReceiveTest.java` (line 2409) calls `consensusService.stop()`, and `ShieldedReceiveTest.java` (line 2537) attempts to restore it with `consensusService.start()` in the `finally` block.

However, `DposTask.stop()` sets `isRunning = false` (`DposTask.java:82`), while `DposTask.init()` does not reset it back to `true` (`DposTask.java:46`). As a result, after `start()` creates a new executor, the newly submitted runnable immediately exits because `isRunning` is still `false`, so block production is not actually resumed.

This means the test can still affect subsequent tests running in the same Spring context.

Consider either:

* resetting `isRunning = true` in `DposTask.init()` before creating the executor, or
* isolating this test in a dedicated context/fixture instead of stopping and restarting a shared singleton service.

framework/src/main/java/org/tron/core/net/TronNetDelegate.java
317787106: It's difficult to cover this with a unit test. Testing it in system integration tests is sufficient.

framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
317787106: Actually, other testcases don't use `consensusService`. But reset DposTask's `isRunning` by reflection in `finally` block is still ok.

framework/src/main/java/org/tron/program/SolidityNode.java
3for: [SHOULD] The `hitDown` shutdown path can still incur a 1-second sleep and emit an error log.

`SolidityNode.java` (line 187) correctly breaks out of `getBlockByNum()` when `!flag || isHitDown()`, but it subsequently throws `"SolidityNode is closing."`.

In the outer `getBlock()` loop (`SolidityNode.java:124`), the `catch` block only checks `!flag` and does not check `tronNetDelegate.isHitDown()`. As a result, if `hitDown == true` while `flag` is still `true`, execution falls into the error path at `SolidityNode.java:129` (error log) and `SolidityNode.java:130` (`exceptionSleepTime` delay).

The existing test (`SolidityNodeTest.java:349`) only covers the `flag=false` race and does not cover the `hitDown=true` shutdown race.

Consider updating the outer catch block as well:

```java
if (!flag || tronNetDelegate.isHitDown()) {
  return;
}
```

so that the `hitDown` shutdown path exits cleanly without the extra error log and sleep.

framework/src/test/java/org/tron/core/zksnark/ShieldedReceiveTest.java
3for: [SHOULD] The test appears to work around a real production issue via reflection rather than fixing the root cause.

`DposTask.stop()` sets `isRunning = false`, but `DposTask.init()` never resets it (`DposTask.java:43,84`). As a result, any `ConsensusService.stop() → start()` cycle leaves block production permanently disabled—not just in tests, but in production as well.

The actual fix seems to be a single line:

```java
isRunning = true;
```

at the beginning of `DposTask.init()` (or `start()`).

That would also allow the test to simply call `stop()`/`start()` without resorting to reflection.

As written, the underlying defect remains, and the test reaches into a private `volatile` field via reflection. This is fairly brittle—a field rename could silently break the test at runtime.

Recommend fixing `DposTask` itself and removing the reflection from the test.

317787106: The main code should not be modified solely to make the test cases pass.

3for: The concern is not about modifying production code for the sake of a test. The test exposed a potential production issue: any `ConsensusService.stop() → start()` cycle leaves block production permanently disabled.

If such a lifecycle transition cannot occur in production, then this can be considered a test-only concern and safely ignored. Otherwise, it may indicate a real bug in the service lifecycle handling.

317787106: This lifecycle transition `ConsensusService.stop() → start()` cannot occur in production. So there is no need to change it.

framework/src/main/java/org/tron/program/SolidityNode.java
317787106: Add it.

3for (approved):

halibobo1205 (approved):
Not a blocker, just a design suggestion for follow-up.
The current `hit-thread` design ties `System.exit(0)` to a long-lived parked thread: anything that causes it to exit before `hitDown` is set (spurious wakeup, early interrupt, an exception in the lambda) silently breaks shutdown — the subsequent `unpark` becomes a no-op, and because `SolidityNode`'s `get-block` / `process-block` executors are non-daemon, the JVM idles forever. This PR closes the spurious-wakeup path, which is good, but the "single long-lived thread must stay alive" invariant remains a fragile pattern.

 The project already has a sturdier mechanism — `ExitManager.logAndExit(TronError)` — which spins up a fresh daemon thread per call to invoke `System.exit(code)` and is already wired into `UncaughtExceptionHandler` and `ExecutorServiceManager.submit`. Migrating the config-driven shutdown to it would let us delete the `hitThread` field, `@PostConstruct init()`, `@PreDestroy close()`, and the `LockSupport.unpark` call:

 ```java
 // in processBlock, when the shutdown condition matches
 hitDown = true;
 if (exit) {
   ExitManager.logAndExit(new TronError(reason, ErrCode.SHUTDOWN_AUTO_STOP)); // new ErrCode
 }
 return;
 ```

 `hitDown` and `exit` stay (still used as gates in `processBlock` / `pushVerifiedBlock` / `SolidityNode.isHitDown()` and as a test hook). No parked thread, no race surface, fewer moving parts.

 Happy to open a separate issue/PR for this if you'd like — doesn't need to block this one.

317787106:
@halibobo1205 `LockSupport.unpark` is not very stable, so using `TronError` maybe an alternative idea. We can adopt this approach when `SolidityNode` is refactored in the future.

PR #6806, opened 2026-06-01
fix(db): fix storage config properties

317787106:
**What does this PR do?**

**`StorageConfig` / `Storage`** — fix `ConfigException$ValidationFailed` when starting with a config file (e.g. `config-nile.conf`) whose `storage.properties` entries only set `name` and `path` (all LevelDB options commented out); clean up dead config fields in the same pass.
- **Bug fix**: `ConfigBeanFactory.create()` iterates every setter on `PropertyConfig` and requires a matching key in each list item; absent keys throw `ValidationFailed` even though `PropertyConfig` has Java-side defaults for all optional fields. Fix: add `@Setter(NONE)` to the `properties` field so `ConfigBeanFactory` skips it, then parse the list manually in `fromConfig()` using `containsKey()` guards — the same pattern already used for `defaultDbOption`/`defaultMDbOption`/`defaultLDbOption`.
- **Dead field removal**: `createIfMissing`, `paranoidChecks`, `verifyChecksums`, and `compressionType` are removed from both `PropertyConfig` and `DbOptionOverride`. These values are fixed constants in `DbOptionalsUtils` (`true`/`true`/`true`/SNAPPY) and are never changed by anyone; exposing them as config fields gave the false impression they were tunable. The corresponding setter calls are also removed from `applyPropertyOptions` and `applyDbOptionOverride` in `Storage`.
- **Refactor**: extract `readLevelDbOptions(ConfigObject)` as a shared parser for the remaining 4 LevelDB fields (`blockSize`, `writeBufferSize`, `cacheSize`, `maxOpenFiles`); `readDbOption` is reduced to 3 lines; `readProperties` reuses it and applies only non-null values to `PropertyConfig`.
- **Validation**: each `properties` entry must have a `name` field; missing `name` now throws `IllegalArgumentException` with a clear message.

Nodes started with a minimal config (e.g. `config-nile.conf`) that sets only `name` and `path` per properties entry were broken by the `ConfigBeanFactory` validation; the refactored manual parser restores the original lenient behaviour while keeping error-handling consistent across all numeric fields. The four removed fields (`createIfMissing`, `paranoidChecks`, `verifyChecksums`, `compressionType`) are global LevelDB defaults fixed in `DbOptionalsUtils` — they were never user-tunable in practice and their presence in the config bean was misleading.

- Unit Tests
  - `StorageTest`: updated to reflect the removed fields; all three test methods pass.
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/common/src/main/java/org/tron/core/config/args/Storage.java
+++ b/common/src/main/java/org/tron/core/config/args/Storage.java
@@ -25,7 +25,6 @@ import java.util.stream.Collectors;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
-import org.iq80.leveldb.CompressionType;
 import org.iq80.leveldb.Options;
 import org.tron.common.cache.CacheStrategies;
 import org.tron.common.cache.CacheType;
@@ -179,11 +178,6 @@ public class Storage {
    * Apply LevelDB options from PropertyConfig bean values.
    */
   private static void applyPropertyOptions(StorageConfig.PropertyConfig pc, Options dbOptions) {
-    dbOptions.createIfMissing(pc.isCreateIfMissing());
-    dbOptions.paranoidChecks(pc.isParanoidChecks());
-    dbOptions.verifyChecksums(pc.isVerifyChecksums());
-    dbOptions.compressionType(
-        CompressionType.getCompressionTypeByPersistentId(pc.getCompressionType()));
     dbOptions.blockSize(pc.getBlockSize());
     dbOptions.writeBufferSize(pc.getWriteBufferSize());
     dbOptions.cacheSize(pc.getCacheSize());
@@ -247,19 +241,6 @@ public class Storage {
   // Apply only user-specified overrides (non-null fields) to LevelDB Options.
   private static void applyDbOptionOverride(
       StorageConfig.DbOptionOverride o, Options dbOptions) {
-    if (o.getCreateIfMissing() != null) {
-      dbOptions.createIfMissing(o.getCreateIfMissing());
-    }
-    if (o.getParanoidChecks() != null) {
-      dbOptions.paranoidChecks(o.getParanoidChecks());
-    }
-    if (o.getVerifyChecksums() != null) {
-      dbOptions.verifyChecksums(o.getVerifyChecksums());
-    }
-    if (o.getCompressionType() != null) {
-      dbOptions.compressionType(
-          CompressionType.getCompressionTypeByPersistentId(o.getCompressionType()));
-    }
     if (o.getBlockSize() != null) {
       dbOptions.blockSize(o.getBlockSize());
     }
--- a/common/src/main/java/org/tron/core/config/args/Storage.java
+++ b/common/src/main/java/org/tron/core/config/args/Storage.java
@@ -169,19 +169,11 @@ public class Storage {
     }
 
     Options dbOptions = newDefaultDbOptions(property.getName());
-    applyPropertyOptions(pc, dbOptions);
+    // PropertyConfig is-a DbOptionOverride: apply only user-specified (non-null) overrides
+    // so unset fields keep the per-tier defaults already applied by newDefaultDbOptions.
+    applyDbOptionOverride(pc, dbOptions);
     property.setDbOptions(dbOptions);
     return property;
-  }
-
-  /**
-   * Apply LevelDB options from PropertyConfig bean values.
-   */
-  private static void applyPropertyOptions(StorageConfig.PropertyConfig pc, Options dbOptions) {
-    dbOptions.blockSize(pc.getBlockSize());
-    dbOptions.writeBufferSize(pc.getWriteBufferSize());
-    dbOptions.cacheSize(pc.getCacheSize());
-    dbOptions.maxOpenFiles(pc.getMaxOpenFiles());
   }
 
   /**
--- a/common/src/main/java/org/tron/core/config/args/StorageConfig.java
+++ b/common/src/main/java/org/tron/core/config/args/StorageConfig.java
@@ -28,6 +28,8 @@ public class StorageConfig {
   private CheckpointConfig checkpoint = new CheckpointConfig();
   private SnapshotConfig snapshot = new SnapshotConfig();
   private TxCacheConfig txCache = new TxCacheConfig();
+  // ConfigBeanFactory requires all bean fields present per item, so we parse manually.
+  @Setter(lombok.AccessLevel.NONE)
   private List<PropertyConfig> properties = new ArrayList<>();
 
   // merkleRoot is a nested object (e.g. { reward-vi = "hash..." }) not a string.
@@ -54,6 +56,7 @@ public class StorageConfig {
   @Getter
   @Setter
   public static class DbConfig {
+
     private String engine = "LEVELDB";
     private boolean sync = false;
     private String directory = "database";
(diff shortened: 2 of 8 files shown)


common/src/main/java/org/tron/core/config/args/Storage.java
```diff
@@ -247,19 +241,6 @@ public Options newDefaultDbOptions(String name) {
...
-    if (o.getParanoidChecks() != null) {
-      dbOptions.paranoidChecks(o.getParanoidChecks());
-    }
-    if (o.getVerifyChecksums() != null) {
-      dbOptions.verifyChecksums(o.getVerifyChecksums());
-    }
-    if (o.getCompressionType() != null) {
-      dbOptions.compressionType(
-          CompressionType.getCompressionTypeByPersistentId(o.getCompressionType()));
-    }
```
3for, line 262: [MUST] `StorageConfig.java:198` and `Storage.java:243` no longer parse or apply `createIfMissing`, `paranoidChecks`, `verifyChecksums`, or `compressionType`. However, the legacy documentation still recommends these fields (common config README:31).

As a result, existing configurations such as:

```hocon
compressionType = 0
```

will continue to start successfully, but the effective behavior silently reverts to the default `SNAPPY` compression without any warning to the user.

framework/src/test/java/org/tron/core/config/args/StorageTest.java
3for: [SHOULD] Remove unused import.

common/src/main/java/org/tron/core/config/args/Storage.java
317787106: Thanks for your catch. Delete the related instruction in `common/src/main/java/org/tron/core/config/README.md`. These four configuration options are silently ignored, as they are rarely modified in practice.

framework/src/test/java/org/tron/core/config/args/StorageTest.java
317787106: Thanks for your review, removed it,

common/src/main/java/org/tron/core/config/args/StorageConfig.java
```diff
@@ -261,6 +242,50 @@ private static DbOptionOverride readDbOption(Config section, String key) {
...
+      return null;
+    }
+    return readLevelDbOptions(section.getObject(key));
+  }
+
+  // Parse storage.properties list manually: ConfigBeanFactory requires every bean field to be
+  // present in each list item, but name+path-only entries (all LevelDB opts commented out) are
+  // valid — missing fields fall back to PropertyConfig Java defaults.
+  private static List<PropertyConfig> readProperties(Config section) {
+    if (!section.hasPath("properties")) {
```
3for, line 255: [SHOULD] Add unit tests for the newly introduced `readProperties()` logic.

common/src/main/java/org/tron/core/config/args/StorageConfig.java
```diff
@@ -261,6 +242,50 @@ private static DbOptionOverride readDbOption(Config section, String key) {
...
+      return new ArrayList<>();
+    }
+    List<? extends ConfigObject> items = section.getObjectList("properties");
+    List<PropertyConfig> result = new ArrayList<>(items.size());
+    for (ConfigObject obj : items) {
+      if (!obj.containsKey("name")) {
+        throw new IllegalArgumentException("[storage.properties] database name must be set.");
+      }
+      PropertyConfig p = new PropertyConfig();
+      p.setName(obj.get("name").unwrapped().toString());
```
3for: [SHOULD] NPE on explicit `name = null`. In `readProperties()` (StorageConfig.java:263-267), `obj.containsKey("name")` is true for `name = null,` then `obj.get("name").unwrapped().toString()` NPEs instead of
  throwing the clean "database name must be set" message.

common/src/main/java/org/tron/core/config/args/Storage.java
```diff
@@ -179,11 +178,6 @@ private Property createPropertyFromBean(StorageConfig.PropertyConfig pc) {
    * Apply LevelDB options from PropertyConfig bean values.
    */
   private static void applyPropertyOptions(StorageConfig.PropertyConfig pc, Options dbOptions) {
-    dbOptions.createIfMissing(pc.isCreateIfMissing());
-    dbOptions.paranoidChecks(pc.isParanoidChecks());
-    dbOptions.verifyChecksums(pc.isVerifyChecksums());
-    dbOptions.compressionType(
-        CompressionType.getCompressionTypeByPersistentId(pc.getCompressionType()));
     dbOptions.blockSize(pc.getBlockSize());
     dbOptions.writeBufferSize(pc.getWriteBufferSize());
```
3for: [SHOULD] The `name/path`-only configuration should preserve the original permissive behavior.

`StorageConfig.java` (line 168) initializes the four LevelDB fields with Java-side defaults, and `Storage.java` (line 180) then unconditionally applies those values to the resulting `Options`.

As a result, a configuration such as:

```hocon
properties = [{ name = "account", path = "..." }]
```

no longer just overrides the path. It also overwrites the tuning already applied by `newDefaultDbOptions()` via `storage.default` / `defaultM` / `defaultL`. For example, `account` would normally inherit `defaultL.maxOpenFiles = 1000`, but it now gets overridden by `PropertyConfig.maxOpenFiles = 100`.

The previous manual parsing logic only called setters when `containsKey()` was true, so omitted fields never overrode the global defaults.

Consider making the four LevelDB fields in `PropertyConfig` boxed nullable types as well, and only applying non-null values in `Storage.applyPropertyOptions()`. This would restore the original behavior where unspecified fields inherit the defaults from `storage.default` / `defaultM` / `defaultL` rather than silently resetting them to `PropertyConfig` defaults.

common/src/main/java/org/tron/core/config/args/StorageConfig.java
317787106: Thanks for your review. It has been optimzed as :
```
      PropertyConfig p = new PropertyConfig();
      if (obj.containsKey("name")) {
        p.setName(obj.get("name").unwrapped().toString());
      }
```
throws NPE is expected if name is null, also it's verifed by following method `createPropertyFromBean`, we don't deal with it here.

common/src/main/java/org/tron/core/config/args/Storage.java
317787106: Good catch. All fields of `PropertyConfig` have been maked boxed nullable type,  only applying non-null values in `readProperties`.

common/src/main/java/org/tron/core/config/args/StorageConfig.java
317787106: Add some testcases in `StorageConfigTest`.

halibobo1205 (approved):
LGTM. One small note on the NPE reply: `createPropertyFromBean()` only checks `isEmpty()`, so `name = null` still NPEs in `readProperties()` before reaching that validation. If the intended behavior is fail-fast on an invalid null config, please just reword the reply to say that explicitly; otherwise, readers may think there is a downstream null guard.

common/src/main/resources/reference.conf
```diff
@@ -44,49 +44,68 @@ net {
...
+  #   cacheSize = 33554432,       // 32 MB
+  #   maxOpenFiles = 100          // recommend 500 for production
   # }
+  # Default for frequently accessed databases: account, delegation, storage-row
   # defaultL = {
-  #   maxOpenFiles = 1000
+  #   blockSize = 4096,           // 4 KB
+  #   writeBufferSize = 67108864, // 64 MB
+  #   cacheSize = 33554432,       // 32 MB
+  #   maxOpenFiles = 100          // recommend 1000 for production
```
3for, line 91: [NIT] Set the default values of `maxOpenFiles` to the recommended settings (e.g., 500 and 1000 respectively) instead of 100.

3for (approved):

common/src/main/resources/reference.conf
317787106: In `reference.conf`, only default values should be written. Recommended values can be placed in `config.conf`

PR #6808, opened 2026-06-01
fix(ci): upgrade actions/cache to v5 and fix container git checkout

317787106:
**What does this PR do?**

- Upgrades `actions/cache` from `v4` to `v5` in `pr-build.yml`, `pr-check.yml`, and `system-test.yml` (7 occurrences) to resolve the Node.js 20 deprecation warning on GitHub Actions runners.
- Reorders the container-based jobs in `pr-build.yml` (`docker-build-rockylinux`, `docker-build-debian11`, `coverage-base`) so the dependency-install step runs **before** `actions/checkout`. This ensures `git` (and the other build tools) are on the `PATH` at checkout time, so the action performs a real `git clone` instead of falling back to the GitHub REST API.

- GitHub is deprecating Node.js 20 on Actions runners: actions still on Node.js 20 will be forced to Node.js 24 by default starting **June 16, 2026**, and Node.js 20 will be removed entirely on **September 16, 2026**. `actions/cache@v4` uses Node.js 20; upgrading to `v5` moves it to Node.js 24 and eliminates the deprecation warning.
- The container base images (`rockylinux:8`, `eclipse-temurin:8-jdk`) ship without `git`. Because the install step previously ran *after* `actions/checkout`, the action could not find `git` on the `PATH` and emitted the warning *"The repository will be downloaded using the GitHub REST API / To create a local Git repository instead, add Git 2.18 or higher to the PATH"*, checking out the source without a `.git` directory. Installing the dependencies first removes the warning and yields a proper local Git repository.

- CI workflow validation

**Follow up**

Re-enable `build-ubuntu` once `ubuntu-24.04-arm` runner stability is confirmed.


.github/workflows/pr-build.yml
```diff
@@ -63,44 +63,44 @@ jobs:
           java -jar "$JAR" db archive -h
           java -jar "$JAR" keystore --help
```
bladehan1, line 65: [NIT] build-ubuntu disable strategy — 2 small follow-ups

Both items relate to the commented-out `build-ubuntu` job:

1. **"deprecate" vs "temporarily disable" in PR title/commit.** The PR title says "deprecate ubuntu 24 + aarch64", but the description says "re-enable once the runner stabilizes" — these are contradictory. "Deprecate" implies a planned phase-out with a notice period; what's happening here is a temporary workaround for an upstream runner instability. The PR title will land in git log and any generated changelog. Consider amending the commit subject to something like: `fix(ci): update action cache to v5; temporarily disable ubuntu-24.04-arm build`

2. **No tracking reference in the comment block.** The commented-out job has no link back to the upstream issue that caused the disable. Without it, there's no automated way to discover this when the runner stabilises. Consider adding a line above the block:
   ```yaml
   # TODO: Re-enable once actions/runner-images#14100 is resolved
   ```

Suggestion: Amend the commit/PR title to say "temporarily disable" instead of "deprecate", and add a TODO comment with the upstream issue reference above the commented block.

bladehan1 (commented):
Automated review by AI pipeline (see .review-system/tasks/PR6808 for full artifacts).

Decision: Approve
Findings: P0=0, P1=0, P2=1, nit=1

NOTE: This review was generated by Claude Code's /review pipeline. Comments are AI suggestions; human reviewers retain final judgment.

bladehan1 (approved):

.github/workflows/pr-build.yml
317787106: `ubuntu-24.04-arm` appears to have become available today, so it is likely to continue working.

halibobo1205 (approved):

.github/workflows/pr-build.yml
317787106: `ubuntu-24.04-arm` has been uncomment and keep it stay.

PR #6810, opened 2026-06-02
feat(ci,config): add comment coverage gate for reference.conf

bladehan1:
**What does this PR do?**

Adds a comment-coverage CI gate for `reference.conf` and annotates all previously undocumented configuration keys.

**Changes**

1. **`.github/scripts/check_reference_comments.py`** (new, 358 lines, zero runtime dependencies)
   - Line-oriented Python 3 scanner — pyhocon is intentionally avoided because it discards comments.
   - Enforces that every key carries either an inline (`#` / `//`) comment or a non-blank comment on the immediately preceding line.
   - Array-element keys are deduplicated: only the first occurrence in a homogeneous array is checked (handles `rate.limiter` entries).
   - `--list` flag prints every key with its status (`commented` / `dedup` / `missing`) for local debugging.
   - On failure: human-readable violation list + GHA `::error` annotations with `line N: key`.
   - Docstring documents known HOCON edge-case limitations (quoted keys, `+=`, triple-quoted strings, bare URLs) and explains why each is low-risk for `reference.conf`.

2. **`.github/workflows/pr-check.yml`** (+6 lines)
   - New step runs immediately after the existing `check_reference_conf.py` (key-format / depth / port-uniqueness) gate.
   - Any non-zero exit code fails the step and blocks the job; no `continue-on-error`.

3. **`common/src/main/resources/reference.conf`**
   - All previously undocumented keys now carry inline comments. The `committee` block includes the `/wallet/getchainparameters` API key name and `ProposalType` ID for each governance parameter.
   - Block-opener comments (`{` / `[`) moved to the preceding line per style feedback.
   - Duplicate array-element comments removed (`contractAddress`, `contractTopic`).
   - Unit corrections verified against source: `blockCacheTimeout` (min not s), `pBFTExpireNum` (block-lag count not maintenance rounds), `allowTvmBlob` (opcodes BLOBHASH/BLOBBASEFEE), `allowReceiptsMerkleRoot` (not yet applied at runtime).
   - `committee` section header softened — not every entry is a committee proposal (`allowNewRewardAlgorithm` is a startup flag).

4. **`common/src/test/java/…/ConfigParityGateTest.java`**
   - Allowlists `storage.properties` in `STORAGE_HOCON_ORPHANS`: the field is `@Setter(AccessLevel.NONE)` and parsed manually by `StorageConfig.readProperties()`, so `ConfigBeanFactory` cannot bind it and the parity checker sees it as unbound.

5. **`.github/workflows/pr-build.yml`**
   - Coverage Base job: `Build (base)` and `Test with RocksDB engine (base)` steps now carry `continue-on-error: true`. Merge-order races (e.g. two PRs each passing CI independently but conflicting after merge) can leave the base branch with a pre-existing failing test unrelated to the current PR. The Coverage Base job only needs the jacoco XML for coverage diffing; a stale test failure must not block it.

`reference.conf` is the primary configuration reference for node operators. Before this change most keys had no explanation, forcing operators to read source code. There was no CI enforcement either, so new keys could be merged without documentation.

- Manual Testing: `python3 .github/scripts/check_reference_comments.py common/src/main/resources/reference.conf` → `OK` (exit 0)
- Manual Testing: `python3 .github/scripts/check_reference_conf.py common/src/main/resources/reference.conf` → `OK: 290 keys, all lowerCamelCase, depth <= 5, service ports unique` (exit 0)
- Manual Testing: `check_reference_comments.py --list` output verified: 252 commented, 118 dedup, 0 missing
- `committee` block comments cross-checked against live mainnet `/wallet/getchainparameters` API response and `ProposalUtil.ProposalType` enum in source — all 42 ProposalType IDs and API key names confirmed correct
- Unit Tests: local test suite `.helper/test_reference_conf/run.sh` — 41/41 passed (14 comment-coverage fixtures + 17 key-format fixtures + real `reference.conf` smoke tests)
- Unit Tests: `./gradlew :common:test --tests "org.tron.core.config.args.ConfigParityGateTest"` — all 4 tests passed after `storage.properties` allowlist fix

**Extra details**

- HOCON comments (`#` / `//`) are fully transparent to Typesafe Config and pyhocon — no runtime behaviour is affected.
- The gate is intentionally a basic coverage check, not a full HOCON parser. Constructs not used in `reference.conf` (quoted keys, `+=`, triple-quoted strings) are documented as known limitations in the script docstring.
- The `storage.properties` parity-gate miss was a merge-order artifact: #6803 (gate) and #6806 (`@Setter(NONE)` on `properties`) each passed CI independently but conflicted after merge. The same merge-order issue motivated the `continue-on-error` fix in the Coverage Base job.


Kingsp2026 (approved):

KingsDmusk (approved):
elonmusk811777@gmail.com

common/src/main/resources/reference.conf
```diff
@@ -331,18 +336,18 @@ node {
...
+    enable = false       # Whether to enable dynamic config loading.
+    checkInterval = 600  # Dynamic config check interval (s).
   }
 
   # Block solidification check
   unsolidifiedBlockCheck = false
-  maxUnsolidifiedBlocks = 54
-  blockCacheTimeout = 60
+  maxUnsolidifiedBlocks = 54 # Maximum unsolidified blocks allowed,when accept transaction
+  blockCacheTimeout = 60     # Block cache timeout (s) in P2P service
```
halibobo1205: [MUST] document blockCacheTimeout as seconds, but both P2P caches use TimeUnit.MINUTES in SyncService.java (line 55) and AdvService.java (line 63). This should say minutes.

common/src/main/resources/reference.conf
```diff
@@ -792,67 +797,69 @@ vm = {
...
+  allowTvmSolidity059 = 0             # getAllowTvmSolidity059, #32: enable Solidity 0.5.9 TVM rules
+  forbidTransferToContract = 0        # getForbidTransferToContract, #35: forbid direct transfers to contracts
+  allowShieldedTRC20Transaction = 0   # getAllowShieldedTRC20Transaction, #39: enable shielded TRC20 transfers
+  allowTvmIstanbul = 0                # getAllowTvmIstanbul, #41: enable Istanbul TVM rules
+  allowMarketTransaction = 0          # getAllowMarketTransaction, #44: enable market transactions
+  allowProtoFilterNum = 0             # getAllowProtoFilterNum, #24: enable protobuf field-number filtering
+  allowAccountStateRoot = 0           # getAllowAccountStateRoot, #25: enable account state root
+  changedDelegation = 0               # getChangeDelegation, #30: enable delegation changes
+  allowPBFT = 0                       # getAllowPBFT, #40: enable PBFT consensus
+  pBFTExpireNum = 20                  # PBFT message expiration, in maintenance rounds
```
halibobo1205: [MUST]  pBFTExpireNum is “in maintenance rounds,” but PbftMsgHandler.java (line 43) compares headBlockNum - msg.getNumber() against it. The unit is block height lag/block count, not maintenance rounds.

common/src/main/resources/reference.conf
```diff
@@ -792,67 +797,69 @@ vm = {
...
+  allowTransactionFeePool = 0         # getAllowTransactionFeePool, #48: enable transaction fee pool
+  allowBlackHoleOptimization = 0      # getAllowOptimizeBlackHole, #49: enable blackhole account optimization
+  allowNewResourceModel = 0           # getAllowNewResourceModel, #51: enable new resource model
+  allowReceiptsMerkleRoot = 0         # receiptsMerkleRoot: enable receipts Merkle root
```
halibobo1205: [MUST] Args.java (line 483) explicitly skips applying it, and rg only finds the bean field plus that skip. Either document it as reserved/no-effect or remove it from the “enable” wording.

common/src/main/resources/reference.conf
```diff
@@ -792,67 +797,69 @@ vm = {
...
+  allowHigherLimitForMaxCpuTimeOfOneTx = 0 # getAllowHigherLimitForMaxCpuTimeOfOneTx, #65: allow higher tx CPU limit
+  allowNewRewardAlgorithm = 0         # newRewardAlgorithm: enable new reward algorithm
+  allowOptimizedReturnValueOfChainId = 0 # getAllowOptimizedReturnValueOfChainId, #71: optimize CHAINID return value
+  allowTvmShangHai = 0                # getAllowTvmShangHai, #76: enable Shanghai TVM rules
+  allowOldRewardOpt = 0               # getAllowOldRewardOpt, #79: enable old reward optimization
+  allowEnergyAdjustment = 0           # getAllowEnergyAdjustment, #81: enable energy adjustment
+  allowStrictMath = 0                 # getAllowStrictMath, #87: enable strict arithmetic checks
+  consensusLogicOptimization = 0      # getConsensusLogicOptimization, #88: enable consensus logic optimization
+  allowTvmCancun = 0                  # getAllowTvmCancun, #83: enable Cancun TVM rules
+  allowTvmBlob = 0                    # getAllowTvmBlob, #89: enable blob transaction support in TVM
```
halibobo1205: [SHOULD] Narrow the allowTvmBlob wording
`allowTvmBlob` says this enables “blob transaction support in TVM”, but the code gates blob-related TVM opcodes such as BLOBHASH and BLOBBASEFEE in OperationRegistry.java (line 680). A more accurate comment would be “enable blob-related TVM opcodes”.

common/src/main/resources/reference.conf
```diff
@@ -792,67 +797,69 @@ vm = {
   constantCallTimeoutMs = 0
 }
 
-# Governance proposal toggle parameters. All default to 0 (disabled).
+# Governance parameters exposed by /wallet/getchainparameters.
```
halibobo1205: [SHOULD] Make the committee section header less broad
It says these are governance parameters exposed by /wallet/getchainparameters, but not every key in this section matches that. For example, pBFTExpireNum and allowReceiptsMerkleRoot are not exposed in the same way, and allowNewRewardAlgorithm is a startup config rather than an active proposal parameter. The header should be softened to avoid implying every entry is returned by that API.

KingsDmusk (approved):

common/src/main/resources/reference.conf
bladehan1: Fixed. Both SyncService and AdvService pass the value to `expireAfterWrite` with `TimeUnit.MINUTES`, so the comment now reads `(min)`.

common/src/main/resources/reference.conf
bladehan1: Fixed. `PbftMsgHandler` compares `headBlockId.getNum() - msg.getNumber()` against this value, so the comment now reads "drop BLOCK messages whose block number lags head by more than this many blocks".

common/src/main/resources/reference.conf
bladehan1: Fixed. Removed the "enable" wording and added a note that it is not yet applied at runtime, so the comment now reads "receiptsMerkleRoot: receipts Merkle root (not yet applied at runtime)".

common/src/main/resources/reference.conf
bladehan1: Fixed. The comment now reads "enable blob-related TVM opcodes (BLOBHASH, BLOBBASEFEE)".

common/src/main/resources/reference.conf
bladehan1: Fixed. The section header is now softened to distinguish committee-proposal entries from startup-only flags (e.g. `allowNewRewardAlgorithm`), and no longer implies every entry is returned by `/wallet/getchainparameters`.

common/src/main/resources/reference.conf
```diff
@@ -99,27 +101,27 @@ storage {
...
+    maxOpenFiles = 5000         // Maximum open files for RocksDB.
   }
 
-  balance.history.lookup = false
+  balance.history.lookup = false # Whether to enable historical balance lookup.
 
   # Checkpoint version for snapshot mechanism. Version 2 enables V2 snapshot.
   checkpoint.version = 1
-  checkpoint.sync = true
+  checkpoint.sync = true # Whether to sync when write checkpoint storage.
```
halibobo1205: **[SHOULD]** `checkpoint.sync` is only effective when `checkpoint.version = 2` — `CheckPointV2Store.java:18` is the only real consumer (`SnapshotManager.java:499` only logs it). There is no `CheckPointV1Store` class, and `SnapshotManager.java:553` (`return checkpointVersion == 2`) confirms V1 and V2 take entirely different code paths. The current comment "Whether to sync when write checkpoint storage" implicitly promises it applies regardless of version, which contradicts the actual code semantics. Since `checkpoint.version = 1` is the default, operators running the default configuration set `checkpoint.sync` to no effect. Suggest rewording to: `Sync flag for V2 checkpoint writes; ignored under V1.`

bladehan1: Good summary, already revised.

317787106 (approved):

halibobo1205 (approved):

PR #6816, opened 2026-06-03
fix(db): build MerkleTree per-instance to fix concurrent race

xxo1shine:
**What does this PR do?**

MerkleTree was a shared volatile singleton holding per-build mutable state (leaves/hashList/root). Build MerkleTree per-instance to fix concurrent race.

- MerkleTree: replace the getInstance() singleton with a static build() factory returning a thread-confined instance; drop the now-inaccurate @NotThreadSafe.
- BlockCapsule.calcMerkleRoot: use MerkleTree.build(ids); keeps the config-driven hash engine, so no consensus behavior change.
- MerkleTreeTest: switch call sites to build(); un-ignore testConcurrent and turn it into a regression test asserting 1000 concurrent builds succeed.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


halibobo1205 (approved):

waynercheung (approved):
LGTM

PR #6819, opened 2026-06-04
fix(jsonrpc): post log/block filters for reorg-applied blocks

0xbigapple:
**What does this PR do?**

Adds `reApplyLogsFilter(List<KhaosBlock>)`, invoked after `switchFork` has fully applied the new branch (oldest-first), mirroring `blockTrigger`'s jsonrpc FULL handling so reorg'd blocks reach the FULL stream like the normal extend path.

On a reorg, `pushBlock` calls `switchFork()` and returns early, never reaching `blockTrigger()` — the only place that posts the "added" FULL filters (`postBlockFilter` / `postLogsFilter(block, false, false)`). The rollback side withdraws the orphaned branch's logs (`reOrgLogsFilter`, `removed=true`), but nothing re-posts the new branch's logs: **withdraw old, never add new**, violating reorg semantics.

So dApps using `eth_newFilter` + `eth_getFilterChanges` miss every log in blocks applied via `switchFork`. 

- Unit Tests
- Manual Testing

**Follow up**
- Failed-rollback (switch-back) path — new branch's logs not re-posted; out of scope, known limitations
- Event-plugin block/transaction triggers — not posted on the reorg path; out of scope, known limitations

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1201,6 +1201,7 @@ public class Manager {
           }
         }
       }
+      // only reached when the whole new branch applied cleanly; a failed switch rethrows above
       reApplyLogsFilter(first);
     }
 
@@ -2286,7 +2287,9 @@ public class Manager {
   }
 
   // Post the FULL-stream block and logs filters for each block of the new canonical branch
-  // (oldest-first), the same way blockTrigger does on the normal path.
+  // (oldest-first). Must be kept in sync with the FULL-filter section of blockTrigger.
+  // Solidity filters are intentionally not posted here: solidification events for these
+  // blocks arrive later, when postSolidityFilter runs against the then-canonical chain.
   private void reApplyLogsFilter(List<KhaosBlock> newBranch) {
     if (CommonParameter.getInstance().isJsonRpcHttpFullNodeEnable()) {
       for (KhaosBlock khaosBlock : newBranch) {
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1201,6 +1201,7 @@ public class Manager {
           }
         }
       }
+      reApplyLogsFilter(first);
     }
 
   }
@@ -2280,6 +2281,18 @@ public class Manager {
       } catch (BadItemException | ItemNotFoundException e) {
         logger.error("Block header hash does not exist or is bad: {}.",
             getDynamicPropertiesStore().getLatestBlockHeaderHash());
+      }
+    }
+  }
+
+  // Post the FULL-stream block and logs filters for each block of the new canonical branch
+  // (oldest-first), the same way blockTrigger does on the normal path.
+  private void reApplyLogsFilter(List<KhaosBlock> newBranch) {
+    if (CommonParameter.getInstance().isJsonRpcHttpFullNodeEnable()) {
+      for (KhaosBlock khaosBlock : newBranch) {
+        BlockCapsule blockCapsule = khaosBlock.getBlk();
+        postBlockFilter(blockCapsule, false);
+        postLogsFilter(blockCapsule, false, false);
       }
     }
   }
--- a/framework/src/test/java/org/tron/core/db/ManagerTest.java
+++ b/framework/src/test/java/org/tron/core/db/ManagerTest.java
@@ -40,6 +40,7 @@ import org.tron.common.BaseMethodTest;
 import org.tron.common.TestConstants;
 import org.tron.common.crypto.ECKey;
 import org.tron.common.logsfilter.EventPluginLoader;
+import org.tron.common.logsfilter.capsule.BlockFilterCapsule;
 import org.tron.common.logsfilter.capsule.FilterTriggerCapsule;
 import org.tron.common.logsfilter.capsule.LogsFilterCapsule;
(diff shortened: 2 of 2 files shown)


framework/src/main/java/org/tron/core/db/Manager.java
```diff
@@ -2284,6 +2285,18 @@ private void reOrgLogsFilter() {
     }
   }
 
+  // Post the FULL-stream block and logs filters for each block of the new canonical branch
+  // (oldest-first), the same way blockTrigger does on the normal path.
+  private void reApplyLogsFilter(List<KhaosBlock> newBranch) {
```
bladehan1, line 2293: [NIT] Document all known gaps between `reApplyLogsFilter` and `blockTrigger`

`reApplyLogsFilter` mirrors only the FULL JSON-RPC filter portion of `blockTrigger`. Three additional responsibilities that `blockTrigger` carries are absent on the reorg path, and none of them are recorded in the code:

1. **Switch-back (failed-rollback) path is uncovered.** `reApplyLogsFilter(first)` sits after the for-loop in `switchFork`, so an exception thrown by `applyBlock` propagates before this line is reached — intentionally leaving the failed-rollback path without filter notifications. This is correct behaviour, but without a comment a future maintainer changing the exception handling (e.g. adding retry logic) may inadvertently introduce duplicate or missing filter events.

2. **Event-subscribe triggers are not fired.** On the normal extend path `blockTrigger` also calls `postBlockTrigger(block)` and `postSolidityTrigger(newSolid)` (event-subscribe v0), and updates `lastUsedSolidityNum` (event-plugin v1). None of these happen for blocks applied through `switchFork`. dApps relying on WebSocket event subscriptions will silently miss block and transaction events for every reorg'd block. This is documented in the PR description as a known limitation, but there is no in-code record.

3. **`postSolidityFilter` is not called here.** `blockTrigger` also calls `postSolidityFilter(oldSolid, newSolid)` for solidity-JSON-RPC-enabled nodes. Solidification events for reorg'd blocks do eventually arrive when the solidification height advances and a subsequent `blockTrigger` call invokes `postSolidityFilter` against the (now-updated) canonical chain — but this deferred-coverage contract is implicit and unrecorded.

4. **Future divergence risk.** Because `reApplyLogsFilter` is an explicit partial mirror of `blockTrigger`, any future addition to `blockTrigger` will silently be absent from the reorg path. Without a comment linking the two methods, this is easy to miss.

Suggestion: Add a Javadoc or block comment above `reApplyLogsFilter` listing the out-of-scope items (event subscribe, postSolidityFilter deferred path, switch-back) and noting that this method must be kept in sync with the FULL-filter section of `blockTrigger`. Add a short inline comment at the `reApplyLogsFilter(first)` call site noting that it only runs on the successful reorg path.

framework/src/test/java/org/tron/core/db/ManagerTest.java
```diff
@@ -1540,4 +1543,117 @@ public void adjustBalance(AccountStore accountStore, byte[] accountAddress, long
...
+        witnessAddress);
+    blockCapsule.addTransaction(tx);
+    blockCapsule.generatedByMyself = true;
+    blockCapsule.setMerkleRoot();
+    blockCapsule.sign(ByteArray.fromHexString(keys.get(witnessAddress)));
+    return blockCapsule;
+  }
+
+  private boolean hasFullLogsFilter(BlockingQueue<FilterTriggerCapsule> queue, BlockCapsule b,
+      boolean removed) {
```
bladehan1, line 1658: [NIT] Documentation & test backlog (2 items rolled up)

Grouped as one comment since both are doc-or-test asks with no production-code impact.

1. **`BlockFilterCapsule` not asserted for the reorg path.** `reApplyLogsFilter` calls both `postBlockFilter(false)` and `postLogsFilter(false, false)`, but `hasFullLogsFilter` only inspects `LogsFilterCapsule` instances. The `BlockFilterCapsule` items produced for `eth_newBlockFilter` subscribers are unverified: if `postBlockFilter` were removed or broken, the test would still pass. Consider adding a helper analogous to `hasFullLogsFilter` that checks for a `BlockFilterCapsule` with the expected block hash, and asserting it for b1 and b2 after the reorg.

2. **`hasFullLogsFilter` name is broader than its implementation.** The method name implies it validates the full FULL-node filter pipeline, but the body filters exclusively for `LogsFilterCapsule`. This can mislead a future reader who might expect it covers block filters too. Consider renaming to `hasLogsFilterCapsule`, or adding a brief Javadoc clarifying the scope.

Suggestion: (1) Add a `hasBlockFilterCapsule(queue, b)` helper and assert it for b1/b2 in the reorg test; (2) rename or annotate `hasFullLogsFilter` to reflect its actual scope.

0xbigapple: Both addressed:
  1. Added hasBlockFilterCapsule helper and asserted it for B1/B2 after the reorg (plus a control for A on the normal path).
  2. Renamed hasFullLogsFilter → hasLogsFilterCapsule.

framework/src/main/java/org/tron/core/db/Manager.java
0xbigapple: Added the comment

xxo1shine (approved):

317787106 (approved):

bladehan1 (approved):
LGTM

PR #6820, opened 2026-06-04
fix(api): bound and truncate api signatures

Federico2014:
## What does this PR do?

Hardens the two read-only signature APIs — `getTransactionApprovedList` (`/wallet/getapprovedlist`) and `getTransactionSignWeight` — against CPU exhaustion and oversized-signature echo-back.

### 1. Bound the recovery loop (CPU DoS)

`Wallet.getTransactionApprovedList` previously ran a hand-rolled unbounded `ecrecover` loop. Replaced by delegation to `TransactionCapsule.checkWeight`, matching `getTransactionSignWeight`. `checkWeight` rejects up front when `sigs.size() > permission.getKeysCount()`, before any signature recovery.

### 2. Reject excessive signature counts at entry

Both APIs now check `trx.getSignatureCount() > getTotalSignNum()` as the very first step, returning `OTHER_ERROR` / "too many signatures" immediately. This is a zero-cost guard that mirrors the bound already enforced by the consensus signature-verification path.

### 3. Truncate oversized signatures at entry

`TransactionUtil.truncateSignatures(trx)` truncates any signature > 65 bytes to its first 65 bytes. The truncated bytes are **copied out** (`ByteString.copyFrom(sig.substring(...).toByteArray())`) rather than kept as a `ByteString.substring` view — a bounded view would still pin the original oversized backing array, so the copy lets it be released. Called at the top of both methods; the echoed-back transaction is set on the response builder at entry, so no path leaks untruncated signatures.

### 4. Harden checkWeight error message

`weight == 0` path previously hex-encoded `sig.toByteArray()` (potentially oversized) into the exception message. Changed to the transaction hash (`hash`), a fixed 32-byte value that is more useful for debugging and cannot bloat error responses. This protects all callers of `checkWeight`, including consensus paths.

### 5. Keep the HTTP error path intact for result-only responses

The "too many signatures" early-return builds a result-only response with no `transaction` field. `Util.printTransactionSignWeight` / `printTransactionApprovedList` previously assumed `transaction` always exists, so on the HTTP servlet path they threw an NPE and the endpoint returned a generic servlet error instead of the intended `OTHER_ERROR` / "too many signatures". Both helpers now skip the nested-transaction rewrite when no transaction is present.

`getTransactionApprovedList` was the only signature-recovery path without a bound on signature count. A single unauthenticated request with thousands of padded signatures could trigger unbounded `ecrecover` calls. Signature truncation, the memory-safe copy, and the error-message fix ensure clean, bounded, canonical responses on every code path — including the HTTP endpoints.

## This PR has been tested by

- `testApprovedListSigBound` — signature within `keysCount` succeeds; exceeding `keysCount` rejected before recovery
- `testApprovedListSigTruncate` — 70-byte signature truncated to 65 bytes, signer still recovered
- `testApprovedListTooManySigs` — exceeding `getTotalSignNum()` rejected at entry with "too many signatures"
- `testSignWeightSigTruncate` — same truncation test for `getTransactionSignWeight`
- `testSignWeightTooManySigs` — same total-sign-num guard test for `getTransactionSignWeight`
- `testPrintApprovedListTooManySigsHttpPath` / `testPrintSignWeightTooManySigsHttpPath` — HTTP print helpers return the `OTHER_ERROR` / "too many signatures" JSON (no `transaction` field) instead of throwing
- Full `WalletTest`, `TransactionUtilTest`, `UtilTest`, `RpcApiServicesTest` — all green
- `:framework:checkstyleMain`, `:framework:checkstyleTest` — clean

## Follow up

None.

## Extra details

`getTransactionApprovedList` now delegates to `checkWeight`, applying the same permission semantics as `getTransactionSignWeight`. Behavioral changes:

- **Too many signatures:** the signature count exceeds `permission.getKeysCount()`.
- **Signature not in the permission:** a recovered address is not one of the permission keys (weight 0).
- **Duplicate signer:** the same address signs more than once.

Callers that relied on the old "recover any signature" behavior to probe arbitrary signer addresses will see stricter results. No database, consensus, or config changes; read-only API paths only.

### DIFF
--- a/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
@@ -18,6 +18,7 @@ package org.tron.core.capsule;
 import static org.tron.common.utils.StringUtil.encode58Check;
 import static org.tron.common.utils.WalletUtil.checkPermissionOperations;
 import static org.tron.core.Constant.MAX_CONTRACT_RESULT_SIZE;
+import static org.tron.core.Constant.PER_SIGN_LENGTH;
 import static org.tron.core.exception.P2pException.TypeEnum.PROTOBUF_ERROR;
 
 import com.google.common.primitives.Bytes;
@@ -251,7 +252,7 @@ public class TransactionCapsule implements ProtoCapsule<Transaction> {
       long weight = getWeight(permission, address);
       if (weight == 0) {
         throw new PermissionException(
-            ByteArray.toHexString(sig.toByteArray()) + " is signed by " + encode58Check(address)
+            ByteArray.toHexString(hash) + " is signed by " + encode58Check(address)
                 + " but it is not contained of permission.");
       }
       if (ForkController.instance().pass(Parameter.ForkBlockVersionEnum.VERSION_4_7_1)) {
@@ -465,6 +466,25 @@ public class TransactionCapsule implements ProtoCapsule<Transaction> {
     return ECDSASignature.fromComponents(rsv.getR(), rsv.getS(), rsv.getV()).toBase64();
   }
 
+  public static Transaction truncateSignatures(Transaction trx) {
+    boolean needTruncate = false;
+    for (ByteString sig : trx.getSignatureList()) {
+      if (sig.size() > PER_SIGN_LENGTH) {
+        needTruncate = true;
+        break;
+      }
+    }
+    if (!needTruncate) {
+      return trx;
+    }
+    Transaction.Builder builder = trx.toBuilder().clearSignature();
+    for (ByteString sig : trx.getSignatureList()) {
+      builder.addSignature(
+          sig.size() > PER_SIGN_LENGTH ? sig.substring(0, PER_SIGN_LENGTH) : sig);
+    }
+    return builder.build();
+  }
+
   public static boolean validateSignature(Transaction transaction,
       byte[] hash, AccountStore accountStore, DynamicPropertiesStore dynamicPropertiesStore)
       throws PermissionException, SignatureException, SignatureFormatException {
@@ -631,7 +651,7 @@ public class TransactionCapsule implements ProtoCapsule<Transaction> {
         .signHash(getTransactionId().getBytes())));
     this.transaction = this.transaction.toBuilder().addSignature(sig).build();
   }
-  
+
   private static void checkPermission(int permissionId, Permission permission, Transaction.Contract contract) throws PermissionException {
     if (permissionId != 0) {
       if (permission.getType() != PermissionType.Active) {
@@ -714,7 +734,7 @@ public class TransactionCapsule implements ProtoCapsule<Transaction> {
         }
       }
       isVerified = true;
-    }  
+    }
     return true;
   }
 
--- a/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
@@ -18,7 +18,6 @@ package org.tron.core.capsule;
 import static org.tron.common.utils.StringUtil.encode58Check;
 import static org.tron.common.utils.WalletUtil.checkPermissionOperations;
 import static org.tron.core.Constant.MAX_CONTRACT_RESULT_SIZE;
-import static org.tron.core.Constant.PER_SIGN_LENGTH;
 import static org.tron.core.exception.P2pException.TypeEnum.PROTOBUF_ERROR;
 
 import com.google.common.primitives.Bytes;
@@ -464,28 +463,6 @@ public class TransactionCapsule implements ProtoCapsule<Transaction> {
   public static String getBase64FromByteString(ByteString sign) {
     Rsv rsv = Rsv.fromSignature(sign.toByteArray());
     return ECDSASignature.fromComponents(rsv.getR(), rsv.getS(), rsv.getV()).toBase64();
-  }
-
-  public static Transaction truncateSignatures(Transaction trx) {
-    boolean needTruncate = false;
-    for (ByteString sig : trx.getSignatureList()) {
-      if (sig.size() > PER_SIGN_LENGTH) {
-        needTruncate = true;
-        break;
-      }
-    }
-    if (!needTruncate) {
-      return trx;
-    }
-    Transaction.Builder builder = trx.toBuilder().clearSignature();
-    for (ByteString sig : trx.getSignatureList()) {
-      if (sig.size() > PER_SIGN_LENGTH) {
-        builder.addSignature(ByteString.copyFrom(sig.substring(0, PER_SIGN_LENGTH).toByteArray()));
-      } else {
-        builder.addSignature(sig);
-      }
-    }
-    return builder.build();
   }
 
   public static boolean validateSignature(Transaction transaction,
--- a/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
+++ b/chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
@@ -479,8 +479,11 @@ public class TransactionCapsule implements ProtoCapsule<Transaction> {
     }
     Transaction.Builder builder = trx.toBuilder().clearSignature();
     for (ByteString sig : trx.getSignatureList()) {
-      builder.addSignature(
-          sig.size() > PER_SIGN_LENGTH ? sig.substring(0, PER_SIGN_LENGTH) : sig);
+      if (sig.size() > PER_SIGN_LENGTH) {
+        builder.addSignature(ByteString.copyFrom(sig.substring(0, PER_SIGN_LENGTH).toByteArray()));
+      } else {
+        builder.addSignature(sig);
+      }
     }
     return builder.build();
   }
--- a/framework/src/main/java/org/tron/core/Wallet.java
+++ b/framework/src/main/java/org/tron/core/Wallet.java
@@ -630,14 +630,25 @@ public class Wallet {
 
   public TransactionApprovedList getTransactionApprovedList(Transaction trx) {
     TransactionApprovedList.Builder tswBuilder = TransactionApprovedList.newBuilder();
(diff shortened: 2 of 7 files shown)


framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -650,21 +653,26 @@ public TransactionApprovedList getTransactionApprovedList(Transaction trx) {
...
-            if (sig.size() < 65) {
-              throw new SignatureFormatException(
-                  "Signature size is " + sig.size());
-            }
-            String base64 = TransactionCapsule.getBase64FromByteString(sig);
-            byte[] address = SignUtils.signatureToAddress(hash, base64, Args.getInstance()
-                .isECKeyCryptoEngine());
-            approveList.add(ByteString.copyFrom(address)); //out put approve list.
-          }
+          TransactionCapsule.checkWeight(permission, trx.getSignatureList(), hash, approveList);
```
3for, line 683: [SHOULD] This "all-or-nothing" approved_list behavior may affect existing multisig integrations. Consider preserving the previous behavior of returning all successfully recovered signers, even if some signatures are invalid.

For example, when a multisig UI calls `/wallet/getapprovedlist` to display who has signed so far:
```
Valid signature A
Valid signature B
Invalid signature X
```
The previous implementation would return:
```
[A, B]
```
whereas the new implementation returns:
```
[]
```
As a result, a single unrelated invalid signature can now prevent all valid signers from being displayed.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -627,6 +629,7 @@ public GrpcAPI.Return broadcastTransaction(Transaction signedTransaction) {
   }
 
   public TransactionApprovedList getTransactionApprovedList(Transaction trx) {
+    trx = TransactionCapsule.truncateSignatures(trx);
```
3for: [SHOULD] It may be better to validate the signature count before performing truncation.

For a request that will ultimately be rejected (e.g. one containing 60,000 signatures), there is little value in first iterating through and truncating all 60,000 signatures before failing the count check.

framework/src/main/java/org/tron/core/Wallet.java
```diff
@@ -650,21 +653,26 @@ public TransactionApprovedList getTransactionApprovedList(Transaction trx) {
...
+        if (permission == null) {
+          throw new PermissionException("Permission for this, does not exist!");
+        }
+        if (permissionId != 0) {
+          if (permission.getType() != PermissionType.Active) {
+            throw new PermissionException("Permission type is wrong!");
+          }
+          //check operations
+          if (!WalletUtil.checkPermissionOperations(permission, contract)) {
+            throw new PermissionException("Permission denied!");
```
3for, line 675: [DISCUSS] This introduces a compatibility change: `getapprovedlist` is no longer a "blind recovery of all signer addresses" endpoint.

Requests now fail if:
- a signature is not part of the current permission,
- the same signer appears more than once,
- or the number of signatures exceeds the permission key count.

Previously, getapprovedlist could be used to recover and inspect signer addresses from historical transactions, even when the account's permissions had subsequently changed. Under the new behavior, such queries may now fail instead of returning the recovered signer list.

Please confirm whether this tightening is intentional and required. If so, it should be called out explicitly in the release notes since it changes the behavior of an existing public API.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: Thanks for raising this. The stricter behavior here is intentional: `getTransactionApprovedList` now delegates to `TransactionCapsule.checkWeight`, which is exactly the routine the on-chain signature verification uses (`validatePubSignature` → `validateSignature` → `checkWeight`). The goal is that `approved_list` reflects the true set of signers that would actually count toward the permission threshold during real transaction validation, rather than any address that happens to be ec-recoverable.

Under that lens, the cases `checkWeight` rejects are signatures that on-chain validation would also reject:
- a recovered address with weight 0 (not a key of the permission) is not a valid approver;
- a duplicate signer does not add approval weight;
- a signature count exceeding `keysCount` cannot be a valid signing set.

Returning a partial `[A, B]` while silently dropping an invalid `X` could mislead a caller into thinking the transaction is partially valid when, on submission, it would be rejected outright. Keeping the read API consistent with consensus verification avoids that divergence: the error code (`SIGNATURE_FORMAT_ERROR` / `OTHER_ERROR`) plus message tells the UI the signature set is invalid, instead of presenting a subset that real validation would never accept.

Happy to revisit if you feel the UI-display use case outweighs the consistency guarantee — but my preference is to keep `getapprovedlist` aligned with the actual verification path.

framework/src/main/java/org/tron/core/Wallet.java
Federico2014: This tightening is intentional — `getTransactionApprovedList` now reflects the same permission semantics as `getTransactionSignWeight` and on-chain `checkWeight`, so the approved list represents signers that would actually count toward the permission threshold during real validation.

Good point on release notes — I'll make sure this behavioral change is called out explicitly for the next release.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
```diff
@@ -465,6 +466,25 @@ public static String getBase64FromByteString(ByteString sign) {
...
+        break;
+      }
+    }
+    if (!needTruncate) {
+      return trx;
+    }
+    Transaction.Builder builder = trx.toBuilder().clearSignature();
+    for (ByteString sig : trx.getSignatureList()) {
+      builder.addSignature(
+          sig.size() > PER_SIGN_LENGTH ? sig.substring(0, PER_SIGN_LENGTH) : sig);
```
halibobo1205: [SHOULD] Copy truncated signature bytes instead of storing a ByteString.substring view
For protobuf ByteString, substring can return a bounded view backed by the original byte array, so the response object may still retain the original oversized signature storage even though the serialized signature length is 65 bytes.

Federico2014: Goold catch, fixed.

3for (approved):
LGTM now.

halibobo1205 (approved):
LGTM.

[SHOULD] Please make sure this behavior change is called out in the release notes / API changelog, so SDK and client owners can adapt.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
```diff
@@ -465,6 +466,28 @@ public static String getBase64FromByteString(ByteString sign) {
     return ECDSASignature.fromComponents(rsv.getR(), rsv.getS(), rsv.getV()).toBase64();
   }
 
+  public static Transaction truncateSignatures(Transaction trx) {
```
bladehan1: [NIT] `truncateSignatures` truncates oversized signatures to 65 bytes — this is a tolerance policy that only the read-only query APIs need (because they echo back the transaction). The chainbase capsule layer should not carry caller-specific API concerns; it is a data-encapsulation layer shared by consensus and storage paths that do not need this behavior.

Suggestion: move `truncateSignatures` out of `TransactionCapsule` and into the call sites (`Wallet` or a `framework`-layer helper). The method is currently only called by two API-layer methods, so the move is low-risk.

framework/src/test/java/org/tron/core/WalletTest.java
```diff
@@ -1440,5 +1440,133 @@ public void testGetSolidBlock() {
...
+    Transaction.Builder overLimit = unsigned.toBuilder();
+    for (int i = 0; i < keysCount + 1; i++) {
+      overLimit.addSignature(oneSig);
+    }
+    GrpcAPI.TransactionApprovedList rejected =
+        wallet.getTransactionApprovedList(overLimit.build());
+    assertEquals(GrpcAPI.TransactionApprovedList.Result.response_code.OTHER_ERROR,
+        rejected.getResult().getCode());
+    assertEquals(0, rejected.getApprovedListCount());
+    Assert.assertTrue(rejected.getResult().getMessage().contains("more than key counts"));
```
bladehan1: [NIT] Two small follow-ups on the new tests and `truncateSignatures`:

1. **`WalletTest.java` line 1489 — assertion depends on internal exception message format.** `Assert.assertTrue(rejected.getResult().getMessage().contains("more than key counts"))` relies on the exact text produced by `checkWeight`'s `PermissionException` propagated through `catch (Exception ex)` message formatting. If the exception message or the catch formatting changes, the test becomes a false pass. Consider asserting only on `response_code == OTHER_ERROR` and that the message is non-empty.

2. **`TransactionCapsule.java` line 469-488 — `truncateSignatures` iterates the signature list twice.** First pass sets `needTruncate`, second pass rebuilds. Since `getTotalSignNum` caps the list at 5 in practice the overhead is negligible, but a single-pass version (always `clearSignature()` + re-add each sig, truncating inline) would be simpler.

Suggestion: tighten the test assertion in item 1; optionally simplify the two-pass loop in item 2.

bladehan1 (commented):
Key design concern: `getTransactionApprovedList` semantic breaking change (all-or-nothing vs partial list). Author acknowledged this is intentional and will add release notes — please ensure that lands before merge.
An error will occur if a signature in the transmitted signature list is not authorized.

chainbase/src/main/java/org/tron/core/capsule/TransactionCapsule.java
Federico2014: Good catch. You're right — `truncateSignatures` is purely an API-level tolerance policy and doesn't belong in the shared chainbase layer.

Moved it to `TransactionUtil` (actuator), where `getTransactionSignWeight` already lives. Both call sites now route through `TransactionUtil.truncateSignatures(trx)`.

framework/src/test/java/org/tron/core/WalletTest.java
Federico2014: Both done:

1. Tightened the assertion to `assertFalse(message.isEmpty())` — no longer coupled to the exception text.
2. Simplified `truncateSignatures` to a single pass (always rebuilds — negligible with the 5-sig cap).

317787106 (approved):

PR #6828, opened 2026-06-09
fix(jsonrpc): harden RPC/HTTP parameter validation

0xbigapple:
**What does this PR do?**

Adds length/format validation to several JSON-RPC and HTTP parameter paths, so malformed,
oversized, or overflowing input is rejected early — a clear `-32602` (or a `null` result for an
out-of-range index) — instead of triggering `O(n²)` work, leaking a `NullPointerException` /
`DecoderException` / Internal error, or silently producing a wrong value. Valid input is unaffected.

1. **`Commons.decodeFromBase58Check`** — reject input whose length `!= 34` before the `O(n²)` `Base58.decode`; single funnel for every HTTP address input.
2. **`JsonFormat.unescapeBytesSelfType`** — an invalid Base58 address (null / illegal char) now returns a clear `invalid address` error instead of a bare `ByteString.copyFrom(null)` NPE; covers all merge-path address fields.
3. **`JsonRpcApiUtil.addressCompatibleToByteArray`** — cap length at `ADDRESS_SIZE + 2` (the `+2` keeps `0x`-prefixed 21-byte addresses valid) before `fromHexString`.
4. **`JsonRpcApiUtil.parseBlockNumber`** — tighten `MAX_BLOCK_NUM_HEX_LEN` 100 → 20, and route the `(String, Wallet)` overload through the single-arg parser so it also gets the length cap checks.
5. **`TronJsonRpcImpl.getStorageAt`** — bound the storage key and parse it before the contract lookup; a `>32`-byte key now returns `-32602 invalid storage key value` instead of `-32001 (json4j default error)`.
6. **`LogFilterWrapper`** — validate `blockHash` via the shared `JsonRpcApiUtil.hashToByteArray` (moved here from `TronJsonRpcImpl`).
7. **`HASH_REGEX`** accepts only hex digits `[0-9a-fA-F]`, so a non-hex hash is rejected as `invalid hash value` rather than slipping through to `fromHexString`.
8. **`JsonRpcApiUtil.parseTxIndex`** — length-capped 0x-hex parse for the tx index; a malformed / oversized / int-overflow index returns `-32602`, and a **negative or out-of-range** index returns a `null` result instead of letting `block.getTransactions(-1)` throw `IndexOutOfBoundsException` (Internal error).
9. **`TronJsonRpcImpl.buildCreateSmartContractTransaction`** — a deeply nested ABI overflows the recursive `JsonFormat` parser; the `StackOverflowError` is caught narrowly around the merge and mapped to `-32602 invalid abi` instead of an `Error` escaping as an Internal error.
10. **`JsonRpcApiUtil.calcFeeLimit`** — compute `feeLimit = gas * energyFee` via `StrictMathWrapper.multiplyExact`, so an oversized `gas` returns `-32602` instead of silently wrapping to a bogus (possibly negative) `feeLimit`.
11. **`JsonRpcApiUtil.parseQuantityValue`** — reject a signed (`0x-..`) `value` / `gas`; `QUANTITY` is unsigned, so a negative parse is `-32602` rather than a negative `callValue` / `feeLimit` reaching the actuator.
12. **`LogFilter` topics** — `topicToByteArray` now validates topics via `TOPIC_REGEX` (`^(0x)?[0-9a-fA-F]{63,64}$`); an invalid-hex / wrong-length / non-string topic returns `-32602` instead of a leaked `DecoderException` / `ClassCastException`, while keeping the odd-length (63-char) leading-`0` compatibility.
13. **`JsonRpcApiUtil.addressToByteArray`** (log-filter address) — cap length before `fromHexString` and wrap the decode, so an oversized or invalid-hex address filter returns `-32602` instead of `O(n)` work or a leaked `DecoderException`;a nested non-string address element (e.g. `{"address":[1]}`) is likewise rejected as `-32602` before the `(String)` cast; the 20-byte output and strict (no `41` prefix) semantics are unchanged.
14. **`getBlockReceipts` + `parseBlockNumber(String, Wallet)`** — a `null` block param dereferenced before validation (`Pattern.matches(HASH_REGEX, null)` / `null.startsWith("0x")`) NPE'd as an Internal error; it now returns `-32602`.

Several **unauthenticated** API entry points accepted unbounded / malformed input, causing two classes of problem:

- **Algorithmic-complexity DoS** — `decodeFromBase58Check` fed the whole string into the `O(n²)`
  `Base58.decode`; since the cost grows quadratically with length, a single oversized address
  payload (well within the 4 MB body cap) can tie up a request thread far out of proportion to
  its size — a cheap DoS the body cap and rate limiter don't stop. Block-number parsing likewise
  had no effective length bound before the `O(n²)` `BigInteger` constructor.
- **Leaked exceptions / unfriendly errors** — an invalid address on the merge path returned a
  bare `NullPointerException`; `eth_getStorageAt` surfaced a `RuntimeException` as
  `-32001 Data word can't exceed 32 bytes: xxx`; a negative tx index hit `getTransactions(-1)`
  (`IndexOutOfBoundsException`); a malformed log topic / address leaked a `DecoderException`; a
  deeply nested ABI exhausted the recursive `JsonFormat` parser's stack (`StackOverflowError`); a
  `null` block param NPE'd on `Pattern.matches` / `startsWith` — all escaping as Internal errors
  instead of `-32602 invalid params`.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

Representative affected endpoints, by guarded parameter (each parameter is a single guarded
funnel, so the lists below are examples, not exhaustive):
- **JSON-RPC**
  - **block number / tag** (length 100 → 20): `eth_getBlockByNumber`,`getBlockReceipts`, …
  - **block hash** (`hashToByteArray`): `eth_getBlockByHash`, `eth_getTransactionByBlockHashAndIndex`, …
  - **tx hash** (`hashToByteArray`): `eth_getTransactionByHash`, `eth_getTransactionReceipt`
  - **storage index** (parsed as a 32-byte word): `eth_getStorageAt`
  - **address** (length capped at `ADDRESS_SIZE + 2`): `eth_call`, `eth_getBalance` …
  - **tx index** (length-capped; negative / out-of-range → `null`): `eth_getTransactionByBlockHashAndIndex`, `eth_getTransactionByBlockNumberAndIndex`
  - **log topic / log address** (strict `hashToByteArray` / length-capped `addressToByteArray`): `eth_getLogs`, `eth_newFilter`
  - **gas / value / abi** (overflow-checked feeLimit, unsigned quantity, ABI-depth guard): `buildTransaction`
  - **block param null** (`null` → `-32602`): `eth_getBlockReceipts`
- **HTTP `/wallet/*`** — all covered by the Base58 length/DoS guard, but they differ on an invalid address:
  - most endpoints pre-convert via `Util.getHexAddress` and just return `{}` (e.g. `getcontract`) — unchanged;
  - endpoints that merge a Base58 field directly (e.g. `getaccount`, `visible=true`) get the NPE → `invalid address` fix;
  - endpoints that resolve the address via `Util.getAddress(request)` (e.g. `getReward`) get null for an invalid address and degrade to `reward 0`.

### DIFF
--- a/chainbase/src/main/java/org/tron/common/utils/Commons.java
+++ b/chainbase/src/main/java/org/tron/common/utils/Commons.java
@@ -48,11 +48,11 @@ public class Commons {
    */
   public static byte[] decodeFromBase58Check(String addressBase58) {
     if (StringUtils.isEmpty(addressBase58)) {
-      logger.warn("Warning: Address is empty !!");
+      logger.debug("address is empty !!");
       return null;
     }
     if (addressBase58.length() != BASE58_ADDRESS_LENGTH) {
-      logger.warn("Warning: invalid Base58 address length !!");
+      logger.debug("invalid Base58 address length");
       return null;
     }
     byte[] address = decode58Check(addressBase58);
--- a/chainbase/src/main/java/org/tron/common/utils/Commons.java
+++ b/chainbase/src/main/java/org/tron/common/utils/Commons.java
@@ -21,6 +21,8 @@ public class Commons {
 
   public static final int ASSET_ISSUE_COUNT_LIMIT_MAX = 1000;
 
+  public static final int BASE58_ADDRESS_LENGTH = 34;
+
   public static byte[] decode58Check(String input) {
     byte[] decodeCheck = Base58.decode(input);
     if (decodeCheck.length <= 4) {
@@ -41,9 +43,17 @@ public class Commons {
     return null;
   }
 
+  /**
+   * Decode a Base58Check address string to its 21-byte form.
+   */
   public static byte[] decodeFromBase58Check(String addressBase58) {
     if (StringUtils.isEmpty(addressBase58)) {
       logger.warn("Warning: Address is empty !!");
+      return null;
+    }
+    if (addressBase58.length() != BASE58_ADDRESS_LENGTH) {
+      logger.warn("Warning: invalid Base58 address length: {} ,expected {} !!",
+          addressBase58.length(), BASE58_ADDRESS_LENGTH);
       return null;
     }
     byte[] address = decode58Check(addressBase58);
--- a/chainbase/src/main/java/org/tron/common/utils/Commons.java
+++ b/chainbase/src/main/java/org/tron/common/utils/Commons.java
@@ -52,8 +52,7 @@ public class Commons {
       return null;
     }
     if (addressBase58.length() != BASE58_ADDRESS_LENGTH) {
-      logger.warn("Warning: invalid Base58 address length: {} ,expected {} !!",
-          addressBase58.length(), BASE58_ADDRESS_LENGTH);
+      logger.warn("Warning: invalid Base58 address length !!");
       return null;
     }
     byte[] address = decode58Check(addressBase58);
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
@@ -424,7 +424,7 @@ public class JsonRpcApiUtil {
   }
 
   /** Matches a 32-byte hash hex string: optional 0x prefix + 64 hex chars (also caps length). */
-  public static final String HASH_REGEX = "(0x)?[0-9a-fA-F]{64}$";
+  public static final String HASH_REGEX = "^(0x)?[0-9a-fA-F]{64}$";
 
   /**
    * Convert a hash hex string (optional 0x prefix) to a byte array, validating
@@ -446,7 +446,7 @@ public class JsonRpcApiUtil {
   /**
    * Matches a 32-byte topic hex string: optional 0x prefix + 63 or 64 hex chars.
    */
-  public static final String TOPIC_REGEX = "(0x)?[0-9a-fA-F]{63,64}$";
+  public static final String TOPIC_REGEX = "^(0x)?[0-9a-fA-F]{63,64}$";
 
   /**
    * Convert a topic hex string (optional 0x prefix, leading zero may be omitted) to a 32-byte
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
@@ -61,6 +61,7 @@ public class JsonRpcApiUtil {
   public static final String TAG_PENDING_SUPPORT_ERROR = "TAG pending not supported";
   public static final String TAG_SAFE_SUPPORT_ERROR = "TAG safe not supported";
   public static final String BLOCK_NUM_ERROR = "invalid block number";
+  public static final String TX_INDEX_ERROR = "invalid index value";
 
   private static final SecureRandom random = new SecureRandom();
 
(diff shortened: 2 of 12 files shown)


framework/src/main/java/org/tron/core/services/jsonrpc/filters/LogFilter.java
waynercheung, line 65: [MUST] This branch still casts nested address/topic array elements to String before validation and only catches `JsonRpcInvalidParamsException`. Inputs like `{"address":[1]}` or `{"topics":[[1]]}` can still throw `ClassCastException` and escape as an internal error. Please validate element types before casting, or catch `ClassCastException` here and map it to the existing `-32602` invalid-params response.

0xbigapple: Good Catch! Fixed.

317787106 (approved):

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -395,46 +398,90 @@ public static long getUnfreezeAssetAmount(byte[] addressBytes, Wallet wallet) {
...
+      throw new JsonRpcInvalidParamsException(e.getMessage());
+    }
+    return bHash;
   }
 
   /**
-   * check if topic is hex string of size 64, padding 0 ahead if length is odd.
+   * Matches a 32-byte topic hex string: optional 0x prefix + 63 or 64 hex chars.
+   */
+  public static final String TOPIC_REGEX = "(0x)?[0-9a-fA-F]{63,64}$";
```
bladehan1: [SHOULD] PR description item 12 contradicts the final implementation

The description says topics are now parsed "via the strict hashToByteArray ... instead of topicToByteArray (now removed)" and that odd-length input is rejected ("no silent leading-0 pad"). The final code does the opposite: `topicToByteArray` is kept, `TOPIC_REGEX` accepts 63 **or** 64 hex chars, and a 63-char topic is accepted and zero-padded (locked in by `topicToByteArrayPadsMissingLeadingZero`). Anyone auditing the behavior change from the description will reach the wrong conclusion about what `eth_getLogs` / `eth_newFilter` accept.

Suggestion: Update description item 12 to match the code: topics still go through `topicToByteArray`, which now enforces a 63/64-char length cap and strict hex, while keeping the odd-length leading-zero compatibility.

chainbase/src/main/java/org/tron/common/utils/Commons.java
```diff
@@ -41,11 +43,18 @@ public static byte[] decode58Check(String input) {
...
+  /**
+   * Decode a Base58Check address string to its 21-byte form.
+   */
   public static byte[] decodeFromBase58Check(String addressBase58) {
     if (StringUtils.isEmpty(addressBase58)) {
       logger.warn("Warning: Address is empty !!");
       return null;
     }
+    if (addressBase58.length() != BASE58_ADDRESS_LENGTH) {
+      logger.warn("Warning: invalid Base58 address length !!");
```
bladehan1: [NIT] Request-driven WARN log on the unauthenticated address funnel

`decodeFromBase58Check` is the single funnel for every HTTP `/wallet/*` address input, so this new length-guard WARN fires once per malformed request. The very scenario this guard defends against (an attacker spraying oversized addresses) would now spam WARN-level logs and pollute alerting; client-input errors are conventionally logged at INFO or below in the API layer. The adjacent empty-input WARN is pre-existing, but the new line can start at the right level.

Suggestion: Log this at info/debug, optionally including the offending length (not the input itself) to keep it useful for config-load debugging (genesis/witness addresses also pass through here).

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
```diff
@@ -395,46 +398,90 @@ public static long getUnfreezeAssetAmount(byte[] addressBytes, Wallet wallet) {
...
+        throw new JsonRpcInvalidParamsException("invalid address");
       }
     } catch (Exception e) {
       throw new JsonRpcInvalidParamsException(e.getMessage());
     }
     return addressByte;
   }
 
+  /** Matches a 32-byte hash hex string: optional 0x prefix + 64 hex chars (also caps length). */
+  public static final String HASH_REGEX = "(0x)?[0-9a-fA-F]{64}$";
```
bladehan1: [NIT] Polish backlog (3 items rolled up)

Grouped as one comment since all are non-blocking polish items:

1. **Regexes lack a `^` anchor.** `HASH_REGEX` (line 427) and `TOPIC_REGEX` (line 449) only have a trailing `$`. All current call sites use `Pattern.matches` (full-string match), so there is no bug today, but both constants are public; a future `Matcher.find`-style use would silently accept partial matches. Adding `^` is free insurance.

2. **Invalid Base58 addresses now have two different outcomes on HTTP endpoints.** In `Commons.decodeFromBase58Check` (chainbase/src/main/java/org/tron/common/utils/Commons.java:54), a wrong-length address with illegal characters returns null (e.g. getReward degrades to reward 0, previously an error response), while a 34-char address with illegal characters still throws from `Base58.decode` and surfaces as an error. Same class of bad input, two user-visible results; clients cannot tell a typo'd address from a genuine zero reward. Worth normalizing in a follow-up.

3. **Cross-cutting concern — commit message hygiene.** Two commits in this branch have a leading-space subject, and one folds the entire message into a single ~300-char subject line (no blank line before the body), which degrades `git log --oneline` and changelog extraction. A squash-merge would also resolve this.

Suggestion: Add `^` anchors to both regexes; track the invalid-address outcome normalization as a follow-up; tidy the two commit messages or rely on squash-merge.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
0xbigapple: Updated the PR description.

framework/src/main/java/org/tron/core/services/jsonrpc/JsonRpcApiUtil.java
0xbigapple: Added a leading ^ to both HASH_REGEX and TOPIC_REGEX.

chainbase/src/main/java/org/tron/common/utils/Commons.java
0xbigapple: Done — lowered to debug.

bladehan1 (approved):
LGTM

waynercheung (approved):
LGTM

PR #6843, opened 2026-06-16
fix(api): hide jetty internals in oversized request 413 response

bladehan1:
**What does this PR do?**

Installs a custom Jetty `ErrorHandler` (private `OversizedRequestErrorHandler` in `HttpService`) so that for an oversized request — the `413` thrown by `SizeLimitHandler` during dispatch — the node logs the detail server-side and returns the short, uniform bad-message page. Previously the default Jetty error page exposed the full exception stack, the Jetty version and the internal request size in the `413` response body.

The handler special-cases `413` only; all other errors fall through to `super` (Jetty's default handling).

The `413` response body leaked Jetty internals (CWE-209 information disclosure / server fingerprinting): exact Jetty version, internal class names and line numbers, thread-pool internals, the JDK build, and the rejected request size, e.g.:

```
org.eclipse.jetty.http.BadMessageException: 413: Request body is too large: 664657339>4194304
    at org.eclipse.jetty.server.handler.SizeLimitHandler.checkRequestLimit(SizeLimitHandler.java:63)
    ...
    at java.lang.Thread.run(Thread.java:750)
```

Only the `413` path is affected, because `SizeLimitHandler` throws during dispatch (`Server.handle`) and is rendered via `ErrorHandler.handle()` -> `writeErrorPageStacks`. Parser-stage errors (`400` / `414` / `431`) already render via `badMessageError()` without a stack, so they are intentionally left untouched.

- Unit Tests
  - `SizeLimitHandlerTest#testHttpBodyExceedsLimit`: `413` body renders the short bad-message page and leaks no Jetty stack / class / size.
  - `SizeLimitHandlerTest#testBadContentLengthReturnsDefault400`, `testOversizedUriReturnsDefault414`: other 4xx stay on Jetty's default path (not rerouted through the `413` branch).
- Manual Testing
  - `curl` with an oversized body returns `413` with the short page; the response no longer contains the exception stack, class name or internal size.

**Follow up**

**Extra details**

Jetty `9.4.58`. The fix lives in the base class `HttpService#initServer()` so it applies to every HTTP service (FullNode HTTP, Solidity, PBFT, JSON-RPC). The `reason` returned to the client is the generic `HttpStatus.getMessage(413)` ("Payload Too Large"); the full detail (including the rejected size) is kept only in the server log.

### DIFF
--- a/framework/src/main/java/org/tron/common/application/HttpService.java
+++ b/framework/src/main/java/org/tron/common/application/HttpService.java
@@ -16,12 +16,22 @@
 package org.tron.common.application;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.concurrent.CompletableFuture;
+import javax.servlet.RequestDispatcher;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
+import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.server.ConnectionLimit;
+import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.ErrorHandler;
 import org.eclipse.jetty.server.handler.SizeLimitHandler;
 import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.util.BufferUtil;
 import org.tron.core.config.args.Args;
 
 @Slf4j(topic = "rpc")
@@ -72,6 +82,7 @@ public abstract class HttpService extends AbstractService {
     if (maxHttpConnectNumber > 0) {
       this.apiServer.addBean(new ConnectionLimit(maxHttpConnectNumber, this.apiServer));
     }
+    this.apiServer.setErrorHandler(new OversizedRequestErrorHandler());
   }
 
   protected ServletContextHandler initContextHandler() {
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/tron/common/application/HttpService.java
```diff
@@ -88,4 +99,30 @@ protected ServletContextHandler initContextHandler() {
...
+   * other errors keep Jetty's default handling.
+   */
+  private static final class OversizedRequestErrorHandler extends ErrorHandler {
+
+    @Override
+    public void handle(String target, Request baseRequest, HttpServletRequest request,
+        HttpServletResponse response) throws IOException, ServletException {
+      if (response.getStatus() == HttpStatus.PAYLOAD_TOO_LARGE_413) {
+        Throwable cause = (Throwable) request.getAttribute(RequestDispatcher.ERROR_EXCEPTION);
+        logger.info("Reject oversized request, uri: {}, detail: {}",
```
halibobo1205, line 116: **Nit (non-blocking):** Every oversized request now emits a `logger.info(...)` line (including the rejected request size). Under a sustained flood of oversized requests, an attacker can use this to amplify log volume — each cheaply-rejected 413 still produces a log entry. Consider dropping it to `debug` or rate-limiting it. That said, keeping it at `info` for visibility into rejected oversized requests is also reasonable — your call.

framework/src/main/java/org/tron/common/application/HttpService.java
```diff
@@ -88,4 +99,30 @@ protected ServletContextHandler initContextHandler() {
...
+   * detail server-side and returns the short, uniform bad-message page, instead of the
+   * default error page that leaks the exception stack and internal request sizes. All
+   * other errors keep Jetty's default handling.
+   */
+  private static final class OversizedRequestErrorHandler extends ErrorHandler {
+
+    @Override
+    public void handle(String target, Request baseRequest, HttpServletRequest request,
+        HttpServletResponse response) throws IOException, ServletException {
+      if (response.getStatus() == HttpStatus.PAYLOAD_TOO_LARGE_413) {
```
0xbigapple, line 114: [SHOULD] `ErrorHandler` defaults to `_showStacks = true`. This PR special-cases `413` only, so an uncaught `RuntimeException` during servlet/filter dispatch (→ `500`) still renders the full stack via `super.handle` → `writeErrorPageStacks`, and the default error page footer still exposes `Powered by Jetty://<version>`.

  Note that `setShowStacks(false)` alone would **not** replace this PR — it wouldn't remove the rejected-size in the `413` message or the version footer, so the custom short page is still needed. But it would be cheap to *additionally* call `setShowStacks(false)` on this handler (or override the footer) to close the same info-disclosure surface for `500`/other dispatch errors, rather than only `413`. Happy to leave it out of this hotfix's scope — just flagging it so it's a conscious decision.

framework/src/main/java/org/tron/common/application/HttpService.java
```diff
@@ -88,4 +99,30 @@ protected ServletContextHandler initContextHandler() {
...
+        Throwable cause = (Throwable) request.getAttribute(RequestDispatcher.ERROR_EXCEPTION);
+        logger.info("Reject oversized request, uri: {}, detail: {}",
+            request.getRequestURI(), cause == null ? "413" : cause.getMessage());
+        baseRequest.setHandled(true);
+        ByteBuffer body = badMessageError(HttpStatus.PAYLOAD_TOO_LARGE_413,
+            HttpStatus.getMessage(HttpStatus.PAYLOAD_TOO_LARGE_413),
+            baseRequest.getResponse().getHttpFields());
+        response.getOutputStream().write(BufferUtil.toArray(body));
```
0xbigapple, line 122: [SHOULD] Content-Length not set explicitly.

bladehan1: image

It will be set automatically.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: Log attacks are extremely inefficient, easily resolved with a single grep filter. In such rare cases, retaining the messages helps operations personnel better understand node performance.

framework/src/main/java/org/tron/common/application/HttpService.java
bladehan1: There are two reasons:

1. Exposing the stack trace helps in troubleshooting when encountering strange problems.

2. In the case of hotfixes, minimize modifications to avoid introducing new problems due to oversights.

halibobo1205 (approved):

0xbigapple (approved):

PR #6845, opened 2026-06-16
refactor(http): improve JsonFormat parser robustness

halibobo1205:
**What does this PR do?**

Improves `JsonFormat` parsing robustness for protobuf JSON used by HTTP wallet APIs.

- Reworks field parsing so comma-separated fields are handled iteratively instead of recursive self-calls.
- Adds nesting-depth checks when parsing object/array structures in `JsonFormat`.
- Adds unit coverage for preserved parser behavior, changed parser behavior, trailing comma handling, nesting limits, and large flat JSON bodies.

The previous parser could consume an excessive call stack when processing very field-heavy or deeply nested JSON input. This PR makes the parser fail predictably with `ParseException` for overly deep input, while keeping normal JSON/protobuf parsing behavior compatible.

- Unit Tests
  - `./gradlew :framework:test --tests org.tron.core.services.http.JsonFormatTest --tests org.tron.core.services.http.TriggerConstantContractServletTest :framework:checkstyleTest -x generateGitProperties`
  - `./gradlew lint`
- Manual Testing
  - Verified compatibility-sensitive parser behavior before and after the change:
    - Existing unknown-field skipping remains supported.
    - Known repeated fields still reject unsupported nested arrays.
    - Unknown nested object/array fields still reject trailing commas.
    - Valid nesting within `Constant.MAX_NESTING_DEPTH` remains accepted.

**Follow up**

None required.

**Extra details**

Compatibility notes:

- Direct `JsonFormat.merge()` callers now reject protobuf JSON nesting beyond `Constant.MAX_NESTING_DEPTH` with `ParseException: Hit recursion limit.`.
- HTTP API compatibility risk is low because the normal HTTP JSON path already applies Jackson's `MAX_NESTING_DEPTH = 100`.
- Deeply nested invalid input now returns a parser error instead of surfacing as a container-level HTTP 500.
- - JSON-RPC create-contract ABI parsing keeps the previous deeply nested ABI behavior: it still returns invalid params with `invalid abi`. Other ABI inputs that fail `JsonFormat` parsing may now also be reported as invalid params instead of internal error; this only affects invalid ABI input.

- Compatibility notes for trailing commas:

| Scenario | Example | Before this PR | After this PR | Notes |
| --- | --- | --- | --- | --- |
| Top-level object trailing comma | `{"address":"61646472657373",}` | Rejected with `ParseException` | Accepted | Behavior is relaxed. This applies to top-level fields after the caller loop consumes the trailing comma. |
| Known nested message trailing comma | `{"genesisBlockId":{"hash":"00","number":1,}}` | Rejected with `ParseException` | Accepted | Behavior is relaxed for known protobuf message fields. |
| Known repeated field trailing comma | `{"approvals":["00","01",]}` | Accepted | Accepted | No behavior change; repeated-field array parsing already tolerated a single trailing comma. |
| Unknown nested object trailing comma | `{"zzz":{"a":1,}}` | Rejected with `ParseException` | Rejected with `ParseException` | No behavior change. Unknown fields are only skipped and are not used by business logic. |
| Unknown nested array trailing comma | `{"zzz":[1,]}` | Rejected with `ParseException` | Rejected with `ParseException` | No behavior change. |
| Repeated commas | `{"address":"61646472657373",,}` | Rejected with `ParseException` | Rejected with `ParseException` | No behavior change; only a single trailing comma may be accepted on known/top-level paths. |

This is an input-acceptance relaxation for top-level fields and known nested protobuf message fields. It is aligned with the existing `org.tron.json.JSON` behavior, which already allows a single trailing comma while rejecting repeated commas.

### DIFF
--- a/framework/src/main/java/org/tron/core/services/http/JsonFormat.java
+++ b/framework/src/main/java/org/tron/core/services/http/JsonFormat.java
@@ -57,6 +57,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.tron.common.utils.ByteArray;
 import org.tron.common.utils.Commons;
 import org.tron.common.utils.StringUtil;
+import org.tron.core.Constant;
 import org.tron.json.JSON;
 import org.tron.protos.contract.BalanceContract;
 
@@ -291,6 +292,7 @@ public class JsonFormat {
     tokenizer.consume("{"); // Needs to happen when the object starts.
     while (!tokenizer.tryConsume("}")) { // Continue till the object is done
       mergeField(tokenizer, extensionRegistry, builder, selfType);
+      tokenizer.tryConsume(",");
     }
     // Test to make sure the tokenizer has reached the end of the stream.
     if (!tokenizer.atEnd()) {
@@ -556,8 +558,9 @@ public class JsonFormat {
   }
 
   /**
-   * Parse a single field from {@code tokenizer} and merge it into {@code builder}. If a ',' is
-   * detected after the field ends, the next field will be parsed automatically
+   * Parse a single field from {@code tokenizer} and merge it into {@code builder}. Exactly one
+   * field is consumed; the caller ({@code merge} / {@code handleObject}) consumes any trailing
+   * ',' and loops over the remaining fields.
    */
   protected static void mergeField(Tokenizer tokenizer,
       ExtensionRegistry extensionRegistry, Message.Builder builder,
@@ -628,11 +631,6 @@ public class JsonFormat {
         handleValue(tokenizer, extensionRegistry, builder, field, extension, unknown, selfType);
       }
     }
-
-    if (tokenizer.tryConsume(",")) {
-      // Continue with the next field
-      mergeField(tokenizer, extensionRegistry, builder, selfType);
-    }
   }
 
   private static void handleMissingField(Tokenizer tokenizer,
@@ -642,18 +640,28 @@ public class JsonFormat {
     if ("{".equals(tokenizer.currentToken())) {
       // Message structure
       tokenizer.consume("{");
-      do {
-        tokenizer.consumeIdentifier();
-        handleMissingField(tokenizer, extensionRegistry, builder);
-      } while (tokenizer.tryConsume(","));
-      tokenizer.consume("}");
+      tokenizer.enterRecursion();
+      try {
+        do {
+          tokenizer.consumeIdentifier();
+          handleMissingField(tokenizer, extensionRegistry, builder);
+        } while (tokenizer.tryConsume(","));
+        tokenizer.consume("}");
+      } finally {
+        tokenizer.exitRecursion();
+      }
     } else if ("[".equals(tokenizer.currentToken())) {
       // Collection
       tokenizer.consume("[");
-      do {
-        handleMissingField(tokenizer, extensionRegistry, builder);
-      } while (tokenizer.tryConsume(","));
-      tokenizer.consume("]");
+      tokenizer.enterRecursion();
+      try {
+        do {
+          handleMissingField(tokenizer, extensionRegistry, builder);
+        } while (tokenizer.tryConsume(","));
+        tokenizer.consume("]");
+      } finally {
+        tokenizer.exitRecursion();
+      }
     } else { //if (!",".equals(tokenizer.currentToken)){
       // Primitive value
       if ("null".equals(tokenizer.currentToken())) {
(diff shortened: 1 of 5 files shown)


framework/src/main/java/org/tron/core/services/http/JsonFormat.java
```diff
@@ -291,6 +292,7 @@ public static void merge(CharSequence input,
     tokenizer.consume("{"); // Needs to happen when the object starts.
     while (!tokenizer.tryConsume("}")) { // Continue till the object is done
       mergeField(tokenizer, extensionRegistry, builder, selfType);
+      tokenizer.tryConsume(",");
```
bladehan1, line 295: [DISCUSS] Trailing comma now accepted on known-field paths

Moving comma handling into the caller loop here means known-field paths now accept a single trailing comma (e.g. {"a":1,} at top level, and inside known message fields via handleObject), whereas unknown nested object/array fields in handleMissingField still reject it. That is a small, observable parsing-behavior change landing on a maintenance branch.

This is actually consistent with the project's JSON layer: org.tron.json.JSON (the fastjson-compat shim) already enables ALLOW_TRAILING_COMMA — a single trailing comma is accepted while repeated commas (,,) are rejected, which matches the new JsonFormat behavior. So the direction looks intentional and aligned.

Question: is the known-vs-unknown asymmetry (known paths tolerate a trailing comma, unknown nested paths reject it) intended? If the goal is parity with the JSON shim, you may want unknown nested paths to tolerate a single trailing comma too.

Suggestion: confirm the asymmetry is deliberate (the PR's Compatibility notes already touch on it), and decide whether handleMissingField should match the known-field trailing-comma tolerance.

bladehan1 (commented):
Findings: 1 [DISCUSS] (trailing-comma behavior — appears intentional, aligns with the fastjson-compat JSON shim) + 1 [NIT] (boundary test + cosmetic loop shape).

framework/src/main/java/org/tron/core/services/http/JsonFormat.java
```diff
@@ -807,20 +815,25 @@ private static Object handleObject(Tokenizer tokenizer,
...
 
-    while (!tokenizer.tryConsume(endToken)) {
-      if (tokenizer.atEnd()) {
-        throw tokenizer.parseException("Expected \"" + endToken + "\".");
-      }
-      mergeField(tokenizer, extensionRegistry, subBuilder, selfType);
-      if (tokenizer.tryConsume(",")) {
-        // there are more fields in the object, so continue
-        continue;
+      while (!tokenizer.tryConsume(endToken)) {
```
bladehan1, line 822: [NIT] Documentation & test backlog (2 items rolled up)

Grouped as one comment since both are minor follow-ups (one test, one cosmetic) on the parser-hardening change; neither affects runtime behavior.

1. **Lock the nesting-limit boundary (test).** JsonFormatTest.java around line 258-276: the depth tests use MAX_NESTING_DEPTH/2 (accepted) and MAX_NESTING_DEPTH+1 (rejected), but the exact boundary is not pinned. An off-by-one change to enterRecursion() would not be caught. Add a pair asserting depth == MAX_NESTING_DEPTH is accepted and depth == MAX_NESTING_DEPTH + 1 is rejected.

2. **Unify sibling-loop shape (cosmetic).** JsonFormat.java handleObject (line 822-831) uses `if (tokenizer.tryConsume(",")) { continue; }`, which is equivalent to `tokenizer.tryConsume(",")` since nothing follows the `continue`. The merge() loop (line 293-296) already uses the simpler form. Optionally align handleObject with merge() for consistency.

Suggestion: add the boundary test pair (==MAX accepted / ==MAX+1 rejected) and optionally simplify handleObject's trailing-comma handling to match merge().

bladehan1 (approved):

Sunny6889 (approved):

framework/src/main/java/org/tron/core/services/http/JsonFormat.java
halibobo1205: Yes, this asymmetry is intentional.

Known fields are the real protobuf JSON contract, so aligning their trailing-comma behavior with `org.tron.json.JSON` is acceptable. Unknown fields are only skipped for compatibility and are not used by business logic, so we do not need to broaden their accepted syntax.

The PR compatibility notes already mention this difference.

framework/src/main/java/org/tron/core/services/http/JsonFormat.java
halibobo1205: Thanks, both points make sense.

Since this PR already covers within-limit parsing and `MAX_NESTING_DEPTH + 1` rejection, I would keep the current tests as-is for now. The `handleObject` loop shape is also functionally equivalent, so I prefer not to add cosmetic churn in this maintenance fix.

waynercheung (approved):

PR #6851, opened 2026-06-22
fix(net): reject unknown inventory type in P2P inv/fetch-inv handling

xxo1shine:
**What does this PR do?**
Validate the inventory type at the inbound entry points and reject any value other than TRX or BLOCK with P2pException(BAD_MESSAGE), before the type is used for cache insertion or outbound fetch construction:
- P2pEventHandlerImpl.checkInvRateLimit: add else branch for unknown type
- InventoryMsgHandler.check: add type allowlist check
- FetchInvDataMsgHandler.check: validate raw getInventoryType() instead of getInvMessageType() (which mapped non-BLOCK types to TRX)

Use getTypeValue() when building the error message to avoid calling getNumber() on an UNRECOGNIZED enum value. Add regression tests for all three entry points.

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
+++ b/framework/src/main/java/org/tron/core/net/P2pEventHandlerImpl.java
@@ -215,7 +215,8 @@ public class P2pEventHandlerImpl extends P2pEventHandler {
     }
   }
 
-  private boolean checkInvRateLimit(PeerConnection peer, InventoryMessage msg) {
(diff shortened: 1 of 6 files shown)


waynercheung (approved):
LGTM

halibobo1205 (approved):

317787106 (approved):

PR #6863, opened 2026-07-02
feat(http): treat null JsonFormat fields as absent

### LINKED ISSUE #6607: [Feature] Replace fastjson with Jackson

halibobo1205:
1. Treat known protobuf fields with literal null as absent so HTTP APIs recover the 4.8.1 fastjson-compatible behavior for null object fields.

2. This relaxes JSON-RPC buildTransaction ABI handling because that path now accepts null ABI fields instead of returning invalid abi.

3. Existing array-element behavior is unchanged: null elements in repeated message/object arrays remain invalid, while null elements in repeated primitive/bytes/enum arrays are consumed and skipped by the legacy parser.
fix: https://github.com/tronprotocol/java-tron/issues/6607#issuecomment-4855703940

### DIFF
--- a/framework/src/main/java/org/tron/core/services/http/JsonFormat.java
+++ b/framework/src/main/java/org/tron/core/services/http/JsonFormat.java
@@ -620,6 +620,10 @@ public class JsonFormat {
 
     if (field != null) {
       tokenizer.consume(":");
+      // Match protobuf JsonFormat: a field whose value is null is treated as absent.
+      if (tokenizer.tryConsume("null")) {
+        return;
+      }
       boolean array = tokenizer.tryConsume("[");
 
       if (array) {
--- a/framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
+++ b/framework/src/test/java/org/tron/core/jsonrpc/JsonrpcServiceTest.java
@@ -1489,6 +1489,52 @@ public class JsonrpcServiceTest extends BaseTest {
   }
 
   @Test
+  public void testBuildCreateSmartContractAcceptsNullAbiOutputsOverHttp() {
+    fullNodeJsonRpcHttpService.start();
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
+      JsonObject buildArgs = new JsonObject();
+      buildArgs.addProperty("from", "0xabd4b9367799eaa3197fecb144eb71de1e049abc");
+      buildArgs.addProperty("data", "608060405234801561001057600080fd5b50");
+      buildArgs.addProperty("gas", "0x3b9aca00");
+      buildArgs.addProperty("abi", "[{\"inputs\":[],\"name\":\"test\",\"outputs\":null,"
+          + "\"type\":\"function\"}]");
+      JsonArray params = new JsonArray();
+      params.add(buildArgs);
+      JsonObject requestBody = new JsonObject();
+      requestBody.addProperty("jsonrpc", "2.0");
+      requestBody.addProperty("method", "buildTransaction");
+      requestBody.add("params", params);
+      requestBody.addProperty("id", 1);
+
+      HttpPost httpPost = new HttpPost("http://127.0.0.1:"
+          + CommonParameter.getInstance().getJsonRpcHttpFullNodePort() + "/jsonrpc");
+      httpPost.addHeader("Content-Type", "application/json");
+      httpPost.setEntity(new StringEntity(requestBody.toString()));
+      try (CloseableHttpResponse response = httpClient.execute(httpPost)) {
+        String resp = EntityUtils.toString(response.getEntity());
+        JSONObject json = JSON.parseObject(resp);
+        Assert.assertNull(resp, json.getJSONObject("error"));
+        JSONObject tx = json.getJSONObject("result").getJSONObject("transaction");
+        Assert.assertNotNull("transaction must be a JSON object", tx);
+
+        JSONArray contracts = tx.getJSONObject("raw_data").getJSONArray("contract");
+        Assert.assertEquals(1, contracts.size());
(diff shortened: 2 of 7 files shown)


bladehan1 (approved):

SeriousCoding789:
Nice fix on the inbound side. One gap worth a follow-up: this only covers the **inbound (parse)** side. The
**outbound (serialize)** side still emits `"field":null` — `org.tron.json`'s `toJSONString()` serializes null fields as `"field":null`, because`@JsonInclude(NON_NULL)` applies to POJO serialization only, not to`JsonNode`/tree-model serialization (`NullNode` is written as-is).

It's functionally fine now (the parser tolerates null), but the response JSON shape still diverges from the old fastjson behavior: fastjson omitted null-valued fields by default, whereas the current output keeps them. Downstream tools that migrated assuming "fastjson omits nulls" will see extra"field":null` entries in node responses.

To fully restore fastjson output compatibility, we'd also need to strip `NullNode` on the `org.tron.json` serialization side (prune null object fields before `writeValueAsString`).

halibobo1205:
@Sunny6889 
Good point in general: `JsonInclude.NON_NULL` does not prune `NullNode` values from Jackson tree nodes.

That said, in the current HTTP response paths, I did not find a production path that inserts `NullNode` into `org.tron.json.JSONObject`, or parses arbitrary explicit-null JSON and returns it directly. Most outbound `JSONObject` usage wraps `JsonFormat.printToString(proto, ...)`, and that printer does not emit JSON `null` fields. Also, `JSONObject.put(key, null)` removes the key.

So I agree this is a valid compatibility caveat for the wrapper. However, I would treat it as a follow-up compatibility-hardening item rather than a blocker for this inbound `JsonFormat` fix.

Sunny6889 (approved):

waynercheung (approved):
LGTM

PR #6864, opened 2026-07-02
fix(db): re-verify all transaction signatures on fork switch

xxo1shine:
**What does this PR do?**
The switchFork new-branch apply loop only re-validated each block's witness signature; the transactions inside were applied through the isVerified cache. That flag caches a signature-verification result computed against a specific account-permission state, but a fork switch re-applies blocks on a rewound, diverged chain state where those permissions may have changed. Trusting the stale flag lets a transaction skip verification and be accepted with a signature no longer valid under the fork-chain state.

- Manager.switchFork: clear isVerified on every transaction of each block on the branch being switched to, before applyBlock, forcing full signature re-validation against the fork-chain state. The switch-back path (original main branch) is left untouched: it reproduces the exact original state, so its cached verifications stay valid and resetting them would only cost perf.
- ManagerTest: add switchForkShouldResetTransactionSignVerifiedOnNewBranch, which drives a real reorg and asserts the new branch's transaction gets its cached verification flag cleared (fails without the fix).

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/db/Manager.java
+++ b/framework/src/main/java/org/tron/core/db/Manager.java
@@ -1146,6 +1146,13 @@ public class Manager {
             throw new ValidateSignatureException(
                 "switch fork: block " + item.getBlk().getNum() + " signature invalid");
           }
+          // The new branch is applied on a rewound, diverged state where account permissions
+          // may have changed, so a cached signature-verification result is no longer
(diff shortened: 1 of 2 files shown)


halibobo1205 (approved):

lvs0075 (approved):

KingsDmusk:
valide

PR #6867, opened 2026-07-07
fix(jsonrpc): don't print stacktrace in jsonrpc when request overload

317787106:
**What does this PR do?**

Let Jetty `BadMessageException` with status `HttpStatus.PAYLOAD_TOO_LARGE_413` propagate back to the existing unified `OversizedRequestErrorHandler` instead of being absorbed inside the servlet/filter chain.

The change has two parts:
- `RateLimiterServlet` no longer handles oversized-request `BadMessageException` locally; it rethrows it.
- `HttpInterceptor` still records failure metrics, but now rethrows only `BadMessageException(413)` so that oversized chunked requests can reach Jetty's existing unified error path. Other exceptions keep the previous swallow-and-meter behavior.

This keeps chunked JSON-RPC oversized requests aligned with the pre-existing unified oversized-request handling path in `HttpService`, instead of introducing a servlet-specific 413 implementation.

For `Transfer-Encoding: chunked`, Jetty only discovers an oversized request while the servlet is reading the request body. In the JSON-RPC path, that expected 413 exception used to be swallowed inside the servlet/filter chain before it could reach Jetty's unified oversized-request error handler, which led to noisy ERROR stack traces and inconsistent handling.

Other HTTP endpoints can hit the same Jetty streaming exception timing. Many HTTP servlets already catch and convert errors locally, but for paths where a 413 does bubble up, this change makes them reuse the same top-level oversized-request handler instead of splitting behavior between ad hoc local handling and Jetty error handling.

- `./gradlew :framework:test --tests "org.tron.core.services.http.RateLimiterServletTest" --tests "org.tron.core.services.filter.HttpInterceptorTest" --tests "org.tron.core.jsonrpc.JsonrpcServiceTest.testJsonRpcSizeLimitIntegration"`
- `./gradlew checkstyleMain checkstyleTest`
- `git diff --check`

**Follow up**

None.

**Extra details**

- `RateLimiterServletTest` now verifies that oversized-request `BadMessageException` is propagated rather than handled in-place.
- `HttpInterceptorTest` locks down the narrower behavior change: rethrow `BadMessageException(413)`, keep swallowing non-413 exceptions.
- `JsonrpcServiceTest` still verifies that oversized chunked JSON-RPC requests return 413 end-to-end.

### DIFF
--- a/framework/src/main/java/org/tron/core/services/filter/HttpInterceptor.java
+++ b/framework/src/main/java/org/tron/core/services/filter/HttpInterceptor.java
@@ -1,6 +1,5 @@
 package org.tron.core.services.filter;
 
-import com.google.common.base.Strings;
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
 import javax.servlet.FilterConfig;
@@ -9,6 +8,8 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
+import org.eclipse.jetty.http.BadMessageException;
+import org.eclipse.jetty.http.HttpStatus;
 import org.tron.common.prometheus.MetricKeys;
 import org.tron.common.prometheus.MetricLabels;
 import org.tron.common.prometheus.Metrics;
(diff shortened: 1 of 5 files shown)


bladehan1 (approved):

0xbigapple (approved):
LGTM

PR #6920, opened 2026-08-17
feat: optimize delegate and undelegate instruction handling

ouy95917:
**What does this PR do?**

This PR optimizes the execution of SELFDESTRUCT, FreezeBalanceV2, UnfreezeBalanceV2, WithdrawExpireUnfreeze, and CancelAllUnfreezeV2 by exiting early in specific edge cases to improve performance.

To streamline the related execution paths and improve processing efficiency in specific scenarios.

- Unit Tests
- Manual Testing

**Follow up**

The release and activation plans will be evaluated separately.

**Extra details**

The version identifier used in this PR is provisional and may be adjusted later.

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
+++ b/actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
@@ -139,6 +139,7 @@ public class RepositoryImpl implements Repository {
   private final HashMap<Key, Value<DelegatedResourceAccountIndex>> delegatedResourceAccountIndexCache = new HashMap<>();
   private final HashBasedTable<Key, Key, Value<byte[]>> transientStorage = HashBasedTable.create();
   private final HashSet<Key> newContractCache = new HashSet<>();
+  private final HashSet<Key> selfDestructCache = new HashSet<>();
 
   public static void removeLruCache(byte[] address) {
   }
@@ -573,6 +574,29 @@ public class RepositoryImpl implements Repository {
   }
 
   @Override
+  public void markSelfDestruct(byte[] address) {
+    selfDestructCache.add(Key.create(address));
+  }
+
+  @Override
+  public boolean isSelfDestructed(byte[] address) {
+    Key key = Key.create(address);
+    if (selfDestructCache.contains(key)) {
+      return true;
+    }
+
+    if (parent != null) {
+      boolean isSelfDestructed = parent.isSelfDestructed(address);
+      if (isSelfDestructed) {
+        selfDestructCache.add(key);
+      }
+      return isSelfDestructed;
+    } else {
+      return false;
+    }
+  }
+
+  @Override
   public void updateAccount(byte[] address, AccountCapsule accountCapsule) {
     accountCache.put(Key.create(address),
         Value.create(accountCapsule, Type.DIRTY));
@@ -780,6 +804,7 @@ public class RepositoryImpl implements Repository {
     commitDelegatedResourceAccountIndexCache(repository);
     commitTransientStorage(repository);
     commitNewContractCache(repository);
+    commitSelfDestructCache(repository);
   }
 
   @Override
@@ -1139,6 +1164,12 @@ public class RepositoryImpl implements Repository {
   public void commitNewContractCache(Repository deposit) {
     if (deposit != null) {
       newContractCache.forEach(key -> deposit.putNewContract(key.getData()));
+    }
+  }
+
+  private void commitSelfDestructCache(Repository deposit) {
+    if (deposit != null) {
+      selfDestructCache.forEach(key -> deposit.markSelfDestruct(key.getData()));
     }
   }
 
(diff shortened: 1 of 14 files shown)


framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
```diff
@@ -1266,6 +1273,31 @@ public void testSuicideAction2() throws ContractValidateException {
     VMConfig.initAllowTvmVote(0);
   }
 
+  @Test
+  public void testSuicide2RejectsSelfDestructedBeneficiaryAfterFork()
```
CodeNinjaEvan, line 1277: Could we also add a pre-fork case asserting that suicide2() succeeds when VERSION_4_8_2_2 is not activated? Since the beneficiary check is invoked unconditionally and gated inside MUtil, this would explicitly protect backward compatibility before the fork.

CodeNinjaEvan, line 1277: Could we also add a pre-fork case asserting that suicide2() succeeds when VERSION_4_8_2_2 is not activated? Since the beneficiary check is invoked unconditionally and gated inside MUtil, this would explicitly protect backward compatibility before the fork.

TonyStank911:
Just to confirm, is the self-destruct marker intentionally maintained at the repository level so that it can be observed across nested executions within the same transaction?

framework/src/test/java/org/tron/common/runtime/vm/OperationsTest.java
yanghang8612: Thanks for pointing this out. Before VERSION_4_8_2_2 is activated, checkCPUTimeForSelfDestructedBeneficiary() is a no-op and execution continues through the existing suicide2() path. The existing testSuicideAction2() still covers the successful legacy flow, while the newly added test focuses specifically on the rejection behavior introduced after activation.

Given that the behavioral change is confined to the activated branch, we would prefer to keep the current test scope unless there is another pre-activation side effect that needs to be covered.

yanghang8612: Thanks. This appears to be the same concern as the discussion above. The pre-activation check returns without throwing, so the existing suicide2() behavior remains unchanged. I have added the detailed explanation in the other thread.

actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
```diff
@@ -572,6 +573,29 @@ public boolean isNewContract(byte[] address) {
...
+  public boolean isSelfDestructed(byte[] address) {
+    Key key = Key.create(address);
+    if (selfDestructCache.contains(key)) {
+      return true;
+    }
+
+    if (parent != null) {
+      boolean isSelfDestructed = parent.isSelfDestructed(address);
+      if (isSelfDestructed) {
+        selfDestructCache.add(key);
```
lily309, line 591: Is caching the positive result from the parent repository intentional here? My understanding is that the self-destruct state is monotonic within the transaction, while a negative result is left uncached so that a later parent update can still be observed.

ouy95917:
@TonyStank911 Yes. Keeping it in the repository allows nested executions to observe the same transaction-scoped state. It also follows the existing repository commit and rollback lifecycle, so the marker is propagated or discarded together with the corresponding state changes.

actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
```diff
@@ -1142,6 +1167,12 @@ public void commitNewContractCache(Repository deposit) {
     }
   }
 
+  private void commitSelfDestructCache(Repository deposit) {
+    if (deposit != null) {
+      selfDestructCache.forEach(key -> deposit.markSelfDestruct(key.getData()));
```
lily309, line 1172: Does propagating the marker only during commit mean that it follows the same lifecycle as the other repository state, so a reverted child repository requires no separate cleanup?

actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
yanghang8612: Yes, that is the intention. Once an address is marked, it remains self-destructed for the current transaction, so caching a positive result avoids repeated parent lookups. Negative results are not cached because the parent repository may receive the marker later.

actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
yanghang8612: Correct. The marker is propagated to the parent only when the child repository commits. If the child execution is reverted, the child repository and its marker are discarded together.

CodeNinjaEvan (approved):

lvs0075 (approved):

PR #6930, opened 2026-08-27
fix(api): incorrect field mappings in GRPC GetStatsInfo and GetNodeInfo

### LINKED ISSUE #6926: Incorrect field mappings in GRPC GetStatsInfo and GetNodeInfo
# Summary
Two GRPC response fields are populated from incorrect source fields due to copy-and-paste mistakes in the POJO → protobuf conversion layer:

- `NetMetricManager.getNetProtoInfo()` assigns UDP inbound traffic to `tcpOutTraffic` instead of `udpInTraffic`. As a result, `tcpOutTraffic` is overwritten with the UDP inbound value, while udpInTraffic is never populated.

- `NodeInfo.transferToProtoEntity()` populates needSyncFromPeer from peerInfo.isSyncFlag() instead of `peerInfo.isNeedSyncFromPeer()`.

Both fields are read-only status/statistics data. Neither defect affects node networking, synchronization, or consensus behavior.

# Affected API Surface
Both defects occur in the protobuf conversion path and are exposed only through the affected GRPC APIs.

## Defect 1 — tcpOutTraffic / udpInTraffic

The defect is in `NetMetricManager.getNetProtoInfo()` in the framework module and is exposed through GRPC `Monitor.GetStatsInfo (Protocol.MetricsInfo)`.

The HTTP `/monitor/getstatsinfo` endpoint is not affected. MetricsServlet serializes the POJO populated by `NetMetricManager.setNetInfo()`, which maps all four traffic directions correctly.

## Defect 2 — needSyncFromPeer

The defect is in `NodeInfo.transferToProtoEntity()` in the common module and is exposed through GRPC `Wallet.GetNodeInfo`.

The HTTP `/wallet/getnodeinfo` and `/monitor/getnodeinfo` endpoints are not affected. `GetNodeInfoServlet` serializes the POJO directly, and `NodeInfoService` assigns `needSyncFromPeer` correctly.

# Root Cause
## 1. NetMetricManager.getNetProtoInfo()

In the UDP block, the target field was not updated after copying the TCP block:
```
// tcp
RateInfo tcpOutTraffic = net.getTcpOutTraffic();
Protocol.MetricsInfo.RateInfo tcpOUTrafficInfo = tcpOutTraffic.toProtoEntity();
netInfo.setTcpOutTraffic(tcpOUTrafficInfo);
// udp
RateInfo udpInTraffic = net.getUdpInTraffic();
Protocol.MetricsInfo.RateInfo udpInTrafficInfo = udpInTraffic.toProtoEntity();
netInfo.setTcpOutTraffic(udpInTrafficInfo);  // <-- should be setUdpInTraffic
```

The second assignment overwrites the first one. Consequently, `tcpOutTraffic` reports the UDP inbound traffic value, while `udpInTraffic` remains unset and is therefore absent from the response or represented by its default value, depending on the client/output format.

## 2. NodeInfo.transferToProtoEntity()

`isSyncFlag()` was copied into the assignment for `needSyncFromPeer`:

```
peerInfoBuilder.setSyncFlag(peerInfo.isSyncFlag());
peerInfoBuilder.setHeadBlockTimeWeBothHave(peerInfo.getHeadBlockTimeWeBothHave());
peerInfoBuilder.setNeedSyncFromPeer(peerInfo.isSyncFlag());  // <-- should be isNeedSyncFromPeer()
peerInfoBuilder.setNeedSyncFromUs(peerInfo.isNeedSyncFromUs());
```

The POJO already carries the correct value: `NodeInfoService` assigns `needSyncFromPeer` from `PeerConnection.isNeedSyncFromPeer()`. The value is therefore lost only during the POJO → protobuf conversion.

For an active peer, `syncFlag` is normally false, while `needSyncFromPeer` can independently be true when the local node needs to synchronize from that peer. Because the protobuf conversion uses `isSyncFlag()` instead of `isNeedSyncFromPeer()`, the GRPC response fails to expose this state.

# Reproduction

## Defect 1 — GRPC Monitor.GetStatsInfo

`grpcurl -plaintext 127.0.0.1:50051 protocol.Monitor/GetStatsInfo`
Compare the result with the HTTP endpoint backed by the same underlying counters:

`curl -s http://127.0.0.1:8090/monitor/getstatsinfo`
The HTTP endpoint already reports the correct field mapping, making the discrepancy directly observable on the same node.

Illustrative response fragment (values are indicative only):
GRPC GetStatsInfo — before fix
```
{
  "net": {
    "tcpInTraffic":  { "count": 120000, "meanRate": 1000.0 },
    "tcpOutTraffic": { "count": 3000, "meanRate": 25.0 },
    "udpOutTraffic": { "count": 2800, "meanRate": 23.0 }
  }
}
```

Here, `tcpOutTraffic` actually contains the UDP inbound traffic value, while `udpInTraffic` is not populated.
GRPC GetStatsInfo — after fix
```
{
  "net": {
    "tcpInTraffic":  { "count": 120000, "meanRate": 1000.0 },
    "tcpOutTraffic": { "count": 98000, "meanRate": 820.0 },
    "udpInTraffic":  { "count": 3000, "meanRate": 25.0 },
    "udpOutTraffic": { "count": 2800, "meanRate": 23.0 }
  }
}
```

## Defect 2 — GRPC Wallet.GetNodeInfo

`grpcurl -plaintext -emit-defaults 127.0.0.1:50051 protocol.Wallet/GetNodeInfo`

Compare with:

`curl -s http://127.0.0.1:8090/wallet/getnodeinfo`

For each entry in `peerInfoList`, the GRPC response currently sets `needSyncFromPeer` to the same value as `syncFlag`, while the HTTP response reports the two fields independently.
For example, consider an active peer that the local node currently needs to synchronize from:
GRPC `GetNodeInfo` — before fix
```
{
  "syncFlag": false,
  "needSyncFromPeer": false,
  "needSyncFromUs": false
}
```

GRPC `GetNodeInfo` — after fix
```
{
  "syncFlag": false,
  "needSyncFromPeer": true,
  "needSyncFromUs": false
}
```

In this case, the peer is actively connected and `needSyncFromPeer` is true. The current GRPC conversion incorrectly reports false because it uses `syncFlag` as the source field.

# Impact

- Statistics returned by GRPC `Monitor.GetStatsInfo` and `Wallet.GetNodeInfo` are inaccurate and may mislead monitoring, capacity planning, and troubleshooting.
- UDP inbound traffic is not reported through the affected GRPC field, while `tcpOutTraffic` reports the UDP inbound value instead of the actual TCP outbound traffic.
- For connected peers, the GRPC `needSyncFromPeer` field can incorrectly remain false, making the actual per-peer synchronization state unobservable through this API.
- The same node can report different values over HTTP and GRPC for the same underlying counters/state, which can be confusing for consumers using both APIs.
- These are read-only reporting defects. Traffic accounting, peer synchronization, and consensus behavior are unaffected because the underlying counters and `PeerConnection` state are correct.

# Backwards Compatibility
The fix changes values returned by two existing GRPC responses. Operators consuming these fields may observe a step change after upgrading:
- `tcpOutTraffic` will change from the previously reported UDP inbound value to the actual TCP outbound traffic.
- `udpInTraffic` will change from being unset/default to reporting the actual UDP inbound traffic.
- `needSyncFromPeer` will stop mirroring syncFlag and will report the actual per-peer synchronization state.

Dashboards or alert rules that rely on historical values of `tcpOutTraffic`, `udpInTraffic`, or `needSyncFromPeer` should be reviewed before upgrading. In particular, consumers that implicitly relied on `udpInTraffic` being unset or zero may need to update their assumptions.

Consumers using the HTTP endpoints require no changes because those values are already correct.

No protobuf schema change is involved. `Protocol.MetricsInfo.NetInfo` and `Protocol.NodeInfo.PeerInfo` already declare all affected fields, so the generated client interfaces remain unchanged and no client recompilation is required.

# Suggested Fix

## 1. Fix the UDP inbound traffic mapping in NetMetricManager.getNetProtoInfo()
```
- netInfo.setTcpOutTraffic(udpInTrafficInfo);
+ netInfo.setUdpInTraffic(udpInTrafficInfo);
```

## 2. Fix the needSyncFromPeer mapping in NodeInfo.transferToProtoEntity()
```
- peerInfoBuilder.setNeedSyncFromPeer(peerInfo.isSyncFlag());
+ peerInfoBuilder.setNeedSyncFromPeer(peerInfo.isNeedSyncFromPeer());
```

## 3. Add unit tests for the POJO → protobuf conversion
Add unit tests that verify each source field is mapped to the corresponding protobuf field.

For the traffic conversion, build a NetInfo object with distinct values for all four traffic directions and assert each target field independently.

For the peer information conversion, use different values for `syncFlag` and `needSyncFromPeer` and verify that both values are preserved independently.

Distinct test values are important: tests that use the same value for multiple source fields would not have caught either of these mapping defects.

xxo1shine:
**What does this PR do?**
Node and network API responses populated two fields from the wrong source values because of copy-and-paste mapping errors.
Map `needSyncFromPeer` from the corresponding peer state and assign UDP inbound traffic to the `udpInTraffic` protobuf field.

Fixes #6926

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/test/java/org/tron/core/metrics/MetricsApiServiceTest.java
+++ b/framework/src/test/java/org/tron/core/metrics/MetricsApiServiceTest.java
@@ -38,6 +38,10 @@ public class MetricsApiServiceTest extends BaseMethodTest {
 
   @Test
   public void testProcessMessage() {
+    MetricsUtil.getMeter(MetricsKey.NET_TCP_IN_TRAFFIC).mark(1000);
+    MetricsUtil.getMeter(MetricsKey.NET_TCP_OUT_TRAFFIC).mark(2000);
+    MetricsUtil.getMeter(MetricsKey.NET_UDP_IN_TRAFFIC).mark(4000);
+    MetricsUtil.getMeter(MetricsKey.NET_UDP_OUT_TRAFFIC).mark(8000);
 
     MetricsInfo m1 = metricsApiService.getMetricsInfo();
 
@@ -79,6 +83,20 @@ public class MetricsApiServiceTest extends BaseMethodTest {
     Assert.assertEquals(m1.getNet().getErrorProtoCount(), m2.getNet().getErrorProtoCount());
     Assert
         .assertEquals(m1.getNet().getValidConnectionCount(), m2.getNet().getValidConnectionCount());
+
+    long tcpIn = m1.getNet().getTcpInTraffic().getCount();
+    long tcpOut = m1.getNet().getTcpOutTraffic().getCount();
+    long udpIn = m1.getNet().getUdpInTraffic().getCount();
+    long udpOut = m1.getNet().getUdpOutTraffic().getCount();
+
+    Assert.assertNotEquals(tcpOut, udpIn);
+    Assert.assertNotEquals(tcpIn, tcpOut);
+    Assert.assertNotEquals(udpIn, udpOut);
+
+    Assert.assertEquals(tcpIn, m2.getNet().getTcpInTraffic().getCount());
+    Assert.assertEquals(tcpOut, m2.getNet().getTcpOutTraffic().getCount());
+    Assert.assertEquals(udpIn, m2.getNet().getUdpInTraffic().getCount());
+    Assert.assertEquals(udpOut, m2.getNet().getUdpOutTraffic().getCount());
   }
 
 }
--- a/common/src/main/java/org/tron/common/entity/NodeInfo.java
+++ b/common/src/main/java/org/tron/common/entity/NodeInfo.java
@@ -146,7 +146,7 @@ public class NodeInfo {
       peerInfoBuilder.setLastBlockUpdateTime(peerInfo.getLastBlockUpdateTime());
       peerInfoBuilder.setSyncFlag(peerInfo.isSyncFlag());
       peerInfoBuilder.setHeadBlockTimeWeBothHave(peerInfo.getHeadBlockTimeWeBothHave());
-      peerInfoBuilder.setNeedSyncFromPeer(peerInfo.isSyncFlag());
+      peerInfoBuilder.setNeedSyncFromPeer(peerInfo.isNeedSyncFromPeer());
       peerInfoBuilder.setNeedSyncFromUs(peerInfo.isNeedSyncFromUs());
       peerInfoBuilder.setHost(peerInfo.getHost());
       peerInfoBuilder.setPort(peerInfo.getPort());
--- a/common/src/test/java/org/tron/common/entity/NodeInfoTest.java
+++ b/common/src/test/java/org/tron/common/entity/NodeInfoTest.java
@@ -1,0 +1,58 @@
+package org.tron.common.entity;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Test;
+import org.tron.protos.Protocol;
+
+public class NodeInfoTest {
+
+  private PeerInfo newPeerInfo(boolean syncFlag, boolean needSyncFromPeer,
+      boolean needSyncFromUs) {
+    PeerInfo peerInfo = new PeerInfo();
+    peerInfo.setSyncFlag(syncFlag);
+    peerInfo.setNeedSyncFromPeer(needSyncFromPeer);
+    peerInfo.setNeedSyncFromUs(needSyncFromUs);
+    // string fields must be non-null, otherwise the protobuf setters throw NPE
+    peerInfo.setLastSyncBlock("");
+    peerInfo.setHost("127.0.0.1");
+    peerInfo.setNodeId("");
+    peerInfo.setHeadBlockWeBothHave("");
+    peerInfo.setLocalDisconnectReason("");
+    peerInfo.setRemoteDisconnectReason("");
+    return peerInfo;
+  }
+
+  /**
+   * The protobuf conversion must map each peer flag from its own source field. A previous
+   * copy-and-paste defect populated needSyncFromPeer from isSyncFlag(); distinct values for
+   * syncFlag and needSyncFromPeer are required so that such a mismatch is detected.
+   */
+  @Test
+  public void testPeerFlagMappingIsIndependent() {
+    NodeInfo nodeInfo = new NodeInfo();
+    nodeInfo.setBlock("");
+    nodeInfo.setSolidityBlock("");
+    List<PeerInfo> peerList = new ArrayList<>();
+    // syncFlag != needSyncFromPeer so the two fields cannot be confused
+    peerList.add(newPeerInfo(false, true, false));
+    peerList.add(newPeerInfo(true, false, true));
+    nodeInfo.setPeerList(peerList);
+    nodeInfo.setCheatWitnessInfoMap(new java.util.HashMap<>());
+
+    Protocol.NodeInfo proto = nodeInfo.transferToProtoEntity();
+
+    Assert.assertEquals(2, proto.getPeerInfoListCount());
+
+    Protocol.NodeInfo.PeerInfo peer0 = proto.getPeerInfoList(0);
+    Assert.assertFalse(peer0.getSyncFlag());
+    Assert.assertTrue(peer0.getNeedSyncFromPeer());
+    Assert.assertFalse(peer0.getNeedSyncFromUs());
+
+    Protocol.NodeInfo.PeerInfo peer1 = proto.getPeerInfoList(1);
+    Assert.assertTrue(peer1.getSyncFlag());
+    Assert.assertFalse(peer1.getNeedSyncFromPeer());
+    Assert.assertTrue(peer1.getNeedSyncFromUs());
+  }
+}
--- a/framework/src/main/java/org/tron/core/metrics/net/NetMetricManager.java
+++ b/framework/src/main/java/org/tron/core/metrics/net/NetMetricManager.java
@@ -181,7 +181,7 @@ public class NetMetricManager {
     // udp
     RateInfo udpInTraffic = net.getUdpInTraffic();
     Protocol.MetricsInfo.RateInfo udpInTrafficInfo = udpInTraffic.toProtoEntity();
-    netInfo.setTcpOutTraffic(udpInTrafficInfo);
+    netInfo.setUdpInTraffic(udpInTrafficInfo);
     RateInfo udpOutTraffic = net.getUdpOutTraffic();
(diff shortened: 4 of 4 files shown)


lvs0075:
Since both issues are caused by errors in manual field mappings, could we also audit similar POJO-to-protobuf conversions for duplicated, swapped, or omitted fields? I noticed two related cases: NodeInfoService maps syncFlag from peerConnection.isDisconnect(), although syncFlag indicates whether the peer is synchronizing, and NetMetricManager#getNetProtoInfo() does not map LatencyDetailInfo.witness to the protobuf builder. Could we verify and fix these cases and add corresponding regression tests?

framework/src/test/java/org/tron/core/metrics/MetricsApiServiceTest.java
```diff
@@ -38,6 +38,10 @@ protected void afterInit() {
 
   @Test
   public void testProcessMessage() {
+    MetricsUtil.getMeter(MetricsKey.NET_TCP_IN_TRAFFIC).mark(1000);
```
bladehan1, line 41: [NIT] Test isolation backlog 

1. **Make the traffic sentinels independent of global meter history.** `MetricsUtil` keeps a static `MetricRegistry`, while `BaseMethodTest` resets the Spring context and `Args` but not that registry. Adding fixed amounts therefore does not guarantee that the four final cumulative counts remain distinct; prior activity in the same test JVM can make one of the `assertNotEquals` checks fail even when the mapping is correct.

Suggestion: derive marks from captured baseline counts so the resulting values are deterministic and distinct, or swap in and restore a test-scoped registry instead of relying on cumulative global totals.

xxo1shine:
> Since both issues are caused by errors in manual field mappings, could we also audit similar POJO-to-protobuf conversions for duplicated, swapped, or omitted fields? I noticed two related cases: NodeInfoService maps syncFlag from peerConnection.isDisconnect(), although syncFlag indicates whether the peer is synchronizing, and NetMetricManager#getNetProtoInfo() does not map LatencyDetailInfo.witness to the protobuf builder. Could we verify and fix these cases and add corresponding regression tests?

I have completed a comprehensive audit of similar manual POJO-to-protobuf field mappings across the codebase. So far, the only additional issues I found are the two you identified: the incorrect source for syncFlag and the missing mapping for LatencyDetailInfo.witness. I will fix both issues in this PR and add the corresponding regression tests. Thank you for the thorough review.

framework/src/test/java/org/tron/core/metrics/MetricsApiServiceTest.java
xxo1shine: I checked how these meters are used in the test suite. Currently, no other tests write non-zero values to these four traffic meters. This test sets four distinct counts immediately after starting the application and reads the results before any actual network traffic can be generated. Therefore, although the `MetricRegistry` is static, there is no historical data in the current test execution path that could cause these counts to collide, so the fixed sentinel values remain deterministic and distinct. Based on this, I do not think the test needs any additional changes at this time.

halibobo1205 (approved):

waynercheung (changes requested):
[MUST] I verified the two original mapping fixes and their targeted tests locally on `49cc8a2` with JDK 17. Reverting either production line makes its corresponding test fail, so the tests detect the two specific regressions originally reported in #6926. The current head, however, still contains only those two fixes and their regression tests.

In the PR discussion, and again in #6926, you confirmed that this PR would also cover:

- `NodeInfoService` populating `syncFlag` from `isDisconnect()`;
- the missing `LatencyDetailInfo.witness` protobuf mapping;
- the corresponding regression tests and the HTTP / gRPC / release-note scope.

Before merging, please either:

1. push the promised changes, update the description accordingly, and request re-review; or
2. open explicitly linked follow-up issues/PRs and update the description to make the split explicit. This is entirely reasonable; it just needs to be stated.

For option 2, `Fixes #6926` may remain if #6926 is considered complete once its two originally reported defects are fixed. If #6926 is intended to remain the umbrella tracker for all four defects, please downgrade it to `Refs #6926` so the remaining two do not lose their tracking when the issue is closed.

For reference on the two outstanding items:

- **`witness`**: the mapping belongs in the latency-detail loop of `getNetProtoInfo()` (`NetMetricManager.java`, before `detailInfo.setTop99(...)` at line 202 in the current head):

    ```java
    detailInfo.setWitness(detail.getWitness());
    ```

Please cover it with a non-empty witness assertion so an omitted mapping cannotpass.

  - syncFlag: when PeerManager.getPeers() builds its result, it filters out peers for which isDisconnect() is true. Consequently, isDisconnect() is normally false again when NodeInfoService reads it, except for a possible disconnect race; either way, it represents connection lifecycle state rather than synchronization state. This affects both HTTP and gRPC.

This defect is in the PeerConnection -> PeerInfo layer, so it should be covered in the existing NodeInfoServiceTest, not only in NodeInfoTest.

Any new commit needs another review, since the existing approval and CI results cover only 49cc8a2.

waynercheung (approved):
Approving. Scoping this PR to the two originally reported mappings and keeping `Fixes #6926` both look right, which resolves the concern from my earlier review.

Verified on `49cc8a2` (JDK 17): reverting either production line makes its test fail, so both mappings are fixed and effectively guarded.

Two follow-up observations for the record, with no action requested here:

- **`witness`** - the latency-detail loop in `getNetProtoInfo()` (`NetMetricManager.java:199-209`) never copies it, although the POJO sets it at line 241 and it is the only identifier in that message. Every returned `net.latency.detail[]` entry therefore carries the protobuf default empty string and cannot be attributed to an individual witness. Filling it needs no schema change and only replaces a default value.

- **`syncFlag`** - `NodeInfoService:162` sources it from `peerConnection.isDisconnect()` while iterating `PeerManager.getPeers()`, which already filters on `!peer.isDisconnect()`, so it reads `false` for every peer and reflects connection lifecycle rather than synchronization state. It has behaved this way since `ecd451402a` (2018-11-12), which replaced `peerConnection.getSyncFlag()` with `isDisconnect()`, so re-sourcing it would itself be a behaviour change for HTTP and gRPC alike. Correcting or deprecating are both worth evaluating, but not in this PR.

Non-blocking: the description has no behaviour-change section. After upgrading, `tcpOutTraffic` reports the actual TCP outbound value instead of UDP inbound, `udpInTraffic` is no longer unset, and `needSyncFromPeer` no longer mirrors `syncFlag`. Worth a release-note line as you mentioned in #6926 on 2026-08-28, noting that no protobuf schema change and no consensus change is involved.

Happy to open follow-up tracking for these if useful; leaving that call to you.

@lvs0075 FYI on the outcome of the audit you raised.

bladehan1 (approved):

PR #6943, opened 2026-09-02
perf: optimize jump table initialization and reuse

ouy95917:
## What does this PR do?

- Removes unused historical version entries and eagerly initialized table mappings while retaining the existing operation-set builders.
- Centralizes the newest operation set selection through `LATEST_VERSION` and `newLatestOperationSet()`.
- Creates dedicated instances of the latest jump table for transactions and constant calls.
- Reuses both tables across executions.
- Applies configuration-dependent adjustments once per top-level execution.
- Reuses the prepared table for nested contract calls and contract creation.

## Why is this needed?

The previous implementation initialized jump tables for multiple historical versions even though only the latest version was selected for execution.

Transactions and constant calls also shared the same table, which allowed adjustments made for constant calls to affect transaction execution. In addition, configuration-dependent adjustments could be applied repeatedly during nested calls and contract creation.

This change centralizes the selection of the latest operation set and separates transaction and constant-call tables. It reduces startup initialization and repeated adjustment overhead while keeping the two execution paths isolated.

## Tests

- Added coverage for transaction and constant-call table isolation.
- Added coverage for table reuse across executions.
- Verified that constant-call adjustments do not affect the transaction table.
- Verified transaction, constant-call, nested-call, and contract-creation execution paths.
- Verified existing instruction and configuration-dependent behavior.

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/Operation.java
+++ b/actuator/src/main/java/org/tron/core/vm/Operation.java
@@ -52,4 +52,12 @@ public class Operation {
   public boolean isEnabled() {
     return enabled.getAsBoolean();
   }
+
+  public Operation adjustCost(Function<Program, Long> newCost) {
+    return new Operation(opcode, require, ret, newCost, action, enabled);
+  }
+
+  public Operation adjustAction(Consumer<Program> newAction) {
+    return new Operation(opcode, require, ret, cost, newAction, enabled);
+  }
 }
--- a/actuator/src/main/java/org/tron/core/vm/OperationRegistry.java
+++ b/actuator/src/main/java/org/tron/core/vm/OperationRegistry.java
@@ -7,12 +7,45 @@ import org.tron.core.vm.config.VMConfig;
 
 public class OperationRegistry {
 
+  private static final Operation DEFAULT_MLOAD = new Operation(
+      Op.MLOAD, 1, 1, EnergyCost::getMloadCost, OperationActions::mLoadAction);
+
+  private static final Operation DEFAULT_MSTORE = new Operation(
+      Op.MSTORE, 2, 0, EnergyCost::getMStoreCost, OperationActions::mStoreAction);
+
+  private static final Operation DEFAULT_MSTORE8 = new Operation(
+      Op.MSTORE8, 2, 0, EnergyCost::getMStore8Cost, OperationActions::mStore8Action);
+
+  private static final Operation ADJUSTED_MLOAD =
+      DEFAULT_MLOAD.adjustCost(EnergyCost::getMloadCost2);
+
+  private static final Operation ADJUSTED_MSTORE =
+      DEFAULT_MSTORE.adjustCost(EnergyCost::getMStoreCost2);
+
+  private static final Operation ADJUSTED_MSTORE8 =
+      DEFAULT_MSTORE8.adjustCost(EnergyCost::getMStore8Cost2);
+
+  private static final Operation DEFAULT_VOTEWITNESS = new Operation(
+      Op.VOTEWITNESS, 4, 1, EnergyCost::getVoteWitnessCost,
+      OperationActions::voteWitnessAction, VMConfig::allowTvmVote);
+
+  private static final Operation ADJUSTED_VOTEWITNESS =
+      DEFAULT_VOTEWITNESS.adjustCost(EnergyCost::getVoteWitnessCost2);
+
+  private static final Operation OSAKA_VOTEWITNESS =
+      DEFAULT_VOTEWITNESS.adjustCost(EnergyCost::getVoteWitnessCost3);
+
+  private static final Operation DEFAULT_SUICIDE = new Operation(
+      Op.SUICIDE, 1, 0, EnergyCost::getSuicideCost, OperationActions::suicideAction);
+
+  private static final Operation ADJUSTED_SUICIDE =
+      DEFAULT_SUICIDE.adjustCost(EnergyCost::getSuicideCost2);
+
+  private static final Operation RESTRICTED_SUICIDE =
+      DEFAULT_SUICIDE.adjustCost(EnergyCost::getSuicideCost3)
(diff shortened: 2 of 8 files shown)


actuator/src/main/java/org/tron/core/vm/OperationRegistry.java
```diff
@@ -74,28 +115,22 @@ public static JumpTable newTronV15OperationSet() {
...
-    }
-
-    if (VMConfig.allowEnergyAdjustment()) {
-      adjustForFairEnergy(table);
-    }
-
-    if (VMConfig.allowTvmSelfdestructRestriction()) {
-      adjustSelfdestruct(table);
-    }
+  public static JumpTable prepareAndGetTable(boolean isConstantCall) {
```
CodeNinjaEvan, line 118: Nice change — just to confirm, are these config-based adjustments expected to stay stable once this table has been prepared?

TonyStank911:
Just to confirm the intended lifecycle: prepareAndGetTable() is called for the top-level execution, while nested executions reuse the table that has already been prepared for that execution context. Is that correct?

actuator/src/main/java/org/tron/core/vm/Operation.java
```diff
@@ -52,4 +52,12 @@ public void execute(Program program) {
   public boolean isEnabled() {
     return enabled.getAsBoolean();
   }
+
+  public Operation adjustCost(Function<Program, Long> newCost) {
```
lily309, line 56: These helper methods preserve the opcode, stack metadata, and enabled supplier while replacing only the cost or action. Is retaining the same enabled supplier important for keeping the activation behavior identical across the cached variants?

actuator/src/main/java/org/tron/core/vm/OperationRegistry.java
```diff
@@ -21,13 +54,21 @@ public enum Version {
...
+  private static final Version LATEST_VERSION = Version.TRON_V1_5;
+
   static {
-    tableMap.put(Version.TRON_V1_0, newTronV10OperationSet());
-    tableMap.put(Version.TRON_V1_1, newTronV11OperationSet());
-    tableMap.put(Version.TRON_V1_2, newTronV12OperationSet());
-    tableMap.put(Version.TRON_V1_3, newTronV13OperationSet());
-    tableMap.put(Version.TRON_V1_4, newTronV14OperationSet());
-    tableMap.put(Version.TRON_V1_5, newTronV15OperationSet());
+    tableMap.put(LATEST_VERSION, newLatestOperationSet());
```
lily309, line 62: The historical operation-set builders remain available, while only the latest set is registered in tableMap. Is this intended to preserve construction and test compatibility without initializing tables that are no longer selected at runtime?

actuator/src/main/java/org/tron/core/vm/program/Program.java
```diff
@@ -914,7 +914,8 @@ this, new DataWord(newAddress), getContractAddress(), value, DataWord.ZERO(),
       if (VMConfig.allowTvmCompatibleEvm()) {
         program.setContractVersion(getContractVersion());
       }
-      VM.play(program, OperationRegistry.getTable());
+      // Reuse the table prepared by the top-level execution.
+      VM.play(program, OperationRegistry.getTable(isConstantCall()));
```
lily309, line 918: For nested contract creation, getTable(isConstantCall()) assumes that the corresponding table was already prepared by the top-level execution. Is the execution mode inherited consistently by all nested frames?

ouy95917:
@TonyStank911 Yes. The top-level execution prepares the configuration-dependent operations, and nested executions retrieve the already prepared table. This avoids repeating the same adjustment work during internal calls.

actuator/src/main/java/org/tron/core/vm/Operation.java
yanghang8612: Yes. Each variant should differ only in the cost or action being adjusted. Reusing the same enabled supplier ensures that opcode activation behavior and the remaining operation metadata stay unchanged.

actuator/src/main/java/org/tron/core/vm/OperationRegistry.java
yanghang8612: Exactly. The builders are retained for compatibility and targeted testing, but only the currently selected version needs a shared runtime table. This avoids initializing unused table instances.

actuator/src/main/java/org/tron/core/vm/program/Program.java
yanghang8612: Yes. Nested frames inherit the execution mode of the parent context, so they should reuse the table prepared for that mode instead of applying the same adjustments again.

actuator/src/main/java/org/tron/core/vm/OperationRegistry.java
```diff
@@ -7,12 +7,45 @@
 
 public class OperationRegistry {
 
+  private static final Operation DEFAULT_MLOAD = new Operation(
```
alan-eth, line 10: Nice work, the table isolation and cached operation variants make this much cleaner. Do you have rough numbers for the startup and per-execution gains after this change? It would be nice to capture them in the PR description.

yanghang8612: Thanks. We have not collected stable end-to-end benchmark numbers for this change yet. The expected gain comes from avoiding repeated table construction and configuration adjustments for nested executions, so the actual improvement varies with the internal call depth and workload.

To avoid presenting environment-specific numbers that may be misleading, we have kept the performance description qualitative for now.

alan-eth: Thanks for the context, and keeping the description qualitative makes sense.

CodeNinjaEvan (approved):

lvs0075 (approved):

PR #6944, opened 2026-09-02
fix(config): remove ineffective RocksDB table setters

### LINKED ISSUE #6939: [Bug] RocksDB table option setters are ineffective in 4.8.2.1
## Bug Description

In `GreatVoyage-v4.8.2.1`, `RocksDbSettings.getOptionsByDbName()` passes a newly created `BlockBasedTableConfig` to `Options.setTableFormatConfig()` before applying the table-level setters.

`setTableFormatConfig()` creates and attaches the native RocksDB table factory immediately. The subsequent setters only modify the Java `BlockBasedTableConfig` object and do not update the native table factory that was already created.

Affected code:

https://github.com/tronprotocol/java-tron/blob/f8b05d40abc949fa588ab64d8fd8fd82845ebeed/common/src/main/java/org/tron/common/setting/RocksDbSettings.java#L214-L220

As a result, the configured block size, shared block cache, index/filter caching, L0 index/filter pinning, and Bloom filter are not reflected in native RocksDB. The Java configuration and startup logging therefore suggest behavior that the native database does not use.

## Environment

**Network**

Mainnet

**Software Versions**

```
OS: Amazon Linux on AWS c6g.4xlarge (ARM64)
JVM: OpenJDK 17
Git Commit: f8b05d40abc949fa588ab64d8fd8fd82845ebeed
Version: 4.8.2.1
Code: 18825
```

**Configuration**

The replay used:

- A snapshot at block height `83,835,006`
- A 500-block input
- A gp3 volume limited to 7,500 IOPS
- The shipped `storage.dbSettings.blocksize = 64`

The affected release uses RocksDB 9.7.4 on ARM64 and RocksDB 5.15.10 on x86_64.

## Expected Behavior

RocksDB table settings exposed through java-tron configuration and applied by `RocksDbSettings` should be reflected in the native table factory and persisted `OPTIONS-*` configuration.

Alternatively, settings intentionally retained only for compatibility should be clearly documented as inactive and should not be applied through ineffective setters.

## Actual Behavior

The table setters execute after native table-factory creation and do not affect the native `OPTIONS-*` configuration.

For the tested ARM64 build, the effective native table settings remain:

| Native option | Effective value |
| --- | --- |
| `block_size` | 4 KiB |
| Block cache | RocksDB 9.7.4 internal default, 32 MiB |
| `cache_index_and_filter_blocks` | `false` |
| `pin_l0_filter_and_index_blocks_in_cache` | `false` |
| `filter_policy` | `nullptr` |
| `whole_key_filtering` | `true` |
| `block_restart_interval` | 16 |

The x86_64 build uses RocksDB 5.15.10, whose internal default block cache is 8 MiB.

In particular:

- The shipped 64 KiB `blocksize` does not change the native 4 KiB `block_size`.
- The shared 1 GiB cache is not attached.
- Index/filter caching and L0 pinning remain disabled.
- The configured 10-bit Bloom filter is not installed.

Changing only the call order is not a safe compatibility fix because it activates several previously inactive settings simultaneously.

In the fixed 500-block replay, the following multi-variable results were observed:

| Configuration under test | Throughput | Change |
| --- | ---: | ---: |
| Pre-change effective native settings | 2.421 blocks/s | Baseline |
| Previously inactive settings activated, first run | 1.969 blocks/s | -18.67% |
| Previously inactive settings activated, repeat run | 1.680 blocks/s | -30.61% |
| Compatibility path restoring effective native settings | 2.418 blocks/s | -0.12% |

The direct-activation runs reached approximately 7,500 IOPS. Average DB Get latency increased from approximately 0.205 ms to 0.325 ms and 0.420 ms.

These are multi-variable results and must not be attributed to any single RocksDB option.

## Frequency

- [x] Always (100%)
- [ ] Frequently (>50%)
- [ ] Sometimes (10-50%)
- [ ] Rarely (<10%)

## Steps to Reproduce

1. Use the `RocksDbSettings.getOptionsByDbName()` implementation from `GreatVoyage-v4.8.2.1`.
2. Open a fresh RocksDB database with the returned `Options`.
3. Inspect the generated `OPTIONS-*` file.
4. Compare the native table settings with the setters invoked after `setTableFormatConfig()`.

The native configuration contains values such as:

```
block_size=4096
cache_index_and_filter_blocks=false
pin_l0_filter_and_index_blocks_in_cache=false
filter_policy=nullptr
whole_key_filtering=true
block_restart_interval=16
```

These values do not reflect the later Java table setters.

## Logs and Error Messages

```
No exception is thrown. The failure is the mismatch between the Java configuration and the native RocksDB OPTIONS-* output.
```

**Possible Solution**

Preserve the effective behavior used by existing nodes rather than moving `setTableFormatConfig()` after the setters:

- Retain `blocksize` in the configuration and Java API for compatibility, but document that it is not currently applied to native table options.
- Remove the ineffective table setter sequence without activating those settings.
- Retain existing effective Options-level settings.
- Verify the behavior using a fresh temporary database and its native `OPTIONS-*` file.
- Evaluate block size, Bloom filters, shared-cache capacity, index/filter caching, and L0 pinning independently before enabling a new table profile.

Acceptance criteria:

- [ ] `blocksize` and its existing API remain compatible and are documented as inactive.
- [ ] Ineffective table setters are removed without activating them through reordering.
- [ ] A fresh database retains the intended compatibility behavior: 4 KiB blocks, no Bloom filter, and no L0 index/filter pinning.
- [ ] Configuration tests, native Options tests, and Checkstyle pass.

bladehan1:
**What does this PR do?**

Remove RocksDB block-table setters that were applied after `setTableFormatConfig()` and therefore never reached the native table factory.

Keep the existing `storage.dbSettings.blocksize` entries and Java API for compatibility, while documenting in `reference.conf` that the value is not currently applied to native RocksDB table options.

Add regression tests for the shipped Options-level configuration and the effective native block-table behavior.

`setTableFormatConfig()` materializes the native table factory from the Java configuration at call time. The later block-size, shared-cache, index/filter-cache, L0-pinning, and Bloom-filter setters only changed the Java object and did not affect the native factory.

Moving those setters before `setTableFormatConfig()` would activate several unverified settings at once. In a fixed 500-block Mainnet replay, activating the combined settings reduced throughput from 2.421 blocks/s to 1.969 and 1.680 blocks/s, while restoring the pre-change effective native behavior produced 2.418 blocks/s.

This change therefore removes the ineffective setter sequence while preserving the behavior that nodes actually used before the change.

- `./gradlew :common:test --tests org.tron.common.setting.RocksDbSettingsTest`
- `./gradlew :framework:test --tests org.tron.core.config.ConfigurationTest`
- `./gradlew :framework:checkstyleMain`
- `./gradlew :framework:checkstyleTest`

**Follow up**

Evaluate block size, Bloom filters, shared-cache capacity, ownership, lifecycle, and observability independently with representative SR and database-specific workloads before enabling another RocksDB table profile.

**Extra details**

- `blocksize` remains configurable and observable for compatibility but is not applied to the native table factory.
- The existing static shared-cache object and accessor remain for separate configuration design and validation.
- No API, protocol, database-format, or migration changes are introduced.

Closes #6939

### DIFF
--- a/common/src/main/java/org/tron/common/setting/RocksDbSettings.java
+++ b/common/src/main/java/org/tron/common/setting/RocksDbSettings.java
@@ -6,7 +6,6 @@ import java.util.Arrays;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.rocksdb.BlockBasedTableConfig;
-import org.rocksdb.BloomFilter;
 import org.rocksdb.ComparatorOptions;
 import org.rocksdb.InfoLogLevel;
 import org.rocksdb.LRUCache;
@@ -211,13 +210,7 @@ public class RocksDbSettings {
     options.setTargetFileSizeBase(settings.getTargetFileSizeBase());
 
     // table options
-    final BlockBasedTableConfig tableCfg;
-    options.setTableFormatConfig(tableCfg = new BlockBasedTableConfig());
-    tableCfg.setBlockSize(settings.getBlockSize());
-    tableCfg.setBlockCache(RocksDbSettings.getCache());
-    tableCfg.setCacheIndexAndFilterBlocks(true);
-    tableCfg.setPinL0FilterAndIndexBlocksInCache(true);
-    tableCfg.setFilter(new BloomFilter(10, false));
+    options.setTableFormatConfig(new BlockBasedTableConfig());
     if (Constant.MARKET_PAIR_PRICE_TO_ORDER.equals(dbName)) {
       ComparatorOptions comparatorOptions = new ComparatorOptions();
       options.setComparator(new MarketOrderPriceComparatorForRocksDB(comparatorOptions));
--- a/common/src/main/resources/reference.conf
+++ b/common/src/main/resources/reference.conf
@@ -108,7 +108,7 @@ storage {
   dbSettings = {
     levelNumber = 7             // Number of RocksDB levels.
     compactThreads = 0          // 0 = auto: max(availableProcessors, 1)
-    blocksize = 16              // n * KB
+    blocksize = 16              // n * KB. Currently retained for compatibility but not applied to native RocksDB table options.
     maxBytesForLevelBase = 256  // n * MB
     maxBytesForLevelMultiplier = 10 // Level size multiplier.
     level0FileNumCompactionTrigger = 2 // L0 files that trigger compaction.
--- a/common/src/test/java/org/tron/common/setting/RocksDbSettingsTest.java
+++ b/common/src/test/java/org/tron/common/setting/RocksDbSettingsTest.java
@@ -1,0 +1,59 @@
+/*
+ * java-tron is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * java-tron is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.tron.common.setting;
+
+import static org.junit.Assert.assertTrue;
+
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Comparator;
+import java.util.stream.Stream;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.rocksdb.Options;
+import org.rocksdb.RocksDB;
+
+public class RocksDbSettingsTest {
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test
+  public void shouldKeepNativeBlockTableDefaults() throws Exception {
+    Path database = temporaryFolder.newFolder("rocksdb").toPath();
+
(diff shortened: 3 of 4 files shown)


common/src/main/java/org/tron/common/setting/RocksDbSettings.java
halibobo1205, line 50: [SHOULD] `cache` is unused now, remove it ?

common/src/main/java/org/tron/common/setting/RocksDbSettings.java
```diff
@@ -211,13 +210,7 @@ protected void log(InfoLogLevel infoLogLevel, String logMsg) {
...
 
     // table options
-    final BlockBasedTableConfig tableCfg;
-    options.setTableFormatConfig(tableCfg = new BlockBasedTableConfig());
-    tableCfg.setBlockSize(settings.getBlockSize());
-    tableCfg.setBlockCache(RocksDbSettings.getCache());
-    tableCfg.setCacheIndexAndFilterBlocks(true);
-    tableCfg.setPinL0FilterAndIndexBlocksInCache(true);
-    tableCfg.setFilter(new BloomFilter(10, false));
+    options.setTableFormatConfig(new BlockBasedTableConfig());
```
halibobo1205, line 213: [NIT] Remove `options.setTableFormatConfig(new BlockBasedTableConfig());` ?

bladehan1: Retain an entry point for modifying the configuration; this is a change that will require separate verification in the future.

common/src/main/java/org/tron/common/setting/RocksDbSettings.java
bladehan1: The existing static shared-cache object and accessor remain for separate configuration design and validation.
This won't actually use 1GB of memory; it only incurs a small object overhead.

halibobo1205 (approved):

317787106 (approved):

PR #6945, opened 2026-09-02
feat: refine contract deployment transaction validation

ouy95917:
## Why is this needed?

This PR refines two existing validation behaviors for contract deployment transactions.

The `code_hash` and `trx_hash` fields supplied during contract deployment are unused. They do not participate in deployment processing or affect the deployment result. Since these fields have no functional meaning as deployment inputs, transactions carrying them should be preventively prohibited at the protocol level.

Contract name length validation already exists, but its current calculation relies on `String.getBytes()`, which depends on the system default charset. The calculation should use the original protobuf byte representation so that the same deployment transaction is evaluated consistently across environments.

## Historical transaction scan

**We performed a complete scan of all historical transactions across the full chain history.**

**No historical contract deployment transaction was found with either `code_hash` or `trx_hash` set. The number of historical transactions carrying either field is zero.**

This confirms that existing contract deployments have never relied on these fields. The restriction does not invalidate any historical transaction or alter existing user behavior.

## Implementation

After activation:

- A deployment transaction containing a non-empty `code_hash` or `trx_hash` is rejected using the existing timeout semantics immediately before its contract code is persisted.
- Contract name length is calculated directly from the original protobuf bytes instead of using the system-dependent string encoding.

Before activation, the existing behavior is preserved.

Normal contract deployments do not populate the hash fields, and ordinary contract names are unaffected by the length-calculation adjustment.

## Compatibility

Both changes are controlled by `VERSION_4_8_2_2`, with an activation threshold of 70%.

This PR currently uses the same fork version introduced by #6920. If the version identifier or activation parameters in #6920 change, this PR should be updated accordingly.

## Tests

- Verified `code_hash` and `trx_hash` behavior before and after activation.
- Verified empty hash fields remain unaffected.
- Verified contract names at the 32-byte boundary.
- Verified contract names exceeding the 32-byte boundary.
- Verified multibyte names are calculated using protobuf bytes.
- Verified pre-activation compatibility.

### DIFF
--- a/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
+++ b/actuator/src/main/java/org/tron/core/actuator/VMActuator.java
@@ -25,6 +25,7 @@ import org.tron.common.runtime.InternalTransaction.ExecutorType;
 import org.tron.common.runtime.InternalTransaction.TrxType;
 import org.tron.common.runtime.ProgramResult;
 import org.tron.common.runtime.vm.DataWord;
+import org.tron.common.utils.ForkController;
 import org.tron.common.utils.StorageUtils;
 import org.tron.common.utils.StringUtil;
 import org.tron.common.utils.WalletUtil;
@@ -33,6 +34,7 @@ import org.tron.core.capsule.AccountCapsule;
 import org.tron.core.capsule.BlockCapsule;
 import org.tron.core.capsule.ContractCapsule;
 import org.tron.core.capsule.ReceiptCapsule;
+import org.tron.core.config.Parameter;
 import org.tron.core.db.EnergyProcessor;
 import org.tron.core.db.TransactionContext;
 import org.tron.core.exception.ContractExeException;
@@ -217,6 +219,9 @@ public class VMActuator implements Actuator2 {
           } else {
             result.spendEnergy(saveCodeEnergy);
             if (VMConfig.allowTvmConstantinople()) {
+              CreateSmartContract createContract =
+                  ContractCapsule.getSmartContractFromTransaction(trx);
+              checkContractHashFields(createContract.getNewContract());
               rootRepository.saveCode(program.getContractAddress().getNoLeadZeroesData(), code);
             }
           }
@@ -330,6 +335,7 @@ public class VMActuator implements Actuator2 {
     if (contract == null) {
       throw new ContractValidateException("Cannot get CreateSmartContract from transaction");
     }
+
     SmartContract newSmartContract;
     if (VMConfig.allowTvmCompatibleEvm()) {
       newSmartContract = contract.getNewContract().toBuilder().setVersion(1).build();
@@ -341,11 +347,7 @@ public class VMActuator implements Actuator2 {
       throw new ContractValidateException("OwnerAddress is not equals OriginAddress");
     }
 
-    byte[] contractName = newSmartContract.getName().getBytes();
-
-    if (contractName.length > VMConstant.CONTRACT_NAME_LENGTH) {
-      throw new ContractValidateException("contractName's length cannot be greater than 32");
-    }
+    checkContractNameLength(contract.getNewContract());
 
     long percent = contract.getNewContract().getConsumeUserResourcePercent();
     if (percent < 0 || percent > VMConstant.ONE_HUNDRED) {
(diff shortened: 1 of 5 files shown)


actuator/src/main/java/org/tron/core/actuator/VMActuator.java
```diff
@@ -455,6 +457,12 @@ private void create()
 
   }
 
+  static void checkContractHashFields(SmartContract contract) {
+    if (!contract.getCodeHash().isEmpty() || !contract.getTrxHash().isEmpty()) {
```
CodeNinjaEvan, line 461: Covering both fields with a single || and relying on proto3 isEmpty() correctly handles unset vs. empty — normal deployments are untouched.

actuator/src/main/java/org/tron/core/actuator/VMActuator.java
```diff
@@ -217,6 +217,9 @@ public void execute(Object object) throws ContractExeException {
           } else {
             result.spendEnergy(saveCodeEnergy);
             if (VMConfig.allowTvmConstantinople()) {
+              CreateSmartContract createContract =
+                  ContractCapsule.getSmartContractFromTransaction(trx);
+              checkContractHashFields(createContract.getNewContract());
```
lily309, line 224: Is placing this check immediately before saveCode intentional so that the restriction is applied at the code-persistence boundary, while the preceding contract validation flow keeps its existing behavior?

actuator/src/main/java/org/tron/core/vm/utils/MUtil.java
```diff
@@ -59,6 +59,12 @@ public static boolean isNotNullOrEmpty(String str) {
     return !isNullOrEmpty(str);
   }
 
+  public static void checkCPUTimeForContractHashFields() {
+    if (ForkController.instance().pass(Parameter.ForkBlockVersionEnum.VERSION_4_8_2_2)) {
```
lily309, line 63: Is the activation check intentionally kept inside this helper so that the call site can remain unconditional once the persistence path is reached, with the helper becoming a no-op before activation?

actuator/src/main/java/org/tron/core/actuator/VMActuator.java
yanghang8612: Yes. The check is intentionally performed when the deployment reaches the code-persistence stage. If either field is populated after activation, execution is stopped before the deployed code is saved, while earlier validation behavior remains unchanged.

actuator/src/main/java/org/tron/core/vm/utils/MUtil.java
yanghang8612: Correct. Keeping the activation decision in the helper centralizes the fork-dependent behavior. Before activation it returns normally, and after activation it applies the new restriction.

actuator/src/main/java/org/tron/core/actuator/VMActuator.java
```diff
@@ -217,6 +217,9 @@ public void execute(Object object) throws ContractExeException {
           } else {
             result.spendEnergy(saveCodeEnergy);
             if (VMConfig.allowTvmConstantinople()) {
+              CreateSmartContract createContract =
+                  ContractCapsule.getSmartContractFromTransaction(trx);
```
alan-eth, line 223: Quick question: since `create()` already unpacked this `CreateSmartContract` from `trx`, does calling `getSmartContractFromTransaction` again here add any measurable cost on the deployment path?

actuator/src/main/java/org/tron/core/actuator/VMActuator.java
yanghang8612: Yes, exactly. In proto3, an unset bytes field is exposed as ByteString.EMPTY, so isEmpty() treats unset and explicitly empty values consistently. Only a non-empty code_hash or trx_hash reaches the activation-gated restriction, leaving normal contract deployments unchanged.

actuator/src/main/java/org/tron/core/actuator/VMActuator.java
yanghang8612: Thanks for checking. We do not have a separate microbenchmark for this parsing step, but it occurs only once when a contract deployment reaches the reaches" and is about to persist its code, rather than inside an execution loop.

Reading it from the original transaction also keeps the check tied to the fields actually supplied by the caller and avoids retaining additional state between the the validation and execution phases. Compared with contract initialization and code persistence writing, the additional protobuf unpacking cost is expected to be negligible.

alan-eth: Thanks for the detailed explanation,  nothing further from me on this line.

lvs0075 (approved):

CodeNinjaEvan (approved):

PR #6946, opened 2026-09-03
fix(toolkit): keep source DB intact when DbMove copy fails

### LINKED ISSUE #6940: [Bug]Toolkit db mv may delete the source database after a copy failure, causing data loss
## Bug Description

When `Toolkit db mv` copies database files, an IOException from an individual `Files.copy` call is printed but not propagated or recorded as an overall migration failure:

```java
try {
  Files.copy(original, destination, StandardCopyOption.REPLACE_EXISTING);
} catch (IOException e) {
  spec.commandLine().getErr().println(e);
}
```

After copying, the command still attempts to delete the source directory. If deletion succeeds, it replaces the original path with a symbolic link to the destination:

```java
if (FileUtils.deleteDir(p.original.toFile())) {
  Files.createSymbolicLink(p.original, p.destination);
}
```

If copying fails because the destination runs out of disk space or inodes, or because of another I/O error, the source database may be deleted and replaced with a symbolic link to an incomplete LevelDB/RocksDB database. The command may still print `move db done.` and exit with status `0`.

The node may then fail to start and require recovery from a backup or a full resynchronization.

## Environment

**Network**

N/A (offline maintenance tool)

**Software Versions**

```
OS: macOS 15 / Linux
JVM: Java 8 / Java 17
Git Commit: f8b05d40abc949fa588ab64d8fd8fd82845ebeed
Version: GreatVoyage-v4.8.2.1
Code: N/A
```

## Expected Behavior

If any file copy fails, the command should:

- Preserve all source databases.  
- Not create a symbolic link to an incomplete destination.  
- Attempt to clean up incomplete destination data created by the migration.  
- Not print `move db done.`.  
- Return a non-zero exit status.

Source directories should be deleted and replaced with symbolic links only after all databases have been copied successfully. A source deletion or symbolic-link creation failure should also return a non-zero status and provide recovery instructions.

## Actual Behavior

A copy exception is printed and ignored. The command still attempts to delete the source directory and, if deletion succeeds, creates the symbolic link. Copy and finalization failures are not reliably propagated to the caller, so the command may still print `move db done.` and return `0`.

## Frequency

- [x] Always (100%)  
- [ ] Frequently (\>50%)  
- [ ] Sometimes (10-50%)  
- [ ] Rarely (\<10%)

The data-loss path is consistently reached when a file copy fails but the subsequent source deletion and symbolic-link creation succeed.

## Steps to Reproduce

> Use only a disposable database from a stopped node and an isolated test filesystem. Do not fill a production disk to reproduce this issue.  
> 

1. Configure the destination on an isolated test filesystem whose available space is smaller than the source database.  
2. Ensure that the destination database path does not exist.  
3. Run:

```shell
java -jar Toolkit.jar db mv -d output-directory -c config.conf
echo $?
```

4. Wait for the copy to fail because the destination filesystem runs out of space, then inspect the source path, destination directory, and exit status.

Based on the current code flow, if the subsequent deletion and symbolic-link creation succeed, the source path will become a symbolic link to an incomplete destination, while the command still reports success and returns `0`.

## Logs and Error Messages

The following is the expected error form for a full destination filesystem. It illustrates the current code flow and is not a captured reproduction log. Exact text depends on the operating system and filesystem.

```
java.nio.file.FileSystemException: <destination file>: No space left on device
move db done.
(exit code 0)
```

**Related Issues**

N/A

**Possible Solution**

Use a two-phase migration:

1. Copy all databases first. If any copy fails, preserve every source database, attempt to remove incomplete destination data, and return a non-zero status.  
2. Delete the source directories and create symbolic links only after all copies succeed.

A source deletion or symbolic-link creation failure should also return a non-zero status and provide recovery instructions.

halibobo1205:
Closes #6940

**What does this PR do?**

Makes `db mv` a fail-safe two-phase migration: validate first (exit 2, nothing written), then copy every database (recursive, link-aware, fail-fast), and only after all copies succeed replace sources with symlinks. Any copy failure keeps all sources, removes the created destinations, and exits 1 — the same command can simply be re-run. Finalization failures keep the complete copy and print recovery instructions. `move db done.` / exit 0 only on full success. Also documents the stop-the-node precondition in `plugins/README.md`.

The previous implementation ignored per-file copy errors, then deleted the source and symlinked it to an incomplete destination while still reporting success — any I/O fault (disk full, permissions, bad sector) meant silent, unrecoverable data loss (#6940).

- Unit Tests: `DbMoveTest` grown to 15 cases (failure/rollback/retry, recursive copy, symlink and permission faults, exit codes).
- Manual Testing: simulated copy failures on real node data — sources kept, destinations rolled back, exit 1; retry completed the migration.

**Follow up**

None.

**Extra details**

Design writeup with sequence diagram and non-goals: see #6940.

### DIFF
--- a/plugins/src/main/java/common/org/tron/plugins/DbMove.java
+++ b/plugins/src/main/java/common/org/tron/plugins/DbMove.java
@@ -4,17 +4,21 @@ import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.file.Files;
+import java.nio.file.LinkOption;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.Arrays;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
 import me.tongfei.progressbar.ProgressBar;
 import org.tron.plugins.utils.FileUtils;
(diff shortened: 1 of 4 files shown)


plugins/src/main/java/common/org/tron/plugins/DbMove.java
```diff
@@ -76,70 +80,143 @@ public Integer call() throws Exception {
...
+      Files.copy(source, destination, StandardCopyOption.REPLACE_EXISTING);
+    } else {
+      throw new IOException(String.format(
+          "%s is neither a regular file nor a directory, can not be moved.", source));
+    }
+  }
+
+  private boolean replaceSourceWithLink(Property p) {
+    try {
+      if (!FileUtils.deleteDir(p.original.toFile())) {
```
warku123, line 169: Great fix for the copy-failure case. One adjacent gap worth flagging while you're in here: `FileUtils.deleteDir` (plugins `utils/FileUtils.java:73-84`) still recurses with `java.io.File`, and `isDirectory()` follows symlinks — so a directory symlink gets traversed and its **target's contents are deleted**.

This composes with this PR: after `db mv`, the original path is a symlink to the moved DB (`DbMove.java:176`). If a later `db lite merge` deletes the archive dbs under that original path (`DbLite.backupArchiveDbs` calls `FileUtils.deleteDir` per db), it wipes the already-migrated DB on the new disk — two legitimate operations, chained data loss, no attacker needed.

Suggested fix: replace the recursion with `Files.walkFileTree` + `NOFOLLOW_LINKS` (delete the link itself, never follow). The copy phase already uses `NOFOLLOW_LINKS` (`:154`/`:198`), so this would make the delete side consistent. The same pattern also exists in `common/.../FileUtil.java:114-126` and the toolkit copy in tron-docker's `tools/toolkit` — happy to split that into a separate PR if you prefer.

halibobo1205: Thanks for flagging this. The symlink traversal issue is valid, but it predates this PR: `db mv` already created symlinks, and this change does not modify `FileUtils.deleteDir` or the `db lite merge` workflow.

I’ll keep this PR focused on preventing source deletion after copy failures and reporting migration failures correctly. The deletion behavior and `lite merge`’s handling of relocated databases should be addressed in a separate PR, with regression coverage for the combined workflow.

PR #6950, opened 2026-09-04
chore(deps): upgrade grpc, jackson, logback, commons and drop joda-time

halibobo1205:
**What does this PR do?**

Upgrade dependencies and remove obsolete dependencies and compatibility code:

| Dependency | From | To |
|---|---|---|
| `io.grpc:grpc-*` | 1.83.0 | 1.83.1 |
| `com.fasterxml.jackson.core:jackson-databind` | 2.18.6 | 2.18.10 |
| `ch.qos.logback:logback-classic` | 1.2.13 | 1.3.16 |
| `org.slf4j:slf4j-api` / `jcl-over-slf4j` / `jul-to-slf4j` | 1.7.36 | 2.0.17 |
| `org.apache.commons:commons-lang3` | 3.4 | 3.20.0 |
| `org.apache.commons:commons-collections4` | 4.1 | 4.6.0 |
| `org.apache.commons:commons-math` | 2.2 | Removed |
| `joda-time:joda-time` | 2.3 | Removed |

Along with these changes:

- Remove the temporary gRPC stream-limit shim and restore the direct`NettyServerBuilder.maxConcurrentCallsPerConnection(...)` call.
- Replace Joda-Time with `java.time`, introducing `Time.getIsoTimeString` for log formatting and migrating date arithmetic in tests.
- Update the bundled `Toolkit` Logback configuration.

gRPC 1.83.1 enforces the concurrent-stream limit at handler startup, before the client acknowledges SETTINGS. This covers the behavior previously supplied by the local shim. See the [gRPC 1.83.1 release notes](https://github.com/grpc/grpc-java/releases/tag/v1.83.1).

Logback 1.3.16 preserves the x86_64 build's JDK 8 requirement while incorporating fixes absent from 1.2.13, including the CVE-2025-11226 backport and removal of `JaninoEventEvaluator`. Logback 1.3 uses the SLF4J 2.0 provider model, so SLF4J is upgraded alongside it. See [Logback's maintenance status and runtime requirements](https://logback.qos.ch/news.html), the [1.3.16 release notes](https://logback.qos.ch/news.html#1.3.16), and the [1.3.15 security backports](https://logback.qos.ch/news-archive.html#1.3.15).

`commons-math` has no direct source references, and the existing Joda-Time usages can be replaced with JDK APIs.

- 56 targeted tests passed .
- CI green.

**Compatibility notes and migration requirements**

The bundled default logging configurations have been verified. Users supplying a custom `--log-config` should review the following changes and migrate affected configurations before upgrading.

1. **Include paths that depend only on XML-defined properties require migration.**

   In the example below, when `includedir` is defined only in the XML and no system property, environment variable, or fallback value supplies it, Logback 1.3.16 fails to resolve the include path:

   ```xml
   <property name="includedir" value="/etc/tron/logging"/>
   <include file="${includedir}/appenders.xml"/>
   ```

   Under these conditions, the path resolves to `includedir_IS_UNDEFINED/appenders.xml`, relative to the process working directory. Logback 1.2.13 resolves the same configuration correctly. In 1.3.16, `doConfigure` returns without throwing, so this failure does not raise `TronError(LOG_LOAD)`. If the failed include supplies all of the application's logging destinations, application logs are discarded. Independently configured destinations can remain operational.

   The failure is reported to the console: `LogService` calls `StatusPrinter.printInCaseOfErrorsOrWarnings`, which prints `WARN in ch.qos.logback.core.joran.action.IncludeAction - Failed to open [.../includedir_IS_UNDEFINED/appenders.xml]` to the console at startup.

   Verified workarounds are to use a literal absolute include path, or to supply the property through a JVM argument such as `-Dincludedir=/etc/tron/logging`. Upstream documented a fix in 1.5.5; the 1.3.16 artifact used here still exhibits this behavior. See the [Logback 1.5.5 release notes](https://logback.qos.ch/news-archive.html#1.5.5).

2. **Janino-based expression evaluators must be migrated.**

   The removal of `JaninoEventEvaluator` was backported in **Logback 1.3.15** to address CVE-2024-12798, so it applies to this upgrade to 1.3.16. See the [Logback 1.3.15 release notes](https://logback.qos.ch/news-archive.html#1.3.15). An `EvaluatorFilter` using the implicit evaluator syntax `<evaluator><expression>…</expression></evaluator>` behaves differently depending on whether Janino is present on the classpath:

   - Without Janino on the classpath, the tested expression configuration throws `NoClassDefFoundError: org/codehaus/janino/ScriptEvaluator` on 1.2.13. This `Error` bypasses `LogService`'s `catch (Exception)` and aborts startup. On 1.3.16, configuration processing instead returns normally with the filter inactive. The project does not declare Janino as a dependency.
   - On a deployment that adds Janino itself, the filter worked on 1.2.13 and is now inert on 1.3.16, so messages the filter previously denied will be logged. Such deployments must migrate the filtering logic to an explicit `EventEvaluator` implementation.

   The bundled default configuration does not use this feature. See the [evaluator migration instructions](https://logback.qos.ch/manual/filters.html#JaninoEventEvaluator).

3. **JMX-based logging management is no longer available.**

   The entire `ch.qos.logback.classic.jmx` package was removed in 1.3.x, so `<jmxConfigurator/>` is now reported as an unknown property and ignored, where 1.2.13 loaded it silently.

4. **The legacy shutdown hook name still works, with a warning.**

   `DelayingShutdownHook` was renamed to `DefaultShutdownHook`. Logback 1.3.16 still accepts the old name through a compatibility mapping and warns before instantiating the new class, so existing operator configurations keep working. This PR does not rely on that mapping: the bundled Toolkit configuration has been migrated to `DefaultShutdownHook`.

5. **Log files may exceed the configured size threshold.**

   `SizeAndTimeBasedRollingPolicy` checks file size less frequently. Under sustained heavy logging, individual files may temporarily exceed `maxFileSize`. Time-based rollover and `maxHistory` / `totalSizeCap` cleanup are unaffected.

6. **Invalid appender classes may affect other logging destinations.**

   A custom configuration referencing a missing or uninstantiable appender class produces console errors and may also prevent other appender references from being attached.

7. **Reflections scan diagnostics are no longer emitted.**

   SLF4J 2.x removes `org.slf4j.impl.StaticLoggerBinder`, which Reflections 0.9.11 probes to determine whether logging is available. Its own diagnostic logging is therefore disabled. Scanning remains functional, all of its logging sites are null-guarded, and actuator registration failures still surface through `TronError(ACTUATOR_REGISTER)`.

**Timestamp formatting**

Ordinary UTC timestamp output has been verified against the previous implementation. Non-UTC output may use different offsets because the JDK and the older Joda-Time release contain different time-zone databases, while still representing the same instant.


GuipaiQigong111 (dismissed):

lxcmyf (approved):
Re-reviewed at 9147ec22e2. Verified dependency resolution, SLF4J 2 provider packaging and release-JAR smoke tests, gRPC pre-SETTINGS_ACK stream enforcement, and the java.time migration. 117 targeted tests passed (2 ignored), and all required CI checks are green. No blocking issues found.

317787106 (approved):

PR #6954, opened 2026-09-08
feat(api): sanitize HTTP API error responses

### LINKED ISSUE #6936: [Feature] Standardize HTTP API Error Messages
# Summary

Some HTTP APIs return Java exception class names and unaudited `Throwable.getMessage()` values to clients when handling exceptions, for example:

```json
{
  "Error": "class java.lang.NullPointerException : null"
}
```

This proposal standardizes client-facing error messages from standard HTTP Servlets in phase 1:

- Existing messages are temporarily preserved for exceptions explicitly covered for compatibility.  
- All other exceptions return the stable message `internal server error`.  
- Java exception class names are no longer returned to clients, nor are raw exception messages that are not explicitly allowed.

Phase 1 changes only the content and format of failure responses. Successful responses, request rules, HTTP status codes, gRPC behavior, and JSON-RPC behavior remain unchanged.

# Problem

### Motivation

Returning runtime exception information directly to clients has the following problems:

- It provides little value **to** clients **in** diagnosing or handling request failures.  
- Exception messages may change with JDK or dependency versions and are unsuitable as a stable, long-term API contract.

### Current State

Before this change, standard HTTP error-response paths behave inconsistently:

- `Util.processError` returns a concatenation of the Java exception class name and raw message.  
- Some Servlets bypass `Util.processError` and directly return exception class names or `Throwable.getMessage()` values.  
- Two Solidity query endpoints return plain-text error bodies instead of standard JSON.

### Limitations and Risks

- Clients that depend on Java exception class names, raw exception messages, or plain-text error bodies will observe compatibility changes.  
- To limit the compatibility impact of phase 1, non-blank raw messages from three exact exception types are temporarily preserved.

# Proposed Solution

#### 1. Standardize HTTP error-response paths

The standard Servlet JSON and text error-response paths governed by phase 1 are routed through `Util.processError`, which centrally determines the final client-facing message.

#### 2. Preserve compatibility messages for three exact exception types

For compatibility with existing clients, `Util.processError` preserves raw messages only for the following **exact runtime types**:

- `JsonFormat.ParseException`  
- `ContractValidateException`  
- `MaintenanceUnavailableException`

These three cases are **legacy residuals** temporarily retained for compatibility with existing behavior.

#### 3. Preserve two exactly matched fixed messages

The following existing fixed responses are preserved only when both the **exact exception type and exact message constant** match:

- `IllegalAccessException("lack of computing resources")`  
  - Used by the RateLimiter.  
- `IllegalArgumentException(EVENTS_DEPRECATED_MSG)`  
  - Used by the Scan Events deprecation check.

#### 4. Fail closed for all other exceptions

Except for the explicitly allowed compatibility branches above, exceptions must not expose Java exception class names or raw messages to clients. This includes, but is not limited to:

- `NullPointerException`  
- Array or collection bounds exceptions  
- Ordinary `IllegalArgumentException`  
- Other unmapped runtime or internal exceptions

They return:

```json
{
  "Error": "internal server error"
}
```

### Key Changes

- The change is limited to the HTTP Servlet layer in the `framework` module.  
- Standard JSON and text error-response paths governed by phase 1 are routed through `Util.processError`.  
- Java exception class prefixes are removed from standard error responses.  
- `Util.processError` applies compatibility rules using exact runtime types and fixed message constants.  
- Exceptions that are not explicitly allowed return `internal server error`.  
- The two Solidity query endpoints change their error bodies from plain text to standard `{"Error":"..."}` JSON.

# Impact

- **Security**  
  - Standard generic and unclassified error-response paths no longer return Java exception class names or uncontrolled raw messages.  
  - The three compatibility branches that retain raw messages are separately registered and tracked as known residuals.  
- **Stability**  
  - Unknown exceptions use a stable message instead of depending on JDK or library exception messages.  
- **Performance**  
  - The change adds only lightweight exception-type and fixed-message comparisons.  
  - Normal request paths are unaffected.  
- **Developer Experience**  
  - Standard HTTP error-response behavior is centralized in `Util.processError`.

# Compatibility

| Item | Result |
| :---- | :---- |
| **Breaking Change** | **Yes, limited to some HTTP failure responses.** Some error messages will change, and the two Solidity query endpoints will return JSON instead of plain-text error bodies. This change must be included in the release notes. |
| **Default Behavior Change** | **Yes.** Only failure responses change; successful responses remain unchanged. |
| **Migration Required** | **Conditional.** Clients that rely only on successful responses or documented response fields require no migration. Clients that depend on Java exception class names, raw exception messages, or plain-text Solidity error bodies must be updated. |

Affected clients should:

- Read the `Error` field from the standard JSON response.  
- Stop depending on Java exception class names.  
- Stop treating JDK or third-party exception messages as a stable API contract.  
- Handle unknown server errors as `internal server error`.

The following behavior remains unchanged:

- Existing HTTP status codes  
- Status codes and response content for successful requests  
- Request parameters and validation rules  
- gRPC API behavior  
- JSON-RPC API behavior

# Acceptance Criteria

- [ ] Standard Servlet error-response paths governed by phase 1 generate responses through `Util.processError`.  
        
- [ ] These paths no longer concatenate Java exception class names or directly return arbitrary `Throwable.getMessage()` values.  
        
- [ ] Only the exact runtime types `JsonFormat.ParseException`, `ContractValidateException`, and `MaintenanceUnavailableException` may preserve non-blank raw messages.  
        
- [ ] For those three types, a null, empty, or whitespace-only message results in `internal server error`.  
        
- [ ] The two fixed messages are returned only when both the exact exception type and exact message constant match.  
        
- [ ] All other exceptions return:

```json
{
  "Error": "internal server error"
}
```

- [ ] If writing the error response to the client fails, at most one additional `debug` log entry is recorded.   
        
- [ ] The two Solidity query endpoints return standard JSON error responses.  
        
- [ ] HTTP status codes, successful responses, request rules, gRPC behavior, and JSON-RPC behavior remain unchanged.

# Follow-up

Future phases will:

- Distinguish client parameter errors from internal server errors based on the specific HTTP input source.  
- Avoid treating JDK or third-party exception messages as part of the long-term API contract.

# Additional Notes

- **Do you have ideas regarding implementation?** Yes. Route the in-scope standard Servlet error paths through `Util.processError`, and use exact runtime-type and fixed-message matching to determine the client-facing error message.  
    
- **Are you willing to implement this feature?** Yes.

halibobo1205:
**What does this PR do?**

Routes the standard HTTP servlet error paths through a single decision point in `Util.processError`, which determines the client-facing text:

- keeps the raw non-blank message only for the exact runtime types `JsonFormat.ParseException`, `ContractValidateException` and `MaintenanceUnavailableException`; a null, empty or whitespace-only message falls back to `internal server error`;
- keeps the events-deprecation message, and its existing HTTP 400, only for the exact `IllegalArgumentException` type carrying the audited message constant;
- writes the fixed rate-limit and `INVALID address` texts, plus the existing GetBlock validation texts, through a package-private `String` overload used at four audited call sites that bypass exception classification;
- returns `{"Error":"internal server error"}` for every other exception, with no Java exception class name.

Matching is on the exact top-level runtime type: subclasses do not inherit the compatibility exemptions, and cause chains are not unwrapped.

The two SolidityNode transaction-query endpoints also change from bare-text error bodies to standard `{"Error":...}` JSON, and `validateaddress`, `getBrokerage` and `getReward` replace leaked library text in their failure branches with the fixed messages they already use elsewhere.

The full before/after compatibility table is in #6936.

Standard HTTP error responses currently return Java exception class names and unaudited `Throwable.getMessage()` values, for example `{"Error":"class java.lang.NullPointerException : ..."}`. That text gives clients nothing actionable, and it changes with JDK and dependency versions, so it is unsuitable as a long-term API contract; the failure-response format is inconsistent across the API.

- Unit Tests

**Follow up**

**Extra details**

- Scope is the `framework` HTTP servlet layer. No gRPC or JSON-RPC code is modified. The only JSON-RPC-visible change is the shared rate-limit rejection body, which all three JSON-RPC servlets inherit from `RateLimiterServlet`.
- Breaking change for failure responses only, and it needs a release-notes entry: clients that parse Java exception class names, raw exception messages, or the Solidity bare-text error bodies must read the `Error` field of the standard JSON body instead and treat unknown server errors as `internal server error`. The `INVALID address` body also loses the space after the colon, which affects clients matching that body literally.
- Successful responses, request parsing and validation rules, and existing HTTP status codes are unchanged.

Closes #6936

### DIFF
--- a/framework/src/main/java/org/tron/core/services/http/GetNodeInfoServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/GetNodeInfoServlet.java
@@ -1,6 +1,5 @@
 package org.tron.core.services.http;
 
-import java.io.IOException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
@@ -24,12 +23,7 @@ public class GetNodeInfoServlet extends RateLimiterServlet {
       response.getWriter().println(JSON.toJSONString(nodeInfo));
 
     } catch (Exception e) {
-      logger.error("", e);
-      try {
-        response.getWriter().println(Util.printErrorMsg(e));
-      } catch (IOException ioe) {
-        logger.debug("IOException: {}", ioe.getMessage());
-      }
+      Util.processError(e, response);
     }
   }
 
--- a/framework/src/main/java/org/tron/core/services/http/GetNodeInfoServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/GetNodeInfoServlet.java
@@ -23,7 +23,7 @@ public class GetNodeInfoServlet extends RateLimiterServlet {
       response.getWriter().println(JSON.toJSONString(nodeInfo));
 
     } catch (Exception e) {
-      Util.processError(e, response);
+      Util.processServerError(e, response);
     }
   }
 
--- a/framework/src/main/java/org/tron/core/services/http/GetBlockServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/GetBlockServlet.java
@@ -77,9 +77,7 @@ public class GetBlockServlet extends RateLimiterServlet {
         response.getWriter().println("{}");
       }
     } catch (IllegalArgumentException e) {
-      JSONObject jsonObject = new JSONObject();
-      jsonObject.put("Error", e.getMessage());
-      response.getWriter().println(jsonObject.toJSONString());
+      Util.writeAuditedError(e.getMessage(), response);
     }
   }
 
--- a/framework/src/main/java/org/tron/core/services/http/GetBrokerageServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/GetBrokerageServlet.java
@@ -1,6 +1,5 @@
 package org.tron.core.services.http;
 
-import java.io.IOException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
@@ -27,12 +26,7 @@ public class GetBrokerageServlet extends RateLimiterServlet {
       }
       response.getWriter().println("{\"brokerage\": " + value + "}");
     } catch (DecoderException | IllegalArgumentException e) {
-      try {
-        response.getWriter()
-            .println("{\"Error\": " + "\"INVALID address, " + e.getMessage() + "\"}");
-      } catch (IOException ioe) {
-        logger.debug("IOException: {}", ioe.getMessage());
-      }
+      Util.writeAuditedError(Util.INVALID_ADDRESS_MSG, response);
     } catch (Exception e) {
       Util.processError(e, response);
     }
--- a/framework/src/main/java/org/tron/core/services/http/GetBurnTrxServlet.java
+++ b/framework/src/main/java/org/tron/core/services/http/GetBurnTrxServlet.java
@@ -1,6 +1,5 @@
 package org.tron.core.services.http;
 
-import java.io.IOException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
@@ -24,12 +23,7 @@ public class GetBurnTrxServlet extends RateLimiterServlet {
           : "{\"burnTrxAmount\": " + value + "}";
       response.getWriter().println(out);
     } catch (Exception e) {
-      logger.error("", e);
-      try {
-        response.getWriter().println(Util.printErrorMsg(e));
-      } catch (IOException ioe) {
-        logger.debug("IOException: {}", ioe.getMessage());
-      }
+      Util.processError(e, response);
     }
   }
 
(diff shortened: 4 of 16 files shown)


framework/src/main/java/org/tron/core/services/http/GetNodeInfoServlet.java
```diff
@@ -24,12 +23,7 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) {
       response.getWriter().println(JSON.toJSONString(nodeInfo));
 
     } catch (Exception e) {
-      logger.error("", e);
-      try {
-        response.getWriter().println(Util.printErrorMsg(e));
-      } catch (IOException ioe) {
-        logger.debug("IOException: {}", ioe.getMessage());
-      }
+      Util.processError(e, response);
```
bladehan1: [SHOULD] Keep server-side failures visible at production log levels

This replacement also removes the previous ERROR-level exception logging; `Util.processError` records the exception only at DEBUG. The same change occurs in `GetBurnTrxServlet`, `GetPendingSizeServlet`, and `GetRewardServlet`. These broad catches include manager, store, and service failures rather than only client input errors, so production configurations that disable DEBUG lose the stack trace and alerting signal. The PR description also does not disclose this observable log-level change.

Suggestion: preserve WARN/ERROR logging for these four server-side catch paths, or classify client versus server failures in the shared helper, and document the chosen behavior in the PR description.

waynercheung (approved):
LGTM

framework/src/main/java/org/tron/core/services/http/GetNodeInfoServlet.java
halibobo1205: fixed

bladehan1 (approved):

PR #6960, opened 2026-09-10
refactor(config): improve startup errors and remove inactive assertions

bladehan1:
**What does this PR do?**

- Replace selected startup `IllegalArgumentException` failures with `TronError(PARAMETER_INIT)`.
- Remove selected historical Java assertions that are inactive under the default production JVM configuration.
- Remove the test-only `Blake2bfDigest` constructor together with its assertion in the Besu-derived `Blake2bfMessageDigest`.
- Preserve the assertions in the EthereumJ-derived `TrieImpl`.
- Add focused unit and integration tests for the affected configuration error paths.
- Remove two tracked 0-byte Java placeholder files from `chainbase`.

The selected failures are parameter initialization errors. Classifying them with `PARAMETER_INIT` makes startup failures consistent while preserving their messages and exit behavior.

The assertion cleanup is limited to historical java-tron code where the assertions are inactive by default. Assertions in `TrieImpl` are retained because its core implementation is derived from EthereumJ and the checks document internal node-type invariants. The Besu-derived `Blake2bfMessageDigest` assertion lived in a test-only `Blake2bfDigest` constructor with no callers in main or test sources, so the constructor is removed entirely instead of being kept for upstream-maintenance reasons.

The two tracked placeholder files are empty and have no source references. Their deletion is moved here so the CI-only PR remains limited to workflow changes.

- Unit tests:
  - `./gradlew :common:test --tests org.tron.core.config.args.CommitteeConfigTest`
  - `./gradlew :framework:test --tests org.tron.core.config.args.ArgsTest`
  - `./gradlew -g /private/tmp/java-tron-gradle-home :framework:test --tests org.tron.core.tire.TrieTest`
- Compilation after the `Blake2bfDigest` constructor removal:
  - `./gradlew :crypto:compileJava :crypto:compileTestJava`
- Coverage reports:
  - `./gradlew :common:jacocoTestReport :framework:jacocoTestReport`
  - Verified zero missed instructions for `CommitteeConfig.java:166` and `Args.java:1048,1092`.
- Checkstyle:
  - `./gradlew :framework:checkstyleTest`
- Whitespace validation:
  - `git diff --check`

**Follow up**

Other startup failures and any replacement of upstream-derived assertions with explicit runtime checks should be evaluated separately according to their semantics and performance impact.

**Extra details**

- Valid configuration and Trie behavior are unchanged.
- No protocol, database, network, or performance impact is expected.
- `TrieImpl` is excluded from the assertion-removal scope.

### DIFF
--- a/actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
+++ b/actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
@@ -924,7 +924,6 @@ public class RepositoryImpl implements Repository {
     }
 
     if (lastTime != now) {
-      assert now > lastTime;
       if (lastTime + windowSize > now) {
         long delta = now - lastTime;
         double decay = (windowSize - delta) / (double) windowSize;
@@ -972,8 +971,6 @@ public class RepositoryImpl implements Repository {
     long energyWeight = frozeBalance / TRX_PRECISION;
     long totalEnergyLimit = getDynamicPropertiesStore().getTotalEnergyCurrentLimit();
     long totalEnergyWeight = getDynamicPropertiesStore().getTotalEnergyWeight();
-
-    assert totalEnergyWeight > 0;
 
     if (hardenResourceCalculation()) {
       return BigInteger.valueOf(energyWeight)
--- a/chainbase/src/main/java/org/tron/common/zksnark/MerklePath.java
+++ b/chainbase/src/main/java/org/tron/common/zksnark/MerklePath.java
@@ -75,7 +75,6 @@ public class MerklePath {
   }
 
   public byte[] encode() throws ZksnarkException {
-    assert (authenticationPath.size() == index.size());
     List<List<Byte>> pathByteList = Lists.newArrayList();
     long indexLong; // 64
     for (int i = 0; i < authenticationPath.size(); i++) {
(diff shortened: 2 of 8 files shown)


halibobo1205 (approved):

actuator/src/main/java/org/tron/core/vm/repository/RepositoryImpl.java
317787106, line 927: There are many other `assert` in framework/src/main/java/org/tron/core/trie/TrieImpl.java. How to deal with it ? 

How about to delete this method in 
crypto/src/main/java/org/tron/common/crypto/Blake2bfMessageDigest.java ? It also has `assert`.

```
    // for tests
    Blake2bfDigest(
        final long[] h, final long[] m, final long[] t, final boolean f, final long rounds) {
      assert rounds <= 4294967295L; // uint max value
      buffer = new byte[MESSAGE_LENGTH_BYTES];
      bufferPos = 0;

      this.h = h;
      this.m = m;
      this.t = t;
      this.f = f;
      this.rounds = rounds;

      v = new long[16];
    }
```

bladehan1: Retain as-is without additional processing.
Preserve the assertions in the EthereumJ-derived TrieImpl and the Besu-derived Blake2bfMessageDigest implementations.
Since the code was migrated from Besu, we are not performing any unnecessary cleanup.

bladehan1: Removed functions used solely for testing.

317787106 (approved):

317787106 (approved):

PR #6966, opened 2026-09-11
fix(net): harden sync message resource controls

### LINKED ISSUE #6942: Harden sync message validation and resource controls
# Summary

Several validation, rate-limiting, and deduplication issues exist in sync message processing:

- `ChainInventoryMessage.remainNum` is not validated against integer overflow, allowing the future block-height upper-bound check to be bypassed.
- `SYNC_BLOCK_CHAIN` rate limiting can be bypassed under a specific combination of fields.
- An off-by-one mismatch exists between the block-fetch deduplication window and the deduplication cache size, allowing a block ID at the window boundary to be requested repeatedly.

These areas should be hardened to ensure stricter validation and more reliable rate limiting and deduplication during synchronization.

# Root Cause

## 1. Future block-height validation bypass through `remainNum` overflow

`ChainInventoryMsgHandler.check()` uses the following comparison to limit the future block height claimed by a peer:

```java
lastNum + remainNum > maxFutureNum
```

However, `remainNum` is a peer-controlled signed 64-bit value and is not range-checked. When it is set close to `Long.MAX_VALUE`, the addition may overflow and wrap to a negative value. The comparison then evaluates to false, bypassing the validation intended to reject the message.

## 2. `SYNC_BLOCK_CHAIN` rate-limit bypass

The rate-limit check in `SyncBlockChainMsgHandler` depends on `peer.getRemainNum() > 0`.

A crafted request combining `remainNum == 0` with a forged trailing block ID can bypass the rate-limit check, causing the limiter not to be applied to the request.

Rate limiting should apply to the incoming request itself rather than depend on peer-controlled or previously derived synchronization state.

## 3. Off-by-one mismatch between the block-fetch window and cache size

`FetchInvDataMsgHandler` permits block requests within the inclusive range:

```text
[lastSyncBlockId - 2 * SYNC_FETCH_BATCH_NUM, lastSyncBlockId]
```

With `SYNC_FETCH_BATCH_NUM = 2000`, this is:

```text
[last - 4000, last]
```

The inclusive window contains 4,001 distinct block numbers. However, `syncBlockIdCache`, which prevents duplicate requests, has a maximum size of:

```text
2 * SYNC_FETCH_BATCH_NUM = 4000
```

The valid request window is therefore one entry larger than the deduplication cache.

After 4,000 entries have filled the cache, inserting the 4,001st valid entry may evict an older entry due to the cache's maximum-size eviction policy. Because the evicted block ID remains inside the valid request window, it can be requested again and pass the `getIfPresent()` deduplication check.

# Reproduction

## 1. `remainNum` overflow

Construct a `ChainInventoryMessage` that satisfies the preceding continuity and linkage checks, but set `remainNum` close to `Long.MAX_VALUE`.

The addition in the upper-bound check overflows to a negative value, allowing the message to pass validation.

## 2. Rate-limit bypass

Construct a `SYNC_BLOCK_CHAIN` request that combines `remainNum == 0` with a specially crafted trailing block ID.

Observe that the rate limiter is not invoked as expected.

## 3. Deduplication off-by-one

1. Fill `syncBlockIdCache` with 4,000 block IDs from the valid request window.
2. Request the 4,001st block ID in the window, causing an older cache entry to be evicted.
3. Request the evicted block ID again.
4. Because the ID is still within the valid range but no longer exists in the cache, it passes the `getIfPresent()` deduplication check and is fetched again.

# Impact

- **Validation and rate limiting:** A peer may cause the node to accept synchronization messages that should have been rejected or process certain requests without applying the intended rate limit. Other continuity and linkage checks still constrain these messages, so the issues do not affect consensus correctness or asset security.
- **Deduplication off-by-one:** A block ID within the valid request window can fall outside the deduplication cache and become requestable again, causing additional redundant block-fetch work. The impact is bounded: because the handler checks every hash and rejects the entire message if any entry is already cached, at most one entry within the valid window can fall outside the cache at a given time. This results in a rolling single-entry bypass rather than bypassing the entire window.
- Overall, the impact is limited to resource consumption and processing efficiency in the synchronization flow. It does not affect consensus correctness or asset security.

# Suggested Fix

1. **`remainNum`:** Reject negative or otherwise out-of-range `remainNum` values. Use `Math.addExact()` with explicit overflow handling, or use a subtraction-based comparison after validating the relevant ranges, to prevent the future block-height check from being bypassed by integer overflow.

2. **`SYNC_BLOCK_CHAIN` rate limiting:** Remove the `peer.getRemainNum() > 0` precondition and apply rate limiting to every received `SYNC_BLOCK_CHAIN` request.

3. **Deduplication off-by-one:** Increase the maximum size of `syncBlockIdCache` so that it covers the entire valid request window:

   ```text
   2 * SYNC_FETCH_BATCH_NUM + 1 = 4001
   ```

   Alternatively, narrow the permitted block-fetch window by one entry so that it matches the existing cache capacity.

xxo1shine:
**What does this PR do?**
Peer-derived sync state could bypass height validation, request rate limiting, and block-fetch deduplication at boundary conditions.

Reject invalid remain counts and overflowed heights, apply sync-chain rate limiting to every request, and size the block ID cache to cover the full valid fetch window.

Fixes https://github.com/tronprotocol/java-tron/issues/6942

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
@@ -142,16 +142,12 @@ public class ChainInventoryMsgHandler implements TronMsgHandler {
       long maxFutureNum =
           maxRemainTime / BLOCK_PRODUCED_INTERVAL + tronNetDelegate.getSolidBlockId().getNum();
       long lastNum = blockIds.get(blockIds.size() - 1).getNum();
-      long declaredHighestNum;
-      try {
-        declaredHighestNum = Math.addExact(lastNum, msg.getRemainNum());
-      } catch (ArithmeticException e) {
+      long declaredHighestNum = lastNum + msg.getRemainNum();
+      if (lastNum < 0 || declaredHighestNum < 0 || declaredHighestNum > maxFutureNum) {
         throw new P2pException(TypeEnum.BAD_MESSAGE,
-            "lastNum and remainNum overflow", e);
-      }
-      if (declaredHighestNum > maxFutureNum) {
-        throw new P2pException(TypeEnum.BAD_MESSAGE, "lastNum: " + lastNum + " + remainNum: "
-            + msg.getRemainNum() + " > futureMaxNum: " + maxFutureNum);
+            "Invalid declared highest block number: " + declaredHighestNum
+                + ", lastNum: " + lastNum + ", remainNum: " + msg.getRemainNum()
+                + ", futureMaxNum: " + maxFutureNum);
       }
     }
   }
--- a/framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
@@ -143,7 +143,7 @@ public class ChainInventoryMsgHandler implements TronMsgHandler {
           maxRemainTime / BLOCK_PRODUCED_INTERVAL + tronNetDelegate.getSolidBlockId().getNum();
       long lastNum = blockIds.get(blockIds.size() - 1).getNum();
       long declaredHighestNum = lastNum + msg.getRemainNum();
-      if (lastNum < 0 || declaredHighestNum < 0 || declaredHighestNum > maxFutureNum) {
+      if (declaredHighestNum < 0 || declaredHighestNum > maxFutureNum) {
         throw new P2pException(TypeEnum.BAD_MESSAGE,
             "Invalid declared highest block number: " + declaredHighestNum
                 + ", lastNum: " + lastNum + ", remainNum: " + msg.getRemainNum()
--- a/framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
+++ b/framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
@@ -108,6 +108,11 @@ public class ChainInventoryMsgHandler implements TronMsgHandler {
       throw new P2pException(TypeEnum.BAD_MESSAGE, "blockIds is empty");
     }
 
+    if (msg.getRemainNum() < 0) {
+      throw new P2pException(TypeEnum.BAD_MESSAGE,
+          "remainNum is negative: " + msg.getRemainNum());
+    }
+
     if (blockIds.size() > NetConstants.SYNC_FETCH_BATCH_NUM + 1) {
       throw new P2pException(TypeEnum.BAD_MESSAGE, "big blockIds size: " + blockIds.size());
     }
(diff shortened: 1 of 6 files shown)


framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
```diff
@@ -137,9 +142,12 @@ private void check(PeerConnection peer, ChainInventoryMessage msg) throws P2pExc
       long maxFutureNum =
           maxRemainTime / BLOCK_PRODUCED_INTERVAL + tronNetDelegate.getSolidBlockId().getNum();
       long lastNum = blockIds.get(blockIds.size() - 1).getNum();
-      if (lastNum + msg.getRemainNum() > maxFutureNum) {
-        throw new P2pException(TypeEnum.BAD_MESSAGE, "lastNum: " + lastNum + " + remainNum: "
-            + msg.getRemainNum() + " > futureMaxNum: " + maxFutureNum);
+      long declaredHighestNum = lastNum + msg.getRemainNum();
```
lxcmyf, line 145: [MUST] The overflow check is still inside the headBlockId > 0 branch. At height 0, a message containing genesis plus continuous heights 1..1999 and remainNum = Long.MAX_VALUE skips the check, so the original overflow bypass remains. Always perform a checked addition or overflow check; only the futureMaxNum bound may remain conditional. Add a height=0 test.

framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
```diff
@@ -146,7 +146,7 @@ public class PeerConnection {
   private volatile long remainNum;
   @Getter
   private Cache<Sha256Hash, Long> syncBlockIdCache = CacheBuilder.newBuilder()
-      .maximumSize(2 * NetConstants.SYNC_FETCH_BATCH_NUM).recordStats().build();
+      .maximumSize(2 * NetConstants.SYNC_FETCH_BATCH_NUM + 1).recordStats().build();
```
lxcmyf, line 150: [MUST] Raising maximumSize to 4001 still does not guarantee that all IDs in the active window remain cached: entries survive across moving windows, and Guava segmented eviction can evict entries before the total size reaches the limit. An evicted block still inside [last-4000,last] can be requested again. Maintain an exact height-aware window and test all 4001 IDs plus advancing, out-of-order windows.

lxcmyf (commented):

framework/src/main/java/org/tron/core/net/messagehandler/ChainInventoryMsgHandler.java
xxo1shine: This change addresses addition overflow that makes the upper-bound comparison ineffective when height validation is enabled. When `headBlockId <= 0`, this validation was already skipped as part of the existing initial-sync behavior; it is not an overflow bypass introduced by this PR. We will preserve that branch behavior in this change. Whether initial sync should have an independent check on the declared highest block number can be evaluated separately.

framework/src/main/java/org/tron/core/net/peer/PeerConnection.java
xxo1shine: The early eviction caused by segmentation is a valid issue. I will add `concurrencyLevel(1)`, retain the capacity of 4001, and add tests covering retention of every ID within a fixed window, out-of-order requests, and rejection of duplicate requests. The goal of this change is to prevent repeated fetching within a fixed window caused by premature cache eviction. Strict deduplication across advancing windows is outside the scope of this change. This path also has request rate limits and a per-request block count limit, so we will not introduce exact height-aware window management at this stage.

PR #6972, opened 2026-09-16
fix(net): prevent hello message replay

### LINKED ISSUE #6675: Add timestamp validation and deduplication to hello handshake messages
# Summary

When TRON nodes process handshake (Hello) messages, although the messages include a timestamp signature, the receiver currently does not validate the timestamp, nor is there a mechanism for message uniqueness or deduplication. Historical messages could be intercepted and reused to establish connections. Adding timestamp validation and a simple deduplication mechanism can more effectively filter out expired or repeated messages, improving handshake stability and node connection reliability.

# Root Cause

Hello messages include a timestamp signature, but the receiver does not validate its freshness, and there is no unique identifier or deduplication. Historical messages could be intercepted and replayed to establish connections, which slightly affects node connection stability.

# Reproduction

- Construct a Hello message containing a historical timestamp or previously sent content.
- Send this message from different peers to the target node.
- The target node processes each Hello message from each peer normally, without verifying timestamp validity or message uniqueness.

# Impact

- A malicious node that intercepts and replays historical messages could successfully establish connections with the target node.
- This may cause unnecessary connection attempts and additional resource consumption on the target node.

# Suggested Fix

Add lightweight validation in the existing handshake process:
- Implement timestamp validation to ensure it falls within a reasonable time window (e.g., 30s).
- Introduce a simple unique identifier (e.g., based on timestamp) and perform short-term deduplication using a cache.

By using the above methods, handshake message processing can be more standardized and stable without affecting the existing signature mechanism.

xxo1shine:
**What does this PR do?**

Hello messages were accepted without checking timestamp freshness, allowing a previously signed witness message to be replayed indefinitely.

Add a configurable freshness threshold and track the latest verified timestamp per witness so stale or repeated messages are rejected before trust is granted. Adapt the setting to the current NodeConfig-based configuration model while retaining signature-length validation.

Fixes https://github.com/tronprotocol/java-tron/issues/6675

- Unit Tests
- Manual Testing

**Follow up**

**Extra details**


framework/src/main/java/org/tron/core/net/service/relay/RelayService.java
```diff
@@ -156,6 +163,20 @@ public boolean checkHelloMessage(HelloMessage message, Channel channel) {
       return false;
     }
 
+    long now = System.currentTimeMillis();
+    if (now - msg.getTimestamp() > HELLO_MESSAGE_TIMESTAMP_THRESHOLD) {
```
lxcmyf: [MUST] This check only rejects old timestamps. A timestamp beyond the future side of the window, or an extreme value that overflows the subtraction, is accepted and cached, causing subsequent normal Hello messages to be rejected for a long time. Use an overflow-safe two-sided comparison (timestamp < now - threshold || timestamp > now + threshold) and add future-timestamp coverage.

lxcmyf (commented):

framework/src/main/java/org/tron/core/net/service/relay/RelayService.java
```diff
@@ -156,6 +163,20 @@ public boolean checkHelloMessage(HelloMessage message, Channel channel) {
...
+      logger.warn("HelloMessage from {}, timestamp age is {} ms",
+          channel.getInetAddress(), now - msg.getTimestamp());
+      return false;
+    }
+
+    Long lastTimestamp = helloReplayCache.getIfPresent(msg.getAddress());
```
lxcmyf, line 175: [MUST] The getIfPresent → signature verification → put sequence is not atomic. Replays of the same address/timestamp on different channels can pass concurrently and both add trust. After signature verification, commit the timestamp with an address-scoped atomic compare-and-update or equivalent lock, and add a concurrency test.

framework/src/main/java/org/tron/core/net/service/relay/RelayService.java
xxo1shine: I agree that we should validate both sides of the timestamp window. If an SR's clock is ahead, a future timestamp with a valid signature can be cached. Even after the clock is corrected, normal Hello messages will still be rejected because their timestamps are lower than the cached value. Block timestamp validation does not resolve this recovery issue. I will use `timestamp < now - threshold || timestamp > now + threshold` and add tests for future timestamps and extreme values, verifying that normal Hello messages are still accepted after invalid timestamps are rejected.

framework/src/main/java/org/tron/core/net/service/relay/RelayService.java
xxo1shine: Exploiting this scenario requires first obtaining a Hello message with a valid witness signature that is still within the freshness window, then replaying it concurrently over multiple connections. Given these prerequisites and the added complexity of synchronization or atomic update logic, we will retain the current implementation without adding concurrency control in this PR.

framework/src/main/java/org/tron/core/net/service/relay/RelayService.java
```diff
@@ -172,6 +195,7 @@ public boolean checkHelloMessage(HelloMessage message, Channel channel) {
         flag = Arrays.equals(sigAddress, witnessPermissionAddress);
       }
       if (flag) {
+        helloReplayCache.put(msg.getAddress(), msg.getTimestamp());
```
317787106, line 198: [MUST] Obtaining a valid signed Hello is already the premise of the replay attack this PR addresses. The freshness window limits its lifetime, but concurrent copies can still pass the cache check and grant trust to multiple IPs.

  With controlled thread scheduling, I reproduced both duplicate acceptance and timestamp rollback: an older request finishes after a newer one, overwrites the cached timestamp, and allows the newer Hello to be replayed successfully.

  A synchronized block around the timestamp comparison and cache update, after signature verification, would keep the critical section small.. Only requests that successfully update the timestamp should grant trust. Given that this directly affects the replay protection being introduced, I think we should fix it and add concurrency coverage in this PR.

Suggest to replace this `put` with following:
```
synchronized (helloReplayCache) {
    Long lastTimestamp = helloReplayCache.getIfPresent(msg.getAddress());
    if (lastTimestamp != null && timestamp <= lastTimestamp) {
      logger.warn("HelloMessage from {}, timestamp {} is not greater than last {}",
          channel.getInetAddress(), timestamp, lastTimestamp);
      return false;
    }

    helloReplayCache.put(msg.getAddress(), timestamp);
}
```
Move the timestamp check from lines 175–180 into this synchronized block so that the check and update are atomic.

PR #6974, opened 2026-09-16
fix(test): isolate cross-test state leaks and stabilize flaky suites

warku123:
## What does this PR do?

Fixes cross-test state isolation defects so the full test chain passes reliably with retries disabled. All changes are test-only (21 files under `framework/src/test/`), organized as five commits by theme.

**Scope constraint:** every defect is fixed from the test side only — production code is deliberately untouched, even where a production change would be shorter. This keeps the patch strictly inside the test source set with zero risk of changing node behavior; the trade-off is that a few fixes are more verbose than their production-side alternatives (reflection-based teardown assertions, test-local state resetters such as `VMConfigRule` / `PeerManagerStateResetter` / `BackupTestUtils`, and retry loops around mock pools). Follow-ups where a production API could simplify the tests are listed below.

| Defect | Symptom | Fix |
|---|---|---|
| VM flags leaked across tests (`ConfigLoader.disable`, `VMConfig` thread-local snapshot, global snapshot); `PeerManager` static peers/counters and a permanently shut-down executor | `AllowTvmLondonTest`/`ValidateMultiSignContractTest` fail on stale VM config; `MetricsApiServiceTest` `RejectedExecutionException`; peer-count assertions off by prior-class residue | `VMConfigRule` restores the whole VM config domain around every test; `PeerManagerStateResetter` restores cold-start static state before each context |
| Backup manager/server/channel/executors never fully released in teardown | `BackupServerTest` times out after 60s stuck in `shutdownAndAwaitTermination`; `BackupServer:WAITING` thread survives the class | Exception-safe teardown with verified termination |
| Class-wide DPoS producer raced with manual Merkle writes; async worker touched a mock concurrently with stubbing | `ShieldedReceiveTest` intermittent `Rt is invalid.`; `TransactionsMsgHandlerTest` intermittent `WrongTypeOfReturnValue` | Witness schedule initialized via balanced start/stop with no background producer; mock pool keeps stubbing single-threaded |
| Metrics HTTP (9527) and native ZMQ queue (5555) bound fixed ports across parallel worker JVMs | `PrometheusApiServiceTest` `BindException`; `EventLoaderTest` ZMQ `Errno 48`; native queue never stopped in `BlockEventGetTest` | Random ports via existing `chooseRandomPort()` convention; queues stopped in teardown |
| expected/actual swapped in one assertion; shielded flag not restored | Misleading failure output; DPS residue | Argument order fix; snapshot/restore |

Test retry (`maxRetries=5`) was masking first-attempt failures: suites passed while state leaks accumulated across Spring contexts sharing worker JVMs (`maxParallelForks=4`). With retries disabled these defects produced deterministic or high-frequency red suites.

**How these were found:** the same commit was run repeatedly with retries disabled across JDK/arch lanes to expose first-attempt failures; failures were correlated with per-class residue of shared state (JVM shutdown diffing), and causality was confirmed by fault injection — injecting a suspected leaked state made the victim suite fail while control arms passed. The port conflicts and the Mockito race were traced from recurring CI failure fingerprints to their call sites.

**One test required more than isolation.** `TransactionsMsgHandlerTest#testProcessMessage` had a vacuous smart-contract section: its assertion checked the previous transaction's request map, the scheduler-submit verification was commented out, and two preloaded `TrxEvent(null, null)` entries were later drained into the real pool by the production scheduler. It now uses a non-executing mock pool (which also removes the Mockito cross-thread race) and drives both queue paths with real transactions; the asserted intent is unchanged or strictly strengthened.

## This PR has been tested by

- Full chain (`clean build` → `:framework:testWithRocksDb` → `jacocoTestReport`) with retries disabled, 3 rounds on JDK8/x86_64 **and** 3 rounds on JDK17/aarch64, all green: https://github.com/warku123/java-tron/actions/runs/35065905472 (identical test-file content to this branch, verified by zero-diff comparison)
- Targeted retry-disabled runs per fixed suite; independent review of all 20 files with no findings at reporting threshold

## Follow up

- Production `Metrics` has no `stop()`/`getPort()` and `NativeMessageQueue` remaps port 0 to the default — test-side random-port probing could be replaced by real ephemeral binds if those APIs are ever added.
- `PeerManager.close()/init()` is not restart-safe in production; making it so would allow the test resetter to be narrowed or removed.

## Extra details

Open question — test retry configuration. All verification for this PR ran with retries disabled `(-PmaxRetries=0)`: six full unit-test runs (3× JDK8/x86_64 + 3× JDK17/aarch64) with zero failures. The current CI default retries failed tests up to 5 times, which can mask precisely the cross-test state leaks this PR eliminates. I kept this change strictly test-side and did not touch CI configuration, so whether to lower (or zero-out) the retry count on some lanes is open for discussion — happy to include it here or split it into a follow-up PR.


framework/src/test/java/org/tron/common/utils/PeerManagerStateResetter.java
```diff
@@ -0,0 +1,90 @@
...
+  private PeerManagerStateResetter() {
+  }
+
+  public static synchronized void reset() {
+    // 1) Drain the old executor first: let running/queued check() tasks die out so they
+    // cannot interleave with the list/counter reset below.
+    ScheduledExecutorService executor = getFieldValue("executor");
+    if (executor != null && !executor.isShutdown()) {
+      ExecutorServiceManager.shutdownAndAwaitTermination(executor, EXECUTOR_NAME);
+    }
```
317787106, line 55: [SHOULD] `isShutdown()` does not guarantee that existing tasks have finished. If the old executor is already shut down but check() is still running, this condition skips waiting and resets the shared state. The old task can then decrement the newly zeroed counters—I reproduced an active-peer count of -1 with a controlled thread interleaving.

Could we wait for any non-terminated executor and verify isTerminated() before replacing it or clearing peers/counters? If termination fails, setup should fail rather than proceed with the reset.

framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
```diff
@@ -86,11 +88,16 @@ public void testProcessMessage() {
...
-      smartContractQueue.offer(new TrxEvent(null, null));
+
+      CountDownLatch smartContractSubmitted = new CountDownLatch(1);
+      ExecutorService mockPool = Mockito.mock(ExecutorService.class);
+      Future<?> submittedTask = Mockito.mock(Future.class);
+      Mockito.when(mockPool.submit(Mockito.any(Runnable.class))).thenAnswer(invocation -> {
+        smartContractSubmitted.countDown();
+        return submittedTask;
+      });
+      originalPool = replaceTrxHandlePool(transactionsMsgHandler, mockPool);
```
317787106: [SHOULD] The mock pool is installed after the first processMessage() call has already submitted work to the real pool. Replacing the pool does not stop that task, so it can still call peer.isBadPeer() while the test re-stubs peer.getAdvInvRequest(). I reproduced ConcurrentHashMap cannot be returned by isBadPeer() with a controlled thread interleaving.

  Could we install the non-executing mock pool before the first submission, or wait for the original task to finish before re-stubbing peer? If moving the mock earlier, please keep the scheduler latch specific to smart-contract submissions so the initial transfer cannot satisfy it.

framework/src/test/java/org/tron/common/utils/PeerManagerStateResetter.java
warku123: Agreed — thanks for the controlled-interleaving repro. `isShutdown()` here was intended as a
fast-path guard, but it conflates "shut down" with "drained". Two details make the -1 outcome
deterministic once the interleaving happens: `ExecutorServiceManager.shutdownAndAwaitTermination`
can return from `PeerManager.close()` while a `check()` is still running (warn-and-continue on
the 60s+60s await, or the InterruptedException path), and `check()` decrements the counters
unconditionally even when its `peers.remove(peer)` is a no-op — so any decrement landing after
step 4's `set(0)` produces a negative count.

Fixed as you suggest, gating on termination instead of shutdown:

```java
ScheduledExecutorService executor = getFieldValue("executor");
if (executor != null && !executor.isTerminated()) {
  ExecutorServiceManager.shutdownAndAwaitTermination(executor, EXECUTOR_NAME);
  if (!executor.isTerminated()) {
    throw new IllegalStateException(
        "peer-manager executor did not terminate; refusing to reset shared state");
  }
}
```

Calling `shutdownAndAwaitTermination` unconditionally on a non-terminated pool costs nothing
when it has already terminated, so the old guard bought no performance. The javadoc/step-4
comments claiming "old tasks can no longer decrement at this point" are corrected as well, and
a new `PeerManagerStateResetterTest` now asserts the cold-start postcondition directly (peers
empty, both counters 0, fresh executor not shut down) so this ordering guarantee is pinned by a
test rather than by comments.

Pushed as 7460d132aa.

framework/src/test/java/org/tron/core/net/messagehandler/TransactionsMsgHandlerTest.java
warku123: Agreed — this is the exact race documented in this test's own `testInvalidSigLength` comment,
and the phase-1 transfer submission still went through the real pool before the mock was
installed. Moving the mock earlier is the right option: the phase-1 assertion (advInvRequest
entry removed) only exercises the synchronous removal inside `processMessage`, not the pool, so
installing the non-executing mock before the first submission weakens no asserted behavior, and
real-pool wiring stays covered by `testDuplicateTransactionRejected` and `testHandleTransaction`.

Fixed: the mock pool is now installed before the first `processMessage`, and the latch stays
specific to the smart-contract scheduler — with the real pool replaced, the only off-main-thread
caller of `submit` is the `smartContractExecutor` thread, so the stub's Answer counts down only
when invoked from a non-test thread. The 3s latch therefore remains a genuine scheduler-path
signal and cannot be satisfied early by the transfer submission.

Pushed as 7460d132aa.

framework/src/test/java/org/tron/common/logsfilter/NativeMessageQueueTest.java
```diff
@@ -6,13 +6,16 @@
...
 import org.zeromq.SocketType;
 import org.zeromq.ZContext;
 import org.zeromq.ZMQ;
 
 public class NativeMessageQueueTest {
 
-  public int bindPort = 5555;
+  // Random port avoids fixed 5555 conflicts; note invalidBindPort/invalidSendLength still
+  // remap to DEFAULT_BIND_PORT (5555) in production start() — known low-risk residual.
+  public int bindPort = PublicMethod.chooseRandomPort();
```
bladehan1, line 18: [NIT] invalidBindPort/invalidSendLength still bind the fixed port 5555

`start(-1111, 0)` and `start(0, -2222)` hit the remap branch in production `NativeMessageQueue.start()` (bindPort <= 0 -> DEFAULT_BIND_PORT = 5555), so these two tests still bind fixed port 5555 across parallel worker JVMs — the same BindException/Errno 48 failure class this PR eliminates elsewhere. The in-code comment already acknowledges this as a known low-risk residual, and the real fix (production-side port-0/ephemeral bind support) is on the follow-up list, so this does not block the PR.

Suggestion: track the production follow-up explicitly (issue or TODO reference) so the residual does not get forgotten; alternatively, make these two tests assert the remap behavior without depending on a real successful bind.

framework/src/test/java/org/tron/common/utils/PeerManagerStateResetter.java
```diff
@@ -0,0 +1,99 @@
...
+ * <p>Wired broadly from BaseTest/BaseMethodTest against unknown prior pollution; the reset is
+ * idempotent and cheap for tests that never use PeerManager. Remove this utility once
+ * production {@code close()}/{@code init()} is restart-safe.
+ */
+public final class PeerManagerStateResetter {
+
+  private static final String EXECUTOR_NAME = "peer-manager";
+
+  private PeerManagerStateResetter() {
+  }
```
bladehan1, line 40: [NIT] Pin the PeerManager static field set so future fields cannot silently skip the reset

This resetter restores a hard-coded list of PeerManager static fields (peers, passivePeersCount, activePeersCount, executor). A field rename fails loudly, but if production later adds a new mutable static field (e.g. another counter), the resetter will silently not restore it and cross-test leakage will regress in a new shape with no signal. VMConfigRule avoids this for Snapshot by iterating getFields(); the equivalent guard here is cheap.

Suggestion: extend PeerManagerStateResetterTest into a field-inventory pin — assert the set/count of PeerManager's static fields matches expectations, so adding one fails the test and forces a resetter update. Long term this utility can be narrowed or removed once production close()/init() is restart-safe, as its javadoc already states.

framework/src/test/java/org/tron/common/VMConfigRule.java
```diff
@@ -0,0 +1,40 @@
+package org.tron.common;
+
+import java.lang.reflect.Field;
+import org.junit.rules.ExternalResource;
+import org.tron.common.parameter.CommonParameter;
+import org.tron.core.vm.config.ConfigLoader;
+import org.tron.core.vm.config.VMConfig;
+
+/** Restores VM flags after each test, including failed setup and assertion paths. */
+public class VMConfigRule extends ExternalResource {
```
bladehan1, line 10: [NIT] VMConfigRule covers method scope only; class-scope VM flag leaks still rely on manual afterClass lists

As a @Rule (method-level), this does not wrap @BeforeClass/static blocks. AllowTvmCompatibleEvmTest sets 6 VM flags in @BeforeClass and still restores them via a hand-maintained @AfterClass list — and the missing initAllowTvmLondon(0) entry fixed in this very PR is proof that the manual-list mechanism keeps leaking. The 13 VMTestBase subclasses inherit the rule, but any class-scope flag mutation remains uncovered.

Suggestion: add a @ClassRule-level VM snapshot restore on VMTestBase (or BaseTest/BaseMethodTest) so @BeforeClass leaks are covered automatically; alternatively, state explicitly in this class's javadoc that class-scope mutations are not covered and document the recommended pattern. Fine as a follow-up, does not block.

framework/src/test/java/org/tron/common/VMConfigRule.java
```diff
@@ -0,0 +1,40 @@
+package org.tron.common;
+
+import java.lang.reflect.Field;
+import org.junit.rules.ExternalResource;
+import org.tron.common.parameter.CommonParameter;
+import org.tron.core.vm.config.ConfigLoader;
+import org.tron.core.vm.config.VMConfig;
+
+/** Restores VM flags after each test, including failed setup and assertion paths. */
```
bladehan1, line 9: [NIT] Follow-up & residual-risk notes (3 items rolled up)

Grouped as one comment since all are doc-or-follow-up asks rather than code defects.

1. **VMConfigRule drift note (framework/src/test/java/org/tron/common/VMConfigRule.java, class javadoc).** The restore domain is complete today (Snapshot fields via getFields(), vmTrace, ConfigLoader.disable, ENERGY_LIMIT_HARD_FORK), and a field rename fails loudly. The one silent drift surface is a future VMConfig static mutable field added outside Snapshot, which this rule would not restore. Suggestion: add one javadoc line stating that new VMConfig static mutable fields (outside Snapshot) must be mirrored here.

2. **chooseRandomPort() TOCTOU window (e.g. framework/src/test/java/org/tron/common/logsfilter/NativeMessageQueueTest.java:18).** The pick-check-release-rebind convention still leaves a race window between parallel worker JVMs, so port flakes are reduced but not zero. Suggestion: once the production follow-ups (Metrics.stop()/getPort(), NativeMessageQueue port-0 support) land, switch tests to real ephemeral binds instead of probing.

3. **CI maxRetries=5 unchanged (framework/build.gradle, not in this diff — cross-cutting concern, documentation/tracking gap).** The "6 green runs with retries disabled" baseline lives only in the fork's CI; after merge, the default retry count can silently mask regressions of exactly the leaks this PR fixes. Suggestion: file a tracked follow-up to lower maxRetries to 0 or 1 on some lanes, as raised in the PR description's open question.

Suggestion: address each item as a small follow-up (javadoc line, ephemeral-bind migration, tracked CI retry change); none blocks this PR.

framework/src/test/java/org/tron/common/logsfilter/NativeMessageQueueTest.java
warku123: We considered this while preparing the PR. For `testInvalidSendLength` the port is incidental, so we can bind a random free port there and remove one of the two fixed-5555 bindings — happy to push that. For `testInvalidBindPort` the remap to `DEFAULT_BIND_PORT` is itself the behavior under test, so a valid random port would change the semantics; that binding stays and is tracked explicitly (in-code comment + PR Follow-up) until production supports port-0. Asserting remap without a real bind would weaken coverage, so we preferred explicit tracking.

framework/src/test/java/org/tron/common/utils/PeerManagerStateResetter.java
warku123: Agreed — added a pin test (`resetterCoversAllPeerManagerStaticFields` in `PeerManagerStateResetterTest`). It reflectively enumerates every static field of `PeerManager` and fails unless the field is either handled by `reset()` (peers, executor, activePeersCount, passivePeersCount) or explicitly allowlisted (thread-name constant, disconnect timeout, logger). A future static field turns the test red until it is classified. Audited the current `reset()` while at it — no missed fields today.

Pushed as 9f4e65d4f1.

framework/src/test/java/org/tron/common/VMConfigRule.java
warku123: Documented in the `VMConfigRule` javadoc: the Snapshot enumeration blind spot (static flags added to `VMConfig`/`ConfigLoader` must be mirrored into `VMConfig.Snapshot` or they leak across tests) and the method-level-rule boundary (`@BeforeClass` writes are outside the baseline and need a manual `@AfterClass` reset — the missed London flag was exactly this gap, fixed in 9c8dac05d1). A `@ClassRule`-based variant that would cover class-level setup mechanically is noted as a follow-up; kept out of this PR to avoid churning shared test infra further.

Pushed as 9f4e65d4f1.

framework/src/test/java/org/tron/common/VMConfigRule.java
warku123: All three tracked:
1. The javadoc note is in — see the reply on the rule comment (9f4e65d4f1).
2. `chooseRandomPort` TOCTOU stays in the PR Follow-up list; it will be replaced by real ephemeral-port binding once production supports it (same dependency as the invalid-port fallback below).
3. The retry-count question is raised in the PR description (open-question section); we'll follow it up as a separate CI-side change/discussion so this PR stays strictly test-side.

317787106 (approved):

bladehan1 (approved):

lxcmyf:
[SHOULD] The documented `-PmaxRetries=0` validation does not disable test retries. `framework/build.gradle` hard-codes `retry.maxRetries = 5` and never reads that project property; with `-PmaxRetries=0`, the effective value remains 5. Please wire the property into `configureTestTask` or add a dedicated zero-retry CI lane, then rerun or update the verification claim so first-attempt failures are not masked.

PR #6975, opened 2026-09-17
refactor(api): remove dead WalletExtension gRPC service and config

### LINKED ISSUE #6931: [Feature] Remove the dead WalletExtension gRPC service and config
# Summary

The `WalletExtension` gRPC service has been registered but not implemented since [Odyssey-v3.7](https://github.com/tronprotocol/java-tron/releases/tag/Odyssey-v3.7) (2020-03): all four of its RPCs return `UNIMPLEMENTED`. This proposal removes the unused Java implementation, the `node.walletExtensionApi` configuration item that controls it, and the `service WalletExtension` definition in `api.proto`. It also removes the messages that are no longer used without this service: `AccountPaginated`, `TransactionList`, `TransactionListExtention`, `TimeMessage`, and `TimePaginatedMessage`.

# Problem

### Motivation

java-tron still includes a gRPC service, a config switch, and a set of proto RPC definitions that do not work. The in-memory index that once supported these RPCs was removed long ago because it used too many node resources for the value it provided; transaction information by account address is now provided by [TronGrid](https://developers.tron.network/reference/get-transaction-info-by-account-address). The config name may make node operators think it controls whether a working API is exposed, while the proto definitions cause SDK code generators to create client stubs for methods that can never succeed.

### Current State

1. **Proto definitions:** `protocol/src/main/protos/api/api.proto` still defines four `WalletExtension` RPCs, all of which take `AccountPaginated`. `GetTransactionsFromThis` and `GetTransactionsToThis` return `TransactionList`; their `*2` versions return `TransactionListExtention`. These RPCs are the only users of the three message types in any `.proto` file in the repository. `TimePaginatedMessage` and the `TimeMessage` it contains are also unused after their `*ByTimestamp` RPCs were removed in #909.
2. **Runtime behavior:** `RpcApiService.WalletExtensionApi` provides no RPC implementation: it extends `WalletExtensionGrpc.WalletExtensionImplBase` but overrides none of its RPC methods, and its only private helper, `transactionList2Extention()`, has no callers. Every request therefore uses the generated default handler and returns `UNIMPLEMENTED`. Registration does not change this behavior. The service is added only on a solidity node when `node.walletExtensionApi = true`; `reference.conf` sets the key to `false` by default, but the bundled `framework/src/main/resources/config.conf` sets it to `true`. The old HTTP endpoints were removed with the RPC implementations in #2533, leaving only unused integration-test helpers.

### Limitations or Risks

This does not add a new runtime risk. The main issue is unnecessary confusion. Node operators may think `node.walletExtensionApi` controls a working API, while generated SDKs still include methods that always return `UNIMPLEMENTED`. The remaining code and proto definitions also require maintenance.

# Proposed Solution

### Proposed Design

1. **Remove the Java implementation:** Delete the `WalletExtensionApi` class, its registration branch in `RpcApiService.addService()`, and the `WalletExtensionGrpc` import.
2. **Delete the proto definitions:** Remove `service WalletExtension` and its four RPCs from `api.proto`, together with the five messages listed above.
3. **Remove the configuration item:** Delete `node.walletExtensionApi` from `CommonParameter`, `NodeConfig`, `Args`, and all bundled `.conf` files. If the old key is still present in an operator config, log a removal warning at startup.
4. **Clean up the remaining unused code:** Remove `Util.printTransactionList()` from main sources, whose only caller is its own unit test; remove the unused WalletExtension test code from `GrpcClient`, `WalletClient`, and `HttpMethed`; and delete the getter/setter assertions for the removed parameter from `ParameterTest`.

### Key Changes

- **Modules:** `framework`, `common`, `protocol`, and related test code
- **API / Behavior Surface:** The four `service WalletExtension` RPCs are no longer registered. Their gRPC status remains `UNIMPLEMENTED`; only the status description changes, from unimplemented-method wording to unregistered-service wording.

# Impact

The change keeps the same gRPC status code and has no performance effect other than registering one fewer service at startup. It makes the node configuration and code easier to understand and maintain: it removes a config item that looks like an API switch but does not control a working service, stops generating SDK stubs for methods that can never succeed, and removes unused code from about a dozen files.

# Compatibility

*   Breaking Change: Yes, for clients that use the removed definitions. Runtime behavior and the wire format do not change because all four RPCs already return `UNIMPLEMENTED`. Clients such as wallet-cli or Trident may keep their own copy of `api.proto` or generated gRPC code; if that code uses `WalletExtensionGrpc` or the five messages listed above, it must be updated when the new proto is adopted.
*   Default Behavior Change: No.
*   Migration Required: Only for clients that use the removed definitions; no action is required for node operators. Existing `walletExtensionApi` entries are ignored with a startup warning.

# Additional Notes

0xbigapple:
**What does this PR do?**

close https://github.com/tronprotocol/java-tron/issues/6931.
Removes the dead `WalletExtension` gRPC service and everything reachable only from it:

- `api.proto`: `service WalletExtension` (`GetTransactionsFromThis/2`, `GetTransactionsToThis/2`), plus messages `AccountPaginated`, `TransactionList`, `TransactionListExtention` (referenced only by these four RPCs) and `TimeMessage` / `TimePaginatedMessage` (request types of the WalletExtension `*ByTimestamp` RPCs deleted in 2018, orphaned ever since)
- `RpcApiService`: the registration branch and the empty `WalletExtensionApi` inner class
- The `node.walletExtensionApi` config item: `CommonParameter` / `NodeConfig` fields, the `Args` binding, and the key in `reference.conf` / `config.conf` / `config-shield.conf`. Following the retirement convention for `node.*` keys, `NodeConfig.fromConfig` now logs a removal warning when the old key is still present in an operator config
- Dead code only reachable from the above: `Util.printTransactionList` (sole caller was its own mock test), the WalletExtension stub and wrappers in test utilities `GrpcClient` / `WalletClient`, `HttpMethed.getTransactions{From,To}ThisFromSolidity` (targets `/walletextension/*` HTTP paths that have no servlet), and commented-out `getTransactionsByTimestamp` / `getAssetIssueListByTimestamp` blocks

`WalletExtension` has had no implementation in any release since v3.7 (2020-03): `RpcApiService$WalletExtensionApi` overrides none of the four RPCs, so every call falls through to the generated `ImplBase` default handlers and returns `UNIMPLEMENTED`. This makes `node.walletExtensionApi` behavior-irrelevant — enabled, a solidity node registers a service with zero implemented methods (advertised via gRPC reflection when `node.rpc.reflectionService` is on); disabled, callers get the same `UNIMPLEMENTED`. Removing it also resolves the default-value inconsistency between `config.conf` (`true`) and `reference.conf` (`false`).

Six years of unconditional `UNIMPLEMENTED` rules out any functional dependency, so the service is removed directly without a deprecation period, following existing practice for dead interfaces.

- Unit Tests
- Manual Test

**Follow up**

Ecosystem code that still compiles against the removed stubs/messages needs a sync: the `tronprotocol/protocol` mirror, the documentation site, and older wallet-cli/trident versions. Compile-time impact only — runtime behavior is unchanged (`UNIMPLEMENTED` before and after).

**Extra details**

None.

### DIFF
--- a/common/src/main/java/org/tron/core/config/args/NodeConfig.java
+++ b/common/src/main/java/org/tron/core/config/args/NodeConfig.java
@@ -26,7 +26,6 @@ public class NodeConfig {
 
   // ---- Flat scalar fields (auto-bound by ConfigBeanFactory) ----
   private String trustNode = "";
-  private boolean walletExtensionApi = false;
   private int syncFetchBatchNum = 2000;
   private int maxPendingBlockSize = 500;
   private int validateSignThreadNum = 0; // 0 = auto (availableProcessors)
@@ -325,6 +324,12 @@ public class NodeConfig {
       logger.warn("Configuring [node.maxActiveNodesWithSameIp] is deprecated and will be removed "
           + "in a future release. Please use [node.maxConnectionsWithSameIp] instead.");
       nc.maxConnectionsWithSameIp = section.getInt("maxActiveNodesWithSameIp");
+    }
+
+    // node.walletExtensionApi (removed): the WalletExtension gRPC service no longer exists
+    if (section.hasPath("walletExtensionApi")) {
+      logger.warn("Configuring [node.walletExtensionApi] has been removed and is ignored. "
+          + "The WalletExtension gRPC service no longer exists.");
     }
 
     // Legacy key fallback: node.allowShieldedTransactionApi wins fullNodeAllowShieldedTransaction
--- a/common/src/main/java/org/tron/common/parameter/CommonParameter.java
+++ b/common/src/main/java/org/tron/common/parameter/CommonParameter.java
@@ -295,9 +295,6 @@ public class CommonParameter {
   public String trustNodeAddr; // clearParam: ""
   @Getter
   @Setter
-  public boolean walletExtensionApi;
-  @Getter
-  @Setter
   public boolean estimateEnergy;
   @Getter
   @Setter
--- a/common/src/main/resources/reference.conf
+++ b/common/src/main/resources/reference.conf
@@ -185,9 +185,6 @@ node {
   # Trust node for solidity node (example: "127.0.0.1:50051").
   trustNode = ""
 
-  # Expose extension api to public or not
-  walletExtensionApi = false
-
   listen.port = 18888   # P2P listen port.
   fetchBlock.timeout = 500 # Block fetch timeout (ms).
 
--- a/framework/src/main/java/org/tron/core/config/args/Args.java
+++ b/framework/src/main/java/org/tron/core/config/args/Args.java
@@ -598,7 +598,6 @@ public class Args extends CommonParameter {
     PARAMETER.trustNodeAddr = nc.getTrustNode();
 
     PARAMETER.validateSignThreadNum = nc.getValidateSignThreadNum();
-    PARAMETER.walletExtensionApi = nc.isWalletExtensionApi();
     PARAMETER.isOpenFullTcpDisconnect = nc.isOpenFullTcpDisconnect();
     PARAMETER.nodeDetectEnable = nc.isNodeDetectEnable();
 
--- a/framework/src/main/java/org/tron/core/services/RpcApiService.java
+++ b/framework/src/main/java/org/tron/core/services/RpcApiService.java
@@ -63,13 +63,10 @@ import org.tron.api.GrpcAPI.TransactionApprovedList;
 import org.tron.api.GrpcAPI.TransactionExtention;
 import org.tron.api.GrpcAPI.TransactionIdList;
 import org.tron.api.GrpcAPI.TransactionInfoList;
-import org.tron.api.GrpcAPI.TransactionList;
-import org.tron.api.GrpcAPI.TransactionListExtention;
 import org.tron.api.GrpcAPI.TransactionSignWeight;
 import org.tron.api.GrpcAPI.ViewingKeyMessage;
 import org.tron.api.GrpcAPI.WitnessList;
 import org.tron.api.MonitorGrpc;
-import org.tron.api.WalletExtensionGrpc;
 import org.tron.api.WalletGrpc.WalletImplBase;
 import org.tron.api.WalletSolidityGrpc.WalletSolidityImplBase;
 import org.tron.common.application.RpcService;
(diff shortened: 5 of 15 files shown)


common/src/main/java/org/tron/core/config/args/NodeConfig.java
```diff
@@ -327,6 +326,12 @@ public static NodeConfig fromConfig(Config config) {
       nc.maxConnectionsWithSameIp = section.getInt("maxActiveNodesWithSameIp");
     }
 
+    // node.walletExtensionApi (removed): the WalletExtension gRPC service no longer exists
+    if (section.hasPath("walletExtensionApi")) {
```
bladehan1, line 330: [NIT] First node.* key removed without a deprecation period — please record the exception in release notes

`node.walletExtensionApi` was a published config key (present in both shipped samples: `config.conf` defaulted it to `true`, `reference.conf` to `false`). The project config convention calls for a full deprecation cycle before removing a published key; this PR removes it in one step, with only this warn-and-ignore fallback. The practical impact is nil — old configs still boot (unknown keys are ignored) and operators get a clear warning, and `ArgsTest.testRemovedWalletExtensionApiKeyIsIgnored` locks that behavior in — so this reads as a documentable exception rather than a contract violation.

Suggestion: list the removal explicitly in the release notes ("Removed: WalletExtension gRPC service (4 RPCs) + node.walletExtensionApi config key") so the skipped deprecation cycle is visible to operators and future reviewers.

common/src/main/java/org/tron/core/config/args/NodeConfig.java
```diff
@@ -327,6 +326,12 @@ public static NodeConfig fromConfig(Config config) {
       nc.maxConnectionsWithSameIp = section.getInt("maxActiveNodesWithSameIp");
     }
 
+    // node.walletExtensionApi (removed): the WalletExtension gRPC service no longer exists
```
bladehan1, line 329: [NIT] PR description reflection claim doesn't match the code

(Cross-cutting note anchored at the removal warning as the most representative added line.)

The PR description says the only observable effect of enabling `node.walletExtensionApi` was that "gRPC reflection advertises a service that always fails", but this repo does not register `ProtoReflectionService` (no grpc-reflection dependency, no registration in `RpcApiService`). The actual observable effect is narrower: the solidity node registers a service whose every method returns UNIMPLEMENTED.

Suggestion: correct that sentence in the PR description, since it may be quoted in future decisions.

0xbigapple: Reflection is registered (`RpcService:116`, behind `node.rpc.reflectionService`), but you're right that the sentence was imprecise — it only applies to solidity nodes with reflection on. Description updated.

common/src/main/java/org/tron/core/config/args/NodeConfig.java
0xbigapple: Thanks for the suggestion. Agreed on the release note, will add it.

PR #6976, opened 2026-09-17
feat(API): refactor merge http servlets

### LINKED ISSUE #6922: [Feature] Deduplicate HTTP servlet stacks with cursor filters and a declarative endpoint registry
# Summary

java-tron currently maintains the same HTTP API surface four times: the default FullNode service, the Solidity service, the PBFT service and the standalone SolidityNode service. The Solidity and PBFT stacks consist of ~96 wrapper servlets whose only job is switching the per-thread read cursor, and every service keeps its own hand-written path-to-servlet registration list.

This proposal removes the duplication in two steps — both are part of this proposal; the ordering only serves a safe rollout:

- **Step 1 — cursor filters**: replace all cursor-switching wrapper servlets with one servlet `Filter` per cursor service, so all services share the single base servlet implementations (removes ~98 classes, ~2,900 lines).
- **Step 2 — endpoint registry**: replace the four hand-written registration lists with one declarative table (endpoint → servlet → access nature → exposed surfaces), from which every service builds its mappings; enforce at startup that non-read endpoints exist only on the FullNode surface.

# Problem

### Motivation

Adding or changing one HTTP endpoint today requires synchronized edits in up to 4 wrapper classes and 4 registration lists. Missing one spot silently produces **feature drift**: the same endpoint behaves differently (or is missing) depending on which service serves it.

### Current State

- `interfaceOnSolidity.http` contains 49 wrapper servlets and `interfaceOnPBFT.http` contains 47, each of the form:

```java
public class GetAccountOnSolidityServlet extends GetAccountServlet {
  @Autowired private WalletOnSolidity walletOnSolidity;
  protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
    walletOnSolidity.futureGet(() -> super.doGet(req, resp));
  }
  protected void doPost(HttpServletRequest req, HttpServletResponse resp) {
    walletOnSolidity.futureGet(() -> super.doPost(req, resp));
  }
}
```

- `WalletOnSolidity` / `WalletOnPBFT` extend `WalletOnCursor`; `futureGet` is a same-thread bracket:

```java
dbManager.setCursor(cursor);   // SOLIDITY or PBFT (PBFT offset computed per call)
try { run(); } finally { dbManager.resetCursor(); }
```

- The underlying read cursor (`Chainbase.cursor`) is a `ThreadLocal`: it only selects the snapshot the current thread starts reading from, and is invisible to every other thread. Block processing, sync and broadcast run on dedicated executors that never touch the cursor, and each Jetty port owns a separate thread pool.
- The standalone SolidityNode's `SolidityNodeHttpApiService` is a wholesale copy of `HttpApiOnSolidityService` — the same wrapper servlets and registration, just with a different cursor wrapper. The PBFT service (`HttpApiOnPBFTService`) checks out the same way: verified endpoint by endpoint, it serves the same read-only interface as those two (bar the drift noted below), just with the PBFT cursor wrapper.
- Registration lists are hand-written per service: 123 mappings on the default FullNode service, 45 on the Solidity service, 47 on the PBFT service, 45 on the standalone service — 128 unique endpoints declared 260 times.

### Limitations or Risks

- **Maintenance burden**: ~98 wrapper/forked classes are pure boilerplate that must track every base servlet change.
- **Feature drift is real, not hypothetical**: the two SolidityNode forks missed later improvements of the base servlets — standard JSON error responses and `visible=true` log address conversion — so the same endpoint already answers differently across deployments:

  ```java
  // fork (GetTransactionInfoByIdSolidityServlet): logs stay hex even when visible=true,
  // and errors are written as raw text
  response.getWriter().println(JsonFormat.printToString(transInfo, visible));
  ...
  } catch (Exception e) {
    response.getWriter().println(e.getMessage());
  }

  // base (GetTransactionInfoByIdServlet): converts log addresses for visible=true,
  // and errors go through the standard JSON error body
  transactionInfo = transactionInfo.toBuilder().clearLog()
      .addAllLog(Util.convertLogAddressToTronAddress(transactionInfo)).build();
  ...
  } catch (Exception e) {
    Util.processError(e, response);
  }
  ```

  The endpoint sets of the Solidity and PBFT services also differ in ways documented nowhere: 2 read-only endpoints — `getpaginatednowwitnesslist` and `gettransactioninfobyblocknum` — exist only on the Solidity service, and 5 dormant shielded-TRX endpoints existed only on the PBFT service:

  | Path (on the PBFT service only) | Servlet |
  |---|---|
  | `/walletpbft/getmerkletreevoucherinfo` | `GetMerkleTreeVoucherInfoServlet` |
  | `/walletpbft/scanandmarknotebyivk` | `ScanAndMarkNoteByIvkServlet` |
  | `/walletpbft/scannotebyivk` | `ScanNoteByIvkServlet` |
  | `/walletpbft/scannotebyovk` | `ScanNoteByOvkServlet` |
  | `/walletpbft/isspend` | `IsSpendServlet` |

  These trace to a 2020 cleanup that commented the shielded-TRX endpoints out on the other services but missed the PBFT service created five weeks earlier, and the leftover went unnoticed for six years because the PBFT service is disabled by default. Building the registry surfaced this immediately; the oversight is confirmed and fixed as part of this work.
- **Convention-only safety**: nothing enforces that cursor services expose read-only endpoints. A registration mistake exposing a write endpoint on the Solidity or PBFT service would let a cursor-switched thread enter the write path (`Chainbase.put` and `SnapshotManager.advance` resolve through the cursor-aware `head()`).

# Proposed Solution

### Proposed Design

**Step 1 — cursor filters (equivalent refactor).**

One abstract filter reproduces the `futureGet` bracket at the transport entry; two `@Component` subclasses parameterize the cursor:

```java
public abstract class WalletCursorFilter implements Filter {
  public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain)
      throws IOException, ServletException {
    try {
      dbManager.setCursor(cursor);   // Solidy/PBFT offset computed inside Manager#setCursor per request
      chain.doFilter(req, resp);
    } finally {
      dbManager.resetCursor();       // always back to HEAD: no residue on pooled jetty threads
    }
  }
}
```

The filter is appended (after the existing `liteFnQueryHttpFilter` / `httpApiAccessFilter`) at `/*` of the Solidity and PBFT contexts, so every request on a cursor service — including the `/wallet/getnodeinfo` alias on the Solidity service — reads the corresponding state view. The services then register the **base** servlet beans on unchanged path strings, and all wrapper servlets are deleted. Equivalence holds because the bracket is same-thread and always resets, exactly like `futureGet`; the standalone SolidityNode service likewise drops its 2 forks in favor of the base servlets.

One deliberate alignment is worth calling out: the `triggerconstantcontract` / `estimateenergy` wrappers wrap `super.doPost` in a `try/catch (IOException)` that logs and swallows the exception — not by design, but because `futureGet(Runnable)` takes a lambda that cannot throw checked exceptions. With the filter there is no lambda, so this workaround disappears: on the Solidity / PBFT services an `IOException` from response writing now propagates to jetty **exactly as it already does on the default FullNode service**. The only affected scenario is a client that dropped the connection mid-response; the three services become uniform instead of subtly different.

**The read-only invariant.** A thread whose cursor is set to SOLIDITY / PBFT must never enter a write path. The cursor does not only redirect reads: `Chainbase.put()/delete()` and `SnapshotManager.advance()/retreat()` all resolve their target snapshot through the cursor-aware `head()`, so a cursor-switched thread executing a write (e.g. a broadcast calling `buildSession()`) would write into the solidified view or advance the shared snapshot chain from the wrong position — corrupting state that every thread sees. This is unchanged from the existing wrapper mechanism; the filter neither widens nor narrows it. Today the invariant holds because both cursor services expose **only read-only endpoints** (audited endpoint by endpoint: all handlers are pure queries, and the two constant-call endpoints buffer VM writes in an in-memory `Repository` that is discarded without commit) — but it holds by convention only. The standalone SolidityNode surface carries the same read-only requirement for a different reason: that process only syncs solidified blocks from a FullNode and has no path to propagate a transaction into the network, so a write endpoint there could only strand transactions (today all of its 45 endpoints are read-only as well). The general rule — **any non-read endpoint may exist only on the FullNode surface** — is exactly what Step 2 turns into a boot-time check.

**Step 2 — endpoint registry (single source of truth).**

A declarative table describes each endpoint once:

```java
public enum HttpApiDef {
  GET_ACCOUNT("getaccount", GetAccountServlet.class,
      Access.READ, Surface.FULL, Surface.SOLIDITY, Surface.PBFT, Surface.SOLIDITY_NODE),
  SCAN_NOTE_IVK("scannotebyivk", ScanNoteByIvkServlet.class,
      Access.READ, Surface.FULL, Surface.PBFT),                  // surface differences become explicit
  BROADCAST_TRANSACTION("broadcasttransaction", BroadcastServlet.class,
      Access.WRITE, Surface.FULL)                                // write endpoints stay FULL-only
  // ... one row per endpoint (123 in total)
}
```

Each service's `addServlet` collapses into one loop over `HttpApiDef.forSurface(...)`, resolving servlet beans from the application context (which also removes the ~40 injected servlet fields per service). Rollout is one service per commit — PBFT (most regular) → Solidity (handles the dual-mount alias) → standalone SolidityNode → FullNode (largest, carries all BUILD/WRITE rows) — each verified by a 1:1 endpoint diff against the previous hand-written list.

Finally the table is armed: at startup the node **refuses to boot** if any non-READ row declares a surface other than FULL (covering both cursor services and the solidified-only standalone SolidityNode), and the endpoint parity test is generated from the table instead of being maintained by hand.

The same table also lets an accidental gap in a read surface be closed. The two endpoints found only on the Solidity service — `getpaginatednowwitnesslist` and `gettransactioninfobyblocknum` — are **read-only and not surface-specific**: each answers a general query against whatever snapshot the current cursor selects, with no behavior tied to a particular service. Read-only means they are safe to serve from a cursor service (no write-path risk); being general-purpose rather than dedicated interfaces means there is no functional reason to scope them to Solidity alone. They are therefore declared on the PBFT surface as well — not on a blanket rule that every read must appear on every surface (a read may legitimately be surface-specific), but because these two have no reason to be withheld from PBFT. This also keeps the two transports in step: the follow-up gRPC merge (see References) serves `RpcApiServiceOnPBFT` from the shared read service, which brings these same two methods onto the PBFT gRPC service by construction. Aligning the HTTP surface here means HTTP and gRPC expose the same PBFT read set instead of the gap reopening on the other transport.

### Key Changes

- **Module**: `framework` only.
  - New: `WalletCursorFilter` + `SolidityCursorFilter` / `PbftCursorFilter` (~60 lines); `HttpApiDef` registry.
  - Changed: the four HTTP service classes register base servlets (Step 1) and then consume the registry (Step 2).
  - Deleted: 96 cursor wrappers, 2 SolidityNode forks, and their orphan tests (~2,900 lines).
- **Configuration**: none (`http.fullNodePort` / `solidityPort` / `PBFTPort` and enable switches keep their semantics).
- **API**: external paths, ports and response semantics unchanged, except two read-only endpoints are additionally exposed on the PBFT surface (see Compatibility).

# Impact

- **Developer Experience**: an endpoint is implemented once and declared once; surface differences and access nature become reviewable data instead of tribal knowledge.
- **Stability**: cursor mechanics are equivalent to the current `futureGet` (same thread, guaranteed reset). Verified by concurrency tests crossing HEAD × SOLIDITY × PBFT requests and by registry-generated endpoint parity tests.
- **Security**: `httpApiAccessFilter` / `liteFnQueryHttpFilter` behavior is untouched; the read-only constraint of cursor services and of the solidified-only standalone SolidityNode is upgraded from review convention to a startup-enforced invariant.
- **Performance**: one filter invocation replaces one wrapper method call per request — neutral.

# Compatibility

- **Breaking Change**: No — paths, ports and response formats are unchanged.
- **Default Behavior Change**: No, with the following deliberate alignments:
  1. `triggerconstantcontract` / `estimateenergy` on the Solidity / PBFT services no longer swallow `IOException` from response writing (the old wrappers had to catch it inside a lambda); they now behave exactly like the default FullNode service.
  2. The two standalone SolidityNode endpoints gain the base servlets' standard JSON error format and `visible=true` log address conversion — i.e., existing drift is fixed, not introduced.
  3. The 5 dormant shielded-TRX endpoints are removed from the PBFT service (a confirmed missed deletion from the 2020 sweep; with the enabling committee parameter never activated, they could only return "not supported" errors).
  4. The two endpoints previously served only on the Solidity service — `getpaginatednowwitnesslist` and `gettransactioninfobyblocknum` — are now also served on the PBFT surface. The change is additive (existing paths and callers are untouched); both are read-only, general-purpose queries with no surface-specific behavior, so serving them on PBFT closes an accidental gap between the two cursor surfaces. It mirrors the follow-up gRPC merge, which serves the same two methods on the PBFT service by construction.
- **Migration Required**: No.

# References (Optional)

- Involved classes: `WalletOnCursor`, `WalletOnSolidity`, `WalletOnPBFT`, `Chainbase` (`ThreadLocal` cursor), `Manager#setCursor/resetCursor`, `SnapshotManager`.
- `WalletOnSolidity` / `WalletOnPBFT` remain in use by the gRPC services and jsonrpc; they are removed only after the follow-up below.
- Follow-up (separate issue/PR): the gRPC twins `RpcApiServiceOnSolidity` / `RpcApiServiceOnPBFT` (~980 lines of `futureGet` delegation) can be merged the same way with a `ServerInterceptor`; gRPC call lifecycle needs its own verification.

# Additional Notes

SeriousCoding789:
**What does this PR do?**

Implements https://github.com/tronprotocol/java-tron/issues/6922.

`develop` serves the four HTTP surfaces (FULL, SOLIDITY, PBFT, and the standalone SolidityNode) with parallel sets of servlets and four hand-written registration lists. This PR collapses them onto one servlet per endpoint:

- **Deletes the 100 per-surface servlets** (`*OnSolidityServlet`, `*OnPBFTServlet`, `http/solidity/*SolidityServlet`). 98 of them carry no logic at all — the whole class body is `walletOnSolidity.futureGet(() -> super.doGet(req, resp))`, i.e. a class per endpoint whose only job is to switch the read cursor.
- **Replaces them with a cursor filter** (`WalletCursorFilter` + `SolidityCursorFilter` / `PbftCursorFilter`) mounted on `/*` of the solidity and pbft ports. Switching the cursor is a per-request, thread-level concern; it does not need a subclass per endpoint.
- **Replaces the four registration lists with `@HttpApi` / `@HttpApiExcluded` declared on the servlet itself**, and derives a read-only `HttpApiRegistry` from them by classpath scan. Adding an endpoint becomes a one-place change instead of a four-place change.
- The registry is **validated at startup, before Jetty binds**. Duplicate or malformed suffixes, a non-`READ` endpoint on a cursor surface, a servlet declaring neither annotation or both, an endpoint declared on a nested or abstract class, and a missing `@Component` all fail the node with `TronError(API_SERVER_INIT)` instead of silently dropping an endpoint.

356 files changed, +2136 / −4751.

Duplicating an endpoint across surfaces is not free — it drifts silently, and two live examples on `develop` came out of this work:

- The standalone SolidityNode's own copy of `gettransactioninfobyid` never picked up the `visible=true` → `convertLogAddressToTronAddress` step the base servlet has, so it returns `log[].address` in hex where FullNode returns base58.
- The PBFT registration list has been out of step with the other three surfaces for years: 5 sapling endpoints that were taken off FULL/SOLIDITY in 2020 stayed active on PBFT, and 2 read endpoints the other three surfaces expose were never mounted there.

Both are "change one place, forget the other" bugs. With one servlet per endpoint and a derived registry, a surface can no longer fall behind on its own.

**Behaviour differences vs `develop`**

Every per-surface servlet was classified by whether its body contains `futureGet`: 98 pure cursor delegations (cannot drift) and 2 hand-copied implementations (can). Per-surface result:

| Surface | Per-endpoint logic | Endpoint set |
|---|---|---|
| FULL | unchanged | 120 → 120, no change |
| SOLIDITY | equivalent (delegation → filter) | 44 → 44, no change |
| SOLIDITY_NODE | 2 endpoints differ, see below | 44 → 44, no change |
| PBFT | equivalent (delegation → filter) | 47 → 44, **−5 / +2** |

Client-visible changes, all deliberate and worth a release note:

1. **Standalone SolidityNode `/walletsolidity/gettransactioninfobyid`** — with `visible=true` on a transaction that has logs, `log[].address` changes from hex to Tron base58, matching FullNode. This is the drift fix above; `visible=false` and log-free transactions are unaffected.
2. **PBFT port drops 5 sapling endpoints** — `getmerkletreevoucherinfo`, `isspend`, `scanandmarknotebyivk`, `scannotebyivk`, `scannotebyovk` now return 404 on `/walletpbft/*`. They were disabled on every other surface in 2020; PBFT is catching up, not regressing.
3. **PBFT port gains 2 read endpoints** — `getpaginatednowwitnesslist` and `gettransactioninfobyblocknum`, which FULL / SOLIDITY / SolidityNode already expose. Pure addition.

Error responses on the two former SolidityNode copies now go through `Util.processError` (the standard `{"Error": ...}` body) instead of writing the raw exception message into the response.

- **Unit Tests** — 24 new tests, all passing:
  - `HttpApiRegistryTest` (17). Twelve drive one validation branch each through a fixture package under `http/regtest/*` and assert the boot failure it produces: a servlet declaring neither annotation or both, an endpoint on a nested or abstract class, a duplicate `(surface, suffix)`, a blank suffix, a `/` in a suffix, a `*` or whitespace suffix, a missing `@Component`, an empty surface list, and a non-`READ` endpoint on a cursor surface. One builds a valid fixture package. The remaining four are the mount-parity tests: each service's mounted path set equals the registry's derived set for its surface, so an endpoint cannot be declared and left unmounted, or mounted without being declared.
  - `CursorFilterInstallationTest` (4) runs each service's real `addFilter` and asserts solidity and pbft install exactly one cursor filter on `/*`, while the FULL port and the standalone SolidityNode install none — a cursor filter on either of those would take the port off HEAD.
  - `WalletCursorFilterTest` (3) asserts the cursor is set before `chain.doFilter` and reset in `finally` including when the servlet throws, and that the PBFT subclass switches to the PBFT cursor rather than SOLIDITY.

  There is no frozen route snapshot: the registry is validated in full when the class is first touched — before Jetty binds, and for every surface whether or not the node enables it — so the invariants are enforced at boot and the tests drive the failure paths rather than re-asserting them over the live table.
- **Manual Testing** — brought up a private chain and checked the mounted endpoint set on every port, including that endpoints marked `@HttpApiExcluded` are unreachable.

**Follow up**

- Grouping the servlet package by function, as raised in the issue discussion. Kept out of this PR so the diff stays a mechanical de-duplication and the endpoint set remains directly diffable; worth doing once this lands.
- Publishing the per-endpoint inventory and the PBFT surface changelog alongside the release note.

**Extra details**

`@HttpApi` / `@HttpApiExcluded` are deliberately **not** `@Inherited`, and the registry reads them with `getDeclaredAnnotation` only. Inheritable exposure is exactly what produced the 100 wrapper classes this PR removes — a subclass must never silently inherit its parent's surface set.


framework/src/test/java/org/tron/core/services/http/HttpApiRegistryTest.java
```diff
@@ -0,0 +1,184 @@
...
+      HttpApiRegistry.buildFromPackage(pkg);
+      Assert.fail("expected build to fail for " + pkg);
+    } catch (IllegalStateException e) {
+      Assert.assertTrue("message '" + e.getMessage() + "' should contain '" + fragment + "'",
+          e.getMessage() != null && e.getMessage().contains(fragment));
+    }
+  }
+
+  @Test
+  public void testFullNodeServiceMountsExactlyTheRegistry() throws Exception {
```
bladehan1, line 121: [SHOULD] Mount-parity tests are self-referential — no frozen route snapshot guards against annotation edits

The four mount-parity tests assert that each service's mounted path set equals the set derived from `HttpApiRegistry` — but mounting is also driven by that same registry. Expected and actual share one source of truth, so a mistaken `@HttpApi` edit (wrong surface list, typo'd suffix) silently drops or moves an endpoint and both sides still agree; the suite stays green while the public API surface changes. The PR description acknowledges there is no frozen route snapshot and argues boot-time validation covers it, but validation checks shape (duplicates, READ-on-cursor, annotations present), not membership of the actual endpoint set.

Suggestion: check in a frozen per-surface suffix snapshot (a literal list per surface that a test diffs against the derived registry) so any endpoint-set change requires an explicit, reviewable test update. Fine as a follow-up, but please track it — this is the one guard that catches accidental surface drift.

framework/src/main/java/org/tron/common/application/HttpService.java
```diff
@@ -80,7 +85,9 @@ protected void initServer() {
     this.apiServer = new Server(this.port);
     int maxHttpConnectNumber = Args.getInstance().getMaxHttpConnectNumber();
     if (maxHttpConnectNumber > 0) {
-      this.apiServer.addBean(new ConnectionLimit(maxHttpConnectNumber, this.apiServer));
+      ConnectionLimit connectionLimit = new ConnectionLimit(maxHttpConnectNumber, this.apiServer);
+      connectionLimit.setIdleTimeout(CONNECTION_LIMIT_IDLE_TIMEOUT_MS);
```
bladehan1, line 89: [NIT] Undeclared behavior change: 5s idle timeout while the connection limit is reached

This `setIdleTimeout(5000)` is a runtime behavior change unrelated to the servlet merge: once `maxHttpConnectNumber` is reached, existing idle connections are now closed after 5s (previously the Jetty default applied). It is not among the client-visible changes declared in the PR description, so it will surprise anyone reviewing the release note or bisecting a connection-churn regression.

Suggestion: declare it in the PR description (and release note) with its rationale, or split it into its own PR — it is small enough to review on its own merits.

framework/src/main/java/org/tron/core/services/filter/PbftCursorFilter.java
```diff
@@ -0,0 +1,15 @@
...
+import org.tron.core.db.Manager;
+import org.tron.core.db2.core.Chainbase;
+
+@Component
+public class PbftCursorFilter extends WalletCursorFilter {
+
+  @Autowired
+  public PbftCursorFilter(Manager dbManager) {
+    super(dbManager, Chainbase.Cursor.PBFT);
+  }
```
bladehan1, line 14: [NIT] Fifth client-visible difference not declared: PBFT-port getBrokerage/getReward now read the PBFT view

On the PBFT port, `getBrokerage` and `getReward` were previously mounted as full-node servlets reading the HEAD view; routing everything through this filter means they now answer from the PBFT cursor. For these two endpoints the returned values can legitimately differ from HEAD, so this is a client-visible behavior change beyond the three declared in the PR description (log address format, PBFT −5/+2 endpoints, error format).

Suggestion: add it to the behavior-differences list in the PR description and the release note — operators watching brokerage/reward values on the PBFT port should know the view changed.

framework/src/main/java/org/tron/core/services/http/HttpApiRegistry.java
```diff
@@ -0,0 +1,238 @@
...
+ *       only — a cursor surface (SOLIDITY / PBFT) must never run a write path on a
+ *       cursor-switched thread, and the standalone SolidityNode surface cannot propagate
+ *       transactions.</li>
+ * </ul>
+ *
+ * <p>Annotations are read with {@link Class#getDeclaredAnnotation} and {@link HttpApi} is not
+ * {@link Inherited}, asserted below, so a servlet subclass can never inherit its parent's
+ * exposure — the failure mode the removed cursor-wrapper subclasses would otherwise reintroduce.
+ */
+public final class HttpApiRegistry {
```
bladehan1, line 56: [NIT] PBFT endpoint removals need the changelog/endpoint-inventory follow-up to land with the release

(Cross-cutting process note anchored at the registry, which now defines the PBFT surface.)

Five endpoints that were live on `/walletpbft/*` (the sapling voucher/scan family) now return 404. The PR description flags this as deliberate and lists "publishing the per-endpoint inventory and the PBFT surface changelog" as a follow-up — but a 404 on a previously live endpoint is exactly the kind of change integrators discover by breakage. If the inventory/changelog slips past the release, the removal becomes an undocumented breaking change.

Suggestion: file a tracking issue for the endpoint inventory + PBFT changelog and link it from the PR description, so the follow-up has an owner and a deadline tied to the release.

framework/src/main/java/org/tron/core/services/http/HttpApi.java
```diff
@@ -0,0 +1,58 @@
...
+ * superclass-walking helper such as Spring's {@code AnnotatedElementUtils#findMergedAnnotation},
+ * any future subclass would silently inherit its parent's suffix, access and surfaces: a
+ * {@code WRITE} endpoint could reach a cursor surface without anyone declaring it.
+ * {@link HttpApiRegistry} asserts both properties at startup so the guarantee cannot be removed
+ * by accident.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface HttpApi {
```
bladehan1, line 29: [NIT] Description & metadata hygiene (4 items rolled up)

Grouped as one comment since all are doc/metadata asks rather than code defects. (Cross-cutting — anchored at the new annotation as a representative added line.)

1. **PR description, behavior-difference item 4 (error format).** It says the two former SolidityNode copies now go through `Util.processError` "instead of writing the raw exception message into the response", but both old copies already used `processError` at the merge base; the error-format behavior is unchanged. Suggestion: drop or reword that bullet so the declared change list stays accurate.

2. **Branch name `refactor_merge_http_servlets`.** Missing the `<type>/` separator used across the repo (`feature/`, `fix/`, …). Not worth re-opening the PR over, but worth matching the convention next time.

3. **First commit message body retains a `# Conflicts:` block.** The merge-conflict scaffolding comment will live in `git log` forever; if the commits are ever rebased or squashed, please drop it.

4. **Validation-timing wording.** The description says the registry is validated "before Jetty binds"; precisely, validation is triggered lazily on first touch of the registry class during `HttpService.start()` (still before bind, as claimed) — a one-line clarification would save future readers a grep.

Suggestion: fix the description bullet and, on the next history rewrite, the commit body; nothing here blocks merge.

PR #6977, opened 2026-09-17
feat(API): refactor merge rpc

### LINKED ISSUE #6927: [Feature] Deduplicate gRPC cursor services with a ServerInterceptor
# Summary

The same gRPC read surface is implemented several times over: `RpcApiServiceOnSolidity` and `RpcApiServiceOnPBFT` are ~490 lines each of methods that do nothing but call the base service and switch the per-thread read cursor. This proposes replacing all of that per-method delegation with a single cursor-parameterized `ServerInterceptor`, and letting the read-only `WalletSolidityApi` handlers delegate to the shared `WalletApi` rather than re-implementing them. About 900 lines go away, with no change to the gRPC service definitions, ports, or responses.

# Problem

### Motivation

A read handler currently lives in up to four places — twice inside `RpcApiService` (`WalletApi` and `WalletSolidityApi`) and once each in the Solidity and PBFT services — so one RPC change may have to be mirrored four times. Miss one and the same RPC behaves differently depending on which port a client hits. This cross-port drift is hard to spot, and it has already happened: an error-handling fix was applied to `WalletApi` only and never reached the copy serving the Solidity and PBFT ports.

### Current State

The Solidity and PBFT services each re-declare the whole read surface, and every method body is the same shape — wrap a call to the base service's singleton in `futureGet` so the thread reads from the solidified / PBFT snapshot instead of HEAD:

```java
@Override
public void getAccount(Account req, StreamObserver<Account> obs) {
  walletOnSolidity.futureGet(
      () -> rpcApiService.getWalletSolidityApi().getAccount(req, obs));
}
```

That is roughly 100 delegations carrying no logic. The cursor itself is a `ThreadLocal`, so switching it is a property of the calling thread, not of the class the handler happens to sit in — which is what makes a single interceptor able to replace all of them.

**How the FullNode and standalone SolidityNode processes use these services.** The standalone SolidityNode has no dedicated service class; the base `RpcApiService` branches by run mode, and the two cursor services only start on a FullNode:

| Service | `enable` condition | FullNode | SolidityNode |
|---|---|---|---|
| `RpcApiService` | `isRpcEnable()` | registers `WalletApi` (HEAD) | registers `WalletSolidityApi` (HEAD) |
| `RpcApiServiceOnSolidity` | `isFullNode() && isRpcSolidityEnable()` | registers `WalletSolidityApi` (SOLIDITY) | not started |
| `RpcApiServiceOnPBFT` | `isFullNode() && isRpcPBFTEnable()` | registers `WalletPBFTApi` (PBFT) | not started |

So on a FullNode the two cursor services already borrow the base service's singletons — the refactor keeps exactly that and only adds the interceptor. The standalone SolidityNode is out of scope: its database holds solidified data only, so it needs no cursor switching and never runs the cursor services.

**A second duplication — the `WalletApi` / `WalletSolidityApi` read handlers.** Inside the base `RpcApiService`, `WalletApi` (`protocol.Wallet`) and `WalletSolidityApi` (`protocol.WalletSolidity`) are two separate proto services whose read methods re-implement the same logic twice. gRPC generates one `ImplBase` per proto service and Java is single-inheritance, so both classes have to exist — but their duplicated handler *bodies* are a second axis of duplication, per proto service rather than per port, and it is where the drift noted in Motivation actually happened.

### Limitations or Risks

- Every RPC addition or change has to be repeated in each copy, and forgetting one silently produces different behavior per port.
- Reviews and tests have to cover the same handler several times over.

# Proposed Solution

### Proposed Design

Add a cursor-parameterized `ServerInterceptor`. The Solidity and PBFT services stop declaring their own handlers, register the base service's `WalletSolidity` / `Database` singletons directly, and attach the interceptor that sets the cursor for the duration of the call and restores it afterwards.

```java
public abstract class CursorServerInterceptor implements ServerInterceptor {
  // sets the cursor before the handler runs and restores it in a finally block
}
```

Two properties decide whether this works, and both should be verified before any code is deleted:

- **The cursor must be switched on the thread that runs the handler.** gRPC dispatches a call's callbacks through the server's executor without pinning them to one thread, so the cursor has to be set where the handler actually runs, not when the call arrives. Getting this wrong fails silently — the port serves HEAD data with no error.
- **Handlers that read the database must be synchronous.** The cursor is restored as soon as the handler returns, so anything deferring its reads to another thread would read after the restore.

Separately, the read-only `WalletSolidityApi` handlers delegate to the shared `WalletApi` singleton instead of repeating the same bodies. The two generated `ImplBase` classes and their signatures stay; only the duplicated bodies go. This is safe for the same reason as the interceptor: which snapshot a read resolves against depends on the calling thread's cursor, not on which class the handler lives in.

### Key Changes

- **Module**: `framework`
- **Add**: a cursor `ServerInterceptor` plus its Solidity and PBFT subclasses, and tests covering the two properties above
- **Modify**: service and interceptor registration in the two cursor services; the read-only handler bodies in `RpcApiService`
- **Delete**: ~100 per-method delegations and the duplicated inner service classes
- **Config / API**: unchanged — same ports, same switches, same proto definitions

# Impact

- **Developer experience**: an RPC change goes from "edit in four places" back to one, removing the drift risk entirely.
- **Security**: the existing interceptor chain (rate limiting, API access, lite-fullnode filtering, metrics) is untouched.
- **Performance**: one interceptor per call replaces one lambda per method; overhead is comparable.

# Compatibility

- **Breaking change**: No. Proto definitions, ports and responses are unchanged.
- **Default behavior change**: One. Because both services now share the same singleton, the PBFT port also exposes `getPaginatedNowWitnessList` and `getTransactionInfoByBlockNum`. PBFT lacked them only because the two services were written separately; both are ordinary read-only queries that resolve against the PBFT snapshot like every other read, so the proposal aligns rather than shields them.
- **Migration required**: No.

# References

- https://github.com/tronprotocol/java-tron/issues/6922 — the equivalent HTTP servlets dedup

# Additional Notes

SeriousCoding789:
**What does this PR do?**

Implements https://github.com/tronprotocol/java-tron/issues/6927 — the gRPC counterpart of the HTTP servlet dedup in #6922.

`RpcApiServiceOnSolidity` and `RpcApiServiceOnPBFT` each re-declare the whole read surface as per-method delegations whose only job is to switch the per-thread read cursor:

```java
@Override
public void getAccount(Account req, StreamObserver<Account> obs) {
  walletOnSolidity.futureGet(() -> rpcApiService.getWalletSolidityApi().getAccount(req, obs));
}
```

There are exactly 100 of these — 51 in the Solidity service, 49 in the PBFT one — spread over four inner service classes. This PR replaces them with a cursor-parameterized `ServerInterceptor`:

- **Adds `CursorServerInterceptor`** with `SolidityCursorInterceptor` / `PbftCursorInterceptor`. It brackets `Listener.onHalfClose()` — the callback gRPC runs a unary handler inline from — setting the cursor before and restoring it in a `finally`.
- **The two cursor services now register the base service's `DatabaseApi` / `WalletSolidityApi` singletons directly**, wrapped in `ServerInterceptors.intercept(...)`. `RpcApiServiceOnSolidity` goes from 488 lines to 36, `RpcApiServiceOnPBFT` from 495 to 36.
- **A second, independent axis of duplication inside `RpcApiService` itself**: `WalletSolidityApi` (serving `protocol.WalletSolidity`) re-implements read handlers `WalletApi` (serving `protocol.Wallet`) already has. 41 of its 47 methods now delegate to the shared `WalletApi` singleton; the remaining 6 already routed through shared `*Common` / `callContract` helpers, so no duplicated handler body is left. Java is single-inheritance and gRPC generates one `ImplBase` per proto service, so both classes have to stay — only the bodies go.
- **Fixes the double-close on the error path**, which the dedup made visible. A handler that calls `responseObserver.onError(...)` and then falls through to `responseObserver.onCompleted()` closes the call twice; the second `close()` hits `checkState(!closeCalled, "call already closed")` in gRPC's `ServerCallImpl` and throws. `RpcApiService` had 32 handlers shaped that way on `develop` — 8 disappear with the duplicated `WalletSolidityApi` bodies, and the remaining 24 get an explicit `return`, the shape the file already used elsewhere. The file now has none.

10 files changed, +945 / −1267.

A read handler currently lives in up to four places, so one RPC change has to be mirrored four times, and missing one makes the same RPC behave differently depending on which port a client hits.

That has already happened. Of the 41 handlers being dedup'd inside `RpcApiService`, 23 are byte-identical between the two copies on `develop` and 18 differ. Of those 18:

- **8 are hardening that only ever landed in `WalletApi`** — it gained a `return` after `responseObserver.onError(...)` and `WalletSolidityApi` did not, so on the error path the copy serving the Solidity and PBFT ports falls through to `onCompleted()` and terminates the call twice (`getMerkleTreeVoucherInfo`, `isSpend`, `scanAndMarkNoteByIvk`, `scanNoteByIvk`, `scanNoteByOvk`, `isShieldedTRC20ContractNoteSpent`, `scanShieldedTRC20NotesByIvk`, `scanShieldedTRC20NotesByOvk`). The last two additionally have a `BadItemException | ZksnarkException` branch with logging that only `WalletApi` ever received.
- 7 are cosmetic — a parameter name, a temporary variable, line wrapping.
- 1 differs only in a log prefix.
- 2 (`getBlockByNum`, `getBlockByNum2`) drifted the other way: `WalletSolidityApi` has a `num >= 0` guard `WalletApi` lacks.

None of that was written deliberately; it is what happens when the same handler exists four times.

Lining the two copies up also showed that the `return` hardening was never finished on `WalletApi` either — 24 more handlers in the same file still fall through — so this PR completes it rather than leaving the file in two states.

**Behaviour differences vs `develop`**

**Method sets, per port.** The base `WalletSolidityApi` (47 methods) and `DatabaseApi` (4) expose exactly the same methods before and after — only bodies changed.

| Port | Before | After | Delta |
|---|---|---|---|
| HEAD (`RpcApiService`) | 47 + 4 | 47 + 4 | none |
| SOLIDITY | 47 + 4 | 47 + 4 | **none** |
| PBFT | 45 + 4 | 47 + 4 | **+2** |

The PBFT port gains `getPaginatedNowWitnessList` and `getTransactionInfoByBlockNum` — the only two methods `RpcApiServiceOnPBFT` never mirrored from `RpcApiServiceOnSolidity`; they returned `UNIMPLEMENTED` there before. Both are ordinary reads and resolve against the PBFT snapshot like every other read on that port.

**Handler bodies.** Only three `WalletApi` bodies changed beyond the added `return`s:

- `getBlockByNum` / `getBlockByNum2` adopt the solidity copy's `num >= 0` guard. **No response change** — `Wallet#getBlockByNum` already catches the `StoreException` and returns `null` for a negative number, so both paths reach `onNext(null)`; the guard only skips a futile store lookup and its log line.
- `getAssetIssueByName` drops the `"FullNode "` prefix from one `logger.debug` line, which was the only difference between the two copies.

**Error paths.** Every handler that used to emit `onError` followed by `onCompleted` now emits a single terminal event, on all three ports. **Not visible to clients** — `onError` had already closed the call with the error status and the second `close()` threw before sending anything; what goes away is one server-side `IllegalStateException` per failed call. Two groups:

- Fixed as a side effect of the dedup, on the SOLIDITY and PBFT ports: the 8 shielded handlers listed above. Worth noting these were not a rare corner — the first statement of the five sapling reads in `Wallet` is `checkAllowShieldedTransactionApi()`, and `node.allowShieldedTransactionApi` defaults to `false`, so on a default node *every* such call took the double-close path.
- Fixed explicitly, on all three ports: 17 handlers in `WalletApi` (`getPaginatedNowWitnessList`, `getTransactionInfoByBlockNum`, `getDelegatedResourceV2`, `getDelegatedResourceAccountIndex`, `getDelegatedResourceAccountIndexV2`, `getCanDelegatedMaxSize`, `getCanWithdrawUnfreezeAmount`, `getAvailableUnfreezeCount`, `getBandwidthPrices`, `getEnergyPrices`, `getMemoFee`, `getNodeInfo`, `getMarketOrderByAccount`, `getMarketOrderById`, `getMarketOrderListByPair`, `getMarketPriceByPair`, `getMarketPairList`) and 7 shared helpers reachable from both services (`getBlockCommon`, `getRewardInfoCommon`, `getBrokerageInfoCommon`, `getBurnTrxCommon`, `getPendingSizeCommon`, `getTransactionFromPendingCommon`, `getTransactionListFromPendingCommon`).

**Cursor semantics are unchanged.** `Manager#setCursor` still computes the `headNum - pbftNum` offset for PBFT, exactly as `WalletOnPBFT.futureGet` did.

**Scope.** gRPC only. `WalletOnCursor` / `WalletOnSolidity` / `WalletOnPBFT` stay, because the HTTP and JSON-RPC servlets still call `futureGet`; removing those is a separate change. Ports, switches, proto definitions and the server-level interceptor chain (rate limiter, api access, lite-fullnode filter, prometheus) are untouched.

- **Unit Tests** — 7 new tests plus two assertions added to the existing end-to-end suite, all passing. Each pins something that can actually go wrong:
  - `CursorInterceptorScopeTest` (2) drives `interceptCall()` on one thread and the returned listener's `onHalfClose()` on another — which is what gRPC's `SerializingExecutor` is free to do. An implementation that scoped the cursor around `interceptCall` fails here by construction rather than by luck. Also covers the `finally` reset when the handler throws.
  - `CursorInterceptorServerTest` (1) runs the production interceptor behind a real gRPC server and asserts the cursor is set and restored exactly once, on the handler's own thread. This is the end-to-end half: the whole design rests on gRPC running the handler inline from `onHalfClose`, and if that stops holding the cursor never reaches the read path and the port serves HEAD data with no error.
  - `CursorInterceptorWiringTest` (2) runs the real `addService` of both cursor services against a mock builder and asserts each shared read service is registered as an intercepted definition. Dropping `ServerInterceptors.intercept` leaves every other test green while the port silently serves HEAD, so this is the gRPC counterpart of `CursorFilterInstallationTest`.
  - `RpcApiServiceErrorPathTest` (2) drives every unary handler of `WalletApi` and `WalletSolidityApi` with collaborators that throw, and asserts none of them terminates the call more than once. Reverting the added `return`s makes it fail on `getDelegatedResourceV2`, `getPendingSize` and `getBlock`, so it reaches the shared `*Common` helpers as well.
  - `RpcApiServicesTest` — the existing suite drives all three ports end to end over 129 tests; it now also calls `getPaginatedNowWitnessList` and `getTransactionInfoByBlockNum` on the PBFT stub, pinning the one intentional behaviour change.
- **Manual Testing**


framework/src/test/java/org/tron/core/services/filter/CursorInterceptorServerTest.java
```diff
@@ -0,0 +1,123 @@
...
+        .build()
+        .start();
+
+    ManagedChannel channel = ManagedChannelBuilder.forAddress("127.0.0.1", port)
+        .usePlaintext().directExecutor().build();
+    try {
+      DatabaseGrpc.newBlockingStub(channel).getNowBlock(EmptyMessage.getDefaultInstance());
+
+      Assert.assertEquals("cursor must be set exactly once per call", 1, setOn.size());
+      Assert.assertEquals("cursor must be restored exactly once per call", 1, resetOn.size());
```
317787106, line 103: [SHOULD] The blocking RPC returning does not guarantee that the server has executed `resetCursor()` in the interceptor’s finally block. The client can receive the response first, so this assertion may observe resetOn.size() == 0 and fail intermittently.

  Please count down a latch after recording the reset, then await it with a timeout before checking the reset count and thread.

framework/src/test/java/org/tron/core/services/filter/CursorInterceptorServerTest.java
```diff
@@ -0,0 +1,123 @@
...
+  public void testCursorIsSetAndRestoredOnTheHandlerThread() throws Exception {
+    final List<String> setOn = new CopyOnWriteArrayList<>();
+    final List<String> resetOn = new CopyOnWriteArrayList<>();
+    final String[] handlerOn = new String[1];
+
+    Manager manager = mock(Manager.class);
+    doAnswer(inv -> setOn.add(Thread.currentThread().getName()))
+        .when(manager).setCursor(any(Chainbase.Cursor.class));
+    doAnswer(inv -> resetOn.add(Thread.currentThread().getName()))
+        .when(manager).resetCursor();
```
317787106, line 75: [SHOULD] `cursorAfterCall` is captured on thread A immediately after `interceptCall()`, before thread B executes `onHalfClose()`. This assertion therefore checks the initial state rather than whether the cursor was restored. The test still passes when the reset mock does nothing.

  Please capture the cursor on thread B after `listener.onHalfClose()` returns. Using a `ThreadLocal<Cursor>` would also better reflect the production behavior.

PR #6979, opened 2026-09-17
feat(framework): support admin JSON-RPC over IPC and HTTP

### LINKED ISSUE #6497: Design of a Secure IPC and JSON-RPC Based Administration Framework for FullNode
# Background
After a FullNode is started, its runtime status must be continuously monitored in real time . Once an abnormal condition is detected, timely manual intervention should be possible, such as dynamically adjusting startup or runtime parameters. Currently, the system lacks an effective mechanism to interact with the node. In most abnormal scenarios, recovery can only be achieved by restarting the node, which inevitably interrupts the services provided by the FullNode and negatively impacts overall stability and availability.

Although the existing configuration file provides the `node.dynamicConfig` capability, its functionality is relatively limited, lacks extensibility, and provides no real-time feedback to users. As a result, it does not adequately meet operational and debugging requirements. To address these limitations, we designed and implemented a real-time interaction mechanism for FullNode based on a UNIX Server Socket combined with a JSON-RPC service bound only to the local network interface. This approach significantly improves system flexibility, operability, and issue-handling efficiency while maintaining strong security guarantees.
# Rationale

This solution enables real-time, bidirectional interaction with a FullNode, rather than being limited to passive reading of node status or information. It greatly improves troubleshooting and debugging efficiency, simplifies node management workflows, and supports rapid intervention in abnormal scenarios, ultimately ensuring the overall stability and availability of the service.

### What are the use-cases?

- Support [dynamic management](https://github.com/tronprotocol/java-tron/issues/6296) of node connections, including operations such as AddPeer, RemovePeer, AddTrustedPeer, and RemoveTrustedPeer, enabling flexible runtime adjustment of peers and related policies. 
- Support querying all runtime non-sensitive parameters, not limited to the information currently exposed by org.tron.core.services.NodeInfoService, thereby improving observability and operational transparency.
- Support dynamic modification of configuration parameters during runtime for debugging and issue diagnosis without restarting the FullNode, reducing service interruption risks and improving debugging efficiency.

# Specification
Users can interact with the FullNode in two ways:

1. an HTTP-based JSON-RPC service, and
2. a local IPC service based on a UNIX Server Socket.

Both approaches share identical interface definitions, parameter validation, and execution logic, differing only in communication channels and security boundaries.

Assume the following JSON-RPC interface is defined in FullNode:
```
@Component
public interface AdminJsonRpc {
  @JsonRpcMethod("admin_example")
  @JsonRpcErrors({
      @JsonRpcError(exception = JsonRpcInvalidParamsException.class, code = -32602, data = "{}"),
  })
  String adminExample(@JsonRpcParam("param1") String param1,
                      @JsonRpcParam("param2") String param2)
      throws JsonRpcInvalidParamsException;
}
```
### Approach 1: HTTP-based JSON-RPC Interaction
In this approach, a dedicated HTTP service and port are started to expose the AdminJsonRpc interface, referred to as AdminRpcHttpService. Users can interact with the node through standard HTTP POST requests, for example:

```
{"jsonrpc":"2.0","id":1,"method":"admin_example","params":["abc","def"]}
```
For security reasons, AdminRpcHttpService listens on localhost by default to prevent external exposure of administrative interfaces. If necessary, the listening address can be adjusted via configuration to allow access from LAN nodes or a wider network scope. This approach is suitable for automated operations, remote management, and integration with external systems.

The overall interaction flow is illustrated below:

Image

### Approach 2: Inter-Process Communication via UNIX Server Socket
The second approach uses a UNIX Server Socket to enable inter-process communication (IPC). When the FullNode starts, it creates a socket file named after the process ID under the `/tmp` directory and starts an `IpcService` bound to this socket. The `IpcService` listens for incoming socket requests in a dedicated thread.

On the same machine, users can bind to the socket file using FullNode.jar and start an `IpcClient` process. Interaction with the IpcService is performed using commands such as:

```
> admin_example abc def
 ```

The IpcClient parses and validates the command-line input, resolves the corresponding JSON-RPC method, and converts it into a standard JSON-RPC request. The request is then sent to the `IpcService` through the UNIX socket, and the execution result is received and displayed.

Upon receiving the request, the `IpcService` delegates execution to the internal JSON-RPC Server and returns the result to the `IpcClient` through the same socket, which is then presented to the user. Functionally and in terms of output, this approach is fully equivalent to the HTTP-based approach.

**Key Characteristics**

- Both interaction methods share the same underlying JSON-RPC Server, ensuring consistent behavior.
- Each method can be independently enabled or disabled via configuration, without mutual dependency or interference.
- The local UNIX Server Socket communication model avoids exposing network ports, significantly enhancing security and reducing the attack surface.

# Test Specification

# Scope Of Impact
Only the services module of the framework is affected.

# Implementation

Yes.

Yes. An example interaction between the IPC Client and IPC Service is shown below:
```
[use@host]$ java -jar build/libs/FullNode.jar --attach /tmp/java-tron.91257.sock
Connected to server: /tmp/java-tron.91257.sock
You can enter the following commands, case insensitive:
admin_example, help, exit, quit
> help
You can enter the following commands, case insensitive:
admin_example, help, exit, quit
> help admin_example
usage: admin_example param1 param2
> admin_example abc def
{"jsonrpc":"2.0","id":2,"result":"abc:def"}
```

# Backwards Compatibility
This solution does not depend on existing JSON-RPC interface definitions; therefore, it introduces no interface evolution or compatibility risks.

Starting from JDK 16, Java provides native support for Unix Domain Sockets. In JDK 1.8 environments, access to Unix Domain Sockets on Linux / Unix platforms can be achieved via the third-party library [junixsocket](https://github.com/kohlschutter/junixsocket). Since FullNode must currently support both JDK 1.8 and JDK 17 runtimes, junixsocket is used as the unified underlying communication mechanism to ensure consistent behavior and implementation stability.

In future versions, once compatibility with JDK 1.8 is no longer required, the implementation can be migrated to Java’s native Unix Domain Socket support. This will reduce third-party dependencies, further simplify the system architecture, and lower long-term maintenance costs.

### PR DESCRIPTION
**Pull request by 317787106**
## What does this PR do?

This PR adds the Admin JSON-RPC transport foundation requested by https://github.com/tronprotocol/java-tron/issues/6497. A single annotated `AdminJsonRpc` interface is shared by an HTTP endpoint and a local Unix-domain-socket service. The initial `admin_example` method verifies typed command dispatch and annotation-based JSON-RPC error handling; additional administrative methods can be added to the same interface.

In this PR, JSON-RPC defines the shared API and message format, while HTTP and IPC are the two transports:

```text
                Admin JSON-RPC API
                 /             \
        HTTP transport     IPC transport
          POST /admin      Unix domain socket
```

## Admin HTTP service

The HTTP service is available to FullNode processes at `POST /admin` and is disabled by default. It binds Jetty to the configured address and port. Enabling it on a non-loopback address emits a warning.

The servlet validates the `Host` header against the configured virtual-host allowlist, while accepting IPv4 and IPv6 literals. It accepts `application/json`, `application/json-rpc`, and `application/*+json` media types, and rejects unsupported content types with HTTP 415. JSON-RPC parsing reuses a constrained object mapper with nesting-depth and token-count limits. JSON-RPC results, including protocol errors, use HTTP 200 responses.

Admin HTTP is intended for trusted node operators. Deployments must restrict access to loopback or a controlled management network. Under this trust model, Admin HTTP intentionally does not apply the public JSON-RPC `node.jsonrpc.maxBatchSize` and `node.jsonrpc.maxResponseSize` limits. The existing HTTP request-size limit and JSON nesting-depth and token-count limits still apply.

The HTTP transport configuration is:

```hocon
node.admin.http {
  enable = false
  listenAddress = "127.0.0.1"
  port = 8575
  virtualHosts = ["localhost"]
}
```

## Admin IPC service

The IPC service is also FullNode-only and disabled by default. By default it creates an endpoint at:

```text
<output-directory>/ipc/<pid>.sock
```

`node.admin.ipc.socketDirectory` can select another socket root and must be an existing absolute directory on a POSIX-compatible filesystem. The service creates the private `ipc` directory with owner-only access and sets the socket to `0600`. It refuses to replace a symbolic link or non-directory at the private directory path.

The complete encoded socket path is limited to 100 bytes for portability across supported Unix-domain-socket implementations. If the path is longer, startup fails with guidance to configure a shorter `node.admin.ipc.socketDirectory`; it does not silently relocate the endpoint.

```hocon
node.admin.ipc {
  enable = false
  socketDirectory = ""
}
```

IPC uses newline-delimited, single-line JSON-RPC messages and the same `AdminJsonRpc`, error resolver, and constrained JSON mapper as HTTP. Request size is bounded by the existing `node.rpc.maxMessageSize` value, whose default is 4 MiB. Each client has a ten-minute idle timeout. The bounded client executor supports concurrent console sessions and immediately closes connections that arrive after all handlers are occupied. Unexpected accept failures use a five-second retry delay.

Startup failures and normal shutdown both clean up owned sockets and the private directory. Shutdown closes active clients before stopping executors so blocked native socket reads do not unnecessarily delay node termination.

## IPC console

The FullNode executable can attach to an active node without initializing another node instance:

```bash
java -jar FullNode.jar --attach <socket-path>
```

A single command can be executed for scripting:

```bash
java -jar FullNode.jar --attach <socket-path> --exec "<command> [arguments]"
```

Attach mode is handled immediately after CLI argument parsing and before `CommonParameter`, Logback, database, witness, or node services are initialized. `--exec` requires `--attach`, an empty socket path is rejected, and attach mode cannot be combined with `--config`.

The JLine console derives command names, parameter names, and parameter types from the annotated Admin API. It supports quoted arguments, typed JSON conversion, sorted help, canonical command completion, formatted JSON results, `help`, `exit`, and `quit`. One-shot execution returns a non-zero process status for invalid commands, JSON-RPC errors, communication failures, disconnection before a response, or the 30-second response timeout.

## Supporting changes

`HttpService` now supports binding a service to a specific listen address. The regular JSON-RPC servlet and the Admin transports share the new constrained `JsonRpcMapper`, and supported JSON media-type matching is centralized in `JsonRpcMediaType`.

The framework adds JLine for the interactive console and junixsocket for Unix-domain-socket support. Dependency verification metadata is updated accordingly.

Administrative operations need a local, scriptable interface without starting a second node or exposing the existing public APIs as privileged management endpoints. The Unix-domain socket provides a private local transport, while the optional HTTP endpoint supports controlled integration when explicitly enabled.

Sharing one typed Admin API across both transports keeps command names, parameters, results, and JSON-RPC errors consistent. Explicit address binding, virtual-host validation, parser limits, filesystem permissions, bounded clients, and deterministic cleanup provide safer operational defaults.

## Testing

The PR adds or updates tests covering:

- Admin configuration defaults, binding, and attach-mode CLI validation.
- HTTP listen-address handling, loopback classification, virtual-host validation, JSON media types, and parser limits.
- IPC path resolution and encoded-length limits, absolute-directory validation, POSIX permissions, stale-path safety, startup rollback, and shutdown cleanup.
- IPC request limits, annotated errors, concurrent clients, idle timeouts, overload rejection, and active-client shutdown.
- Console command discovery, typed arguments, quoted whitespace, completion, help, response formatting, one-shot exit codes, timeouts, and disconnect behavior.
- FullNode startup ordering to ensure attach mode does not initialize node logging.

The related Admin HTTP, IPC, CLI, configuration, and FullNode tests, together with production and test checkstyle checks, passed during development.

## Follow-up

Runtime parameter export  and Peer management commands will be implemented separately using the Admin transports introduced here.


framework/src/main/java/org/tron/core/services/admin/ipc/server/IpcSocketFiles.java
```diff
@@ -0,0 +1,115 @@
...
+      }
+      deleteDirectoryWithDirectEntries(socketDirectory);
+    }
+    Files.createDirectory(socketDirectory, PosixFilePermissions.asFileAttribute(
+        EnumSet.of(PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE,
+            PosixFilePermission.OWNER_EXECUTE)));
+  }
+
+  private void deleteDirectoryWithDirectEntries(Path directory) throws IOException {
+    try (DirectoryStream<Path> entries = Files.newDirectoryStream(directory)) {
```
waynercheung, line 91: **[MUST]** Starting a second node with the same `socketDirectory` deletes the first node's live socket path, so new attach connections to the first node fail. This loop also deletes unrelated files.

Please either give each node its own IPC directory, or support a shared directory safely, and limit startup/shutdown cleanup to the entries the node owns. If the directory is shared, leave it in place while other entries remain.

Please add a regression test showing that starting and stopping a second node does not break the first node's endpoint.

framework/src/main/java/org/tron/core/services/admin/http/AdminRpcServlet.java
```diff
@@ -0,0 +1,106 @@
...
+      resp.sendError(HttpServletResponse.SC_FORBIDDEN, "Invalid Host header");
+      return;
+    }
+    // Require JSON media types so browser form and text/plain submissions are rejected.
+    if (!JsonRpcMediaType.isSupported(req.getContentType())) {
+      resp.setStatus(HttpServletResponse.SC_UNSUPPORTED_MEDIA_TYPE);
+      resp.setContentLength(0);
+      return;
+    }
+    rpcServer.handle(req, resp);
```
waynercheung, line 104: **[MUST]** `jsonrpc4j` swallows parser-limit exceptions here and returns HTTP 200 with an empty body, while IPC turns the same failure into `-32603`. Malformed JSON also gets the string `"null"` as the response id.

Please handle these failures in the shared Admin request handling, so that both transports return valid, consistent JSON-RPC errors, with a JSON `null` id when the request id cannot be determined.

Please also change the depth/token tests to assert the error body, and to check that no Admin method was invoked.

framework/src/main/java/org/tron/core/services/admin/ipc/server/IpcRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    return new String(bytes, 0, length, StandardCharsets.UTF_8);
+  }
+
+  String handleCommand(String jsonRequest) {
+    ByteArrayInputStream input =
+        new ByteArrayInputStream(jsonRequest.getBytes(StandardCharsets.UTF_8));
+    ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+    try {
+      jsonRpcServer.handleRequest(input, output);
```
waynercheung, line 70: **[MUST]** Batch handling is left to `jsonrpc4j` and does not follow JSON-RPC 2.0. HTTP and IPC both behave this way:

- `[]` returns `[]`
- `[{"jsonrpc":"2.0","method":"admin_example","params":["a","b"]}]` returns `[null]`
- `[{"jsonrpc":"2.0","method":"admin_example","params":["x","y"]},{"jsonrpc":"2.0","method":"admin_example","params":["a","b"],"id":2}]` returns `[null,{"jsonrpc":"2.0","id":2,"result":"a:b"}]`

Please handle batches in the shared Admin dispatch layer: return a single `-32600` error for an empty batch, leave notifications out of the response, and return no body when every entry is a notification.

Please add empty-batch, notification-only, and mixed-batch tests for both HTTP and IPC.

framework/src/main/java/org/tron/core/services/admin/http/AdminRpcServlet.java
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * <p>This endpoint is intended for trusted node operators. Deployments must restrict access to
+ * loopback or a controlled management network. It intentionally does not apply the public JSON-RPC
+ * batch-size or response-size limits. HTTP request-size limits are enforced by
+ * {@link org.tron.common.application.HttpService}; JSON parser limits come from
+ * {@link JsonRpcMapper}.
+ */
+@Component
+@Slf4j(topic = "API")
+public class AdminRpcServlet extends RateLimiterServlet {
```
waynercheung, line 34: **[SHOULD]** Extending `RateLimiterServlet` puts Admin HTTP on the same global quota as the public API, so public traffic can delay admin requests, or get them rejected when `apiNonBlocking` is enabled. If a local reverse proxy makes public requests appear to come from the same loopback address as admin requests, they also share the per-IP quota.

Please give Admin HTTP an independent rate-limit policy, and decide what an overloaded request should return. The inherited `{"Error":"lack of computing resources"}` body is not a JSON-RPC response.

framework/src/main/java/org/tron/core/config/args/Args.java
```diff
@@ -146,19 +152,32 @@ public class Args extends CommonParameter {
...
       printVersion();
       exit(0);
     }
     if (cmd.help) {
-      Args.printHelp(jc);
+      Args.printHelp(arguments.getCommander());
       exit(0);
     }
+    List<ParameterDescription> assignedParameters = arguments.getAssignedParameters();
+    if (arguments.isAttachMode()) {
```
waynercheung, line 173: **[SHOULD]** `FullNode.main()` already handles attach mode and returns before calling `Args.setParam()`, so this branch and the attach fields stored here are unused in the current FullNode startup path. They also let a direct caller skip node configuration without any error.

Please keep attach dispatch in `FullNode`, remove this duplicate state from `Args`, and reject attach options explicitly if they ever reach this method.

framework/src/main/java/org/tron/core/services/admin/ipc/client/IpcClient.java
```diff
@@ -0,0 +1,247 @@
...
+        } catch (SocketTimeoutException e) {
+          System.err.println("Timed out waiting for IPC response.");
+          return EXIT_FAILURE;
+        }
+      } while (response != null && response.trim().isEmpty());
+      if (response == null) {
+        System.err.println("Disconnected from server before receiving a response.");
+        return EXIT_FAILURE;
+      }
+      IpcResponse parsedResponse = IpcResponse.parse(response);
```
waynercheung, line 122: **[SHOULD]** The response is never checked against the request that was sent. `{"result":true}`, or a response with a different id, still counts as success and exits 0.

Please check that `jsonrpc` is `"2.0"`, that the response id matches the request, and that exactly one of `result` or `error` is present. An invalid or mismatched response should exit with a non-zero status, since scripts use this status to tell whether the command succeeded.

common/src/main/resources/reference.conf
```diff
@@ -442,6 +442,32 @@ node {
     maxMessageSize = 4194304
   }
 
+  # Administrative API settings. Disabled by default.
```
waynercheung, line 445: **[SHOULD]** Please also add disabled Admin HTTP/IPC examples to `framework/src/main/resources/config.conf`. The defaults from `reference.conf` apply at runtime, but the config file we ship to operators doesn't show any of the new settings.

Next to the examples, please note that:

- HTTP must only be exposed on a trusted network.
- For IPC, a non-empty `node.admin.ipc.socketDirectory` must be an existing absolute directory, and the full socket path must not exceed 100 bytes once encoded.
- IPC request size is controlled by `node.rpc.maxMessageSize`, so operators know that changing it affects both gRPC and IPC.

waynercheung (changes requested):

framework/src/main/java/org/tron/core/services/admin/http/AdminRpcServlet.java
317787106: Addressed in 0e624ea. Given its trusted-operator, low-frequency usage, Admin HTTP now extends `HttpServlet` and does not participate in public API rate limiting. This removes both the shared-quota dependency and the inherited REST-style rate-limit response.

  Existing Host, media-type, request-size, and parser checks remain in place, along with HTTP latency metrics. Added regression coverage through `service()` confirming that Admin requests succeed when the public limiter rejects, under both blocking and non-blocking configurations. All 37 related tests and Checkstyle passed locally.

framework/src/main/java/org/tron/core/config/args/Args.java
317787106: Good catch ! Updated in 6152d3e. Removed the unused attach fields and the attach branch from `Args`.

  We kept attach validation and dispatch in `FullNode`, which handles attach mode before calling `Args.setParam()`. To keep this responsibility in one place, we removed the duplicate check without adding a separate rejection guard in `Args`; direct callers are expected to supply node-mode arguments.

  Existing FullNode tests cover valid and invalid attach options. All 34 related tests and Checkstyle passed locally.

common/src/main/resources/reference.conf
317787106: Updated in 280653e. Added a disabled Admin IPC example to `config.conf`, documenting the existing absolute-directory requirement, the output-directory fallback, and the 100-byte limit for the encoded full socket path.

  IPC now uses `node.jsonrpc.maxMessageSize`, sharing the limit with public JSON-RPC HTTP instead of gRPC. Both configuration files document this, and the handler preserves the configuration’s long type.

  This update adds only IPC settings to `config.conf`; Admin HTTP settings remain in `reference.conf`. All 113 related tests and Checkstyle passed locally.

PR #6980, opened 2026-09-18
fix(framework): handle shielded admission and validate note logs

Federico2014:
**What does this PR do?**

Propagate local transaction admission results to RPC and P2P callers, and add defensive length validation to Shielded TRC20 note scanning.

- When `Manager.pushTransaction` returns `false`, RPC returns `result=false` with `SERVER_BUSY` and skips broadcast. The response message and logs describe local admission rejection without assuming a specific Manager rejection reason.
- Return the admission result from `TronNetDelegate.pushTransaction`; the P2P handler drops rejected transactions without rebroadcasting or penalizing the peer.
- After contract-address and topic identification, require exactly 800 bytes for `MintNewLeaf`, `TransferNewLeaf`, and `BurnNewLeaf` logs in IVK and OVK scanning. Use consistent length conditions and preserve the separate `TokenBurn` validation.
- Re-enable Shielded TRC20 builder tests and correct their ABI encoding, value balances, outgoing viewing keys, change outputs, and frontier updates.

A full native shielded pending pool makes `Manager.pushTransaction` return `false`. Previously, RPC and P2P callers ignored that result and continued broadcasting; RPC could report success even though the transaction had not been admitted locally.

Note scanning already identifies supported events by contract address and topics. The explicit length check is defensive hardening that makes the expected ABI layout clear: four 32-byte fields plus `bytes32[21]` ciphertext data total 800 bytes, including 12 bytes of ciphertext padding.

- Local targeted tests passed: `WalletMockTest.testBroadcastTransactionShieldedPendingPoolFull`, `ShieldedTRC20BuilderTest.testScanShieldedTRC20NotesSkipsInvalidLogLengths`, and all four `NoteEncDecryTest.testGetNoteTxFromLogListByOvk*` tests.
- Admission tests verify `SERVER_BUSY`, retained txid cache entries, duplicate rejection on retry, and skipped broadcast. Existing delegate and P2P regression tests cover result propagation and no peer penalty on local rejection.
- Note scanning tests cover lengths 0, 788, 799, 800, 801, and 832 for all three note events through IVK and OVK. The TokenBurn tests cover successful decoding, short records, missing nullifiers, and cursor pairing.
- Local `:framework:checkstyleMain` passed.
- Manual data verification: sampled mainnet mint, transfer, and burn receipts from contract `TQEuSEVRk1GtfExm5q9T8a1w84GvgQJ13V`; their note logs were each 800 bytes. No live transaction submission was performed.

**Follow up**

Txid cache cleanup after admission rejection or validation exceptions, concurrent reservation ownership, and retry control will be handled in a separate PR. This PR preserves the existing cache lifecycle: with caching enabled, rejected transactions retain their txid entries, so retries can return `DUP_TRANSACTION_ERROR` even after pending capacity recovers.

Rejection metrics and delayed retry for re-push transactions can be considered separately.

**Extra details**

Admission handling applies to native `ShieldedTransferContract` transactions when that feature is enabled; log validation applies to Shielded TRC20 contract events.

RPC rejection now uses the existing `SERVER_BUSY` code instead of incorrectly reporting success. The Java return type of `TronNetDelegate.pushTransaction` changes from `void` to `boolean`: source callers may ignore the result, overriding subclasses must adapt, and previously compiled Java consumers must be recompiled. HTTP/gRPC schemas are unaffected by that Java signature change.

Re-push behavior is unchanged. No database schema, consensus, protocol, or configuration changes are included.


framework/src/main/java/org/tron/core/net/messagehandler/TransactionsMsgHandler.java
```diff
@@ -183,7 +183,11 @@ private void handleTransaction(PeerConnection peer, TransactionMessage trx) {
 
     try {
       trx.getTransactionCapsule().checkExpiration(chainBaseManager.getNextBlockSlotTime());
-      tronNetDelegate.pushTransaction(trx.getTransactionCapsule());
+      if (!tronNetDelegate.pushTransaction(trx.getTransactionCapsule())) {
+        logger.debug("Drop trx {} from {}, local admission rejected",
```
317787106, line 187: [NIT] Could we replace `local admission rejected` with `transaction was rejected locally`? “Admission” can sound like a permission check, while the proposed wording more clearly describes the result without assuming a specific rejection reason. 

Also the same problem:
```
return builder.setResult(false).setCode(response_code.SERVER_BUSY)
            .setMessage(ByteString.copyFromUtf8(
                "Transaction was not admitted to the pending pool."))
            .build();
```

Issue #6923, opened 2026-08-24
[Feature] Remove the legacy Monitor API and non-Prometheus metrics implementation


warku123:
# Summary

java-tron maintains two parallel metrics stacks with independent switches: a legacy Dropwizard-based implementation (`org.tron.core.metrics`, gated by `node.metricsEnable`) exposed only through the gRPC `MonitorApi.getStatsInfo` endpoint and the HTTP `/monitor/getstatsinfo` servlet, and the Prometheus stack (`org.tron.common.prometheus`) that has become the standard monitoring path.

This proposal removes the legacy stack together with `MonitorApi`, leaving Prometheus as the single supported monitoring backend. It also adds a `tron:node_info{version, genesis_block_id}` info metric so that node information stays easy to observe after the legacy payload is gone.

This is the implementation issue for item 7 of the tracking issue #6921.

# Problem

### Motivation

Prometheus has become the standard monitoring solution for java-tron (see #6590 and the tron-docker `metric_monitor` reference stack). Keeping a second, non-Prometheus implementation alongside it means duplicated instrumentation and extra maintenance, and it misleads operators: enabling `node.metricsEnable` without Prometheus produces no scrapeable output at all. The legacy endpoints have no known production consumers.

### Current State

- The legacy stack writes into a Dropwizard `MetricRegistry` and is served only via `MonitorApi.getStatsInfo` (gRPC) and `/monitor/getstatsinfo` (HTTP). Most of its fields already have Prometheus equivalents.
- The legacy registry is not purely observational: fetch-block peer selection (`FetchBlockService.getPeerTop75`) consumes the per-peer `net.latency.fetch.block.<peerIP>` 75th-percentile histogram as functional failover input. These histograms are written via `histogramUpdateUnCheck` — bypassing the metrics-enable gate — use one unbounded key per peer IP that is never cleaned on disconnect, and mix latency samples from different block-fetch paths, so the choice of metrics backend silently dictates scheduling behavior.
- A few node-level fields exist only in the legacy payload — notably the node version — and would be lost on removal unless replaced.

### Limitations or Risks

- The fetch-block peer-selection read must be migrated before the legacy registry can be removed.
- Operators with `node.metricsEnable=true` would silently lose monitoring after upgrade unless they switch to the Prometheus flag.

# Proposed Solution

### Proposed Design

[...]

PR #6982, opened 2026-09-20
refactor(api): replace response copy wrappers with Jetty content count

halibobo1205:
**What does this PR do?**

Replaces response-body copying for HTTP traffic metrics with Jetty's native `Response.getContentCount()`. Removes `CharResponseWrapper`, `ServletOutputStreamCopy`, and the unused response wrapping in `HttpApiAccessFilter`.

The old wrappers retained response bodies in memory solely to measure their size. Full-node wallet requests could pass through two copies, and bulk writes fell back to a per-byte loop. Jetty already maintains a byte count, including content still in its output buffer, so these copies are unnecessary. `HttpApiAccessFilter` built a wrapper it never read, on four HTTP services — three of which register no interceptor at all.

Using Jetty's native response writer also fixes incomplete responses caused by bytes remaining in the old wrapper's unflushed `OutputStreamWriter` buffer. Unflushed `print()` calls could lose an entire small response or the trailing bytes of a larger response. This is latent rather than live: every servlet under `services/http` ends with `println`, and the one non-`println` writer call runs upstream of any wrapper.

Before/after checks found matching response bodies and byte counts for the tested synchronous production write patterns, including `println`, JSON-RPC output-stream writes, and UTF-8 responses.

Metrics are still recorded after the filter chain returns. They measure response-body bytes written to Jetty, excluding HTTP headers and chunk framing; they do not guarantee delivery to the client. Existing exclusions for disabled-API and lite-node responses remain. Requests without a Jetty base request record zero bytes, and the new lookup introduces no cast. Production filter registration and request parsing are unchanged.

- Unit Tests
  - Real-Jetty tests cover `println`, unflushed `print`, output-stream writes with `Content-Length`,
    UTF-8, a body past the output buffer, and HTTP 400, using the access-filter-before-interceptor
    ordering of full-node wallet endpoints. An outer completion latch keeps the metric assertions
    deterministic. Unit tests cover the non-Jetty fallback and propagation to both traffic meters.
  - On the pre-change classes, the `print` and large-body tests fail; mutation testing confirms the
    remaining assertions are not vacuous.
  -  Focused tests passed.
- Manual Testing
  - Test by private chain.

**Follow up**

Heads-up for whoever lands #6923: it removes the Codahale `MetricsUtil.meterMark` calls from this
same method, so the two will conflict textually. The byte count itself is still needed afterwards
for the Prometheus `tron:http_bytes` histogram.


317787106 (approved):

317787106:
LGTM.

Issue #6951, opened 2026-09-08
[Feature] Standardize JSON-RPC null parameter handling


waynercheung:
# Summary

Some JSON-RPC methods dereference a `null` parameter without a null check, throw `NullPointerException`, and answer the client with jsonrpc4j's fallback error. On Java 8 the response is:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":null,"data":"java.lang.NullPointerException"}}
```

This issue defines the `null` handling of 10 parameter positions and 7 optional DTO fields:

- A required object parameter that is null returns `-32602`.
- A null filter ID or `fullTransactionObjects` follows go-ethereum's result.
- An optional DTO field explicitly set to `null` is treated as omitted.

This issue handles the listed null inputs and aligns `eth_uninstallFilter` lookup-miss results with go-ethereum. Other behavior of non-null inputs, HTTP status codes, gRPC and HTTP API behavior remain unchanged.

This issue covers how these methods themselves check and answer a null, together with the `eth_uninstallFilter` lookup-miss result; it does not touch the framework layer. The shape of the fallback response for unmapped exceptions (`-32001`, echoed exception class name) is a separate problem, tracked in #6941.

# Problem

### Motivation

- `message: null` violates JSON-RPC 2.0 section 5.1, which defines `message` as "A String providing a short description of the error" (`null` is not a String). On JVMs where helpful NullPointerException messages are enabled, which is the default from JDK 15 onwards, `message` instead carries a diagnostic string naming internal fields and method signatures.
- `data` echoes the Java exception class name, which clients should not depend on.
- `-32001` is registered in the public error catalog as a server-side internal error, so clients cannot tell that they passed a bad parameter.
- The same class of input gets different results across methods (an error, a success, or a different code), so clients cannot handle it uniformly.

### Current State

10 parameter positions dereference the argument without a null check:

- null object parameter: `eth_getLogs`, `eth_newFilter`, `eth_estimateGas`, `eth_call`, `buildTransaction`
- null filter ID: `eth_uninstallFilter`, `eth_getFilterChanges`, `eth_getFilterLogs`
- null `fullTransactionObjects` (`Boolean` auto-unboxing): `eth_getBlockByHash`, `eth_getBlockByNumber`

[...]

Issue #6941, opened 2026-09-01
[Feature] Standardize JSON-RPC error mapping and exception boundaries


waynercheung:
# Summary

For an exception without an `@JsonRpcErrors` mapping, JSON-RPC falls back to jsonrpc4j's `-32001` and returns the Java exception class name and the raw exception message to the client, for example:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":null,"data":"java.lang.NullPointerException"}}
```

This issue standardizes JSON-RPC error mapping and the exception boundaries:

- Unmapped non-fatal exceptions return `-32603 "Internal error"`; Java exception class names and raw messages are no longer echoed.
- Four fatal `Error` categories, including java-tron's `TronError`, propagate instead of being disguised as JSON-RPC error responses. That classification governs the resolver and the two dispatch boundaries; the outer `doPost` guard is a last-resort cleanup that covers any `Error` raised outside the conversion path, so it is not a guarantee that only these four categories can escape every stage of the servlet. Before rethrowing, the servlet makes a best-effort attempt to commit an empty HTTP 500 so the container does not render exception details.
- An invalid request ID type or non-null scalar `params` returns `-32600 "Invalid Request"` instead of being swallowed on the single-request path or aborting the whole batch.

Only failure-path responses change. Whenever a normal JSON-RPC response is produced, its HTTP status remains 200; before propagating a fatal `Error`, the servlet best-effort commits an empty HTTP 500, while a failed attempt may leave a closed connection. Successful responses, gRPC and non-JSON-RPC HTTP API behavior remain unchanged.

This issue only covers the framework-level fallback and exception boundaries; it does not change how any method validates its own parameters. The null parameter of `eth_getLogs` in the example is a separate problem; even once it gets a null check, any other unmapped exception still takes this path.

# Problem

### Motivation

- `message: null` violates JSON-RPC 2.0 section 5.1, which defines `message` as "A String providing a short description of the error" (`null` is not a String); on Java 17 it becomes a diagnostic string containing internal method signatures; `data` echoes the Java exception class name. None of these should be depended on by clients.
- `-32001` is registered in the public error catalog as a server-side internal error, yet the fallback files every unmapped exception there, including client input errors, so clients cannot tell them apart.
- `OutOfMemoryError` / `StackOverflowError` are converted into ordinary error responses, masking an unrecoverable process state.
[...]

PR #6984, opened 2026-09-21
fix(jsonrpc): handle null parameters consistently

waynercheung:
**What does this PR do?**

Handles 10 identified top-level parameter positions that can trigger unchecked dereferences, normalizes seven optional DTO fields whose explicit JSON `null` currently behaves differently from omission, and aligns `eth_uninstallFilter` lookup misses with go-ethereum. The per-parameter decision basis is recorded in #6951.

- Handle the 10 affected positions without unchecked dereferences. For source-gated methods the request-source check (`disableInPBFT` / FullNode-only) stays first. Object and filter-ID arguments are validated before business processing, while `fullTransactionObjects` is normalized at its existing use site after the block lookup, preserving the non-existent-block behavior:
  - `eth_getLogs` / `eth_newFilter` with a null filter object -> `-32602 "invalid filter request"`
  - `eth_call` / `eth_estimateGas` / `buildTransaction` with a null argument object -> `-32602 "invalid params"`
  - `eth_uninstallFilter(null)` -> `false`; `eth_getFilterChanges(null)` / `eth_getFilterLogs(null)` -> `-32000 "filter not found"` (go-ethereum's results for a null filter ID)
  - `eth_getBlockByHash` / `eth_getBlockByNumber` with a null `fullTransactionObjects` -> treated as `false` (`Boolean.TRUE.equals(...)` instead of auto-unboxing)
- `eth_call` validates its required transaction argument before the block selector, so `[null, null]` returns `-32602` instead of `-32600`.
- `@JsonSetter(nulls = Nulls.SKIP)` on the six optional `BuildArguments` fields (`tokenId`, `tokenValue`, `consumeUserResourcePercent`, `originEnergyLimit`, `permissionId`, `extraData`) and on `CallArguments.from`, so an explicit JSON `null` keeps the field default exactly as omitting the field does.
- A defensive null check at the `LogFilter(FilterRequest)` constructor, plus a shared `INVALID_FILTER_REQUEST` message constant.
- `eth_uninstallFilter` returns `false` for null, unknown or already removed IDs, including empty and non-hexadecimal strings that miss the lookup. It returns `true` only when `remove` actually removes an installed event or block filter, without a separate `containsKey` check. ID normalization and request-source selection are unchanged; no format validation or locking is added.
- Remove this method's obsolete `ItemNotFoundException` declaration from `TronJsonRpc` and `TronJsonRpcImpl`, and its corresponding `-32000` annotation. Other methods' mappings are unchanged.

Remaining method-level error responses keep their existing `@JsonRpcErrors` mappings: `data` stays `"{}"` and the request `id` is echoed. Uninstall lookup misses now produce a Boolean result instead of the removed error mapping.

Before this PR the ten positions dereferenced the argument without a null check. The resulting `NullPointerException` is not covered by `@JsonRpcErrors`, so jsonrpc4j's fallback returned:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":null,"data":"java.lang.NullPointerException"}}
```

`message: null` violates JSON-RPC 2.0 (on JVMs where helpful NullPointerException messages are enabled, the default from JDK 15 onwards, it becomes a diagnostic string echoing internal class and method names), `data` exposes the Java exception type, and a required-parameter error such as this example is reported under the code java-tron documents for internal errors. The seven DTO fields were inconsistent among themselves: an explicit `null` failed with `-32001` or `-32000` depending on the path, while `CallArguments.from` returned `-32602` where omitting the field continues with the zero address.

The per-parameter decisions follow the basis recorded in #6951: the Execution API `required` / schema first, then whether a zero value can be inferred from `null`, then go-ethereum / Besu behavior as a reference. Details and reproduction steps are in #6951.

- [x] Unit Tests
- [ ] Manual Testing

- `JsonRpcArgumentNullBindingTest` - Jackson binding through `ObjectMapper`: an explicit `null` and an omitted field produce the same `BuildArguments` / `CallArguments`.
- `JsonRpcCallAndEstimateGasTest` / `JsonRpcTest` - service-level results for null argument objects and null filter requests, and the `eth_call` argument-before-block ordering.
- `JsonrpcServiceTest` - wire-level contract through a real `JsonRpcServer`: `code` / `message` / `data` for every case above, filter-object null wildcard semantics unchanged (`address: null`, `topics: null`, positional `[A, null]`), optional DTO null fields, and an assertion that no response contains `java.` or an exception class name. Includes wire regressions for the four hash-parameter methods fixed by #6828.
- `JsonrpcServiceTest` also covers event/block filter creation and removal through the real server, repeated removal, missing/empty/non-hexadecimal IDs, strict Boolean results without an `error` field, preserved ID normalization, unchanged get-filter errors and concurrent removal of one installed filter.
- `WalletCursorTest` - FullNode/Solidity filter stores stay isolated; both filter types can be removed only from the selected source. PBFT rejects null, installed, unknown and malformed string IDs before removing anything.

On 2026-09-08, `:framework:cleanTest :framework:test --no-build-cache` for `org.tron.core.jsonrpc.*` and `org.tron.core.services.jsonrpc.*` passed **24 classes / 242 tests** (0 failures, 0 errors, 0 skipped) on JDK 17 (arm64), including all eight new lookup-miss regressions. On 2026-09-15 the same suite was re-run on the branch rebased onto `release_v4.8.3` (`0d19485318`, which already carries the slf4j 2.0.17 / logback 1.3.16 / jackson 2.18.10 upgrade from #6950 and the HTTP error sanitization from #6954) with the same result. `:framework:checkstyleMain`, `:framework:checkstyleTest` and `git diff --check` also pass. This validates compilation of the repository's callers after the checked-exception removal; external Java source callers may still need adjustment. Java 8/x86 execution was not performed.

**Compatibility**

Breaking, limited to the affected null-input and filter lookup-miss responses. Successful valid calls remain unchanged. Observable changes are:

- required object parameters change from the unmapped `-32001` fallback to `-32602`, while the two get-filter methods change from `-32001` to `-32000`;
- `eth_uninstallFilter(null)` and `eth_getBlockBy*(..., null)` go from an error to a normal response;
- `eth_uninstallFilter` with an unknown, already removed, empty or non-hexadecimal string ID changes from `-32000 "filter not found"` to `result: false`. Existing filters still return `true` on removal; simultaneous removals no longer both report success for the same installed entry;
- explicit null DTO fields recover the omitted-field semantics;
- `eth_call([null, null])` goes from `-32600` to `-32602`;
- `eth_call([null, "earliest"])` and similar shapes where parameter 0 and the block selector are both invalid keep `-32602`, but the message changes from the block error to `"invalid params"` because the required argument is now validated first.

Clients matching on the old codes or messages for these paths, or branching on an uninstall error versus a result, need to adjust.

Removing the checked `ItemNotFoundException` declaration is Java binary-compatible, but source callers that specifically catch it around this call, or implementations that still declare it, may need adjustment when recompiled. This PR removes one obsolete mapping; it does not change the other filter methods' errors.

Unchanged: successful valid-call results, HTTP status, request-source precedence, `data` on remaining mapped errors, `result: null` for a non-existent block, filter-object null wildcards, unknown-ID errors from `eth_getFilterChanges` / `eth_getFilterLogs`, gRPC and non-JSON-RPC HTTP APIs.

**Follow up**

- The unmapped-exception fallback path itself (any other unmapped exception) is addressed in #6941. If that change lands first, the observed fallback for these paths becomes `-32603 "Internal error"` without `data`, which moves only the "before" side of the comparison; the target behavior of this PR is the same either way.
- Follow-up items: `web3_sha3(null)`, unconditional transaction index validation.
- `params` missing / `null` / `[]` retain existing dispatch and arity behavior and are not universally rejected: whether they are accepted depends on how many parameters the method declares. They are not touched here. Request-envelope validation is tracked by #6676, which overlaps this PR in `JsonRpcApiUtil` and `TronJsonRpcImpl`; there is no dependency between them (see Extra details).

**Extra details**

This PR overlaps the request-envelope validation planned in #6676 in `JsonRpcApiUtil` and `TronJsonRpcImpl`, and shares `TronJsonRpc` and `TronJsonRpcImpl` with the error-mapping change tracked by #6941. This PR does not require #6676 or #6941 to land first; whichever lands second rebases and re-runs the related regression tests.

Pre-submit checklist:

- [x] Google Java Style; Checkstyle passes on main and test sources
- [x] No debug code, temporary comments or TODOs
- [x] No numeric computation or narrowing casts introduced
- [x] No logging added or changed
- [x] No DB, consensus, config or dependency changes
- [x] Comments explain the non-obvious points: `@JsonSetter` protects Jackson input only, and positional vs. OR-list null topics in `LogFilter`

Closes #6951
Refs #6676


framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpc.java
```diff
@@ -307,10 +307,9 @@ String newFilter(FilterRequest fr) throws JsonRpcInvalidParamsException,
   @JsonRpcErrors({
       @JsonRpcError(exception = JsonRpcMethodNotFoundException.class, code = -32601, data = "{}"),
       @JsonRpcError(exception = JsonRpcInvalidParamsException.class, code = -32602, data = "{}"),
-      @JsonRpcError(exception = ItemNotFoundException.class, code = -32000, data = "{}"),
   })
   boolean uninstallFilter(String filterId) throws JsonRpcInvalidParamsException,
-      JsonRpcMethodNotFoundException, ItemNotFoundException;
+      JsonRpcMethodNotFoundException;
```
bladehan1, line 312: [NIT] Preserve source compatibility for uninstallFilter

Returning `false` on a lookup miss only requires the implementation to stop throwing `ItemNotFoundException`; removing the checked exception from this public interface also makes existing source callers that catch it and third-party implementations that still declare it fail to recompile. That is a separate compatibility cost from the intended JSON-RPC wire change.

Suggestion: Keep the existing `ItemNotFoundException` declaration and error mapping in this release while leaving the implementation non-throwing, and remove them only in a separately announced compatibility cleanup.

PR #6986, opened 2026-09-21
fix(config,toolkit): harden keystore, grpc config and build supply chain

warku123:
**What does this PR do?**

Six independent hardening changes, one commit each.

- **Keystore password input fails fast on EOF** — `WalletUtils.inputPassword()` now terminates with `TronError(WITNESS_KEYSTORE_LOAD)` instead of crashing with a raw NPE (`Console.readPassword()` returns null) or leaking `NoSuchElementException` (`Scanner.nextLine()`) when stdin closes mid-prompt. Covered by `WalletUtilsInputPasswordTest`.

- **Keystore KDF parameter bounds** — create/decrypt/validate now enforce bounds on untrusted keystore KDF parameters before any KDF runs: scrypt n power-of-2 in [4096, 1048576], r<=8, p<=8, 128*r*n<=1 GiB, dklen in [32,128]; pbkdf2 c<=2^20. `Wallet.create(...)` applies the same checks so a keystore cannot be created that a later decrypt would reject. Prevents huge allocations/OOM, integer-overflow exceptions, and multi-hour CPU hangs from hostile keystore files. Covered by `WalletDecryptBoundsTest`.

- **CLI secret flag deprecated** — `--private-key` is marked `@Deprecated` and logs a WARN at startup when used (it exposes the witness key to the process list and shell history); README and `docs/configuration.md` point to `localwitnesskeystore` instead. `--password` is left untouched pending a real non-interactive alternative.

- **gRPC connection defaults** — `maxConnectionIdleInMillis`/`maxConnectionAgeInMillis` = 0 now selects a built-in 60 s default instead of `Long.MAX_VALUE` (unbounded); negative values fail fast with `TronError`; a half-configured RST window pair logs a loud WARN that flood protection is disabled. Covered by `NodeConfigTest`.

- **Toolkit password/key file permissions** — `--password-file`/`--key-file` must now be owner-only (no group/other access) on POSIX systems; otherwise the tool refuses the file with an actionable error message. Docs note the required `chmod 600`. Covered by `KeystoreCliUtilsTest` and the keystore command tests.

- **Build supply chain** — Gradle distribution pinned via `distributionSha256Sum`; `gradle/actions/wrapper-validation` added to all build workflows; JitPack removed and `mavenLocal()` moved after Central; blanket SNAPSHOT/javadoc/sources trust rules removed from `verification-metadata.xml` so strict checksum verification applies to every artifact.

- Untrusted keystore files are attacker-controlled input: unbounded KDF cost parameters turn a single decrypt call into memory exhaustion or multi-hour CPU consumption, and malformed inputs escape as raw runtime exceptions.
- Password entry must fail with an actionable error, not an NPE, and secrets should not be passed on the command line.
- A gRPC endpoint whose configured lifetime is actually unbounded cannot shed stale connections; a half-configured RST window silently disables flood protection.
- Build inputs (wrapper zip, dependency repositories, verification metadata) must be pinned or checksum-verified.

- Unit Tests
  - `WalletDecryptBoundsTest` — hostile keystores (oversized scrypt, overflow, huge pbkdf2), boundary-valid cases, and create-side bounds
  - `WalletUtilsInputPasswordTest` — piped stdin EOF
  - `NodeConfigTest` — defaults, explicit zero, half-configured pair, negative values
  - `KeystoreCliUtilsTest` / `KeystoreImportTest` / `KeystoreNewTest` / `KeystoreUpdateTest` — permission gate and command flows
  - `:common:test :crypto:test :plugins:test` full suites green
- Manual Testing — clean-clone `:framework:compileTestJava` plus targeted tests green

**Moved out of this PR during review**

- `protoc-gen-grpc-java` 1.60.0 → 1.83.0 unification — dedicated dependency-upgrade PR (1.60 stays pinned for CentOS 7 compatibility).
- `--password` deprecation + WARN — deferred until a non-interactive alternative (e.g. `--password-file`) exists.

**Follow up**

- Not included: Toolkit `DbMove` failure-path/symlink handling, `start.sh` hardening — proposed separately.

**Extra details**

- No behavior change for keystores within the bounds, for explicitly positive gRPC settings, for interactive password entry, or for successful builds.


gradle/verification-metadata.xml
```diff
@@ -3,11 +3,6 @@
    <configuration>
       <verify-metadata>true</verify-metadata>
       <verify-signatures>false</verify-signatures>
-      <trusted-artifacts>
-         <trust file=".*-javadoc[.]jar" regex="true"/>
-         <trust file=".*-sources[.]jar" regex="true"/>
-         <trust group=".*" name=".*" version=".*-SNAPSHOT" regex="true"/>
-      </trusted-artifacts>
```
317787106, line 10: [MUST] Don't remove it.

plugins/src/test/java/org/tron/plugins/KeystoreNewTest.java
```diff
@@ -27,6 +28,7 @@ public void testNewKeystoreWithPasswordFile() throws Exception {
     File dir = tempFolder.newFolder("keystore");
     File pwFile = tempFolder.newFile("password.txt");
     Files.write(pwFile.toPath(), "test123456".getBytes(StandardCharsets.UTF_8));
+    makeOwnerOnly(pwFile);
```
317787106, line 31: [SHOULD] Please avoid using `Assume.assumeTrue()` inside this shared helper. On filesystems without POSIX attributes, it skips the entire calling test, including otherwise platform-independent tests for keystore creation, import, and password updates. Production code explicitly supports these filesystems by skipping the permission check.

  Please make `makeOwnerOnly()` a no-op when POSIX attributes are unavailable, and keep the assumption only in tests specifically verifying POSIX permissions. The same change applies to the helpers in `KeystoreCliUtilsTest`, `KeystoreImportTest`, and `KeystoreUpdateTest`.

gradle/verification-metadata.xml
warku123: Restored per review. One correction: the first restore commit 03c88f1ed5 mistakenly placed the block as a sibling of `<configuration>`, which broke Gradle verification-schema parsing and failed CI across the board. Fixed in 6b4f194a7d — `trusted-artifacts` now sits inside `<configuration>` right after `verify-signatures`, and the file is byte-identical to upstream. Verified locally with a Gradle run before pushing.

plugins/src/test/java/org/tron/plugins/KeystoreNewTest.java
warku123: Fixed in c30640555a. The shared makeOwnerOnly helpers no longer call Assume.assumeTrue — on non-POSIX filesystems they now skip only the chmod step and return, mirroring the production fail-open path in KeystoreCliUtils (posixView == null → permission check skipped). This un-skips the ~69 platform-independent cases that consume the helper (CRLF/BOM/SM2/large-file/password-format across the 4 keystore test classes) on Windows. Tests asserting POSIX permission behavior keep their own method-level assumes, and the missing one was added to testReadRegularFileRejectsGroupReadable, which chmods unconditionally after the helper. Behavior on Linux CI is unchanged.

gradle/verification-metadata.xml
warku123: Follow-up on the build supply-chain hardening scope: after re-evaluating the invasiveness of the broader changes, they have been migrated out of this PR — `build.gradle` is restored to upstream (JitPack repositories and `mavenLocal()` ordering), the `distributionSha256Sum` pin in `gradle-wrapper.properties` has been removed, and `verification-metadata.xml` is byte-identical to upstream with `trusted-artifacts` restored (094834c692, 6b4f194a7d).

The only supply-chain measure remaining in this PR is the CI-level `wrapper-validation` step in the workflows, which validates the `gradle-wrapper.jar` itself against Gradle's official checksums and carries no upgrade-maintenance coupling.

The other remediation approaches (repository allow-listing, distribution checksum pinning, tightening dependency verification) proved too risky to land safely in a patch-release hardening PR — they interact with the release build and local developer workflows in ways that need a dedicated discussion. We will address them in a separate follow-up rather than in this PR.

PR #6990, opened 2026-09-22
refactor(framework): decouple Manager from TronJsonRpcImpl

### LINKED ISSUE #6963: [Feature] decouple json-rpc filter processing from Manager
# Summary

The json-rpc filter event-processing thread currently is located within `Manager` (core layer), forcing it to hold `TronJsonRpcImpl` (API layer) via a `@Lazy` injection — a reverse circular dependency. This proposal moves that thread into `TronJsonRpcImpl`; `Manager` only posts filter events into a standalone queue bean, and the reverse edge disappears.

# Problem

### Motivation

#6732 turned `handleLogsFilter`/`handleBLockFilter` from static into instance methods, so `Manager`'s event-processing thread needs a `TronJsonRpcImpl` instance; the review deferred the resulting `@Lazy` cycle to a follow-up ([discussion](https://github.com/tronprotocol/java-tron/pull/6732#discussion_r3193518283)) — this issue serves as the dedicated follow-up to address that technical debt.

### Current State

```
Manager ──@Lazy @Autowired field──▶ TronJsonRpcImpl ──@Autowired field──▶ Manager
```

Each arrow reads "depends on"; the labels are injection styles. `TronJsonRpcImpl`'s constructor parameters `Wallet`/`NodeInfoService` each depend on `Manager` as well, closing the same cycle indirectly.

`Manager` owns the filter queue, the consumer thread, and its shutdown; the consumer calls back into the API layer.

### Limitations or Risks

- `FullNode` sets `allowCircularReferences(false)`, so a directly injected cycle is rejected at startup; `@Lazy` (like `ObjectProvider` or a runtime `getBean()`) only makes Spring tolerate the cycle — the reverse edge stays, obscuring the component graph and complicating lifecycle changes.
- The reverse core → API edge risks encouraging future changes to repeat the same `@Lazy` pattern, widening the breach.
- One latent defect remains in the current implementation: the consumer stop flag is a non-volatile boolean (the shutdown-time write has no visibility guarantee for that thread), and the loop does not exit on `InterruptedException` — under the Java memory model, neither the normal nor the forced stop path is reliable.

# Proposed Solution

### Proposed Design

Eliminate the reverse dependency entirely rather than suppressing it via @Lazy: a standalone `FilterCapsuleQueue` bean is introduced as an intermediary between the two layers — `Manager` only produces into it; `TronJsonRpcImpl` owns the consumer thread (start on `@PostConstruct`, stop via idempotent `close()`). `ApplicationImpl.shutdown()` calls `close()` explicitly after producers stop and before `dbManager.close()` — Spring destroys beans only after the database closes, so bean destruction alone would let the consumer outlive it (`ApplicationImpl` is already the shutdown-orchestrating composition root). Filter API behavior stays unchanged (unbounded queue; pending entries are not drained on shutdown, while an in-flight capsule may complete).

### Key Changes

- Module: `framework` only
- `org.tron.common.logsfilter.queue.FilterCapsuleQueue` (new bean, beside the existing `org.tron.common.logsfilter.capsule` package)
- `Manager`: drop the `@Lazy` field, the consumer loop, and its thread lifecycle
- `TronJsonRpcImpl`: constructor gains a `Manager` parameter, consumer lifecycle, rewritten `close()` ordering (stop and await the consumer first, then shut down `logsFilterPool`)
- `ApplicationImpl.shutdown()`: explicit consumer stop
- No configuration or API changes

# Impact

- Stability: fixes the latent stop-path defect above along the way; graceful shutdown stops the consumer before the database closes.
- Performance: no steady-state change expected — processing path, queue, and thread count are unchanged.
- Developer Experience: the core layer no longer references the API layer; `framework/src/main` becomes `@Lazy`-free.

# Compatibility

- Breaking Change: No
- Default Behavior Change: No — filter delivery, queue semantics, and shutdown semantics are preserved
- Migration Required: No

# Additional Notes

0xbigapple:
**What does this PR do?**
close #6963 
Removes the `Manager → TronJsonRpcImpl` reverse dependency left by #6732, where core-layer `Manager` holds the json-rpc filter consumer through a `@Lazy` injection.

- Adds a standalone `FilterCapsuleQueue` bean; `Manager.postBlockFilter/postLogsFilter` produce into it, so Manager no longer references anything under `org.tron.core.services.jsonrpc` (repo-wide `framework/src/main` is now `@Lazy`-free).
- Moves the consumer loop into `TronJsonRpcImpl`: started by `@PostConstruct` when `isJsonRpcFilterEnabled()`, single daemon thread, stopped by `close()`.
- `TronJsonRpcImpl` constructor becomes `(NodeInfoService, Wallet, Manager)`; `setManager()` is removed and the direct-construction test sites are migrated.
- `close()` is rewritten: an `AtomicBoolean` guard makes it idempotent and serves as a visibility-safe stop flag, the loop exits on interrupt, and the consumer is awaited before `logsFilterPool` shuts down so an in-flight capsule completes on graceful shutdown.
- `ApplicationImpl.shutdown()` now closes the consumer after producers stop and before `dbManager.close()`; the later Spring-destruction `close()` is a no-op.
- The `instanceof` dispatch logs a warning for unknown `FilterTriggerCapsule` subtypes instead of silently dropping them.

Runtime behavior of the filter API is unchanged: same unbounded queue, same discard-on-shutdown semantics, one consumer shared by the FullNode/solidity/PBFT json-rpc services.

Follow-up agreed in the #6732 review (see [discussion](https://github.com/tronprotocol/java-tron/pull/6732#discussion_r3193518283)). `FullNode` sets `allowCircularReferences(false)`; `@Lazy`, like `ObjectProvider` or a runtime `getBean()`, only makes Spring tolerate the cycle. The reverse core → API edge stays, obscuring the component graph, and later json-rpc cleanups keep copying the pattern. This PR removes the edge instead.

- Unit Tests
- Manual Testing
  - private chain: block production with registered filters; `eth_getFilterChanges` delivers; SIGTERM log order confirms consumer → `logs-filter-pool` → `dbManager` → context close, single shutdown episode, no errors
  - Nile and mainnet lite nodes: 5 restart cycles each with 100 filters created and polled per cycle (~2,500 queries/cycle, zero rpc errors); on mainnet the log path delivered 20k+ real contract log entries through the new queue; filters correctly report `filter not found` after restart; no `RejectedExecutionException`, no bean-cycle errors, shutdown order identical in every cycle

**Follow up**

**Extra details**

### DIFF
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
@@ -236,6 +236,7 @@ public class TronJsonRpcImpl implements TronJsonRpc, Closeable {
       } catch (InterruptedException e) {
         logger.error("FilterProcessLoop get InterruptedException, error is {}.", e.getMessage());
         Thread.currentThread().interrupt();
+        return;
       } catch (Throwable throwable) {
         logger.error("Unknown throwable happened in filterProcessLoop. ", throwable);
       }
--- a/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
+++ b/framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
@@ -36,7 +36,9 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
+import javax.annotation.PostConstruct;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
@@ -53,7 +55,9 @@ import org.tron.common.crypto.Hash;
 import org.tron.common.es.ExecutorServiceManager;
 import org.tron.common.logsfilter.ContractEventParser;
 import org.tron.common.logsfilter.capsule.BlockFilterCapsule;
+import org.tron.common.logsfilter.capsule.FilterTriggerCapsule;
 import org.tron.common.logsfilter.capsule.LogsFilterCapsule;
+import org.tron.common.logsfilter.queue.FilterCapsuleQueue;
 import org.tron.common.parameter.CommonParameter;
 import org.tron.common.runtime.vm.DataWord;
 import org.tron.common.utils.ByteArray;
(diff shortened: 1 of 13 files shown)


framework/src/main/java/org/tron/core/services/jsonrpc/TronJsonRpcImpl.java
```diff
@@ -1613,7 +1646,13 @@ public Object[] getFilterResult(String filterId, Map<String, BlockFilterAndResul
...
 
   @Override
-  public void close() throws IOException {
+  public void close() {
+    if (!closed.compareAndSet(false, true)) {
+      return;
+    }
+    // The consumer loop submits to logsFilterPool (over-threshold path), so it must
+    // terminate before the pool shuts down.
+    ExecutorServiceManager.shutdownAndAwaitTermination(filterEs, filterEsName);
```
317787106, line 1655: [SHOULD] `shutdownAndAwaitTermination()` does not guarantee that the consumer has terminated: if the closing thread is interrupted, it calls `shutdownNow()` and returns without waiting again. We then shut down `logsFilterPool` while the consumer may still submit work to it.

  I reproduced `close()` returning with `filterEs.isTerminated() == false` and `logsFilterPool.isShutdown() == true`; resuming the consumer then caused a RejectedExecutionException.

  Could we preserve the consumer-before-pool shutdown ordering on this interruption path and add a regression test? The existing interrupted-close test registers no filters, so it never exercises submission to `logsFilterPool`.

PR #6992, opened 2026-09-22
chore(p2p): internalize libp2p v2.2.9 as a local `p2p` module

barbatos2011:
## What

Internalize `io.github.tronprotocol:libp2p:2.2.9` source as a new `p2p/` Gradle submodule, replacing the external Maven dependency. No functional change to P2P behaviour, protocol, or consensus.

| | |
|---|---|
| Base | `tronprotocol/java-tron` `release_v4.8.3` @ `880eef6d5` |
| Vendored | `tronprotocol/libp2p` tag `v2.2.9` @ `c564f263d310d7a964035d3b597634aba6bda86d` |

**Commit 1 is byte-comparable to the upstream tag.** It contains upstream's `src/main` verbatim — all 94 Java files, 90 under `src/main/java` and the 4 reference classes split into a separate `src/example` sourceSet — plus the new build files. A reviewer can diff it directly against `c564f263` and see that no vendored line was altered. Everything this PR changes about that source lands in commit 2, separately, for exactly this reason.

## Why

Upstream `tronprotocol/libp2p` is maintained in bursts. Over the last 25 months (2024-09 through 2026-09) **11 months had no commits at all**, including four consecutive months (2025-12 through 2026-03), and there has been no commit since 2026-07:

| Release | Date |
|---|---|
| v2.2.9 | 2026-07-30 |
| v2.2.8 | 2026-06-29 |
| v2.2.7 | 2025-11-11 |
| release_v2.2.6 | 2025-03-14 |

Cross-repo maintenance creates friction that shows up on every P2P change:

- **Version coordination overhead** — modifying P2P behaviour means publishing libp2p first, then bumping java-tron's dependency
- **Debugging difficulty** — the IDE cannot step into an external artifact without downloading source jars
- **CI dependency** — the build depends on external Maven/JitPack artifact availability
- **Industry alignment** — go-ethereum keeps its `p2p/` package in the monorepo; this follows the same pattern

Internalizing makes a P2P change a single PR instead of a cross-repo dance, and does not preclude back-porting individual commits to upstream libp2p when appropriate.

### 1. `chore(p2p)`: add libp2p v2.2.9 source as p2p module

Upstream `src/main` verbatim: 90 files under `p2p/src/main/java` (including the embedded `org.web3j` crypto utilities), 2 `.proto` files, the 4 reference classes under `p2p/src/example`, plus `p2p/build.gradle`, `p2p/.gitignore` and the `settings.gradle` registration. Protobuf-generated sources are gitignored and regenerated at build time, so they are not in the diff.

### 2. `style(p2p)`: conform vendored source to project conventions

Three mechanical rewrites plus formatting, kept separate so commit 1 stays diffable against the tag:

- `log.` → `logger.` (178 call sites, 158 under `src/main` and 20 in the example sourceSet) — the root `lombok.config` sets `lombok.log.fieldName=logger`
- `Math.` → `StrictMath.` (6 sites) — CI's `check-math` rule rejects `java.lang.Math` tree-wide
- `toLowerCase()` / `toUpperCase()` → `Locale.ROOT` (4 sites) — compile-forced by errorprone, see **Key design decisions**

Formatting: `google-java-format` over the 9 vendored `org/web3j/**` files, which came in AOSP 4-space style, plus import reordering. `:p2p:checkstyleMain` passes with 0 violations.

`BasicThreadFactory` is deliberately left as upstream writes it — twelve `builder()` calls and the one `new BasicThreadFactory.Builder()` in `DiscoverServer`. The root provides commons-lang3 3.20.0, where `builder()` exists and the no-arg `Builder()` constructor is `@Deprecated`, so rewriting either way would add deviation from the tag and, in one direction, thirteen deprecation warnings.

### 3. `build(common)`: switch from external libp2p to local p2p module

Replaces the Maven dependency with `api project(":p2p")`, collapsing 17 lines of dependency plus excludes into one.

The dom4j exclusion tail that used to sit on the libp2p dependency here (`jaxen`, `stax-api`, `xsdlib`, `pull-parser`, `xpp3`) does not disappear: it arrives via the Aliyun and Route53 SDKs, which are now `:p2p`'s own dependencies, so a `configurations.configureEach` block in `p2p/build.gradle` (added with the module in commit 1) carries it. Dropping it would silently re-admit artifacts the project has excluded for years. That block also names `msv` and `relaxngDatatype`, which `common/build.gradle` did not — the same dom4j tail, excluded completely rather than partially.

Adding a project to the dependency graph also needs task-dependency edges that an external jar did not. `:framework`'s `buildFullNodeJar` and `:plugins`' `binaryRelease` both zip up `runtimeClasspath` and maintain a hand-written `dependsOn` list of project jars; `p2p-1.0.0.jar` is now on both classpaths, and without the edge a parallel build could assemble the shipped fat jar before `:p2p:jar` exists.

`gradle/verification-metadata.xml` gains three components that resolve once p2p compiles in-tree: `bcutil-jdk18on:1.84`, `gson:2.9.0` and `gson-parent:2.9.0`. Checksums taken from Maven Central and cross-checked against the published `.sha1`.

### 4. `build`: track `netty-codec-protobuf` via a root `nettyVersion`

Netty 4.2 split `io.netty.handler.codec.protobuf` out of `netty-codec`, so `:framework` and `:p2p` both declare it explicitly — each puts the varint32 framing codecs on its channel pipelines. Netty itself is declared nowhere; it arrives transitively through `grpc-netty`, which `:p2p` tracks as `rootProject.grpcVersion`. Extracting `nettyVersion` next to `grpcVersion` makes that coupling visible in one place instead of leaving two literals to drift.

### 5. `build(framework)`: declare the direct `:p2p` dependency

`:framework` uses `org.tron.p2p` in 17 files under `src/main/java` but declared nothing, relying on a three-hop transitive through `:common -> :crypto -> :chainbase`. Declared as **`api`, not `implementation`**, because framework re-exports p2p types: `P2pEventHandlerImpl extends org.tron.p2p.P2pEventHandler`, `HelloMessage.getFrom()` returns `org.tron.p2p.discover.Node`, `PeerManager.add/remove` take `org.tron.p2p.connection.Channel`, and `Args.loadDnsPublishConfig` returns `PublishConfig`. No resolution change either way.

### 6. `test(p2p)`: port and extend the module's test suite

All 23 of v2.2.9's own test files, plus new ones, in `p2p/src/test/java` next to the code they cover — 64 files in total.

Two upstream tests were unreliable by construction and are fixed rather than carried over as-is:

- `NetUtilTest.testGetIP` called three public IP-echo services and asserted all three returned the same string: a network dependency, and a coin flip on any host with more than one egress address. It now runs against a loopback `HttpServer`, covering the same fetch/parse/validate path plus the rejection branches. (libp2p's own CI never ran its tests, so this had not surfaced.)
- `ConnPoolServiceTest.getNodes_orderByUpdateTimeDesc` asserted `getNodes()` returns nodes ordered by `updateTime` descending. `getNodes()` sorts, truncates to `max(limit * 10, 50)` candidates, then calls `Collections.shuffle()` — so with two nodes that assertion passes about half the time. It now asserts membership, and a new test covers the descending sort above the candidate bound, where it is actually observable.

Two further defects made coverage depend on fork scheduling rather than on what the tests assert:

- `ConnPoolServiceTest` and `SocketTest` bound fixed ports (10000 / 10001). `PeerServer.start` only logs on bind failure, so a collision let them pass while exercising nothing. Both now take a free port from `TestPort.choose()`. framework's own tests use `org.tron.common.utils.PublicMethod.chooseRandomPort` for this, but `:p2p` cannot depend on `:framework` without creating a cycle, so the same few lines live in the module.
- `NodeTableTest` read `Parameter.p2pConfig` without ever setting it, so it depended on an earlier class in the same fork having done so. Running the class on its own failed all eleven methods. It now sets up and restores its own config.

### 7. `test(p2p)`: replace the example sourceSet with a usage contract test

`DnsExample1`, `DnsExample2` and `ImportUsing` documented how an embedder configures the module, but they only compiled — each ended in a `while (true)` loop, bound a fixed port and pointed at live seed nodes, so nothing they showed was checked. Two carried real defects (`TestMessage` is not serializable, so `ByteArray.fromObject` returns null and `Channel.send` closes the channel; `DnsExample1` held a signing private key in copyable code). `ExampleUsageTest` pins the contract they advertised instead: those configuration shapes are still accepted and still mean what the comments said.

### 8. `docs(p2p)`: add a README, drop the vendored logback sample

The README is promoted from `src/example/resources/README.md`, where no reader would find it, and rewritten — it still read as upstream's document, with four source links pointing at github.com/tronprotocol/libp2p and prose describing libp2p as a standalone project.

`StartApp` moves to `src/main/java` as the standalone entry point and gets the fixes that made it worth documenting, including a real bug: `--trust-ips` is declared `ip[,ip[...]]` but resolved the whole comma-separated value as a single hostname, so with more than one address none of the listed peers became trusted.

`logback.xml.example` came from libp2p as a standalone project; inside java-tron `framework/src/main/resources/logback.xml` is what applies, and the sample was being packaged into the jar for no reason.

### 9. `test(p2p)`: cover StartApp's arg parsing, correct the coverage surface

Moving `StartApp` out of the checkstyle- and coverage-exempt `example` sourceSet puts 981 instructions at 0% into the denominator. It is argument parsing, option declarations and a `main()` that starts services and blocks — not module logic, and not code the node runs — so `org/tron/p2p/example/**` is excluded from this module's jacoco report alongside `**/protos/**`. The two parsing helpers it does own are package-private now and covered by `StartAppArgsTest`, so the exclusion does not take regression protection with it.

`:p2p:jacocoTestReport` also reads `:framework`'s exec data, because framework's own tests execute p2p code and that coverage is real.

### 10. `build(p2p)`: ship a runnable standalone jar, stop repeating versions

The plain jar is thin, so declaring `Main-Class` on it would be worse than declaring nothing — the entry point resolves and then dies on `NoClassDefFoundError: org/apache/commons/cli/ParseException`. It declares none and fails honestly with "no main manifest attribute"; `buildStandaloneJar` produces `p2p-standalone.jar` with the runtime classpath bundled, wired exactly like `:framework`'s `FullNode.jar` and `:plugins`' `Toolkit.jar` — same `artifacts { archives(...) }`, same `-PbinaryRelease=false` opt-out, same Bouncy Castle signature and dnsjava SPI exclusions. Verified by running both jars.

`protobufVersion` 3.25.8 was declared in both `:protocol` and `:p2p`; checkstyle 8.7 sat in a local `versions` map in `:framework` and `:plugins` and as a literal in `:p2p`. Both move to the root `ext` beside `grpcVersion` and `nettyVersion`, and all four modules reference them — removed rather than relocated.

### 11. `build(p2p)`: put `.proto` files where the protobuf plugin looks for them

`src/main/protos` with an explicit `srcDir` override builds fine, but IDEA's protobuf plugin does not read that override — it resolves against the plugin's default path — so `import "Discover.proto"` and every type it brings in showed as unresolved in the editor. The override is gone; generated sources still land in `src/main/java/org/tron/p2p/protos` and are still gitignored.

Flagging for a follow-up: `:protocol` has the identical setup and presumably the identical symptom. Aligning it is the same one-directory rename, but it is core java-tron with a `src/main/gen` interplay and does not belong here.

## Key design decisions

| Decision | Choice | Rationale |
|---|---|---|
| Module name | `p2p` | Matches package `org.tron.p2p`, geth's `p2p/` convention, and the project's short-name style (`common`, `crypto`) |
| Proto files | Stay in `:p2p` | Moving them to `:protocol` would make `:p2p` depend on `:protocol` and cost the module its leaf position — it currently has zero project dependencies, which is what keeps vendored code from reaching back into java-tron internals and keeps upstream re-contribution possible |
| Unit test location | `p2p/src/test` | Same as `common` and `plugins`. Keeping them in the module also avoids a `:framework` test-classpath SDK re-declaration plus dom4j mirror, and a jacoco `additionalClassDirs` bolt-on |
| Reference examples | Replaced by `ExampleUsageTest` | The originals only compiled; a test pins the same contract and actually runs |
| commons-lang3 | No declaration in `:p2p` | The root already hands every subproject `implementation "org.apache.commons:commons-lang3:3.20.0"`, and an `implementation` dependency is on the declaring project's own compile classpath. Confirmed by compiling `:p2p` main and test sources with no `commons-lang3` line |
| `BasicThreadFactory` | Left exactly as upstream | `builder()` exists at 3.20.0 and the no-arg `Builder()` constructor is `@Deprecated`; rewriting either way adds deviation from the tag |
| `grpc-netty` | Tracks `rootProject.grpcVersion` | libp2p pinned 1.81.0; tracking the root's 1.83.1 keeps it from drifting from the Netty the rest of the build resolves |
| `netty-codec-protobuf` | Root `nettyVersion` | Declared by both `:framework` and `:p2p`; a shared version makes the coupling to `grpcVersion` visible |
| `:framework -> :p2p` | `api` | framework re-exports p2p types in its own public signatures |
| `Math` → `StrictMath` | All 6 sites | CI `check-math` rule; integer results identical |
| `Locale.ROOT` | All 4 sites | errorprone `StringCaseLocaleUsage` is ERROR on every subproject except `protocol`/`errorprone`. **The one non-mechanical change in this PR** — behaviour is identical for ASCII but differs under a Turkish locale. Matches the project's own idiom in `Args.java:1273` |
| Checkstyle/jacoco excludes | `**/protos/**`, `**/example/**` | Generated code and the CLI entry point |

## Scope

- No protocol or message-format change
- No consensus change
- No proto change — the `.proto` files move verbatim
- No hard fork
- Package names preserved (`org.tron.p2p.**`), so imports in dependent code are untouched
- Existing nodes upgrading need only `git pull` and rebuild
- `:p2p` remains usable standalone — `java -jar p2p-standalone.jar` runs the module on its own; see `p2p/README.md`

## Defects found in the vendored source (not fixed here)

Two instances of the same shape: a DNS parse helper throws an *unchecked* exception past a `catch (DnsException)` that was clearly written to tolerate unparseable input, so one malformed TXT record aborts the whole operation instead of being skipped.

1. **Short root entry.** `RootEntry.parseEntry` (`dns/tree/RootEntry.java:67`) does `e.substring(rootPrefix.length())` with no length guard, so any value shorter than the 13-character `tree-root-v1:` prefix throws `StringIndexOutOfBoundsException`. The caller at `dns/update/AwsClient.java:334` catches only `DnsException`, so it escapes and aborts `computeChanges`, failing the entire publish.

2. **Malformed base64 in a nodes entry.** `Algorithm.decode64` (`dns/tree/Algorithm.java:121`) calls `Base64.getUrlDecoder().decode()` directly, which throws `IllegalArgumentException`. `NodesEntry.parseEntry` converts only `InvalidProtocolBufferException` and `UnknownHostException` into `DnsException`, so it escapes both that and the `DnsException`-only catch at `dns/update/AliClient.java:138`, aborting the whole `collectRecords` and the `deploy()` that called it.

A third, latent instance: `BranchEntry.parseEntry` (`dns/tree/BranchEntry.java:19`) does the same unguarded `substring`, and unlike its siblings does not declare `throws DnsException`. Its only caller checks the prefix first, so it is not currently reachable — a hazard for the next caller, not a live bug. `LinkEntry.parseEntry` shows the correct shape: prefix check, length check, and `catch (RuntimeException)` around the decode.

### Reachability — operator-side, not peer-reachable

| Path | Guard | Result |
|---|---|---|
| TCP wire decode | `Message.parse` `catch (Exception)` | contained |
| UDP discovery decode | `P2pPacketDecoder` `catch (Exception)` | contained |
| DNS sync / iteration | `RandomIterator.next()` `catch (Exception)` around `syncRandom()` | contained |
| **DNS publish / collect** | only `catch (DnsException)` | **escapes** |

**No remote peer can trigger these.** Both live instances are on the operator's own publish path: they abort a DNS publish for whoever runs it, they do not give an attacker anything.

All three are pre-existing in libp2p and out of scope for a no-functional-changes PR, so they are reported rather than fixed. Instance 2 is pinned by a test that asserts the *defect* — `AwsClientRecordsTest.malformedBase64InANodesEntryAbortsTheWholeCollection` fails loudly if the escape is ever closed.

## Known security issues — pre-existing, deferred

These exist in libp2p today and are unchanged by this PR. Internalizing the source is what makes them fixable in-tree; each gets its own follow-up:

- Weak PRNG for node ID generation
- `compressPubKey` drops leading zeros
- Unauthenticated handshake
- Unbounded neighbour injection
- `AwsClient` swallows `InterruptedException` without restoring the interrupt flag

Mixing any of them in would break the "no functional changes" claim, which is the only thing that makes a diff this size reviewable.

## Test

| Gate | Result |
|---|---|
| Build — ubuntu24 / macos26 / debian11 / rockylinux | PASS |
| `:p2p:test` | **PASS — 357 tests, 0 failures, 3 skipped** |
| `:framework:test` | PASS |
| `:p2p:checkstyleMain` / `checkstyleTest` | PASS, 0 violations |
| `check-math` | PASS |
| Integration test — single node smoke | PASS |
| `java -jar p2p-standalone.jar --help` | PASS — prints help |
| `project :p2p` on framework `runtimeClasspath` | PASS |
| Zero external `libp2p` artifacts | PASS |
| `gson:2.9.0 -> 2.14.0`, `commons-lang3` at 3.20.0 | PASS |
| No new `implicit_dependency` warning vs base | PASS |

### Coverage

Vendoring puts **15,893 instructions** into the repo-wide denominator that were not there before, and the overall-delta gate is sensitive to exactly that. The choice is between hiding vendored code from the metric and actually testing it; this PR does the latter.

| | Value | Threshold | |
|---|---:|---:|---|
| p2p instruction coverage | **79.15%** (12,580/15,893) | | |
| p2p line coverage | 78.44% (2,983/3,803) | | |
| Changed-line coverage | **78%** | > 60% | PASS |
| Base overall | 79.78% | | |
| PR overall | 79.77% | | |
| Overall delta | **−0.0100%** | ≥ −0.1% | PASS |

The tests added here target what needs no live connection: the web3j copy, every kad discovery message through its own wire bytes, `P2pPacketDecoder`'s drop paths for hostile datagrams, `Channel`'s send guard and exception classification, `ChannelManager.processPeer`'s admission branches, `Tree`'s signing and TXT output, `AwsClient`'s Route53 batching limits, `AliClient.deploy`'s threshold decision, and `P2pService`, which had no test at all.

What remains uncovered needs a real socket: `ConnPoolService.onConnect/onDisconnect/onMessage`, `NodeDetectService`, `PeerClient`, `Channel.init/send`. Upstream's own `SocketTest` for exactly that is entirely commented out, so closing the gap means integration tests with real channels rather than more unit tests.

### Known flakiness

Three `org.tron.common.runtime.vm` tests fail on their first attempt in a full local run and pass on retry — `ValidateMultiSignContractTest.testTip854RejectsMalformedCalldata`, `AllowTvmLondonTest.testBaseFee`, `AllowTvmLondonTest.testStartWithEF`. A full `./gradlew test` on an unmodified `release_v4.8.3` checkout produces the same three; both runs execute an identical set of 3,307 tests, and this PR adds none to `:framework`.

Two ported upstream tests need live DNS against a third-party zone and are left as-is rather than disabled, since they exercise real discovery behaviour: `RandomTest.testRandomIterator` and `SyncTest` both sync `tree://…@nile.trondisco.net` through hard-coded public resolvers. On a runner with restricted egress, or if that DNS tree is re-published or retired, they fail permanently. `LookUpTxtTest` in the same package already `@Ignore`s its network tests, so that is the precedent if reviewers would rather these be skipped than retried; disabling them costs roughly 1.5 points of coverage, which still clears both gates.

A `BindException` in `Metrics.init` can cascade through four framework test classes (`SRMetricsTest`, `PrometheusApiServiceTest`, `JsonrpcServiceTest`, `RpcApiServicesTest`) that each bind the same hard-coded Prometheus port **9527** while the test task runs up to 4 parallel forks. Verified pre-existing by running those four classes together with no p2p tests in the run: it reproduces, and hits a different class each time. The race is java-tron's, not p2p's.


plugins/build.gradle
317787106: [SHOULD] Please update the existing libp2p exclusion in `plugins/build.gradle` to match the local `:p2p` module. The exclusion still targets `io.github.tronprotocol:libp2p`, so it no longer applies. Consequently, Toolkit and ArchiveManifest now bundle P2P and its AWS, Aliyun, and other transitive dependencies, which were previously excluded.

  Building the base and PR locally increased `Toolkit.jar` from 107.85 MiB to 120.80 MiB. Please preserve the original exclusion and remove the corresponding `project(':p2p').jar` dependency from the packaging task, so this migration keeps the tools’ dependency footprint unchanged.

barbatos2011: Thanks, confirmed and fixed in ae791fddb.

The `libp2p` exclude on `:crypto` now matched nothing, so `:crypto -> :common` (`api project(":p2p")`) pulled `:p2p` and its DNS-publish SDKs into both tool jars. It now excludes the project (`java-tron:p2p`), and the `project(':p2p').jar` entry is gone from `binaryRelease`'s `dependsOn`, so that list matches the base again. Nothing under `plugins/src/main` references `org.tron.p2p`; tests still get it via `testImplementation project(":framework")`.

Rebuilt both jars on base, on the previous head and on this commit:

| | `Toolkit.jar` | Entries |
|---|---:|---:|
| Base | 113,086,237 B (107.85 MiB) | 25,068 |
| Previous head | 126,664,753 B (120.80 MiB) | 33,397 |
| ae791fddb | 113,086,237 B (107.85 MiB) | 25,068 |

The entry list after the fix is identical to base, and `ArchiveManifest.jar` gives the same result. The 8,329 extra entries were mostly the AWS SDK v2, BouncyCastle `bcpkix`/`bcutil`, JAXB, the Aliyun SDK, dom4j and dnsjava.

This also corrects my commit 3, which had added `:p2p:jar` to that `dependsOn` list to quiet the `implicit_dependency` warning instead of fixing the exclude. The PR description is updated to match.

